PR #2, opened 2025-12-23
feat(doctor): add hook-singleton check

steveyegge:
## Summary
- Add `gt doctor` check to ensure each agent has at most one handoff bead
- Multiple handoff beads for the same role create ambiguity about which work is actually assigned

## What it checks
- Scans all pinned beads looking for titles matching `{role} Handoff` pattern
- Errors if multiple beads share the same handoff title for a role

## Auto-fix
- Running `gt doctor --fix` closes duplicate handoff beads, keeping the most recent one

## Test plan
- [x] Unit tests added and passing
- [x] Manual testing with `gt doctor`

Closes gt-h6eq.1

### DIFF
--- a/internal/cmd/doctor.go
+++ b/internal/cmd/doctor.go
@@ -100,6 +100,7 @@ func runDoctor(cmd *cobra.Command, args []string) error {
 	d.Register(doctor.NewLifecycleHygieneCheck())
 
 	// Hook attachment checks
+	d.Register(doctor.NewHookSingletonCheck())
 	d.Register(doctor.NewHookAttachmentValidCheck())
 	d.Register(doctor.NewHookSingletonCheck())
 
--- a/internal/doctor/hook_check.go
+++ b/internal/doctor/hook_check.go
@@ -4,10 +4,186 @@ import (
 	"fmt"
 	"os/exec"
 	"path/filepath"
+	"sort"
 	"strings"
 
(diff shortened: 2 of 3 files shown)


aphexcx:
## Post-review fixes (commit 01af1f34)

Addressed all findings from the Codex PR review:

**P0 (Critical):** Routing table keys are now lowercased at construction time in `buildRoutingTable`, matching the case-insensitive `Resolve` contract.

**P1 (Duplicate names):** `buildRoutingTable` now detects duplicate display names across rigs and logs a WARNING to stderr with both addresses. First-writer wins; the operator sees the collision immediately on daemon start.

**P1 (Atomic file delivery):** Text + files are now treated as one atomic delivery unit in the publisher. If ANY file upload fails (transient or permanent), the whole message goes through `handlePostError` — transient unclaims for retry, permanent dead-letters. No more silently dropped attachments.

**P2 (Malformed JSON loop):** Added `quarantineRawClaim` helper that moves unparseable claim files directly to `failed/` via `os.Rename`, bypassing `DeadLetterOutboxMessage` which would re-read the broken JSON and loop.

**P2 (Directory permissions):** `LoadConfig` now checks the parent directory permissions (must be ≤0700), not just the file. Added test coverage.

All 40 tests pass. Build and vet clean.

PR #36, opened 2026-01-03
fix(doctor): use route path directly for agent beads check

newideas99:
## Summary

- Fix agent-beads-exist check to use the route path directly instead of hardcoding `/mayor/rig` suffix
- The check was constructing incorrect beads paths when routes.jsonl uses simple rig names like `my-saas` instead of `my-saas/mayor/rig`

## Problem

When routes.jsonl contains:
```json
{"prefix":"ms-","path":"my-saas"}
```

The doctor check was looking for beads at:
```
~/gt/my-saas/mayor/rig/.beads  (wrong)
```

Instead of:
```
~/gt/my-saas/.beads  (correct)
```

This caused false "agent beads missing" errors even when the beads existed in the correct location.

## Solution

Store the full route path in a `rigInfo` struct and use it directly when creating the beads client, rather than reconstructing an assumed path structure.

## Testing

- [x] `go build ./...` passes
- Manually verified the fix addresses the issue in local Gas Town installation

### DIFF
--- a/internal/doctor/agent_beads_check.go
+++ b/internal/doctor/agent_beads_check.go
@@ -33,6 +33,12 @@ func NewAgentBeadsCheck() *AgentBeadsCheck {
 	}
 }
 
+// rigInfo holds the rig name and its beads path from routes.
+type rigInfo struct {
+	name      string // rig name (first component of path)
+	beadsPath string // full path to beads directory relative to town root
+}
+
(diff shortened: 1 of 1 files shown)


steveyegge (approved):
LGTM. Clean fix for the hardcoded /mayor/rig assumption - using the route path directly is the right approach.

PR #40, opened 2026-01-03
Update installation and workspace setup instructions

jakehemmerle:
```
~ ❯ gt rig add tallstuffnearme https://github.com/jakehemmerle/tallstuffnearme
Error: not in a Gas Town workspace: not in a Gas Town workspace
Usage:
  gt rig add <name> <git-url> [flags]

Flags:
  -h, --help            help for add
      --prefix string   Beads issue prefix (default: derived from name)

~ ❯ cd ~/gt
~/gt main* ❯ gt rig add tallstuffnearme https://github.com/jakehemmerle/tallstuffnearme
Creating rig tallstuffnearme...
  Repository: https://github.com/jakehemmerle/tallstuffnearme
  Warning: Could not create agent beads: creating ta-tallstuffnearme-witness: bd create --json --id=ta-tallstuffnearme-witness --type=agent --title=Witness for tallstuffnearme - monitors polecat health and progress. --description=Witness for tallstuffnearme - monitors polecat health and progress.

role_type: witness
rig: tallstuffnearme
agent_state: idle
hook_bead: null
role_bead: gt-witness-role
cleanup_status: null
active_mr: null
notification_level: null: Error: prefix mismatch: database uses 'gm' but you specified 'ta' (use --force to override)
```

Second error is unrelated to these changes, but first command demonstrates that `gt rig add` must be run in `~/gt`

PR #50, opened 2026-01-03
Papercuts: fixing little things on the golden path

maphew:
# First Round of Golden Path Papercuts

This PR addresses issues encountered while following the golden path (following Readme instructions, and recommendations from `gt` console output). They're not stoppers, not bugs, but annoying and slow things down.

## Issues Fixed

1. **Install Fingerprint Warning** - "Could not verify repo fingerprint" when running `gt install`
2. **Rig Add Beads Error** - "database not initialized" when creating agent beads during `gt rig add`
3. **Unclear Documentation** - Updated Quick Start to clarify correct startup command

## Key Changes

- Added `ensureTownBeadsInitialized()` helper to handle lazy initialization of town beads
- Fixed beads prefix from 'hq' to 'gt' for consistency with agent bead naming
- Skip fingerprint migration if not in a git repository (optional operation)
- Silence benign role slot warnings when role beads don't exist yet

## Related Work

- Complements PR #38 (mayor/rig path fixes) - both PRs address prefix mismatch issues at different levels
- Supersedes documentation-only approach in PR #40 by actually fixing the root cause
- Incorporates feedback and patterns from PR #44 (Quick Start documentation fixes)

## Credits

- Investigation and pattern analysis through collaborative debugging
- Work tracked in Amp thread: https://ampcode.com/threads/T-019b833a-8d57-765a-8d9e-a1391e16f879


maphew:
Note: This PR addresses what is documented in PR #40. Instead of just documenting that `gt rig add` must run in ~/gt, we fix the root cause so it works from anywhere.

Key improvements in this PR:
- **Town beads initialization** - Ensures .beads/ is properly initialized with 'gt' prefix (not 'hq'), handles edge cases where initialization failed
- **Workspace detection** - Improves the check for valid Gas Town workspaces so `gt rig add` works correctly
- **Prefix consistency** - Changes all references from 'hq' prefix to 'gt' prefix for agent coordination

This supersedes the documentation-only approach in #40.

maphew:
## Cooperation with PR #38

PR #38 addresses the same "prefix mismatch" issue at a different level:
- **PR #38** changes where agent bead creation looks for the beads database (always use mayor/rig path)
- **This PR #50** ensures the town-level beads database is properly initialized in the first place

Both fixes are needed and don't conflict:
- PR #38 prevents the wrong path from being used
- PR #50 prevents the database from being misconfigured

**Recommendation:** Both PRs can merge. PR #38's changes are surgical fixes to polecat/crew code. Our changes are about initialization and prefix consistency, which is orthogonal work.

**Credit:** Work in this PR builds on investigation and pattern analysis that would be part of collaborative debugging.

maphew:
## Collaboration with PR #38

### Overlap Analysis

**PR #38** (use mayor/rig path for beads) addresses similar issues but at a different code level:
- Changes  to always use  path
- Changes  to use  path

**This PR #50** doesn't change those files because the fixes work at a different level:
- Ensures town beads database is properly initialized before use
- Fixes beads prefix configuration (gt vs hq)
- Handles lazy initialization for edge cases

### Why Both PRs Are Needed

- **PR #38** prevents incorrect paths from being used
- **PR #50** ensures the database exists and is configured correctly at the source

They're orthogonal fixes that complement each other.

### Recommendation

Both PRs can merge without conflicts:
- PR #38 is surgical path fixes in specific functions
- PR #50 is initialization and configuration at the manager level

**Suggested merge order**: Either order works, but #50 first provides better initialization foundation.

steveyegge:
Thank you for identifying this prefix mismatch issue! The problem you spotted is real - there is indeed inconsistency between the expected prefix and what's being used.

However, the fix direction in this PR is incorrect. Changing the town beads prefix from 'hq' to 'gt' would break the two-level beads architecture:

**Current Design (from architecture docs):**
- **Town-level beads** (`~/gt/.beads/`): Use `hq-*` prefix for Mayor mail, HQ coordination, cross-rig issues
- **Rig-level beads** (`<rig>/.beads/`): Use project prefix (e.g., `gt-*` for gastown) for implementation work

The correct fix is to update the **agent bead creation** to use `hq-*` prefix for town-level agents, not to change town beads to use `gt-*` prefix.

This is being addressed in epic gt-4r1ph: 'Align agent/role beads with two-level architecture' which:
1. Adds `hq-*` prefix helpers for town-level agent beads
2. Updates `gt install` to create `hq-mayor`, `hq-deacon`, `hq-*-role` beads in town beads
3. Updates agent lookups to use the correct tier
4. Provides migration for existing installations

Thanks again for catching this - your investigation helped clarify the architecture!

maphew:
Ahh, thanks for clarifying the purpose of the two prefixes! That makes much more sense to me now.

PR #61, opened 2026-01-03
feat: allow local repo reference clones to save disk

danshapiro:
## Summary
- add `--local-repo` to `gt rig add` and persist it in rig config
- use git reference clones for rigs and crew when a local repo is provided
- add tests covering reference clone behavior and config round-trips

## Test plan
- [x] go test -v -race -short ./...

### DIFF
--- a/internal/cmd/rig.go
+++ b/internal/cmd/rig.go
@@ -251,6 +251,7 @@ Examples:
 // Flags
 var (
 	rigAddPrefix       string
+	rigAddLocalRepo    string
 	rigResetHandoff    bool
 	rigResetMail       bool
 	rigResetStale      bool
@@ -279,6 +280,7 @@ func init() {
 	rigCmd.AddCommand(rigStopCmd)
 
 	rigAddCmd.Flags().StringVar(&rigAddPrefix, "prefix", "", "Beads issue prefix (default: derived from name)")
+	rigAddCmd.Flags().StringVar(&rigAddLocalRepo, "local-repo", "", "Local repo path to share git objects (optional)")
 
 	rigResetCmd.Flags().BoolVar(&rigResetHandoff, "handoff", false, "Clear handoff content")
 	rigResetCmd.Flags().BoolVar(&rigResetMail, "mail", false, "Clear stale mail messages")
@@ -330,6 +332,9 @@ func runRigAdd(cmd *cobra.Command, args []string) error {
 
 	fmt.Printf("Creating rig %s...\n", style.Bold.Render(name))
 	fmt.Printf("  Repository: %s\n", gitURL)
+	if rigAddLocalRepo != "" {
+		fmt.Printf("  Local repo: %s\n", rigAddLocalRepo)
+	}
 
 	startTime := time.Now()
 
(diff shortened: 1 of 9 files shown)


danshapiro:
Note: the CI lint failure appears pre-existing on main. Running golangci-lint locally with new-issue filtering is clean for this PR:\n\n`/home/user/go/bin/golangci-lint run --timeout=5m --new-from-rev=origin/main`\n\nThat reports: "0 issues."

michaellady (commented):
## 7-Point Code Review

### ✅ CORRECTNESS
The implementation correctly uses git's `--reference-if-able` flag to create alternate object stores without changing remotes. The graceful fallback pattern (try reference clone → warn → regular clone) ensures robustness.

Key validation in `resolveLocalRepo`:
- Converts to absolute path with symlink resolution
- Verifies it's a git repository via `IsRepo()`
- Validates origin URL matches the target git URL

One potential issue: the origin URL comparison at line 189 in `rig/manager.go` uses exact string matching. This could fail for equivalent URLs (e.g., `git@github.com:foo/bar` vs `https://github.com/foo/bar`). Consider normalizing URLs before comparison, or accepting the mismatch with a warning rather than rejecting.

### ✅ PERFORMANCE
This is the correct solution for the disk space problem. Using `--reference-if-able`:
- Shares pack files between the local repo and clones
- Creates `.git/objects/info/alternates` pointing to shared objects
- Significantly reduces disk usage when adding multiple rigs for the same project

### ✅ SECURITY
No concerns. The `--reference-if-able` flag is safe:
- It only creates hardlinks/symlinks to objects, no credential sharing
- The fallback behavior handles missing/invalid references gracefully
- Path validation uses `filepath.Abs` and `EvalSymlinks` to prevent traversal

### ✅ ELEGANCE
Good separation of concerns:
- Git operations in `git/git.go`
- Config types updated consistently
- Validation logic isolated in `resolveLocalRepo`
- Fallback pattern is clean and consistent across all clone sites

Minor suggestion: The fallback pattern is repeated 3 times (bare repo, mayor clone, crew clone). Consider extracting a helper:
```go
func (g *Git) CloneWithFallback(url, dest, reference string, bare bool) error {
    // Try with reference first, fallback to regular clone
}
```

### ✅ RESILIENCE
Excellent error handling:
- `--reference-if-able` (vs `--reference`) means git won't fail if reference is invalid
- Explicit fallback to regular clone on reference failure
- Warnings printed but don't block operation
- Tests verify alternates file is created

### ✅ STYLE
Code follows existing patterns:
- Consistent flag naming (`--local-repo`)
- JSON tags include `omitempty` for optional fields
- Test names are descriptive
- Comments explain the purpose of reference clones

### ⚠️ SMELLS (Minor)

1. **URL matching strictness** (mentioned above): Consider:
   ```go
   if origin != gitURL {
       return absPath, fmt.Sprintf("local repo origin %q differs from %q; using anyway", origin, gitURL)
   }
   ```

2. **Missing test for URL mismatch**: Add a test verifying behavior when local repo has different origin.

3. **No cleanup of alternates on rig removal**: If a rig is removed but the local repo remains, this is fine. But if the local repo is deleted, the alternates file will point to a missing location. Git handles this gracefully, but worth documenting.

---

## Summary
Clean implementation of a useful disk-saving feature. The `--reference-if-able` approach is the right choice, and the fallback behavior ensures reliability. **LGTM** ✅

Consider the URL normalization suggestion as a follow-up improvement.

🤖 Reviewed with Claude Code

PR #77, opened 2026-01-03
fix: Remove vestigial rigs/ directory creation

dannomayernotabot:
## Summary
- Removes the unused `rigs/` directory creation from `gt install`
- `gt rig add` creates rigs at town root (e.g., `~/gt/sat/`), not inside `rigs/`
- Updates docs, gitignore, and tests to reflect actual behavior

## Changes
- `internal/cmd/install.go`: Remove rigs/ directory creation
- `internal/cmd/gitinit.go`: Remove `/rigs/*/` gitignore pattern
- `docs/INSTALLING.md`: Update directory structure docs
- `internal/cmd/install_integration_test.go`: Remove rigs/ assertion

Fixes #74

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #79, opened 2026-01-03
fix(rig): Create rigs in rigs/ subdirectory of town root

### LINKED ISSUE #74: gt rig add creates rigs at town root instead of rigs/ directory
## Summary

`gt rig add` creates new rigs at the town root level (`~/gt/<rigname>/`) instead of inside the `rigs/` subdirectory (`~/gt/rigs/<rigname>/`).

## Steps to Reproduce

1. Initialize a town (creates empty `rigs/` directory)
2. Run `gt rig add myrig https://github.com/user/repo.git`
3. Observe rig is created at `~/gt/myrig/` not `~/gt/rigs/myrig/`

## Current Behavior

```
~/gt/
├── .beads/
├── mayor/
├── deacon/
├── rigs/        ← empty, never used
├── sat/         ← rig created here
├── gastown/     ← rig created here
```

## Expected Behavior

Either:
- `gt rig add` should create rigs inside `rigs/` directory, OR
- The empty `rigs/` directory shouldn't be created during init

## Environment

- gt version 0.1.1
- macOS Darwin 24.6.0

dannomayernotabot:
## Summary

Fixes #74 by ensuring `gt rig add` creates rigs inside the `rigs/` subdirectory that `gt install` creates, rather than at the town root level.

This is the **opposite direction** from PR #77 which removed the `rigs/` directory - here we keep `rigs/` and fix the code to actually use it.

### Changes

- Add `RigsDir` constant ("rigs") and `RigPath()` helper to `types.go`
- Add `TownRoot` field to `Rig` struct for explicit town root access
- Update `loadRig` and `AddRig` in `manager.go` to use `rigs/` subdirectory
- Fix code that derived townRoot from `filepath.Dir(rig.Path)`
- Update `daemon.go` path constructions for polecats and witnesses
- Add regression test `TestRigsInSubdirectory` to prevent recurrence

### After this change

```
~/gt/
├── mayor/          # Town-level config
├── rigs/           # All rigs go here
│   └── gastown/    # Rig directory
│       ├── crew/
│       ├── polecats/
│       └── ...
├── .beads/         # Town beads
└── plugins/        # Town plugins
```

## Test plan

- [x] All existing tests pass
- [x] Added regression test `TestRigsInSubdirectory` that verifies:
  - `RigPath()` returns path with `rigs/` subdirectory
  - Rigs are NOT at town root level
  - Rigs ARE in `rigs/` subdirectory
  - Manager loads rigs from correct location
  - `TownRoot` field is set correctly

### DIFF
--- a/internal/cmd/rig.go
+++ b/internal/cmd/rig.go
@@ -361,11 +361,11 @@ func runRigAdd(cmd *cobra.Command, args []string) error {
 	// The conditional routing is necessary because initBeads creates the database at
 	// "<rig>/.beads", while repos with tracked beads have their database at mayor/rig/.beads.
 	if newRig.Config.Prefix != "" {
-		routePath := name
-		mayorRigBeads := filepath.Join(townRoot, name, "mayor", "rig", ".beads")
+		routePath := rig.RigsDir + "/" + name
+		mayorRigBeads := filepath.Join(townRoot, rig.RigsDir, name, "mayor", "rig", ".beads")
 		if _, err := os.Stat(mayorRigBeads); err == nil {
 			// Source repo has .beads/ tracked - route to mayor/rig
-			routePath = name + "/mayor/rig"
+			routePath = rig.RigsDir + "/" + name + "/mayor/rig"
 		}
 		route := beads.Route{
 			Prefix: newRig.Config.Prefix + "-",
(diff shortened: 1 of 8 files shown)


steveyegge:
This PR is superseded by commit 5186cd9 which takes the simpler approach: remove the vestigial `rigs/` directory creation from `gt install` rather than restructure the codebase.

The flat structure works well (similar to Android's top-level layout) and avoids a breaking change for existing installations. Rigs at town root is fine.

Recommend closing this PR.

PR #108, opened 2026-01-04
fix(daemon): prevent orphan daemons via file locking

jakehemmerle:
I had an orphaned process that was chomping through tokens when I thought it was shutdown (150m tok/day idle!) Did confirm this fix manually by creating additional towns as described below.

## Summary

- Add `syscall.Flock()` exclusive lock in `daemon.Run()` to prevent TOCTOU race condition
- Detect race losers in `runDaemonStart()` and report "already running" instead of false success
- Lock is per-town (`townRoot/daemon/daemon.lock`) so multiple GT instances work independently

## Problem

The `gt daemon start` command had a TOCTOU race condition: multiple concurrent starts could all pass the `IsRunning()` check before any wrote the PID file, resulting in orphaned daemons that `gt daemon stop` couldn't kill.

See `BUG_REPORT_daemon_orphan.md` for detailed root cause analysis.

## Solution

Use `syscall.Flock()` with `LOCK_EX|LOCK_NB` to acquire an exclusive lock before writing the PID file. The lock is automatically released by the OS when the process exits (including crashes), so stale locks are not a concern.

## Manual Testing

**Race condition fix (single town):**
```
$ cd ~/gt && gt daemon start & gt daemon start & wait
✓ Daemon started (PID 22803)
● Daemon already running (PID 22803)

$ pgrep -lf "gt daemon"
22803 /Users/jake/go/bin/gt daemon run   # Only ONE process
```

**Multi-town independence:**
```
$ gt install ~/gt-test-one --name test-one
$ gt install ~/gt-test-two --name test-two

$ cd ~/gt-test-one && gt daemon start
✓ Daemon started (PID 19860)

$ cd ~/gt-test-two && gt daemon start  
✓ Daemon started (PID 20216)

$ pgrep -lf "gt daemon"
19860 /Users/jake/go/bin/gt daemon run
20216 /Users/jake/go/bin/gt daemon run   # Both running independently
```

**Double-start in both towns:**
```
$ cd ~/gt-test-one && (gt daemon start & gt daemon start & wait)
✓ Daemon started (PID 22803)
● Daemon already running (PID 22803)

$ cd ~/gt-test-two && (gt daemon start & gt daemon start & wait)
✓ Daemon started (PID 22903)  
● Daemon already running (PID 22903)

$ pgrep -lf "gt daemon"
22803 /Users/jake/go/bin/gt daemon run
22903 /Users/jake/go/bin/gt daemon run   # Exactly 2 daemons, one per town
```

### DIFF
--- a/internal/cmd/daemon.go
+++ b/internal/cmd/daemon.go
@@ -117,7 +117,7 @@ func runDaemonStart(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("starting daemon: %w", err)
 	}
 
-	// Wait a moment for the daemon to initialize
+	// Wait a moment for the daemon to initialize and acquire the lock
 	time.Sleep(200 * time.Millisecond)
 
 	// Verify it started
@@ -127,6 +127,15 @@ func runDaemonStart(cmd *cobra.Command, args []string) error {
 	}
 	if !running {
 		return fmt.Errorf("daemon failed to start (check logs with 'gt daemon logs')")
+	}
+
+	// Check if our spawned process is the one that won the race.
+	// If another concurrent start won, our process would have exited after
+	// failing to acquire the lock, and the PID file would have a different PID.
+	if pid != daemonCmd.Process.Pid {
+		// Another daemon won the race - that's fine, report it
+		fmt.Printf("%s Daemon already running (PID %d)\n", style.Bold.Render("●"), pid)
+		return nil
 	}
 
 	fmt.Printf("%s Daemon started (PID %d)\n", style.Bold.Render("✓"), pid)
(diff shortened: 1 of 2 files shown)


steveyegge (approved):
Solid fix for daemon orphan race condition.

**Reviewed:**
- File locking logic is correct (LOCK_EX|LOCK_NB before PID write)
- Proper cleanup via defers
- Race detection in starter (PID comparison) handles concurrent starts gracefully
- All daemon tests pass
- Merges cleanly with current main

**Note:** syscall.Flock is Unix-only, which is fine for gastown's target platforms.

LGTM - ready to merge.

PR #117, opened 2026-01-04
fix: Commit embedded formulas for go install @latest

rawwerks:
## Summary

- Commits the `internal/formula/formulas/` directory (previously gitignored)
- Fixes `go install github.com/steveyegge/gastown/cmd/gt@latest` failing with:
  ```
  pattern formulas/*.formula.json: no matching files found
  ```
- Updates `.gitignore` with a note to run `go generate ./...` after modifying `.beads/formulas/`

## Problem

The `go:embed` directive in `internal/formula/embed.go` requires the formulas directory at build time. When users install via `go install @latest`, the source is fetched fresh without running `go:generate`, so the embedded formulas don't exist.

## Test plan

- [x] Verified build fails without formulas: `rm -rf internal/formula/formulas && go install ./cmd/gt`
- [x] Verified build succeeds after committing formulas
- [x] `gt version` works after install


akatz-ai:
Just ran into this same issue and independently created a fix before finding this PR. Can confirm this is the correct fix - the gitignored formulas directory breaks `go install @latest` for everyone. +1 for merging.

PR #120, opened 2026-01-05
fix: track embedded formulas for go install to work

akatz-ai:
## Summary

- Track `internal/formula/formulas/` in git so `go install` works from the module proxy

## Problem

Running `go install github.com/steveyegge/gastown/cmd/gt@latest` fails with:

```
go/pkg/mod/github.com/steveyegge/gastown@v0.2.0/internal/formula/embed.go:14:36: pattern formulas/*.formula.json: no matching files found
```

The `//go:embed` directive requires the formula files at build time. The `go:generate` directive copies them from `.beads/formulas/`, but since `internal/formula/formulas/` was gitignored, these files weren't included in the published module.

## Fix

1. Removed the gitignore entry for `internal/formula/formulas/`
2. Committed the generated formula files

After merging and tagging a new release, `go install @latest` will work for everyone.

## Test plan

- [x] Verified `go build ./cmd/gt` succeeds


akatz-ai:
Closing as duplicate of #117 which addresses the same issue.

PR #140, opened 2026-01-05
docs(mayor): Add Polecat Operations section

greghughespdx:
## Problem

Users look for `gt polecat spawn` when they should use `gt sling`.

## Changes

Add "Polecat Operations" section to Mayor template clarifying that `gt sling` is THE command for spawning polecats with work.

Searching for polecat spawn you are? Here it is not. Hrmmm.


steveyegge (approved):
LGTM - helpful clarification on polecat operations and the gt sling workflow.

PR #155, opened 2026-01-05
fix: log rig discovery errors instead of silently swallowing (gt-rsnj9)

michaellady:
## Summary
- Logs rig discovery errors instead of silently swallowing them
- Improves debugging by making error conditions visible

## Test plan
- [x] All tests pass (`go test ./...`)
- Rebased on current main


michaellady:
Consolidated into michaellady/gastown#3

PR #208, opened 2026-01-06
Claude/add auditor role

nikspatel007:


nikspatel007:
Apologies. I thought I was creating PR in my forked version. I will continue to work on it in a forked version. I can create a request again when it is ready. Thanks for checking.

PR #201, opened 2026-01-05
[DO NOT MERGE] feat: Add Cursor Agent support for multi-agent Gas Town

michaellady:
## Summary

Adds support for Cursor Agent as a compatible agent in Gas Town's multi-agent infrastructure.

**Changes:**
- Add `AgentCursor` preset configured for headless mode (`cursor-agent -p -f`)
- Introduce `ProcessNames` field on `AgentPresetInfo` for flexible agent detection
- Add `IsAgentRunning(session, processNames)` helper in tmux package
- Add `GetProcessNames(agentName)` to retrieve detection patterns for any agent

This enables Gas Town to orchestrate Cursor Agent alongside Claude Code in hybrid multi-agent workflows.

## Test Plan

- [x] Unit tests for agent preset configuration
- [x] Unit tests for GetProcessNames helper
- [ ] Manual verification that cursor preset is recognized by `gt` commands

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #168, opened 2026-01-05
[DO NOT MERGE] feat: add --model flag to gt sling for specifying Claude model

michaellady:
## Summary
- Adds `--model` flag to `gt sling` for specifying Claude model when spawning polecats
- Accepts model aliases ('sonnet', 'opus') or full model names ('claude-sonnet-4-5-20250929')
- Model is passed through to claude CLI via `--model` flag

## Example
```
gt sling gt-abc greenplace --model opus
gt sling issue-123 gastown --model claude-sonnet-4-5-20250929
```

## Test plan
- [x] Build succeeds (`go build ./...`)
- [x] Tests pass (`go test ./internal/cmd/... ./internal/session/...`)
- [x] Help text shows new flag (`gt sling --help`)
- [ ] Manual test: spawn polecat with `--model` flag and verify Claude starts with correct model

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #110, opened 2026-01-04
Feature: Allow specifying Claude model when assigning work to polecats


sandoak:
## Feature Request

When assigning work to a polecat (via `gt mail send` or `gt sling`), it would be useful to specify which Claude model the polecat should use.

### Use Case
- Some tasks benefit from Opus (complex reasoning, architecture)
- Some tasks are fine with Sonnet (routine implementation)
- Some quick tasks could use Haiku (simple fixes, formatting)

### Proposed Solution
Add a `--model` flag when dispatching work:

```bash
gt sling <issue> <rig> --model opus
gt mail send <polecat> --model sonnet -s "WORK: ..."
```

Or configure default model per polecat in town config.

### Benefits
- Cost optimization (use cheaper models for simple tasks)
- Quality optimization (use Opus for critical work)
- Flexibility in multi-agent workflows

---
*Suggested via Claude Code overseer session*

julianknutsen:
This is implemented via the `role_agents` configuration system. Closing as implemented.

### How to control which model polecats use

**Per-role at the town level** — in your town's `settings.json`:
```json
{
  "role_agents": {
    "polecat": "claude --model opus",
    "deacon": "claude --model sonnet"
  }
}
```

**Per-role at the rig level** — rig settings override town settings, so you can run opus polecats in one rig and sonnet polecats in another.

**Custom agent presets** — define named agent configs with `gt config agent set`:
```bash
gt config agent set claude-opus "claude --model claude-opus-4"
```

Then reference them in `role_agents`:
```json
{
  "role_agents": {
    "polecat": "claude-opus"
  }
}
```

This covers the core use cases from the issue (cost optimization, quality optimization per task complexity). There's no `gt sling --model` flag for one-off overrides, but the config-based approach gives persistent per-role and per-rig control.

Issue #102, opened 2026-01-04
IsClaudeRunning detection fails - agents restart every 3 minutes


johann-taberlet:
## Problem

Witness and Refinery agents are detected as "not running" and restarted every 3 minutes by the daemon, even though Claude Code is actively running in the sessions.

## Symptoms

Daemon log shows repeated restarts:
```
2026/01/04 18:24:25 Witness for pixelforge not running per agent bead, starting...
2026/01/04 18:24:25 Witness session for pixelforge started successfully
2026/01/04 18:27:30 Witness for pixelforge not running per agent bead, starting...
2026/01/04 18:27:30 Witness session for pixelforge started successfully
```

This happens every 3 minutes (daemon heartbeat interval).

## Root Cause

The `IsClaudeRunning()` function in `internal/tmux/tmux.go` checks if `pane_current_command == "node"`:

```go
func (t *Tmux) IsClaudeRunning(session string) bool {
    cmd, err := t.GetPaneCommand(session)
    if err != nil {
        return false
    }
    return cmd == "node"
}
```

However, Claude Code 2.0.76 reports the **version number** as `pane_current_command`, not "node":

```bash
$ tmux display-message -t gt-pixelforge-witness -p "#{pane_current_command}"
2.0.76
```

The actual process tree shows Claude IS running:
```
zsh (pane pid)
└── claude (child process)
```

## Environment

- Claude Code version: 2.0.76
- macOS (Darwin 25.2.0)
- tmux with default configuration

## Workaround

Added detection for "claude" command name and version patterns in `internal/tmux/tmux.go`:

```go
import "regexp"

// IsClaudeRunning checks if Claude appears to be running in the session.
// Only trusts the pane command - UI markers in scrollback cause false positives.
func (t *Tmux) IsClaudeRunning(session string) bool {
    // Check pane command - Claude can report as "node", "claude", or version like "2.0.76"
    cmd, err := t.GetPaneCommand(session)
    if err != nil {
        return false
    }
    // Direct matches for known command names
    if cmd == "node" || cmd == "claude" {
        return true
    }
    // Check for version pattern (e.g., "2.0.76") - Claude Code shows version as pane command
    matched, _ := regexp.MatchString(`^\d+\.\d+\.\d+`, cmd)
    return matched
}
```

After applying this fix and restarting the daemon:
```
2026/01/04 18:33:13 Witness for pixelforge session healthy (Claude running), skipping restart despite stale bead
[...]

PR #235, opened 2026-01-07
feat: add per-command --agent overrides

joshuavial:
PR summary: Per-command runtime overrides (`--agent`) + runtime-aware liveness + daemon/role config fixes

This PR makes Gas Town’s “which LLM runtime/model do we run?” decision controllable per command, without forcing you to change town/rig defaults. It also fixes a couple of downstream assumptions that were Claude-specific (tmux liveness checks, permissions dialog handling) so non-Claude agents don’t get treated as “dead” and restarted in loops.

## Why / Context

Gas Town already supports:
- Town-level agent aliases (`gt config agent set …`) and a town default agent (`gt config default-agent …`).
- Rig-level default agent selection via `RigSettings.Agent`.

But it was hard to do **one-off overrides** like:
- “Sling this one issue using `codex --thinking low`” without changing town/rig config.
- “Start Mayor in gemini for a session” without resetting defaults.

Also, various parts of the system assumed “Claude == node process in tmux”, which breaks when you use other agents (and even Claude behavior can vary across installs). This was tracked upstream.

### 1) `--agent` per-command overrides (new CLI surface)

You can now override the runtime on key commands:

- `gt sling <bead> <rig> --agent <alias>`
  - Threads the override into polecat spawn and session start.
  - Works in both tmux-start and `--naked` mode (printed command reflects the override).

- `gt start --agent <alias>`
  - Starts Mayor + Deacon using the override for that invocation.

- `gt mayor start|attach|restart --agent <alias>`
- `gt deacon start|attach|restart --agent <alias>`

- Crew:
  - `gt start crew <name> --agent <alias>`
  - `gt crew start … --agent <alias>`
  - `gt crew at <name> --agent <alias>`

Important: the override is strictly per invocation; it does not mutate town/rig settings.

### 2) Centralized override resolution and command building (config-layer)

To avoid “every call site implements its own rules”, this PR adds a single config-level resolver and helper builders:

- `ResolveAgentConfigWithOverride(townRoot, rigPath, agentOverride)`
  - If `agentOverride` is set: must resolve to built-in or custom alias; otherwise error.
  - If not set: uses rig settings first (when rigPath provided), then town default.

- Builders that use the resolver and return full startup commands:
  - `BuildAgentStartupCommandWithAgentOverride(...)`
  - `BuildCrewStartupCommandWithAgentOverride(...)`
  - `BuildPolecatStartupCommandWithAgentOverride(...)`
  - `GetRuntimeCommandWithAgentOverride(...)`

This keeps behavior consistent across sling/start/daemon paths.

### 3) Fix: rig agent settings honored for witness/refinery start paths

Previously, several witness/refinery start paths called startup builders with `rigPath=""`, which causes resolution to fall back to town-level defaults. That meant `RigSettings.Agent` could be ignored for these roles.

Now, witness/refinery starts consistently pass the rig path so rig agent settings are honored by default.

### 4) Runtime-aware tmux liveness checks (replaces Claude-only assumptions)

Previously: liveness was effectively “pane_current_command == node” via `IsClaudeRunning`, and a few places used that to decide “zombie session → kill/restart”.

Now:
- Added `tmux.IsAgentRunning(session, expectedPaneCommands...)`
  - If expected commands provided: exact match on `pane_current_command`.
  - If not: “any non-shell foreground command” counts as running (anything not in `constants.SupportedShells`).
- Added `config.ExpectedPaneCommands(rc)` to map runtime → expected tmux command:
  - `claude` → `node`
  - everything else → `basename(command)` (e.g. `codex`, `gemini`)

Callers that know the configured runtime (daemon start logic, crew/witness/refinery ensure logic) pass the expected pane command list. Generic helpers that don’t know (or are intentionally generic) use the “non-shell” fallback.

Also: the Claude-specific bypass-permissions dialog dismissal is now gated so it only runs when the runtime is actually Claude, preventing garbage keystrokes sent to other agents.

### 5) Fix: role bead ID canonicalization in daemon lifecycle

Daemon lifecycle role-config lookup was using legacy role bead IDs (`gt-<role>-role`). Canonical role beads are town-level `hq-<role>-role`. This mismatch can cause “missing role bead” confusion and config not being applied.

Now daemon lifecycle:
- Prefers canonical `hq-<role>-role`.
- Falls back to legacy `gt-<role>-role` if the canonical doesn’t exist (backward compatibility / migration-friendly).

## Tests added / updated

Unit tests:
- Extended `internal/config/loader_test.go`:
  - override resolution behavior
  - startup command builders with overrides
  - rigPath-aware resolution behavior
  - `ExpectedPaneCommands` mapping

tmux tests:
- Updated `internal/tmux/tmux_test.go` to cover `IsAgentRunning` behavior in a “shell-only” session.

Integration tests:
- Added `internal/daemon/role_config_integration_test.go` (tag `integration`):
  - Verifies daemon role config lookup prefers canonical role bead IDs and falls back to legacy in the expected situations.
  - Skips cleanly if `bd` is not installed.

Validation run locally:
- `go test ./...`
- `go test -tags=integration ./...`
- `go build ./...`
- `go install ./cmd/gt`

## Documentation updates (onboarding + reference)

- `docs/INSTALLING.md`: new “Configure Agents (Optional)” section:
  - `gt config agent list/set/default-agent`
  - examples showing aliases that embed model/thinking flags
  - examples of per-command `--agent` overrides

- `docs/reference.md` and `README.md`: updated examples to mention `--agent` overrides.

`gt install` output now includes an explicit “Configure agents” next step (`gt config agent list`) to make discovery easier.

## Decisions / Non-goals

- No new `--model` flag in this PR.
  - Rationale: agent aliases already express “model choice” in a runtime-specific way (`claude --model haiku`, `codex --thinking low`, etc.).
  - This avoids building a cross-runtime model abstraction prematurely.

- Override errors on unknown agent names (explicit failure).
  - Rationale: silent fallback would be confusing and could launch expensive/default models unintentionally.

- Liveness checks still rely on `pane_current_command`.
  - This PR makes it runtime-aware and avoids Claude-only restarts, but it doesn’t fully solve every tmux/reporting oddity in the general case; it’s the minimal, consistent abstraction that unblocks multi-agent usage.

## Upstream references consulted (issues / PRs / beads)

GitHub issues:
- #102 “IsClaudeRunning detection fails …” — drove the tmux liveness rework and the “don’t assume Claude/node everywhere” changes.
- #110 “Allow specifying Claude model when assigning work to polecats” — informed the decision to lean on agent aliases for model selection rather than add `--model` in this PR.
- #10 “Agent Agnostic Engine … multi-provider support” — general direction; this PR is a small enabling step.

GitHub PRs (reviewed for direction):
- PR #168 (`--model` flag for sling) — noted review concerns and “parking” status.
- PR #201 (agent detection abstractions) — informed approach for liveness checks; also parked upstream.

Upstream beads referenced:
- `gt-dc2fs` (stevey, closed): runtime configuration foundation.
- `gt-qao` (open): CLI design mentions `--agent` for mayor commands.
- `gt-th7` (open): agent abstraction layer direction.
- `gt-otgn3` (closed): agent registry caching/testing utilities.

## Implementation highlights (for reviewers)

Key files:
- Config plumbing: `internal/config/loader.go`
- `--agent` for sling: `internal/cmd/sling.go`, `internal/cmd/polecat_spawn.go`
- `--agent` for start/mayor/deacon/crew: `internal/cmd/start.go`, `internal/cmd/mayor.go`, `internal/cmd/deacon.go`, `internal/cmd/crew.go`, `internal/cmd/crew_at.go`, `internal/cmd/crew_lifecycle.go`
- Rig agent honored for witness/refinery: `internal/cmd/witness.go`, `internal/cmd/start.go`, `internal/refinery/manager.go`, `internal/daemon/daemon.go`, `internal/daemon/lifecycle.go`
- tmux runtime-aware checks: `internal/tmux/tmux.go`
- Integration test: `internal/daemon/role_config_integration_test.go`

### DIFF
--- a/README.md
+++ b/README.md
@@ -414,4 +414,4 @@ MIT License - see LICENSE file for details
 
 ---
 
-**Getting Started:** Run `gt install ~/gt --git && cd ~/gt && gt mayor attach` and tell the Mayor what you want to build!
+**Getting Started:** Run `gt install ~/gt --git && cd ~/gt && gt config agent list && gt mayor attach` (or `gt mayor attach --agent codex`) and tell the Mayor what you want to build!
--- a/docs/INSTALLING.md
+++ b/docs/INSTALLING.md
@@ -130,6 +130,29 @@ gt doctor              # Run health checks
 gt status              # Show workspace status
 ```
 
+### Step 5: Configure Agents (Optional)
+
+Gas Town supports built-in runtimes (`claude`, `gemini`, `codex`) plus custom agent aliases.
+
+```bash
+# List available agents
+gt config agent list
+
+# Create an alias (aliases can encode model/thinking flags)
+gt config agent set codex-low "codex --thinking low"
+gt config agent set claude-haiku "claude --model haiku --dangerously-skip-permissions"
+
+# Set the town default agent (used when a rig doesn't specify one)
+gt config default-agent codex-low
+```
+
+You can also override the agent per command without changing defaults:
+
+```bash
+gt start --agent codex-low
+gt sling issue-123 myproject --agent claude-haiku
+```
+
 ## Minimal Mode vs Full Stack Mode
 
 Gas Town supports two operational modes:
--- a/internal/cmd/install.go
+++ b/internal/cmd/install.go
@@ -253,6 +253,8 @@ func runInstall(cmd *cobra.Command, args []string) error {
 	}
 	fmt.Printf("  %d. Add a rig: %s\n", step, style.Dim.Render("gt rig add <name> <git-url>"))
 	step++
+	fmt.Printf("  %d. (Optional) Configure agents: %s\n", step, style.Dim.Render("gt config agent list"))
+	step++
 	fmt.Printf("  %d. Enter the Mayor's office: %s\n", step, style.Dim.Render("gt mayor attach"))
 
 	return nil
@@ -337,9 +339,9 @@ func ensureRepoFingerprint(beadsPath string) error {
 
 // initTownAgentBeads creates town-level agent and role beads using hq- prefix.
 // This creates:
-// - hq-mayor, hq-deacon (agent beads for town-level agents)
-// - hq-mayor-role, hq-deacon-role, hq-witness-role, hq-refinery-role,
-//   hq-polecat-role, hq-crew-role (role definition beads)
+//   - hq-mayor, hq-deacon (agent beads for town-level agents)
+//   - hq-mayor-role, hq-deacon-role, hq-witness-role, hq-refinery-role,
+//     hq-polecat-role, hq-crew-role (role definition beads)
 //
 // These beads are stored in town beads (~/gt/.beads/) and are shared across all rigs.
 // Rig-level agent beads (witness, refinery) are created by gt rig add in rig beads.


steveyegge:
Merged manually after resolving conflicts with main's Manager.Start() refactoring. Thanks @joshuavial\!

joshuavial:
I just noticed that the witness start and restart stuff might need some work - I had to resort to manually editing the town settings to get them on different agents.

PR #239, opened 2026-01-07
fix: Isolate Claude settings outside source repos, unify agent startup

julianknutsen:
Remaining fixes on top of c2451b85e77a33bdc67e763dd99f2fae39ac7014

## Summary

Fixes critical bugs where Claude Code settings and context files were placed in wrong locations, causing:
- **Hooks not executing** - Agents missing required SessionStart/Stop hooks for deacon nudges and cost tracking
- **Wrong prompts/instructions** - Source repo CLAUDE.md files overriding Gas Town agent context
- **Settings pollution** - Town-root settings inherited by ALL agents via Claude's upward directory traversal
- **Leaked credentials/config** - Source repo `.claude/settings.json` and `.mcp.json` affecting agent behavior

### The Bug

Claude Code searches for configuration files (`.claude/settings.json`, `CLAUDE.md`) starting from the working directory and traversing upward to parent directories. Gas Town was placing files in locations that caused unintended inheritance:

1. **Mayor files at town root (`~/gt/`)** - Settings and CLAUDE.md here were inherited by every agent (crew, polecats, refinery, etc.) because all agent paths are descendants of the town root.

2. **Settings inside git clones** - Placing `.claude/settings.json` inside worktrees (e.g., `refinery/rig/.claude/`) meant they could be committed/pushed, and conflicted with source repos that have their own Claude configuration.

3. **No sparse checkout** - Source repos were cloned without any exclusions, so repos with their own `.claude/`, `CLAUDE.md`, `CLAUDE.local.md`, or `.mcp.json` would override Gas Town agent configuration.

### The Fix

- **Move mayor files to `~/gt/mayor/`** - Isolated from all other agents
- **Place settings in parent directories** - e.g., `refinery/.claude/settings.json` is found by `refinery/rig/` via upward traversal, without polluting the git repo
- **Expand sparse checkout** - Now excludes all Claude context files: `.claude/`, `CLAUDE.md`, `CLAUDE.local.md`, `.mcp.json`
- **Add comprehensive doctor checks** - Detect and auto-fix all misconfiguration scenarios

## Test plan

- [x] All existing tests pass
- [x] New doctor check tests cover all failure modes
- [x] Sparse checkout tests verify all patterns and upgrade scenarios
- [x] Manual test: `gt doctor` detects stale settings in existing town
- [x] Manual test: `gt doctor --fix` migrates settings and restarts agents
- [x] Manual test: New rig creation places settings in correct locations

### `claude-settings` Check

[...]

PR #243, opened 2026-01-07
fix(doctor): warn instead of killing sessions for stale town-root settings

julianknutsen:
## Summary

  When `gt doctor --fix` detects stale Claude settings at town root, it was automatically killing ALL Gas Town sessions (gt-* and hq-*). This is too disruptive because:

  1. Deacon runs `gt doctor` automatically, creating a restart loop
  2. Active crew/polecat work could be lost mid-task
  3. Settings are only read at startup, so running agents already have the config loaded

  Now it warns the user instead:
    ⚠ Town-root settings were moved. Restart agents to pick up new config:
        gt up --restart

  Addresses PR #239 feedback.

  ## Testing

  - [x] Added `TestClaudeSettingsCheck_TownRootSettingsWarnsInsteadOfKilling`

  ## Checklist

### DIFF
--- a/internal/doctor/claude_settings_check.go
+++ b/internal/doctor/claude_settings_check.go
@@ -10,6 +10,7 @@ import (
 
 	"github.com/steveyegge/gastown/internal/claude"
 	"github.com/steveyegge/gastown/internal/session"
+	"github.com/steveyegge/gastown/internal/style"
 	"github.com/steveyegge/gastown/internal/templates"
 	"github.com/steveyegge/gastown/internal/tmux"
 	"github.com/steveyegge/gastown/internal/workspace"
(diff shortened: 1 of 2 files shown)


steveyegge (approved):
**Post-merge review** ✓

Good change. The rationale is solid:
1. Deacon auto-running `gt doctor` would create restart loops
2. Active work could be lost mid-task
3. Running agents already have config loaded in memory

The fix is clean and focused - replaces destructive auto-kill with a clear warning message. The test adequately covers the new behavior (verifying files are cleaned up without session killing).

One minor observation: the warning format uses `%s` for the warning emoji, but this is fine since `style.Warning.Render()` handles the styling.

PR #244, opened 2026-01-07
  test(polecat): skip beads-unavailable test when bd is installed

julianknutsen:
## Summary

  `TestGetReturnsWorkingWithoutBeads` assumes bd is not available and expects polecat state to default to `StateWorking`. When bd is installed locally, it actually queries beads and returns the real state, causing the test to fail.

  Skip the test when bd is detected to avoid environment-dependent failures.

  ## Related Issue

  N/A - flaky test fix

  ## Testing

  - [x] Lint passes

  ## Checklist

### DIFF
--- a/internal/polecat/manager_test.go
+++ b/internal/polecat/manager_test.go
@@ -2,6 +2,7 @@ package polecat
 
 import (
 	"os"
+	"os/exec"
 	"path/filepath"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


steveyegge (approved):
LGTM. Clean fix for environment-dependent test flakiness.

The skip condition correctly handles the case where `bd` is installed locally - the test's assumption (bd unavailable → default to StateWorking) doesn't hold when bd is present and can query actual beads state.

Code quality:
- Import added in correct alphabetical order
- Clear skip message explains the rationale
- Uses standard `t.Skip()` pattern
- Minimal, focused change

PR #284, opened 2026-01-08
fix: auto-inject GT_ROOT in agent startup commands

vessenes:
## Problem

Agents running in tmux sessions (deacon, polecats, etc.) couldn't find formulas when running `bd formula list`:

```
⏺ Bash(bd formula list)
  ⎿  No formulas found.

     Search paths:
       /Users/pv/src/gt/deacon/.beads/formulas
       /Users/pv/.beads/formulas
```

The `bd` tool searches for formulas in:
1. `.beads/formulas/` (project directory)
2. `~/.beads/formulas/` (user directory)  
3. `$GT_ROOT/.beads/formulas/` (town directory, if GT_ROOT is set)

The town's formulas live in `$GT_ROOT/.beads/formulas/`, but GT_ROOT wasn't being exported to the agent process.

## Root Cause

Commit 18f9c28 ("feat: add GT_ROOT environment variable to all agent session starts") added GT_ROOT to tmux sessions via `SetEnvironment()`, but **tmux SetEnvironment only affects new panes** - it doesn't propagate to processes started via `SendKeys()`.

The startup command sent to Claude exports `GT_ROLE`, `BD_ACTOR`, and `GIT_AUTHOR_NAME`, but not `GT_ROOT`.

## Solution

Auto-inject `GT_ROOT` into the exported environment variables in `BuildStartupCommand()` and `BuildStartupCommandWithAgentOverride()`. 

Both functions already calculate `townRoot` internally to resolve agent configuration - now they also add it to `envVars` when available. This is a minimal change (1 file, +20/-4 lines) that fixes the issue without requiring changes to any callers.

## Testing

After rebuilding `gt` and restarting the deacon:
- `bd formula list` now correctly finds formulas in the town's `.beads/formulas/` directory
- All existing tests pass

### DIFF
--- a/internal/config/loader.go
+++ b/internal/config/loader.go
@@ -1062,20 +1062,28 @@ func findTownRootFromCwd() (string, error) {
 // envVars is a map of environment variable names to values.
 // rigPath is optional - if empty, tries to detect town root from cwd.
 // prompt is optional - if provided, appended as the initial prompt.
+// GT_ROOT is automatically added to envVars if town root can be determined.
 func BuildStartupCommand(envVars map[string]string, rigPath, prompt string) string {
 	var rc *RuntimeConfig
+	var townRoot string
 	if rigPath != "" {
 		// Derive town root from rig path
-		townRoot := filepath.Dir(rigPath)
+		townRoot = filepath.Dir(rigPath)
 		rc = ResolveAgentConfig(townRoot, rigPath)
 	} else {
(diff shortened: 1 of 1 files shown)


julianknutsen:
dup of https://github.com/steveyegge/gastown/pull/268

steveyegge:
Thanks for the fix! This was addressed by #268 which takes the same approach but also adds a doctor check for GT_ROOT verification. Appreciate you digging into this issue.

PR #287, opened 2026-01-09
Add mandatory work completion checklist to polecat template

bkoltai:
## Summary
- Adds a clearly visible 🚨 MANDATORY section at the end of the polecat context
- Overrides any other session end instructions to ensure polecats always run `gt done`
- Addresses the failure mode where polecats commit but don't land their work

## Test plan
- [ ] Verify template renders correctly with `gt prime` in a polecat session
- [ ] Verify the checklist appears at the end of the generated polecat context

### DIFF
--- a/internal/templates/roles/polecat.md.tmpl
+++ b/internal/templates/roles/polecat.md.tmpl
@@ -270,3 +270,27 @@ gt mail send mayor/ -s "Need coordination" -m "..."
 Polecat: {{ .Polecat }}
 Rig: {{ .RigName }}
 Working directory: {{ .WorkDir }}
+
+---
+
+## 🚨 MANDATORY: Work Completion Checklist
+
+**This checklist overrides any other session end instructions.**
+
+When your work is complete:
+
+```bash
+# 1. Verify clean state
+git status                    # Should show nothing to commit
+
+# 2. Close your issue
+bd close <issue-id>           # Mark the bead complete
+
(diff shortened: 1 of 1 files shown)


steveyegge:
Thanks for catching this gap in our polecat completion workflow! You're right that polecats need clearer reminders about landing their work.

I've addressed this with a slightly different approach in c2c6ddea that avoids conflicting with our Refinery workflow:

**What I added to the polecat template:**
- A prominent "⚠️ CRITICAL" warning that work isn't complete until `gt done`
- A "Pre-Submission Checklist" with `git status` and `git log` verification
- Explicit note that polecats should NOT manually close the root issue

**Why the different approach:**
The `bd close <issue-id>` instruction in the PR conflicts with our Refinery workflow. From `mol-polecat-work.formula.toml`:

> Do NOT close the issue. The Refinery will close it after successful merge.
> This enables conflict-resolution retries without reopening closed issues.

If polecats close their root issue, the Refinery can't dispatch a fresh polecat for conflict resolution if needed.

The new template text gives polecats the visibility you were after while keeping the Refinery workflow intact. Thanks again for the improvement idea!

bkoltai:
Thanks for the response @steveyegge. This PR was auto-generated by my `gt` to address [this issue](https://github.com/steveyegge/gastown/issues/288) (also generated by `gt`).

Can you provide any guidance on the issue? Do you need more info to do so?

PR #293, opened 2026-01-09
Fix witness start/restart to honor role start_command (agent override wins)

joshuavial:
## Summary
- Add --agent override support for gt witness start/restart.
- Honor hq-witness-role start_command + env vars for manual start/restart.
- Ensure --agent override takes precedence over role start_command.
- Add tests for start_command behavior and override precedence.

### DIFF
--- a/internal/formula/formulas/mol-deacon-patrol.formula.toml
+++ b/internal/formula/formulas/mol-deacon-patrol.formula.toml
@@ -23,7 +23,7 @@ Witnesses detect it and escalate to the Mayor.
 The Deacon's agent bead last_activity timestamp is updated during each patrol
 cycle. Witnesses check this timestamp to verify health."""
 formula = "mol-deacon-patrol"
-version = 7
+version = 6
 
 [[steps]]
 id = "inbox-check"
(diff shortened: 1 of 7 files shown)


internal/cmd/rig.go
```diff
@@ -759,7 +759,7 @@ func runRigBoot(cmd *cobra.Command, args []string) error {
 	} else {
 		fmt.Printf("  Starting witness...\n")
 		witMgr := witness.NewManager(r)
-		if err := witMgr.Start(false); err != nil {
+		if err := witMgr.Start(false, ""); err != nil {
```
julianknutsen, line 762: do we want to pass in the config values here as a default if they exist?

internal/cmd/start.go
```diff
@@ -233,7 +233,7 @@ func startRigAgents(t *tmux.Tmux, townRoot string) {
 			fmt.Printf("  %s %s witness already running\n", style.Dim.Render("○"), r.Name)
 		} else {
 			witMgr := witness.NewManager(r)
-			if err := witMgr.Start(false); err != nil {
+			if err := witMgr.Start(false, ""); err != nil {
```
julianknutsen, line 236: same question about parsing config

internal/daemon/daemon.go
```diff
@@ -378,15 +378,15 @@ func (d *Daemon) ensureWitnessRunning(rigName string) {
...
+	// startup readiness waits, and crucially - startup/propulsion nudges (GUPP).
 	// It returns ErrAlreadyRunning if Claude is already running in tmux.
 	r := &rig.Rig{
 		Name: rigName,
 		Path: filepath.Join(d.config.TownRoot, rigName),
 	}
 	mgr := witness.NewManager(r)
 
-	if err := mgr.Start(false); err != nil {
+	if err := mgr.Start(false, ""); err != nil {
```
julianknutsen, line 389: same question

internal/formula/formulas/mol-deacon-patrol.formula.toml
```diff
@@ -23,7 +23,7 @@ Witnesses detect it and escalate to the Mayor.
 The Deacon's agent bead last_activity timestamp is updated during each patrol
```
julianknutsen, line 23: looks like unintentional changes in the formulas folder

internal/witness/manager.go
```diff
@@ -172,6 +169,18 @@ func (m *Manager) Start(foreground bool) error {
...
 
+	roleConfig, err := m.roleConfig()
+	if err != nil {
+		_ = t.KillSession(sessionID)
+		return err
+	}
+
+	townRoot := m.townRoot()
+	// Apply role config env vars if present (non-fatal).
+	for key, value := range roleConfigEnvVars(roleConfig, townRoot, m.rig.Name) {
```
julianknutsen, line 180: see https://github.com/steveyegge/gastown/pull/294 that cleans this up. Since the role bead is never updated, I'm not sure we want to use it as the source of truth

steveyegge:
Merged to main with --env flag enhancement. Thanks for the contribution!

joshuavial:
> Seems like a general architecture question about the source of truth between role_beads and config that we need to figure out.

agree @julianknutsen , I wasn't sure if the role bead start_command is the right approach or if we should funnel everything through named agents. Personally I think there is a case for ditching role beads and going with cli param > env var > config

PR #296, opened 2026-01-09
add the ability to configure the branch naming convention at the rig level

veenified:
## Summary
Some organizations have standards around how branches need to be formatted. To suit those branch naming standards, I suggest adding a rig-level configuration option that allows a user to set the branch naming format.

## Related Issue
none

## Changes
- Add branch_naming configuration to rig settings with a polecat_branch_template option
- Support template variables: {name} (polecat name), {timestamp} (base36 milliseconds), {rig} (rig name)
- Default template remains polecat/{name}-{timestamp} for backwards compatibility
- Add branch name validation to catch invalid git branch characters
- Include comprehensive tests for branch name building and validation

##  Example Configuration

```
  {
    "type": "rig-settings",
    "version": 1,
    "branch_naming": {
      "polecat_branch_template": "work/{rig}/{name}-{timestamp}"
    }
  }
```

## Testing
- Unit tests for BuildPolecatBranchName with various template patterns
- Unit tests for ValidateBranchName covering invalid git branch characters
- Manual test: Create polecat with default template
- Manual test: Create polecat with custom template in rig settings

## Checklist

### DIFF
--- a/internal/config/types.go
+++ b/internal/config/types.go
@@ -182,13 +182,14 @@ type RigConfig struct {
 
 // RigSettings represents per-rig behavioral configuration (settings/config.json).
 type RigSettings struct {
-	Type       string            `json:"type"`                  // "rig-settings"
-	Version    int               `json:"version"`               // schema version
-	MergeQueue *MergeQueueConfig `json:"merge_queue,omitempty"` // merge queue settings
-	Theme      *ThemeConfig      `json:"theme,omitempty"`       // tmux theme settings
-	Namepool   *NamepoolConfig   `json:"namepool,omitempty"`    // polecat name pool settings
-	Crew       *CrewConfig       `json:"crew,omitempty"`        // crew startup settings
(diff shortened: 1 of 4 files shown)


steveyegge:
Thanks for the contribution! After reviewing, we're going to pass on this one.

The key consideration is that polecat branches are ephemeral and never pushed to origin - they exist only locally in worktrees and get cleaned up by git GC. Since CI/CD pipelines, PR reviewers, and remote tooling never see these branches, org naming conventions don't really apply to them.

The current `polecat/{name}-{timestamp}` naming is unique, descriptive, and working well. Adding configurable templates would introduce new config surface area, template parsing, and validation logic for a use case that seems speculative.

If you have a concrete scenario where local-only branch naming conventions matter (specific local tooling, etc.), happy to revisit - but without that, we'd prefer to keep this simple.

Thanks again for thinking about Gas Town!

PR #313, opened 2026-01-10
feat: add --start flag to gt sling for auto-starting idle polecat sessions

cstar:
## Summary

This PR adds `--start` flag support to `gt sling`, enabling automatic startup of idle (existing but not running) polecat sessions. This improves the polecat reuse workflow by allowing users to quickly restart existing workers without spawning new ones.

### Base `--start` flag
- `gt sling <rig> --start` reuses idle polecats instead of always spawning new ones
- Checks for uncommitted work before starting (unless `--force` is used)
- Falls back to spawning new polecat if no idle ones exist

### Selection Preferences (via `--start-pref`)
- `any` (default): First found idle polecat
- `newest`: Most recently created polecat
- `oldest`: Least recently created polecat  
- `cleanest`: Polecat with cleanest git state

### Targeted Restart
- `gt sling <rig> --start <name>` starts a specific idle polecat by name

## Implementation Details

- Added `IdlePolecatInfo` struct with metadata (name, path, session, created_at, clean)
- Added `StartPreference` type and `FindIdlePolecatOptions` for flexible selection
- Enhanced `FindIdlePolecat()` to support preference-based selection and specific-name targeting
- Session startup logic reuses existing `StartPolecatSession()` function

## Testing

All existing tests pass with the new functionality.

## Philosophically
I'm actually wondering if this is required in some form or if we (Claude and I) misunderstood something in Gastown. Claude tells me it will make its work easier, so here we are. What do you think?

### DIFF
--- a/internal/beads/beads_agent.go
+++ b/internal/beads/beads_agent.go
@@ -133,6 +133,7 @@ func (b *Beads) CreateAgentBead(id, title string, fields *AgentFields) (*Issue, 
 	description := FormatAgentDescription(title, fields)
 
 	args := []string{"create", "--json",
+		"--type=agent",
 		"--id=" + id,
 		"--title=" + title,
 		"--description=" + description,
--- a/internal/cmd/root.go
+++ b/internal/cmd/root.go
@@ -100,7 +100,7 @@ func warnIfTownRootOffMain() {
 // checkBeadsDependency verifies beads meets minimum version requirements.
 // Skips check for exempt commands (version, help, completion).
 // Deprecated: Use persistentPreRun instead, which calls CheckBeadsVersion.
-func checkBeadsDependency(cmd *cobra.Command, args []string) error {
+func checkBeadsDependency(cmd *cobra.Command, _ []string) error {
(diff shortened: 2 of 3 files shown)


steveyegge:
Thanks for the contribution! After discussion, we're going in a different direction here.

The core issue is that polecats shouldn't be sitting idle waiting to be reused - they should self-terminate once their work is complete. The existence of 'idle polecats' indicates a gap in our cleanup flow rather than a resource to exploit.

We're going to focus on:
1. Making polecats exit immediately when done (like other roles do with `gt handoff`)
2. Having polecats self-delete after completing their work unit (whether that's submitting an MR, getting it merged, or any defined completion point)
3. Reducing reliance on external watchers (Deacon/Witness) for cleanup

The `--start` flag solves a symptom rather than the root cause. Appreciate you surfacing this though - it highlighted that our polecat lifecycle needs tightening.

cstar:
Understood, and that makes it way more aligned with the blog post — my initial expectation, but not my experience.
Thanks for your feedback and your work !

PR #321, opened 2026-01-10
fix: prevent inherited BEADS_DIR from causing prefix mismatch

easel:
## Summary

- Fix `beads.run()` to always explicitly set `BEADS_DIR` based on the working directory or explicit override
- This prevents inherited environment variables (e.g., from mayor session with `BEADS_DIR=/home/erik/gt/.beads`) from causing prefix mismatch errors when creating agent beads for rigs

**Root cause**: When `BEADS_DIR` was set in the parent environment, all `bd` commands used the town database (`hq-` prefix) instead of the rig database (`gt-` prefix), causing "prefix mismatch: database uses 'hq' but you specified 'gt'" errors during polecat spawn.

## Changes

1. **internal/beads/beads.go**: Always resolve and set `BEADS_DIR` in `run()` method
2. **internal/polecat/manager.go**: Use `NewWithBeadsDir` for explicit beads directory
3. **internal/rig/manager_test.go**: Add `assertBeadsDirLog` helper and assertions to verify BEADS_DIR is correctly set
4. **internal/cmd/beads_routing_integration_test.go**: Convert log-only tests to asserting tests
5. **internal/doctor/orphan_check.go**: Add `SessionLister` interface for dependency injection
6. **internal/doctor/orphan_check_test.go**: Add deterministic orphan session check test

## Test plan

- [x] Verify `go build ./...` passes
- [x] Manual test: polecat spawn with inherited BEADS_DIR no longer fails
- [ ] Run `go test ./...` to verify all tests pass

### DIFF
--- a/internal/polecat/manager.go
+++ b/internal/polecat/manager.go
@@ -170,7 +170,10 @@ func (m *Manager) repoBase() (*git.Git, error) {
 	bareRepoPath := filepath.Join(m.rig.Path, ".repo.git")
 	if info, err := os.Stat(bareRepoPath); err == nil && info.IsDir() {
 		// Bare repo exists - use it
-		return git.NewGitWithDir(bareRepoPath, ""), nil
+		// Set workDir to rig root so git commands run in a neutral context,
+		// not from the current working directory which might be in a different rig's worktree.
+		// This prevents git from using the wrong repository when creating worktrees.
+		return git.NewGitWithDir(bareRepoPath, m.rig.Path), nil
 	}
 
 	// Fall back to mayor/rig (legacy architecture)


julianknutsen:
As an FYI, I'm working on a patch to completely remove BEADS_DIR from the system. It has been a source of pain for a lot of existing issues. Better to get it cleaned up once and for all.

steveyegge:
Merged manually via cherry-pick to resolve conflicts. Changes included in commit 598a39e7.

easel:
Sorry for the mess here, still figuring out how to play nice with gastown and polecats.

PR #366, opened 2026-01-12
fix(install): allow --wrappers in existing town without recreating HQ

easel:
## Summary

- Allow `gt install --wrappers` to work in an existing Gas Town installation
- Previously, `--wrappers` required recreating the entire HQ which was destructive
- Now wrapper installation can be done independently

## Test plan

- [x] Verify `go build ./...` passes
- [ ] Manual test: run `gt install --wrappers` in existing town


steveyegge (approved):
LGTM! Clean fix that:
- Allows --wrappers to work in existing town without recreating HQ
- Returns early after installing wrappers if that's all that's requested
- Good test coverage verifying HQ isn't modified

PR #386, opened 2026-01-12
fix: spawn polecats from rig's configured default_branch

carmandale:
## Problem

When `gt sling` spawns a new polecat, `AddWithOptions()` uses inline branch resolution. This PR adds a centralized `resolveDefaultBranch()` helper for consistency across all branch resolution points.

## Changes

1. **Add `resolveDefaultBranch()` helper** - Centralizes the pattern used in `AddWithOptions()`, `RepairWorktreeWithOptions()`, and `DetectStalePolecats()`

2. **Use helper in all three functions** - DRY principle, single source of truth for default branch resolution

3. **Fix test setup** - Added `git branch -m main` after `git init` in tests to ensure `origin/main` exists (git may default to `master` depending on configuration)

## Testing

```bash
go build ./...     # ✅ Passes
go test ./internal/polecat/...   # ✅ Passes (with test fix)
```

## Context

This is a refactoring to improve maintainability. The helper consolidates the default branch lookup pattern that was duplicated across multiple functions.

### DIFF
--- a/internal/polecat/manager.go
+++ b/internal/polecat/manager.go
@@ -180,6 +180,17 @@ func (m *Manager) repoBase() (*git.Git, error) {
 	return git.NewGit(mayorPath), nil
 }
 
+// resolveDefaultBranch returns the configured default branch for the rig.
+// Falls back to "main" if not configured. This centralizes branch resolution
+// to ensure consistent behavior across Add, Repair, and other operations.
+func (m *Manager) resolveDefaultBranch() string {
+	defaultBranch := "main"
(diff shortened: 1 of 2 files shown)


julianknutsen:
Closing — while the duplication is real (the default branch resolution pattern appears in AddWithOptions, RepairWorktreeWithOptions, and DetectStalePolecats), it's a 4-line pattern that's simple and stable. The maintenance risk of the duplication is low relative to the codebase's current priorities.

If the branch resolution logic becomes more complex in the future, centralizing it would make more sense at that point. Thanks for the clean refactoring proposal!

PR #401, opened 2026-01-12
fix(mail): use SendNotificationBanner for Claude Code notifications

sauerdaniel:
## Summary
Changed mail notification delivery to use `SendNotificationBanner` instead of `NudgeSession`. This displays notifications in the message history rather than injecting them into the input buffer.

## Test Plan
- [x] Build passes
- [ ] Mail notifications appear in conversation history, not input buffer


sauerdaniel:
Closing - this would revert #347 which was the correct fix. I mistakenly cherry-picked an older iteration.

PR #410, opened 2026-01-12
fix(startup): use WaitForRuntimeReady before sending propulsion nudges

julianknutsen:
## Summary

Fixes a timing issue where propulsion nudges were being sent before Claude's input prompt was ready to receive input. This caused agents (especially the deacon) to start but hang waiting for an Enter key, as the nudges were lost during Claude's initialization.

## Related Issue

Fixes agent startup reliability issue where deacon and other agents would not auto-start properly.

## Changes

- Replace `WaitForCommand` with `WaitForRuntimeReady` in agent startup flows
- Add `runtime.SleepForReadyDelay` and `runtime.RunStartupFallback` calls after prompt detection
- Add `runtime` package import to affected files
- Rename `config` variable to `roleConfig` in `daemon/lifecycle.go` to avoid shadowing the config package

### Root Cause

`WaitForCommand` only checks that the pane command is not a shell (bash/zsh/sh). It returns as soon as the `node` process starts, but Claude's TUI may not be fully initialized and ready to accept input yet.

`WaitForRuntimeReady` polls for Claude's `"> "` prompt prefix in the captured pane content, which is the reliable indicator that Claude is actually ready to receive input.

### Files Modified

| File | Change |
|------|--------|
| `internal/deacon/manager.go` | Replace `WaitForCommand` with `WaitForRuntimeReady` pattern |
| `internal/mayor/manager.go` | Replace `WaitForCommand` with `WaitForRuntimeReady` pattern |
| `internal/witness/manager.go` | Replace `WaitForCommand` with `WaitForRuntimeReady` pattern |
| `internal/daemon/lifecycle.go` | Replace `WaitForCommand` with `WaitForRuntimeReady` pattern in `restartSession()` |

### Reference Implementation

The `refinery/manager.go` already had the correct pattern and served as the reference:

```go
// Wait for Claude to start and show its prompt (non-fatal)
// WaitForRuntimeReady waits for the runtime to be ready
if err := t.WaitForRuntimeReady(sessionID, runtimeConfig, constants.ClaudeStartTimeout); err != nil {
    // Non-fatal - try to continue anyway
}

// Accept bypass permissions warning dialog if it appears.
_ = t.AcceptBypassPermissionsWarning(sessionID)

// Wait for runtime to be fully ready
runtime.SleepForReadyDelay(runtimeConfig)
_ = runtime.RunStartupFallback(t, sessionID, "refinery", runtimeConfig)
```

## Testing

- [x] Manual testing performed (build successful, all existing tests pass)
- [ ] Integration testing with actual tmux sessions (requires manual verification)

## Checklist

- [x] Documentation updated (if applicable) - N/A, internal fix

### DIFF
--- a/internal/daemon/lifecycle.go
+++ b/internal/daemon/lifecycle.go
@@ -13,6 +13,7 @@ import (
 	"github.com/steveyegge/gastown/internal/config"
 	"github.com/steveyegge/gastown/internal/constants"
 	"github.com/steveyegge/gastown/internal/rig"
+	"github.com/steveyegge/gastown/internal/runtime"
 	"github.com/steveyegge/gastown/internal/session"
 	"github.com/steveyegge/gastown/internal/tmux"
 )
@@ -335,7 +336,7 @@ func (d *Daemon) identityToSession(identity string) string {
 // Uses role bead config if available, falls back to hardcoded defaults.
 func (d *Daemon) restartSession(sessionName, identity string) error {
 	// Get role config for this identity
-	config, parsed, err := d.getRoleConfigForIdentity(identity)
+	roleConfig, parsed, err := d.getRoleConfigForIdentity(identity)
 	if err != nil {
 		return fmt.Errorf("parsing identity: %w", err)
 	}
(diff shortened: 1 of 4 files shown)


julianknutsen:
Closing in favor of #418

PR #368, opened 2026-01-12
Different roles to different models

abhijit360:
## Summary
This implementation allows assigning different LLM agents/models to different roles (Mayor, Deacon, Witness, Refinery, Polecat, Crew) for cost optimization.

## Related Issue
#335 

## Changes
  internal/config/types.go
  - Added RoleAgents map[string]string field to TownSettings
  - Added RoleAgents map[string]string field to RigSettings
  - Updated NewTownSettings() to initialize RoleAgents map

  internal/config/loader.go
  - Added ResolveRoleAgentConfig(role, townRoot, rigPath) - resolves agent config for a role
  - Added ResolveRoleAgentName(role, townRoot, rigPath) - returns agent name and whether role-specific
  - Resolution order: rig's RoleAgents → town's RoleAgents → default agent

  internal/daemon/lifecycle.go
  - Modified getStartCommand() to use ResolveRoleAgentConfig() for role-based agent selection

  internal/config/loader_test.go
  - Added TestResolveRoleAgentConfig
  - Added TestResolveRoleAgentName
  - Added TestRoleAgentsRoundTripzation

  ---

## Testing
- [x] Unit tests pass (`go test ./...`)
- [ ] Manual testing performed

## Checklist
- [ ] Code follows project style
- [ ] Documentation updated (if applicable)
- [ ] No breaking changes (or documented in summary)

PR #418, opened 2026-01-12
fix(agents): use CLI prompt for propulsion instead of unreliable nudges

sauerdaniel:
## Summary

Fixes agent startup reliability by passing `"gt prime"` as a CLI argument to Claude instead of using post-startup nudges. The CLI prompt is queued before Claude starts, avoiding timing issues entirely.

## Problem

Post-startup nudges are unreliable in Claude Code v2.1.4+ - text arrives before input is ready, causing only Enter to make it through (empty input). This causes agents to hang or enter restart cycles.

## Related

- Related to #410 which uses `WaitForRuntimeReady` + nudges (different approach)
- This PR removes nudges entirely, which is more robust

## Why CLI Prompt is Superior

| Approach | Timing Issue? | Complexity |
|----------|---------------|------------|
| **Post-startup nudges** (#410) | Yes - must wait for prompt | Polling + sleeps |
| **CLI prompt** (this PR) | No - queued before start | Simple |

The CLI argument is processed before Claude's TUI initializes, so there's no race condition.

## Changes

**Affected agents:** Deacon, Witness, Refinery, Polecat

For each:
1. Pass `"gt prime"` as CLI prompt to `BuildAgentStartupCommand*`
2. Remove post-startup nudges (StartupNudge, PropulsionNudge)
3. Remove timing sleeps
4. Keep SessionStart hook as backup

## Test Plan

- [x] `go build ./...` passes
- [x] `go test ./...` passes
- [x] Manual testing: agents start and execute gt prime immediately

### DIFF
--- a/internal/deacon/manager.go
+++ b/internal/deacon/manager.go
@@ -79,9 +79,10 @@ func (m *Manager) Start(agentOverride string) error {
 		return fmt.Errorf("ensuring Claude settings: %w", err)
 	}
 
-	// Build startup command first
+	// Build startup command with initial prompt for propulsion.
+	// The CLI prompt is more reliable than post-startup nudges (which arrive before input is ready).
 	// Restarts are handled by daemon via ensureDeaconRunning on each heartbeat
-	startupCmd, err := config.BuildAgentStartupCommandWithAgentOverride("deacon", "", m.townRoot, "", "", agentOverride)
+	startupCmd, err := config.BuildAgentStartupCommandWithAgentOverride("deacon", "", m.townRoot, "", "gt prime", agentOverride)
 	if err != nil {
 		return fmt.Errorf("building startup command: %w", err)
 	}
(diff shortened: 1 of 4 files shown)


mrm007:
How would this work when different agents are configured (#368)? How would this work once #424, #433 and #438 land?

sauerdaniel:
@mrm007 Great question! The CLI prompt approach is fully compatible with different agent configurations:

**#368 (different agents per role):** The prompt is agent-agnostic. `BuildCommandWithPrompt()` resolves the RuntimeConfig for the role via `ResolveAgentConfigForRole()` (which respects the `role_agents` map), builds the base command from the resolved config (could be `claude`, `gemini`, `codex`, etc.), then appends `"gt prime"` as a quoted shell argument. Each agent runtime can define its own `PromptMode` config (e.g., "none" to ignore prompts), so non-Claude runtimes that don't support CLI prompts can opt out.

**#424 (--agent flag for refinery):** PR #418 uses `BuildAgentStartupCommandWithAgentOverride()` which accepts agent overrides. The prompt is appended to whatever command resolves.

**#433 (daemon doesn't respect role_agents):** This is a bug in daemon's agent resolution, not in prompt delivery. Once fixed, the CLI prompt will work because it's appended *after* the agent command is resolved.

**#438 (missing --agent flag for refinery/boot):** PR #424 addresses refinery. Boot would need similar work. PR #418 is compatible because it passes the prompt to `BuildStartupCommand*` functions that already support agent overrides.

**Summary:** PR #418 is about *how* the prompt is delivered (CLI arg vs post-startup nudge), not *which* agent runs. Agent selection happens in the config resolution layer, and the prompt is just appended to whatever command that layer produces. Fixing #433 and adding --agent flags (#424/#438) are orthogonal to this PR.

mrm007:
Thanks ClauDan 👍

aleiby:
## Fix Applied: FormatStartupNudge beacon instead of bare "gt prime"

The original fix used bare `"gt prime"` strings as CLI prompts, but this confuses agents because they see a raw command instead of the structured `[GAS TOWN]` beacon format.

### The Problem
Agents receiving `gt prime` don't understand it as a beacon - they just see a command without context about who sent it or what topic to patrol.

### The Solution
Use `session.FormatStartupNudge()` which creates a proper beacon like:
```
[GAS TOWN] witness <- daemon • 2026-01-16T19:33 • patrol
```

This provides:
- **Sender/recipient context** for agent discovery and `/resume`
- **Structured topic** for propulsion (patrol, work)
- **Consistent format** with commit 86739556 (crew_at.go, handoff.go, start.go)

### Files Fixed
- `internal/deacon/manager.go`
- `internal/witness/manager.go`
- `internal/refinery/manager.go`
- `internal/polecat/session_manager.go`

### Testing
Verified on gastown rig - witness and refinery now start with proper beacon prompts and begin patrol without hanging.

PR #419, opened 2026-01-12
fix(sling): check GT_ROLE before GT_POLECAT to allow coordinators

### LINKED ISSUE #664: Sling incorrectly blocks coordinator roles with stale GT_POLECAT env
## Problem

When Mayor (or other coordinator) inherits a stale `GT_POLECAT` environment variable from a previous shell session, `gt sling` fails with: `"polecats cannot sling (use gt done for handoff)"`

Example scenario:
```
GT_POLECAT=furiosa  # stale from previous session
GT_ROLE=mayor       # current role
```

## Expected Behavior

Coordinator roles (mayor, deacon, witness, refinery) should be able to sling regardless of stale environment variables.

## Actual Behavior

Sling checks `GT_POLECAT` first and blocks the operation, even though `GT_ROLE=mayor` indicates this is a coordinator session.

## Proposed Solution

Check `GT_ROLE` first. If the role is a coordinator (mayor, deacon, witness, refinery), allow sling regardless of `GT_POLECAT` value.

sauerdaniel:
## Summary

Fixes sling incorrectly blocking coordinator roles when a stale `GT_POLECAT` environment variable is inherited.

## Problem

When Mayor (or other coordinator) inherits `GT_POLECAT` from a previous shell session:

```
GT_POLECAT=furiosa  # stale from previous session
GT_ROLE=mayor       # current role
```

Running `gt sling` fails with: `"polecats cannot sling (use gt done for handoff)"`

## Fix

Check `GT_ROLE` first. If the role is a coordinator (mayor, deacon, witness, refinery), allow sling regardless of `GT_POLECAT` value.

## Changes

- `internal/cmd/sling.go`: Add role check before polecat check

## Test Plan

- [x] `go build ./...` passes
- [x] Manual test: Mayor with stale GT_POLECAT can now sling

Fixes #664

### DIFF
--- a/internal/cmd/sling.go
+++ b/internal/cmd/sling.go
@@ -141,9 +141,15 @@ func init() {
 }
 
 func runSling(cmd *cobra.Command, args []string) error {
-	// Polecats cannot sling - check early before writing anything
-	if polecatName := os.Getenv("GT_POLECAT"); polecatName != "" {
-		return fmt.Errorf("polecats cannot sling (use gt done for handoff)")
+	// Polecats cannot sling - check early before writing anything.
+	// However, coordinator roles (mayor, deacon, witness, refinery) CAN sling,
+	// even if GT_POLECAT is set from a stale environment.
+	role := os.Getenv("GT_ROLE")
+	isCoordinator := role == "mayor" || role == "deacon" || role == "witness" || role == "refinery"
+	if !isCoordinator {
+		if polecatName := os.Getenv("GT_POLECAT"); polecatName != "" {
+			return fmt.Errorf("polecats cannot sling (use gt done for handoff)")
+		}
 	}
 
 	// Disable Dolt auto-commit for all bd commands run during sling (gt-u6n6a).


internal/cmd/sling.go
```diff
@@ -115,9 +115,15 @@ func init() {
 }
 
 func runSling(cmd *cobra.Command, args []string) error {
-	// Polecats cannot sling - check early before writing anything
-	if polecatName := os.Getenv("GT_POLECAT"); polecatName != "" {
-		return fmt.Errorf("polecats cannot sling (use gt done for handoff)")
+	// Polecats cannot sling - check early before writing anything.
```
julianknutsen, line 144: Why are polecats spawning coordinator agents?

sauerdaniel: The issue was that a polecat env variable blocked the mayor from slinging, this change ignores the variable for the mayor and the other non-polecats.

julianknutsen: I understand that, I'm just confused as to how a mayor gets a GT_POLECAT env variable set in the first place.

sauerdaniel:
Answer by Claude:

Julian, you raise a good point. After investigating:

**Root cause**: `tmux new-session` inherits all environment variables from the parent shell. When coordinator startup commands export `GT_ROLE=mayor BD_ACTOR=mayor ...`, they don't explicitly unset `GT_POLECAT`. If the shell that ran `gt mayor start` (or `gt up`) had `GT_POLECAT` exported from a previous debugging session, the tmux session inherits it.

**Reproduction**:
```bash
# In terminal, user was debugging polecat issues:
export GT_POLECAT=furiosa

# Later, same terminal:
gt mayor start  # tmux inherits GT_POLECAT=furiosa

# Mayor tries to sling and fails
```

**Two fix options**:
1. This PR (defensive): Check `GT_ROLE` first - if coordinator role, allow sling regardless of `GT_POLECAT`
2. Proper fix: Non-polecat startup commands should include `unset GT_POLECAT` to clear any inherited value

This PR is the safer immediate fix. We could also add a follow-up to unset stale env vars during coordinator startup.

julianknutsen:

Thanks for tracking down this stale-env bug — Mayor getting blocked by inherited `GT_POLECAT` is a real pain point, and the direction of checking `GT_ROLE` first is the right instinct. Both reviewers found issues worth addressing before merge, but the core idea is sound.

---

### Blocker: Canonical role addresses bypass the check

`GT_ROLE` is often set in canonical address form like `gastown/witness` or `gastown/refinery` (see `identity.go:133`, `handoff.go:369`). The PR's check uses exact string comparison against `"witness"`, `"refinery"`, etc., so these canonical forms won't match — leaving the stale `GT_POLECAT` bug unfixed for witness/refinery sessions running with address-form roles.

**Suggested fix:** Use the existing `parseRoleString()` from `role.go:213` to extract the base role before comparing, rather than raw string equality.

### Major: Over-broad coordinator allowlist

The issue (#664) only describes Mayor hitting this bug, but the fix allows deacon, witness, and refinery to bypass the guard too. Deacon coordinates cleanup, witness observes, refinery processes merges — it's unclear any of them legitimately call `gt sling`.

**Suggested fix:** Either narrow to `role == "mayor"` (or parsed mayor), or document which coordinator roles actually sling and why they need the bypass.

### Major: `GT_ROLE` has the same staleness problem

The PR fixes stale `GT_POLECAT` by trusting `GT_ROLE`, but `GT_ROLE` can also be stale. If a polecat inherits `GT_ROLE=mayor` from a parent shell, the sling guard is completely bypassed — a worse violation than the original bug.

**Suggested fix:** Add a code comment explaining why `GT_ROLE` is less susceptible to staleness (e.g., set per-session by spawn machinery), or validate against the current tmux session identity rather than trusting the env var.

### Major: No test coverage for new path

Existing sling tests set `GT_POLECAT` to empty (`sling_test.go:266,428,618,797`) but none exercise the new coordinator-override logic. The canonical-address regression above would be caught immediately with a test.

**Suggested fix:** Add table-driven tests covering: (1) `GT_ROLE=mayor` + stale `GT_POLECAT` → allowed, (2) `GT_ROLE=gastown/witness` + stale `GT_POLECAT` → allowed, (3) `GT_ROLE=polecat` + `GT_POLECAT=foo` → blocked.

### Minor: Silent override with no diagnostic trace

When `GT_POLECAT` is set but overridden by `GT_ROLE`, it's swallowed silently. Logging a warning would make stale-env issues visible during debugging.

**Suggested fix:** Add a debug log line when the override triggers.

---

### Pre-Merge Checklist
- [ ] Use `parseRoleString()` for canonical address handling
- [ ] Narrow allowlist to roles that actually sling (or document why all coordinators need it)
- [ ] Add tests for the new guard logic
- [ ] Consider logging when `GT_POLECAT` override triggers

---

julianknutsen:
Thanks for this contribution! I'm closing this PR as part of a backlog triage pass. Closing old PRs to keep our attention on the highest value items. If you'd like to revive this, please reopen it.

PR #421, opened 2026-01-13
fix(beads): handle .beads paths and absolute routes correctly

timyarm:
## Summary

Three related fixes for beads database routing that caused `gt hook show` to fail when querying hooked beads across rigs:

1. **ResolveBeadsDir**: Detect if `workDir` already ends with `.beads` to prevent double paths like `/path/.beads/.beads` when called with a beads directory directly

2. **findTownRoot**: Use `workspace.FindFromCwd()` instead of nonexistent `gt root` command (currently returns "unknown command")

3. **scanAllRigsForHookedBeads**: Check `filepath.IsAbs()` before joining paths since Go's `filepath.Join` doesn't handle absolute paths like Python - it concatenates instead of replacing

## Problem

When `routes.jsonl` contains absolute paths (e.g., `/home/user/gt/rig/mayor/rig/.beads`), the scan would produce invalid doubled paths like `/home/user/gt/home/user/gt/...` causing beads lookups to fail.

## Test plan

- [ ] `gt hook show <polecat>` correctly finds hooked beads
- [ ] `gt sling <bead> <rig>` updates bead status to hooked
- [ ] Works with both absolute and relative paths in routes.jsonl

### DIFF
--- a/internal/beads/beads_redirect.go
+++ b/internal/beads/beads_redirect.go
@@ -23,7 +23,14 @@ import (
 // this indicates an errant redirect file that should be removed. The function logs a
 // warning and returns the original beads directory.
 func ResolveBeadsDir(workDir string) string {
-	beadsDir := filepath.Join(workDir, ".beads")
+	// If workDir is already a .beads directory, use it directly
+	// (prevents double .beads paths like /path/.beads/.beads)
+	var beadsDir string
+	if strings.HasSuffix(workDir, ".beads") {
+		beadsDir = workDir
+	} else {
+		beadsDir = filepath.Join(workDir, ".beads")
+	}
 	redirectPath := filepath.Join(beadsDir, "redirect")
 
 	// Check for redirect file
--- a/internal/cmd/hook.go
+++ b/internal/cmd/hook.go
@@ -5,13 +5,13 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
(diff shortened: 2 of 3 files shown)


julianknutsen:
Hey @timyarm — thanks for these fixes! I've checked all three against the current codebase:

**1. `ResolveBeadsDir` double `.beads` path — Already fixed**
The current code uses `filepath.Base(workDir) == ".beads"` to strip the suffix before joining, which prevents the double-path issue. Different approach, same result.

**2. `findTownRoot` uses nonexistent `gt root` — Still broken!**
`hook.go:437` still shells out to `exec.Command("gt", "root")`, which returns "unknown command." Your fix to use `workspace.FindFromCwd()` is correct. This is called from `hook.go:381` during `gt hook show` when scanning town beads as a fallback for non-polecat agents.

**3. `scanAllRigsForHookedBeads` absolute path handling — Still broken!**
`molecule_status.go:1015` still does `filepath.Join(townRoot, route.Path)` without checking `filepath.IsAbs(route.Path)`. When `routes.jsonl` contains absolute paths, this produces invalid doubled paths. Your fix is correct.

Since the codebase has diverged (this PR was against an older `main`), we'll close this and apply fixes 2 and 3 in a fresh commit. The bugs you identified are real and we'll credit you in the fix. Thanks for catching these!

PR #425, opened 2026-01-13
Fix go-19z: updateAgentHookBead uses wrong directory for rig-level beads

Atmosphere-Aviation:
## Summary

This PR includes multiple fixes:

### 1. Fix updateAgentHookBead directory resolution
- Fixed `updateAgentHookBead()` to use correct working directory for rig-level agent beads
- Agent beads with prefixes like `go-` live in rig databases, not the town database
- Uses `beads.ResolveHookDir()` to resolve the correct path based on the agent bead's prefix

### 2. Fix orphaned Claude processes (Critical CPU bug)
- **Problem:** When `gt down` kills tmux sessions, Claude/node processes may not receive signals properly. They get reparented to launchd (PPID=1) and run forever as zombies, causing 500%+ CPU consumption.
- **Fix A (gt up):** Before starting services, detect and kill orphaned Claude/node processes (PPID=1)
- **Fix B (gt down):** Before killing tmux session, find Claude/node children via `pgrep -P` and kill them properly

Key commits:
- `bc6666c8` fix(up): add orphan Claude process cleanup before starting services
- `5d3c5eca` fix(down): kill orphan Claude/node processes before stopping tmux sessions
- `0782fcb5` fix(up): remove TTY filter from orphan detection

## Test Plan

**Hook bead fix:**
- [x] Build succeeds
- [x] Verified `bd slot show go-global_opt-crew-spencer` works from rig directory
- [x] Verified `bd slot set` succeeds from rig directory

**Orphan fix:**
- [x] Verified orphan detection works (PPID=1)
- [x] Verified cleanup runs on `gt up`
- [x] Verified prevention runs on `gt down`
- [x] Verified user's active Claude sessions are NOT killed (they have PPID ≠ 1)


julianknutsen:
Hey @Atmosphere-Aviation — thanks for these fixes! Both issues you identified were real and important. The good news is that both have since been addressed in the current codebase:

**1. `updateAgentHookBead` directory resolution — Merged**
`beads.ResolveHookDir()` exists at `internal/beads/routes.go:273` with full test coverage. It resolves the correct rig directory from the bead's prefix, which is exactly what your fix targeted.

**2. Orphaned Claude process cleanup — Merged (and expanded)**
The orphan handling is now comprehensive:
- **`gt up` / `gt start`**: Calls `cleanupOrphanedClaude()` (Phase 5 of shutdown, with Unix/Windows platform-specific implementations in `start_orphan_unix.go` / `start_orphan_windows.go`)
- **`gt down`**: Uses `KillSessionWithProcesses` to explicitly kill child processes before stopping tmux sessions (prevents orphans at the source), plus `findOrphanedClaudeProcesses` in post-shutdown verification, plus `KillTrackedPIDs` for PID-file-tracked orphans

Both of your contributions were clearly on the right track — the codebase landed equivalent (and in some cases expanded) fixes through other paths. Closing this since the work is covered. Thanks for the contribution!

PR #431, opened 2026-01-13
Fix race condition in agent startup by replacing sleep with polling

sigfawn:
## Summary
This PR replaces the fragile `time.Sleep(8 * time.Second)` in `ensureAgentReady` with a robust polling mechanism that waits for the agent's prompt to appear.

## Problem
The previous implementation used a fixed 8-second sleep to wait for Claude to initialize after bypassing permissions. This was unreliable:
- **Fast machines/agents**: Wasted up to 7 seconds waiting unnecessarily.
- **Slow machines**: Failed if startup took >8 seconds, causing subsequent commands to be sent to a non-ready shell (often resulting in lost input or errors).

## Fix
Introduced `waitForPrompt(session, timeout)` which:
1. Polls the tmux pane content every 500ms.
2. Checks for standard prompt indicators (`>`, `$`, `%`).
3. Returns immediately when ready.
4. Times out gracefully (logging a warning but proceeding, to maintain fallback behavior).

This significantly improves the reliability of `gt sling` and reduces latency for fast agents.

### DIFF
--- a/internal/cmd/sling_helpers.go
+++ b/internal/cmd/sling_helpers.go
@@ -241,14 +241,41 @@ func ensureAgentReady(sessionName string) error {
 	if t.IsClaudeRunning(sessionName) {
 		_ = t.AcceptBypassPermissionsWarning(sessionName)
 
-		// PRAGMATIC APPROACH: fixed delay rather than prompt detection.
-		// Claude startup takes ~5-8 seconds on typical machines.
-		time.Sleep(8 * time.Second)
+		// Wait for prompt (e.g. ">") to appear, indicating readiness.
+		// This replaces the fragile fixed 8s sleep with polling.
+		if err := waitForPrompt(t, sessionName, 15*time.Second); err != nil {
+			// If we timed out, just proceed - it might be ready but using a non-standard prompt.
+			// Falling back to "hope it works" is safer than hard failing here.
+			// In debug mode, this would be logged.
+		}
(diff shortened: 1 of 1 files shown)


julianknutsen:
Possibly related:

https://github.com/steveyegge/gastown/pull/410
https://github.com/steveyegge/gastown/pull/417

sigfawn:
417 is closed and didn't even fix the problem.

We actually need these 2 and close 410.

- #418 is best for agent self-startup - eliminates the problem entirely
- #431 is still needed for CLI commands - they use nudges, so timing must be handled

#410 becomes redundant if #418 is merged (both fix agent layer, but #418 is better).

julianknutsen:
Closed #410 after locally verifying both of these. Thanks!

julianknutsen:
Hey @sigfawn — thank you for this contribution! You correctly identified that the fixed sleep in `ensureAgentReady()` was a real problem: too slow on fast machines, too short on slow ones, and completely unaware of whether the agent was actually ready.

Your PR was the foundation for #1397, which cherry-picks your commit, rebases onto current main (this branch was ~90 commits behind), and extends the approach to handle edge cases we discovered during review — zero-delay presets, empty `GT_AGENT` normalization, and swarm worker readiness.

Closing this in favor of #1397. Your authorship is preserved in the cherry-picked commit. Thanks again for the fix! 🎉

PR #440, opened 2026-01-13
fix(hook): handle error from events.LogFeed

sigfawn:
## Summary

Previously the error from  was silently ignored with . Now we log the error to stderr at warning level but don't fail the operation since the primary hook action succeeded.

## Problem

The activity feed logging error was silently ignored with:

If the activity feed is down or misconfigured, this could result in lost visibility into hook operations without any indication to the user.

## Impact

- **Severity**: Medium - improves observability
- **Effort**: Minimal - 4 line change
- **Risk**: None - non-fatal error handling

## Note

This is part of a broader pattern - there are 20 files that silently ignore LogFeed errors. This fix addresses hook.go as a starting point. Future PRs can address the other files.

### DIFF
--- a/internal/cmd/hook.go
+++ b/internal/cmd/hook.go
@@ -230,8 +230,10 @@ func runHook(_ *cobra.Command, args []string) error {
 	fmt.Printf("  Use 'gt handoff' to restart with this work\n")
 	fmt.Printf("  Use 'gt hook' to see hook status\n")
 
-	// Log hook event to activity feed
-	_ = events.LogFeed(events.TypeHook, agentID, events.HookPayload(beadID))
+	// Log hook event to activity feed (non-fatal)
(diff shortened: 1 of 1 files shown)


steveyegge (approved):
LGTM. Four reasonable fixes bundled together:

1. **beads_version.go**: Good - cache version check with sync.Once, add 2s timeout, pre-compile regex. Prevents CLI lag.

2. **hook.go**: Good - log LogFeed errors to stderr instead of silent ignore. Improves observability.

3. **mail_queue.go**: Good defensive coding - nil check prevents potential panic.

4. **migrate_agents_test.go**: Correct - aligns test expectations with actual output.

Note for future: ideally these would be separate PRs for cleaner history, but all changes are sound.

PR #458, opened 2026-01-13
feat: Add worktree setup hooks for injecting local configurations

### LINKED ISSUE #220: Feature: worktree setup hook for injecting local configurations
## Problem Statement

When Gas Town creates polecat worktrees or crew clones, there's no way to inject local, gitignored configuration files. Projects that rely on per-user configs (which are intentionally not committed) end up with broken builds in new worktrees.

## Proposed Solution

Add a worktree_setup field to rig settings (settings/config.json):

```
  {
    "worktree_setup": "./scripts/setup-worktree.sh"
  }
```

  The script runs after worktree/clone creation, receiving the new path as an argument. It can copy, symlink, or generate whatever local configs are needed.

  Affected code paths:
  - internal/polecat/manager.go - AddWithOptions() ~line 238
  - internal/crew/manager.go - Add() ~line 114

## Alternatives Considered

Template directory: A per-rig .gt-worktree-template/ directory that gets auto-copied into new worktrees. Simpler but can't handle dynamic generation or conditional logic.

## Use Case

Our project uses Bazel with remote build caching. The cache credentials and settings live in user.bazelrc, which is gitignored. Without this file, builds fall back to slow local compilation. Currently there's no way to automatically provision this in new polecat worktrees.

sauerdaniel:
## Summary

When polecats are spawned, their worktrees are created fresh from the rig's repository. Currently, there's no way to inject custom configurations during this process, forcing users to manually configure each worktree or accept default settings.

This PR implements GitHub issue #220 by adding a hook system that runs custom scripts during worktree creation.

## Problem

- Polecats spawn with default git config (no signing keys, default user settings)
- Secrets/credentials must be manually copied to each worktree
- Project-specific setup (pre-commit hooks, local overrides) must be repeated
- No automation path for worktree initialization

## Solution

Add a setup-hooks directory at `<rig>/.runtime/setup-hooks/` where users can place executable scripts. These hooks:
- Execute in alphabetical order (e.g., `01-git-config.sh`, `02-copy-secrets.sh`)
- Run with the new worktree as working directory
- Receive `GT_WORKTREE_PATH` and `GT_RIG_PATH` environment variables
- Fail gracefully (warnings only, don't block polecat spawn)

### Example Use Case

```bash
# rig/.runtime/setup-hooks/01-git-config.sh
#!/bin/sh
git config --local user.signingkey ~/.ssh/key.asc
git config --local commit.gpgsign true
```

## Changes

- `internal/rig/setuphooks.go` - New file implementing hook discovery and execution
- `internal/polecat/manager.go` - Calls `RunSetupHooks()` after worktree creation

## Test Plan

- [x] Build passes
- [x] `go test ./internal/rig/... ./internal/polecat/...` passes

Closes #220

### DIFF
--- a/internal/cmd/polecat_identity.go
+++ b/internal/cmd/polecat_identity.go
@@ -407,11 +407,7 @@ func runPolecatIdentityShow(cmd *cobra.Command, args []string) error {
 	sessionRunning, _ := polecatMgr.IsRunning(polecatName)
 
 	// Build CV summary with enhanced analytics
-	cv, err := buildCVSummary(r.Path, rigName, polecatName, beadID, clonePath)
-	if err != nil {
-		// Continue without CV if there's an error
-		cv = &CVSummary{Identity: beadID}
-	}
+	cv := buildCVSummary(r.Path, rigName, polecatName, beadID, clonePath)
 
 	// JSON output - include both identity details and CV
 	if polecatIdentityShowJSON {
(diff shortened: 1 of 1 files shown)


sauerdaniel:

Fix: #466

Issue #345, opened 2026-01-11
`gt down --all` kills tmux server


johnbindel:
## Bug Description
When I run `gt down --all` it now kills the tmux server. I was running other things in tmux, so I expected this would not happen.

## Steps to Reproduce
1. Any gastown setup already present
2. `gt up`
3. `gt down --all`

## Expected Behavior
tmux server is still running, if previously running before `gt up`.

## Actual Behavior
tmux server was killed

## Environment
- **OS**: Ubuntu
- **Go version**: 1.24.4 linux/amd64
- **Gas Town version**: 0.2.4
- **tmux version** (if applicable): 3.5a

steveyegge:
**Investigation complete.**

**Root cause:** Since tmux 3.2, the `exit-empty` option defaults to `on`, causing the tmux server to exit when the last session ends. When `gt down --all` kills all GT sessions and no other non-GT sessions exist, the server exits.

**Decision:** This is expected and acceptable behavior. If the user runs `gt down --all` with only GT sessions, the server exiting is reasonable. `gt start` will create a new server when restarting.

The command progression already distinguishes this:
- `gt down` → Stop infrastructure
- `gt down --all` → Also stop bd daemons/activity
- `gt down --nuke` → Explicitly kill tmux server (DESTRUCTIVE)

No code changes needed.

PR #460, opened 2026-01-13
fix(shutdown): Improve gastown shutdown reliability

### LINKED ISSUE #291: gastown is very hard to kill / shutdown / stop - needs design decision
## Bug Description
On the latest `gastown` head, shutdown-related commands do not reliably terminate all processes.

**Main request: I think this needs a decision about the interface of shutdown/down/stop. I think they should all be aliases, and they should all shutdown everything unless specific args are given (like `gt shutdown polecats`). We should not need --nuclear because it should just work plainly. We should not need `--all` because that should be default. I don't want to submit a huge PR to change this without discussion around expected API and unification.**

I'm constantly having sessions come back to life after `gt shutdown --nuclear` and `gt down --all`. It seems like some daemons will get running and lost, then respawn the whole town. 

When this happens I mostly see the `gt feed` still running in the background, but also lots of `bd activity` and `bd daemon` processes.

Is there a clear shutdown system? I know it's not `stop` and it seems to not be `shutdown` or even `--nuclear` and `down` doesn't seem to do it reliably. 

## Steps to Reproduce
1. Start Gastown normally and use for a while (attach, detach, sling, etc)
2. Run `gt shutdown`
3. Observe that processes stop briefly but then respawn.
4. Run `gt shutdown --nuclear`
5. Observe tmux sessions being recreated and new claudes reconnecting.
6. Running `gt down --all` seems like it might work in some cases after shutdown has killed other things, but not reliably?

## Expected Behavior
- A single, clear, unified shutdown that works
- Pid tracking or some other ways to prevent respawning or losing processes

sauerdaniel:
## Summary

Fixes shutdown reliability by coordinating between `gt down` and the daemon.

## Problem

When running `gt down`, the daemon's heartbeat loop races against shutdown:
1. `gt down` kills an agent session
2. Daemon heartbeat detects "agent down" and restarts it
3. User runs `gt down` again, cycle repeats

Users reported needing to run shutdown multiple times, with sessions "coming back to life" after `gt shutdown --nuclear`.

## Solution

**Shutdown coordination**: `gt down` creates `shutdown.lock` before killing sessions. The daemon checks for this lock and skips auto-restart logic when shutdown is in progress.

**Orphan detection**: Added `findOrphanedClaudeProcesses()` to `verifyShutdown()` to catch Claude/node processes that survive tmux session termination.

## Changes

- `internal/daemon/daemon.go`: Add `isShutdownInProgress()` check at heartbeat start
- `internal/cmd/down.go`: Add orphaned process detection in shutdown verification

## Test Plan

- [x] Build passes
- [x] go test ./internal/daemon/... passes

Fixes #291

### DIFF
--- a/internal/cmd/down.go
+++ b/internal/cmd/down.go
@@ -393,8 +393,8 @@ func stopSession(t *tmux.Tmux, sessionName string) (bool, error) {
 		time.Sleep(100 * time.Millisecond)
 	}
 
-	// Kill the session
-	return true, t.KillSession(sessionName)
+	// Kill the session (with explicit process termination to prevent orphans)
+	return true, t.KillSessionWithProcesses(sessionName)
 }
 
 // acquireShutdownLock prevents concurrent shutdowns.


sauerdaniel:

Fix: #466

Avyukth:
have you  tried `GT_NUKE_ACKNOWLEDGED=1 gt down --nuke` ?

sauerdaniel:
@Avyukth Good suggestion! Yes, `GT_NUKE_ACKNOWLEDGED=1 gt down --nuke` works as a workaround for forced cleanup.

However, this PR addresses the root causes of why graceful shutdown fails in the first place:

1. **Daemon respawning during shutdown** - The daemon would restart processes while we're trying to shut them down, creating a race condition
2. **Missing shutdown coordination** - Components didn't signal each other properly during teardown
3. **#345 (tmux server killed)** - `gt down --all` was too aggressive

The goal is to make `gt down` and `gt down --all` work reliably without needing `--nuke`. The `--nuke` flag is a "break glass" option that should rarely be needed if graceful shutdown works correctly.

That said, I'll verify this PR doesn't break the `--nuke` path for cases where it's still needed.

Avyukth:
https://github.com/steveyegge/gastown/pull/330 check this one  maybe the docs  updated I have missed , but look into it

sauerdaniel:
@Avyukth Thanks for pointing to #330 - I reviewed it carefully. PR #460 actually **builds on top of** #330's work rather than duplicating it:

**What #330 added (already merged):**
- Phased shutdown with bd daemon/activity stopping
- The `shutdown.lock` mechanism to prevent concurrent shutdowns
- `--nuke` flag (replacing old `--all` behavior)
- Verification phase to detect respawns

**What #460 adds (this PR):**
1. **Daemon shutdown awareness** - The daemon now checks for `shutdown.lock` and skips heartbeat if shutdown is in progress. This prevents the race condition where daemon auto-restarts agents that `gt down` just killed. Without this, #330's lock only prevents concurrent `gt down` runs, but doesn't stop the daemon from fighting the shutdown.

2. **Orphaned process detection** - Adds `findOrphanedClaudeProcesses()` to `verifyShutdown()` to catch Claude/node processes that survive tmux session kills.

**The gap #460 fills:**
```
#330 creates: shutdown.lock file during gt down
#460 adds:    daemon checks for shutdown.lock before restarting agents
```

Without #460, even with #330 merged, the daemon can still respawn agents during the shutdown window because it doesn't know a shutdown is happening.

Avyukth:
ok , update corresponding docs also , skill (if has reference )

sauerdaniel:
@Avyukth Good question. I reviewed the docs - **no documentation updates needed** for this PR.

PR #460's changes are internal implementation fixes:
- Daemon shutdown awareness (`shutdown.lock` check) - users don't interact with this
- Orphaned process detection - automatic and transparent

The existing `gt down --help` already documents the behavior:
- `--all`: "Stop bd daemons/activity and verify shutdown"

PR #460 just makes this work more reliably. The user-facing behavior and expectations remain the same - we're fixing bugs, not adding features that need explaining.

If there were docs to update, it would have been in #330 (which added `--nuke` and the phased shutdown). That's already merged with its documentation.

sauerdaniel:
Thanks for the review! Sorry, #345 was added to this PR by mistake and has since been removed.

A quick clarification on the #345 point:

**#345 (tmux server killed by `gt down --all`)** is a separate issue with a different root cause: tmux has `exit-empty on` by default, meaning the server automatically exits when all sessions are closed. If gastown sessions were the user's only tmux sessions, killing them causes tmux to exit - this is tmux's default behavior, not something gastown explicitly does.

#345 can be addressed separately, likely by setting `exit-empty off` during `gt up` to prevent tmux from auto-exiting when gastown sessions are closed.

Regarding the other review points - those have been resolved during the rebase:
- ✅ Unused constants removed
- ✅ Duplicate polecat_identity.go changes dropped (already merged via #466)

sauerdaniel:
**Re: CI status** - The integration test failures are unrelated to this PR.

The failing tests (`TestBeadsDbInitAfterClone`, `TestHookSlot_*`, `TestBeadsRoutingFromTownRoot`, etc.) are beads-related integration tests that are also [failing on the main branch](https://github.com/steveyegge/gastown/actions/runs/21038933376).

This PR only modifies `down.go` and `daemon.go` for shutdown coordination - none of the beads integration test failures are related to these changes.

**Passing checks:**
- ✅ Build
- ✅ Lint  
- ✅ Unit Tests
- ✅ Check embedded formulas

The build/lint issues from the review have been fixed (removed duplicate `isProcessRunning` function, removed unused constants, dropped already-merged polecat_identity.go changes).

steveyegge (changes requested):
The daemon shutdown awareness (`isShutdownInProgress`) is a good fix for #291.

However, the change from `KillSessionWithProcesses` to `KillSession` in `stopSession()` is a regression:

**What's removed:**
- `KillSessionWithProcesses` recursively finds ALL descendant processes
- Sends SIGTERM then SIGKILL to ensure clean termination  
- This was specifically added to prevent orphan processes that ignore SIGHUP

**What's added instead:**
- `findOrphanedClaudeProcesses()` only **detects** orphans after shutdown
- It doesn't kill them - just reports them
- Only looks for `node` processes (misses other Claude-related processes)

**Suggested fix:**
1. Keep `KillSessionWithProcesses` in `stopSession()`
2. Keep the `isShutdownInProgress()` daemon awareness (this is good)
3. The orphan detection can stay as a helpful warning, but shouldn't replace proper process termination

The session cache removal is fine - simplifying that code is reasonable.

steveyegge (commented):
Reviewed this PR for potential merge. Here's my assessment:

**The Good:**
- `isShutdownInProgress()` in the daemon is a solid fix for #291. Checking for `shutdown.lock` before heartbeat prevents the race condition where daemon restarts agents that `gt down` just killed.
- Orphan detection in `verifyShutdown()` is helpful for visibility.

**Blocker (echoing steveyegge's second review):**

The change from `KillSessionWithProcesses` to `KillSession` in `stopSession()` is a regression:

```go
// Before (proper cleanup):
return true, t.KillSessionWithProcesses(sessionName)

// After (regression):
return true, t.KillSession(sessionName)
```

`KillSessionWithProcesses` recursively terminates all descendant processes with SIGTERM/SIGKILL. The new `findOrphanedClaudeProcesses()` only **reports** orphans after shutdown - it doesn't kill them. This will leave orphan processes.

**Suggested path forward:**
1. Keep `KillSessionWithProcesses` in `stopSession()`
2. Keep the `isShutdownInProgress()` daemon awareness (this is the valuable fix)
3. Orphan detection can stay as a helpful warning, but shouldn't replace proper process termination
4. Session cache removal is fine

Happy to re-review once the regression is addressed.

sauerdaniel:
@steveyegge Applied the requested fix:

**Change:** Reverted `stopSession()` to use `KillSessionWithProcesses` instead of `KillSession`.

```go
// Kill the session (with explicit process termination to prevent orphans)
return true, t.KillSessionWithProcesses(sessionName)
```

This ensures proper termination of all descendant processes with SIGTERM/SIGKILL, preventing orphaned Claude/node processes.

**Summary of what remains in the PR:**
- ✅ `isShutdownInProgress()` daemon awareness (the valuable fix for #291)
- ✅ `KillSessionWithProcesses` for proper process termination (restored)
- ✅ Orphan detection in `verifyShutdown()` (as a helpful warning)

Ready for re-review.

PR #462, opened 2026-01-13
fix(errors): use errors.As for wrapped error handling

sauerdaniel:
## Summary

`IsSilentExit()` uses a type assertion to detect `SilentExitError`, which fails when the error is wrapped (e.g., via `fmt.Errorf("%w", err)`).

## Problem

```go
// This works:
err := NewSilentExit(1)
IsSilentExit(err) // returns 1, true

// This fails (returns 0, false):
wrapped := fmt.Errorf("context: %w", NewSilentExit(1))
IsSilentExit(wrapped) // returns 0, false ← BUG
```

## Solution

Use `errors.As()` instead of type assertion to properly unwrap error chains.

## Changes

- `internal/cmd/errors.go` - Use `errors.As()` in `IsSilentExit()`

## Test Plan

- [x] Build passes

### DIFF
--- a/internal/cmd/errors.go
+++ b/internal/cmd/errors.go
@@ -1,6 +1,9 @@
 package cmd
 
-import "fmt"
+import (
+	"errors"
+	"fmt"
+)
 
 // SilentExitError signals that the command should exit with a specific code
 // without printing an error message. This is used for scripting purposes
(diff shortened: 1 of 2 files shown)


sauerdaniel:

Fix: #466

steveyegge (approved):
Code is correct - using `errors.As()` instead of type assertion properly handles wrapped errors.

The integration test failure (`TestBeadsDbInitAfterClone/TrackedRepoWithPrefixMismatchErrors`) is **unrelated** to this PR's changes - it's testing rig prefix handling, not error unwrapping. This appears to be a flaky test or affected by other recent merges.

Unit tests and lint pass. Approved.

PR #463, opened 2026-01-13
test: Add test coverage for 16 files (40.3% → 45.5%)

### LINKED ISSUE #237: Functional System Tests eventually

sauerdaniel:
## Summary

Gas Town's test coverage has gaps in several core packages. This PR adds comprehensive tests for 16 previously untested or under-tested files, improving overall coverage from 40.3% to 45.5%.

## Problem

- Multiple packages had 0% test coverage
- Low test coverage reduces confidence in changes and makes refactoring risky

## Solution

Add test files for:
- `internal/agent/state_test.go`
- `internal/cmd/errors_test.go`
- `internal/crew/types_test.go`
- `internal/doctor/errors_test.go`
- `internal/dog/types_test.go`
- `internal/mail/bd_test.go`
- `internal/opencode/plugin_test.go`
- `internal/rig/overlay_test.go`
- `internal/runtime/runtime_test.go`
- `internal/session/town_test.go`
- `internal/style/style_test.go`
- `internal/ui/markdown_test.go`
- `internal/ui/terminal_test.go`
- `internal/wisp/io_test.go`
- `internal/wisp/types_test.go`
- `internal/witness/types_test.go`

## Changes

- 16 new test files (+2659 lines)

## Test Plan

- [x] Build passes
- [x] All new tests pass
- [x] `go test ./...` passes (except pre-existing upstream failures)

Closes #237

### DIFF
--- a/internal/cmd/polecat_identity.go
+++ b/internal/cmd/polecat_identity.go
@@ -407,11 +407,7 @@ func runPolecatIdentityShow(cmd *cobra.Command, args []string) error {
 	sessionRunning, _ := polecatMgr.IsRunning(polecatName)
 
 	// Build CV summary with enhanced analytics
-	cv, err := buildCVSummary(r.Path, rigName, polecatName, beadID, clonePath)
-	if err != nil {
-		// Continue without CV if there's an error
-		cv = &CVSummary{Identity: beadID}
-	}
+	cv := buildCVSummary(r.Path, rigName, polecatName, beadID, clonePath)
 
 	// JSON output - include both identity details and CV
 	if polecatIdentityShowJSON {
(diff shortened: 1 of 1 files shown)


sauerdaniel:

Fix: #466

steveyegge (approved):
Reviewed and approved. Well-structured tests covering 16 files with comprehensive coverage of happy paths, error cases, JSON marshaling, and edge cases. CI passes.

PR #479, opened 2026-01-13
fix(security): prevent SQL injection in isTrackedByConvoy

KeithWyatt:
## Summary
- Add `escapeSQLString` helper function to prevent SQL injection in convoy tracking queries
- Apply escaping to beadID parameter used in SQL WHERE and LIKE clauses
- Add comprehensive test coverage for injection edge cases (16 test cases)

## Details
The `isTrackedByConvoy` function in `sling_convoy.go` accepts beadID from external input and uses it directly in a SQL query. Without proper escaping, a malicious beadID containing single quotes could break out of the string literal and execute arbitrary SQL.

The fix follows the same escaping pattern already used in:
- `convoy.go` (safeConvoyID, safeID)
- `daemon/convoy_watcher.go` (safeIssueID)
- `web/fetcher.go` (safeConvoyID)

## Test plan
- [x] All existing tests pass
- [x] New tests cover injection attempts (DROP TABLE, UNION SELECT, OR 1=1)
- [x] New tests verify query safety with escaped strings
- [x] Build passes with no warnings

### DIFF
--- a/internal/cmd/mail_queue.go
+++ b/internal/cmd/mail_queue.go
@@ -35,7 +35,7 @@ func runMailClaim(cmd *cobra.Command, args []string) error {
 	}
 
 	queueCfg, ok := cfg.Queues[queueName]
-	if !ok || queueCfg == nil {
+	if !ok {
 		return fmt.Errorf("unknown queue: %s", queueName)
 	}
 


KeithWyatt:
This PR is failing CI due to a build error in `internal/cmd/mail_queue.go:38` (invalid nil comparison for value type). PR #492 or PR #508 fixes this issue. Once that fix is merged into main, this PR should be rebased to pick up the fix.

julianknutsen:
@KeithWyatt — thanks for flagging the SQL injection risk in `isTrackedByConvoy`! You were right that the raw `fmt.Sprintf` into `sqlite3` was unsafe.

This was independently resolved by commit 30e65b5c ("refactor(convoy): replace SQLite queries with bd dep list commands"), which removed the raw SQL queries entirely in favor of `bd dep list` — eliminating the injection surface rather than escaping it.

Closing as superseded, but appreciate you catching this and putting together the fix with thorough test coverage.

PR #504, opened 2026-01-14
fix(cmd): cache bd version check to reduce subprocess contention

### LINKED ISSUE #503: bd version check causes contention under high concurrency
## Problem

Every `gt` command runs `bd version` to verify the minimum beads version requirement. Under high concurrency (17+ agent sessions), this check times out and blocks gt commands from running.

## Root Cause

While `bd version` is in the "noDbCommands" list and skips database initialization, it still runs expensive operations in `PersistentPreRun`:

1. **`ensureForkProtection()`** spawns multiple git subprocesses:
   - `git rev-parse --show-toplevel`
   - `git config --get beads.fork-protection`
   - `git remote get-url origin`
   - `git remote -v`
   - `git rev-parse --git-dir`

2. **`signalOrchestratorActivity()`** (deferred) calls `getActorWithGit()`:
   - `git config user.name`

3. **`resolveBranch()`** in version.go:
   - `git symbolic-ref --short HEAD`

So `bd version` spawns **5-7 git subprocesses** per invocation.

## Impact

With 17+ concurrent sessions each running gt commands:
- Each gt command spawns `bd version`
- Each `bd version` spawns 5-7 git processes
- This creates 85-120+ git processes competing for resources
- The 2-second timeout in gt is exceeded
- gt commands fail with "bd version check timed out"

## Reproduction

Run many concurrent gt commands (e.g., during `gt init` bringing up multiple rigs, or with many active agent sessions).

## Current Timeout Handling

The timeout is treated as a fatal error that blocks the command:

```go
// internal/cmd/beads_version.go
if ctx.Err() == context.DeadlineExceeded {
    return "", fmt.Errorf("bd version check timed out")  // blocks command
}
```

## Related

- Separate from `gt doctor` hang issue (#380 - git fetch without timeout)
- Part of broader pattern of subprocess contention under high concurrency

technicalpickles:
## Summary

- Cache bd version check based on binary path and mtime to avoid spawning subprocesses
- Under high concurrency (17+ sessions), eliminates 85-120+ competing git processes
- Cache stored at `~/.gt/beads-version-cache.json`, auto-invalidates on bd upgrade
- Also fixes pre-existing compile error in mail_queue.go

## Test plan

- [x] Unit tests for cache save/load/invalidation
- [x] Unit tests for version validation
- [x] Verified code compiles
- [ ] Manual test with high concurrency scenario

Fixes #503

### DIFF
--- a/internal/cmd/beads_version.go
+++ b/internal/cmd/beads_version.go
@@ -3,8 +3,11 @@ package cmd
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
+	"os"
 	"os/exec"
+	"path/filepath"
 	"regexp"
 	"strconv"
 	"strings"
@@ -90,6 +93,69 @@ func (v beadsVersion) compare(other beadsVersion) int {
 // Pre-compiled regex for beads version parsing
 var beadsVersionRe = regexp.MustCompile(`bd version (\d+\.\d+(?:\.\d+)?(?:-\w+)?)`)
 
+// beadsVersionCache stores cached version info keyed by binary path and mtime.
+// This avoids spawning bd subprocess on every gt command when the binary hasn't changed.
+type beadsVersionCache struct {
+	BdPath  string    `json:"bd_path"`
+	BdMtime time.Time `json:"bd_mtime"`
+	Version string    `json:"version"`
+}
+
+// getCacheFilePath returns the path to the version cache file.
+func getCacheFilePath() string {
+	home, err := os.UserHomeDir()
+	if err != nil {
+		return ""
+	}
+	return filepath.Join(home, ".gt", "beads-version-cache.json")
+}
+
+// loadVersionCache loads the cached version info from disk.
+func loadVersionCache() *beadsVersionCache {
+	path := getCacheFilePath()
+	if path == "" {
+		return nil
+	}
+
+	data, err := os.ReadFile(path)
+	if err != nil {
+		return nil
+	}
+
+	var cache beadsVersionCache
+	if err := json.Unmarshal(data, &cache); err != nil {
+		return nil
+	}
+
+	return &cache
+}
+
+// saveVersionCache saves the version info to disk.
+func saveVersionCache(cache *beadsVersionCache) {
+	path := getCacheFilePath()
+	if path == "" {
+		return
+	}
+
+	// Ensure directory exists
+	dir := filepath.Dir(path)
+	if err := os.MkdirAll(dir, 0755); err != nil {
+		return
+	}
+
+	data, err := json.Marshal(cache)
+	if err != nil {
+		return
+	}
+
+	// Write atomically by writing to temp file then renaming
+	tmpPath := path + ".tmp"
+	if err := os.WriteFile(tmpPath, data, 0644); err != nil {
+		return
+	}
+	_ = os.Rename(tmpPath, path)
+}
+
 func getBeadsVersion() (string, error) {
 	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
 	defer cancel()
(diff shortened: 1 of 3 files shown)


KeithWyatt:
CI needs to be triggered for this PR.

badri:
Tried this fix and found an issue with the time comparison. The cache wasn't being used because Equal() requires exact nanosecond match, which fails after JSON round-tripping.

**The problem:**
`  cache.BdMtime.Equal(bdMtime)  // returns false even when times are the same second
`
  When time.Time is marshaled to JSON and unmarshaled, nanosecond precision can differ, causing cache misses.

**The fix:**
`  cache.BdMtime.Unix() == bdMtime.Unix()  // compare to second precision
`
Second-precision is sufficient for detecting binary changes - if someone rebuilds bd within the same second, a cache miss is acceptable.

Debug output confirmed this was the issue:
`  DEBUG: cache.BdMtime.Unix()=1768468661, bdMtime.Unix()=1768468661, timeMatch=true
`
After this change, the cache works correctly and gt commands run instantly even under high concurrency.

julianknutsen:

Thanks for tackling the bd version check contention, @technicalpickles! This is a real problem — under high concurrency those 85-120+ competing subprocesses are a significant bottleneck, and a file-based mtime cache is a clean, well-targeted solution. The cache invalidation strategy (keyed on binary path + mtime) is solid, and the fail-open design is exactly right for a warn-only check.

---

### Major: Hand-rolled atomic write races under 17+ concurrent writers

The PR uses a fixed `.tmp` suffix (`path + ".tmp"`) for atomic writes, but under the exact scenario this PR targets (17+ concurrent sessions all cold-starting), every process races to write the same temp path. The codebase already provides `util.AtomicWriteFile` (at `internal/util/atomic.go:58-95`) which uses `os.CreateTemp` with random suffixes for proper concurrent write safety and handles crash cleanup via deferred remove. Several existing callsites use the `path + ".tmp"` pattern too, but the concurrency profile here is explicitly higher.

**Suggested fix:** Replace the hand-rolled atomic write block in `saveVersionCache` with `util.AtomicWriteFile` (or `util.EnsureDirAndWriteJSON` if available). This eliminates the write race, handles crash cleanup, and reduces ~15 lines to one call.

---

### Minor: Stale cache entry causes persistent validation failure without self-healing

If a cache entry matches on path+mtime but contains a corrupted or stale version string (e.g., written by a buggy previous version), `validateBeadsVersion` fails and emits a warning on every `gt` command. The cache-hit path returns directly without refreshing, so the stale entry persists indefinitely until the user upgrades `bd` (changing mtime) or manually deletes the cache file.

**Suggested fix:** On cache hit, if `validateBeadsVersion(cache.Version)` returns an error, treat it as a cache miss — call `getBeadsVersion()`, rewrite the cache, then validate the fresh value.

---

### Minor: No test coverage for cache-hit decision logic in checkBeadsVersionInternal

The new tests cover parsing, serialization, and unit-level cache save/load, but don't exercise the actual decision logic in `checkBeadsVersionInternal` — the path that checks cache validity, falls back to subprocess, and saves the result. The stale-cache bug above would ship untested.

**Suggested fix:** Add tests that exercise `checkBeadsVersionInternal` for: (1) valid cache hit skips subprocess, (2) invalid cached version triggers fallback + cache rewrite, (3) path/mtime mismatch triggers refresh.

---

### Minor: Orphaned .tmp files on process crash

If the process crashes between `os.WriteFile(tmpPath, ...)` and `os.Rename(tmpPath, path)`, the `.tmp` file is left behind indefinitely. This is subsumed by adopting `util.AtomicWriteFile` which handles cleanup.

**Suggested fix:** Resolved automatically if the atomic write fix above is adopted.

---

### Nit: Symlink path normalization for cache key

`exec.LookPath("bd")` may return a symlink path. If users have multiple symlinks or reorder PATH, the cache key changes unnecessarily, causing cache misses. Not a correctness bug — just cache inefficiency under an uncommon scenario.

**Suggested fix:** Consider `filepath.EvalSymlinks(bdPath)` before stat/caching to normalize. Low priority.

---

### Pre-Merge Checklist
- [ ] Replace hand-rolled atomic write with `util.AtomicWriteFile`
- [ ] Add self-healing fallback on cache-hit validation failure
- [ ] Add test coverage for `checkBeadsVersionInternal` cache-hit/miss paths
- [ ] Verify `mail_queue.go` compile fix matches current `main` (reviewers could not validate this hunk against upstream worktree state)

---

steveyegge:
Closing — the underlying issue (#503) has been fully resolved on main through three subsequent fixes:

1. cdcde82c — --no-daemon bypass (Jan 25, closed #503)
2. 637581bc — non-blocking version check (Feb 2)
3. 4cd8a776 — bd removed daemon mode entirely (Feb 10), eliminating the root cause

The mail_queue.go nil-check fix was also already applied separately (58820397).

Thanks for identifying the real problem — the concurrency contention was a genuine issue that prompted the fixes above. Appreciate the contribution!

PR #509, opened 2026-01-14
fix(config): add --hook flag to gt prime in SessionStart templates

aleiby:
## Summary
- Add `--hook` flag to `gt prime` commands in SessionStart hook templates
- The flag was added in v0.2.0 (commit e7994d98) but templates were never updated
- This broke `gt seance` which relies on session ID persistence for transcript lookup

## Related
- **See also #510** - Fixes seance to run `claude --resume` from the session's original cwd (both fixes needed for seance to work)

## Test plan
- [x] Verify new sessions persist session ID via `gt prime --hook`
- [x] Verify `gt seance` can find transcripts for recent sessions

### DIFF
--- a/internal/claude/config/settings-autonomous.json
+++ b/internal/claude/config/settings-autonomous.json
@@ -9,7 +9,7 @@
         "hooks": [
           {
             "type": "command",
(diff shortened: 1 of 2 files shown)


KeithWyatt:
CI needs to be triggered for this PR.

aleiby:
Already committed in 8c200d4a

PR #515, opened 2026-01-14
fix(mail): remove invalid nil check for value type QueueConfig

KeithWyatt:
## Summary
- Removes invalid nil check for `QueueConfig` (a value type struct, not a pointer)
- This was previously fixed in 58820397 but accidentally re-introduced in PR #445

The `!ok` check from the map lookup is sufficient to determine if the queue exists. Value types in Go can never be nil.

## Test plan
- [x] `go vet ./...` passes
- [x] `go build ./...` passes

### DIFF
--- a/internal/cmd/mail_queue.go
+++ b/internal/cmd/mail_queue.go
@@ -35,7 +35,7 @@ func runMailClaim(cmd *cobra.Command, args []string) error {
 	}
 
 	queueCfg, ok := cfg.Queues[queueName]
-	if !ok || queueCfg == nil {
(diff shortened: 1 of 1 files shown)


steveyegge:
This PR has a merge conflict. Could you rebase on main to resolve? The fix looks correct, just needs the conflict resolved.

steveyegge:
Closing as obsolete - the mail_queue.go file has been refactored and no longer uses the cfg.Queues map. The code this PR was fixing no longer exists. Thanks for the contribution!

PR #526, opened 2026-01-15
fix: witness patrol deacon check for town-level service

easel:
## Problem
Witness patrol pings deacon by filtering agent beads with "deacon" in description, which misses the town-level deacon (rig: null) and causes false alerts.

## Fix
- Match on "role_type: deacon" so the town-level service is detected
- Clarify alert subject/body to mention Town Deacon (hq-deacon)

## Testing
Not run (formula-only change).


easel:
Superseded by #527, which contains a single commit with just the mol-witness-patrol formula fix.

PR #527, opened 2026-01-15
fix: witness patrol deacon check for town-level service

easel:
## Problem
Witness patrol pings deacon by filtering agent beads with "deacon" in description, which misses the town-level deacon (rig: null) and causes false alerts.

## Fix
- Match on "role_type: deacon" so the town-level service is detected
- Clarify alert subject/body to mention Town Deacon (hq-deacon)
- Sync .beads formulas so embedded formulas stay in sync

## Testing
- go generate ./internal/formula/...
- go test ./...


easel:
Correctly closed: This PR was rebased to create a clean commit history and reopened as #561.

PR #528, opened 2026-01-15
fix(seance): run claude --resume from session's original cwd

aleiby:
## Summary
- Store the working directory when starting a seance session
- Use that stored cwd when running `claude --resume` to continue the session

This ensures the resumed session runs from the same directory as the original, rather than from wherever the seance command happens to be invoked.

## Test plan
- [x] Run `go test ./internal/cmd/... -run Seance` - all tests pass


aleiby:
Already merged in commit 11051b3d

PR #533, opened 2026-01-15
Readme beads installation (#1)

pRizz:
I ran into an issue installing gastown the first time; I got an error about beads not being installed; added that to the readme


pRizz:
I realize beads is in the prerequisites; moving too fast haha

PR #543, opened 2026-01-15
fix(beads): Fix test isolation and env override for TestCloseAndClearAgentBead

aleiby:
## Summary
- Fix `New(beadsDir)` → `New(tmpDir)` in 8 test locations (New() expects working directory, not .beads path)
- Filter out existing `BEADS_DIR` from environment before appending override (prevents duplicate env vars where first occurrence wins)
- Use alphabetic test suffixes instead of numeric to avoid bd prefix parsing issues

## Test plan
- [x] Run `go test ./internal/beads/...` - all tests pass
- [x] Verify BEADS_DIR override works correctly in subprocess calls


steveyegge:
Changes already in main via #557. Closing.

PR #561, opened 2026-01-15
fix(formulas): witness patrol deacon check for town-level service

easel:
## Summary
Fix witness patrol formula to correctly identify Deacon as a town-level service.

## Changes
- `internal/formula/formulas/mol-witness-patrol.formula.toml`:
  - Update `bd list` filter to check for `role_type: deacon` instead of just `deacon`
  - Update escalation message to reference "Town-level Deacon" and "Town Deacon (hq-deacon)"

## Context
The Deacon is a town-level service (not per-rig), so the Witness patrol should reference it appropriately when checking health and escalating issues.

## Test plan
- [ ] Verify formula syntax is valid
- [ ] CI tests pass


easel:
Thanks for the merge!

PR #501, opened 2026-01-14
fix(done): use KillSessionWithProcesses to prevent orphaned Claude processes

julianknutsen:
## Summary
Fix orphaned Claude processes after `gt done` by using proper process tree cleanup.

## Related Issue
Follow-up to f32a63e (feat: complete self-cleaning by killing tmux session) and 1043f00d (fix: kill entire process tree)

## Changes
- Replace `exec.Command("tmux", "kill-session")` with `tmux.KillSessionWithProcesses()`
- The original code only sent SIGHUP, which Claude ignores, causing orphaned processes
- `KillSessionWithProcesses` properly:
  1. Recursively finds all descendant processes via `pgrep -P`
  2. Sends SIGTERM (deepest first to avoid orphaning grandchildren)
  3. Waits 100ms for graceful shutdown  
  4. Sends SIGKILL to survivors
  5. Then kills the tmux session
- The original comment about "import cycles" was incorrect - many cmd files already import tmux

## Testing
- [x] Build passes (`go build -o gt ./cmd/gt`)
- [x] Unit tests pass for done command (`go test ./internal/cmd/... -run Done`)
- [ ] Manual testing: verify no orphaned claude processes after `gt done`

## Checklist
- [x] Code follows project style
- [ ] Documentation updated (if applicable) - no documentation changes needed
- [x] No breaking changes (or documented in summary)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #567, opened 2026-01-16
fix(tmux): kill pane process explicitly to prevent setsid orphans

### LINKED ISSUE #513: session_manager uses wrong kill method, orphaning Claude processes
## Problem

The polecat `Stop()` method in `session_manager.go` uses `KillSession()` instead of `KillSessionWithProcesses()`. This mismatch allows Claude (Node.js) child processes to become orphaned when polecats are stopped.

## Impact

- Orphaned Node.js processes survive session termination
- On WSL2, these orphaned processes cause `vmmemWSL` CPU spikes and system unresponsiveness
- Users must run `wsl --shutdown` to recover system stability
- The 100ms delay between SIGTERM and SIGKILL may be insufficient for WSL signal handling

This was often observed after running `gt down` (which stops polecats), though it can occur in any context where polecats are terminated.

## Fix

In `session_manager.go`, change `Stop()` to use `KillSessionWithProcesses()` instead of `KillSession()` to ensure child processes are properly terminated.

## Related

- #291 (difficulty killing gastown)
- #336 (shutdown model consolidation)

aleiby:
## Summary

- Fixes orphaned Claude processes that survive `gt done` by escaping via `setsid()`
- Adds explicit kill of the pane PID itself after killing descendants
- Completes the fix started in #501, which only killed descendant processes

## Root Cause

`KillSessionWithProcesses` was only killing descendant processes (found via `pgrep -P`), assuming the tmux session kill would terminate the pane process itself. However, Claude CLI spawns processes that call `setsid()`, which:
- Creates a new session ID and process group
- Detaches from the controlling terminal (TTY shows as `?`)
- Breaks the parent-child relationship that `pgrep -P` relies on

When the session was killed, these detached processes survived.

## Fix

After killing descendants, explicitly send SIGTERM/SIGKILL to the pane PID itself before killing the tmux session. This catches processes that have escaped the process tree.

## Testing

1. Spawned polecat, ran `gt done` - orphan process remained (before fix)
2. Applied fix, nuked polecat - no orphan processes (after fix)

Fixes #513

### DIFF
--- a/internal/tmux/tmux.go
+++ b/internal/tmux/tmux.go
@@ -176,6 +176,11 @@ func (t *Tmux) KillSessionWithProcesses(name string) error {
 		for _, dpid := range descendants {
 			_ = exec.Command("kill", "-KILL", dpid).Run()
 		}
+
+		// Kill the pane process itself (may have called setsid() and detached)
+		_ = exec.Command("kill", "-TERM", pid).Run()
+		time.Sleep(100 * time.Millisecond)
+		_ = exec.Command("kill", "-KILL", pid).Run()
 	}
 
 	// Kill the tmux session


julianknutsen:
> Breaks the parent-child relationship that pgrep -P relies on

Of course it does... thanks for finding and fixing this

boshu2:
## Analysis: Session Lifecycle PR Stack

This addresses the `setsid()` escape path where Claude subagents detach from the controlling terminal (TTY shows as `?`) and break the parent-child relationship that `pgrep -P` relies on.

### Technical Notes

- Minimal change: +5 lines
- Adds explicit SIGTERM/SIGKILL to pane PID after killing descendants
- Manual testing confirms orphan reproduction before fix, none after

### Questions for Review

1. What's the timeout between SIGTERM and SIGKILL for the pane process?
2. Is there any risk of PID reuse between getting the pane PID and sending the signal?

### Merge Order Context

This builds on #585 and should merge second:

```
#585 → #567 (this) → #586 → #590 → #588 → #562
```

groblegark:
Cherry-picked to groblegark/gastown fork (with minor fix for test compatibility). Thanks for the contribution!

steveyegge (approved):
LGTM. Simple 5-line fix for setsid orphan processes - SIGTERM, 100ms grace period, then SIGKILL on the pane PID.

PR #568, opened 2026-01-16
fix(sling): increase pane lookup retries to prevent session start race condition

groblegark:
## Summary

When `gt sling` spawns a polecat, the tmux session creation succeeds but `getSessionPane` could fail with 'exit status 1' due to a race condition where `tmux new-session` returns before the pane is fully queryable via `list-panes`.

**Changes:**
- Increase `maxRetries` from 10 to 30 (1s → 3s total wait time)
- Capture stderr from tmux command for better error diagnosis
- Add documentation explaining the race condition

This is especially important on slower systems or under load where tmux initialization may take longer than 1 second.

## Test plan

- [x] Build passes (`go build ./...`)
- [ ] Manual testing: spawn polecats repeatedly under load to verify no pane lookup failures
- [ ] Verify existing tests pass

Fixes: gt-07k

### DIFF
--- a/internal/cmd/handoff.go
+++ b/internal/cmd/handoff.go
@@ -6,6 +6,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/spf13/cobra"
 	"github.com/steveyegge/gastown/internal/config"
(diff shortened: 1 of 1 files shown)


groblegark:
Closing: These changes are already in main (commit ff6ef1dc). Thank you for the contribution!

Issue #280, opened 2026-01-08
When attaching to Mayor, the first command to the shell doesn't land


normtown:
## Bug Description
In a new installation, when I start and then run `gt mayor attach`, I get the following:

```bash
export BD_ACTOR=mayor GIT_AUTHOR_NAME=mayor GT_ROLE=mayor && claude --dangerously-skip-permissions
% [GAS TOWN] mayor <- human • 2026-01-08T11:22 • cold-s
tart
zsh: bad pattern: [GAS
% Run `gt prime` to check mail and begin coordination.
zsh: command not found: Run
```

Seems like the first line doesn't execute and then '[GAS TOWN] mayor <- human • 2026-01-08T11:22 • cold start' is passed to zsh, followed by 'Run \`gt prime\` to check mail and begin coordination.'

## Steps to Reproduce
1. `gt install ~/gt --git`
2. `cd ~/gt`
3. `gt start`
4. `gt mayor attach`

## Expected Behavior
A tmux session is opened and a Claude Code session is started with the Mayor agent.

## Actual Behavior
A tmux session is opened, the command to start the Claude Code session doesn't seem to execute and the first couple instructions to the Mayor seem to get fed to my shell. These fail because they are not valid shell commands.

## Environment
- **OS**: macOS Sequoia 15.7.3
- **Go version**: go version go1.25.5 darwin/arm64
- **Gas Town version**: gt version 0.2.1 (dev)
- **tmux version** (if applicable): tmux 3.6a

## Logs / Error Output
```
export BD_ACTOR=mayor GIT_AUTHOR_NAME=mayor GT_ROLE=mayor && claude --dangerously-skip-permissions
% [GAS TOWN] mayor <- human • 2026-01-08T11:22 • cold-s
tart
zsh: bad pattern: [GAS
% Run `gt prime` to check mail and begin coordination.
zsh: command not found: Run
```

normtown:
Bead filed by local mayor based on its diagnosis:

> Bug: Claude sessions fail to start due to tmux send-keys race condition
> Status: open
> Priority: P1
> Type: bug
> Created: 2026-01-08 22:04
> Created by: mayor
> Updated: 2026-01-08 22:04
> 
> Description:
> ## Summary
> 
> Agent sessions (polecats, witnesses, refineries, deacon) frequently end up at shell prompts instead of running Claude. This requires manual intervention to restart.
> 
> ## Root Cause
> 
> Race condition in the tmux send-keys startup pattern:
> 
[...]

PR #571, opened 2026-01-16
fix(tmux): serialize nudges to prevent interleaving

aleiby:
## Summary

Adds per-session mutex serialization to `NudgeSession()` and `NudgePane()` to prevent concurrent nudges from interleaving and corrupting tmux input.

## Problem

When multiple agents start simultaneously (e.g., `gt up`), each runs `gt nudge deacon session-started` in their SessionStart hook. These nudges arrive concurrently and interleave in the tmux input buffer, causing:

1. Text from one nudge mixing with another
2. Enter keys not properly submitting messages  
3. Garbled input requiring manual intervention to unstick
4. **Agent crashes or unexpected exits** - corrupted input can cause Claude to misinterpret commands or enter bad states

**Symptoms observed:**
- Deacon stuck waiting for input (nudges queued but not submitted)
- Deacon crashes or exits unexpectedly after concurrent nudges
- Polecats exiting early without completing work
- Manual Enter key required to unstick queued nudges

**Example corruption from [#280](https://github.com/steveyegge/gastown/issues/280):**
```
zsh: bad pattern: [GAS
zsh: command not found: Run
```
Where `[GAS TOWN] mayor <- human...` and `Run gt prime...` were fed to the shell instead of Claude due to interleaved input.

Example from logs - 4 nudges arriving within 8 seconds:
```
21:26:35 [nudge] deacon nudged with "[from deacon] session-started"
21:26:42 [nudge] deacon nudged with "[from beads/witness] session-started"
21:26:42 [nudge] deacon nudged with "[from da/witness] session-started"
21:26:43 [nudge] deacon nudged with "[from gastown/witness] session-started"
```

## Why Corruption Occurs

Claude Code is designed to allow typing messages while it's processing, which should queue and either steer Claude or add context. However, this functionality has known reliability issues:

- [Messages queued while working get "lost"](https://github.com/anthropics/claude-code/issues/12047) - "Claude keeps ignoring the messages I queue up while it's working"
- [Messages not processed between queries](https://github.com/anthropics/claude-code/issues/10892) - "Previous query still processing" error, typed message disappears
- [Queued prompts no longer steering](https://github.com/anthropics/claude-code/issues/11543) - Feature "no longer seems to be working"

We would like to take advantage of Claude's message queuing to handle concurrent nudges gracefully, but this functionality is currently unreliable. Until it stabilizes upstream, we need to serialize nudges ourselves to ensure robustness.

## Solution

Add a `sync.Map` of per-session mutexes. Before sending a nudge, acquire the lock for that session. This ensures concurrent nudges queue and execute one at a time.

## Alternatives Considered

| Alternative | Pros | Cons |
|-------------|------|------|
| **Delay nudges** - Add sleep in SessionStart | Simplest (one line) | Doesn't prevent concurrent nudges from multiple agents |
| **Debounce session-started** | Medium complexity | Only helps one nudge type |
| **File-based signaling** | Avoids tmux entirely | Major architectural change |
| **Rely on Claude queuing** | No code changes | Currently unreliable (see above) |

## Tradeoffs

- ✅ Prevents all nudge interleaving
- ⚠️ Adds latency when nudges queue (but prevents corruption)
- ⚠️ Memory: one mutex per session name (bounded, acceptable)
- ❌ Does not fix Claude Code's underlying queuing issues

## Future Work

- Revisit this fix if Claude Code's message queuing becomes reliable
- Consider deduplicating redundant session-started nudges (tracked in hq-u8l3)

## Testing

- [x] Build passes
- [x] All tmux package tests pass
- [x] Manual testing: started deacon + multiple witnesses concurrently, nudges processed correctly

### DIFF
--- a/internal/tmux/tmux.go
+++ b/internal/tmux/tmux.go
@@ -9,11 +9,17 @@ import (
 	"os/exec"
 	"regexp"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/steveyegge/gastown/internal/config"
 	"github.com/steveyegge/gastown/internal/constants"
 )
+
+// sessionNudgeLocks serializes nudges to the same session.
+// This prevents interleaving when multiple nudges arrive concurrently,
+// which can cause garbled input and missed Enter keys.
+var sessionNudgeLocks sync.Map // map[string]*sync.Mutex
 
 // versionPattern matches Claude Code version numbers like "2.0.76"
 var versionPattern = regexp.MustCompile(`^\d+\.\d+\.\d+$`)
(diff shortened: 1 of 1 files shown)


groblegark:
Cherry-picked to groblegark/gastown fork. Thanks for the contribution!

PR #593, opened 2026-01-16
fix(session): prevent orphaned Claude instances on session termination

jonabe:
## Summary
- Fixed orphaned Claude Code instances (2.1.9) accumulating when polecats complete work
- `selfKillSession()` now uses `KillSessionWithProcesses()` to recursively terminate all child processes
- This also cleans up the MCP servers spawned by Claude instances
- Added tests for the process termination logic

## Problem
When polecats completed work via `gt done`, the tmux session was killed but child processes were not terminated. Claude Code instances (2.1.9) remained running, along with their spawned MCP server processes. This caused processes to accumulate over time, with each worker leaving behind orphaned Claude instances and their MCP servers.

## Solution
Use `KillSessionWithProcesses()` everywhere sessions are terminated. This function:
1. Recursively finds all descendant PIDs (deepest-first order)
2. Sends SIGTERM, waits 100ms, then SIGKILL to remaining processes
3. Finally kills the tmux session

## Test plan
- [x] Verified baseline processes (10) → spawn polecat (11) → gt done (10) - processes return to baseline
- [x] Bulk tested with 5 polecats - all completed cleanly, no orphans
- [x] Added unit tests for `KillSessionWithProcesses` and nested child processes

### DIFF
--- a/internal/boot/boot.go
+++ b/internal/boot/boot.go
@@ -161,8 +161,9 @@ func (b *Boot) Spawn(agentOverride string) error {
 // spawnTmux spawns Boot in a tmux session.
 func (b *Boot) spawnTmux(agentOverride string) error {
 	// Kill any stale session first
+	// Use KillSessionWithProcesses to prevent orphaned MCP servers
 	if b.IsSessionAlive() {
-		_ = b.tmux.KillSession(SessionName)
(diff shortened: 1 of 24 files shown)


jonabe:
Closing - the fix doesn't fully address the orphan process issue. There are still 2.1.9 processes remaining after session termination. Will investigate further.

PR #597, opened 2026-01-16
Fix stop hook failing in role subdirectories

aarden-jasper:
The stop hook runs `gt costs record` from role subdirectories (e.g., `mayor/`) that lack a `.beads` database, causing bd to fail.

Fix: Use `workspace.FindFromCwd()` to locate town root and run bd from there.


groblegark:
Cherry-picked to groblegark/gastown fork. Thanks for the contribution!

PR #601, opened 2026-01-16
fix(install): set allowed_prefixes for convoy beads during gt install

aleiby:
## Summary
- Convoy beads use `hq-cv-*` IDs for visual distinction from other town beads
- `gt install` was adding `hq-cv-` to `routes.jsonl` but not setting `allowed_prefixes` config
- This caused `bd create --id=hq-cv-xxx` to fail prefix validation after fresh install

## Fix
Add `allowed_prefixes` config (hq,hq-cv) during `initTownBeads()` so convoy creation works out of the box.

## Test plan
- [x] Verified convoy creation works after applying config fix
- [x] Build passes


aleiby:
Related: #608 adds a `gt doctor` check for this config, catching cases where the database is recreated without running `gt install` again.

PR #608, opened 2026-01-17
fix(doctor): add allowed_prefixes check for convoy beads

aleiby:
## Summary
- Add `beads-allowed-prefixes` doctor check to verify and fix `allowed_prefixes` config
- Convoy beads use `hq-cv-*` IDs which require `allowed_prefixes` to include both `hq` and `hq-cv`
- Without this config, `bd create --id=hq-cv-xxx` fails prefix validation

## Related
- Companion to #601 which sets `allowed_prefixes` during `gt install`
- This check catches cases where the config is missing (e.g., database recreated without running `gt install`)

## The check
- Detects missing required prefixes and reports them
- Can be auto-fixed with `gt doctor --fix`
- Follows the same pattern as `CustomTypesCheck`

## Test plan
- [x] Verified check detects missing `hq-cv` prefix
- [x] Verified `--fix` restores the correct config
- [x] Build passes

### DIFF
--- a/internal/cmd/doctor.go
+++ b/internal/cmd/doctor.go
@@ -134,6 +134,7 @@ func runDoctor(cmd *cobra.Command, args []string) error {
 	d.Register(doctor.NewBootHealthCheck())
 	d.Register(doctor.NewBeadsDatabaseCheck())
 	d.Register(doctor.NewCustomTypesCheck())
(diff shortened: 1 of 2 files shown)


aleiby:
Related: #601 sets this config during `gt install`. This check is the safety net for when the database is recreated.

aleiby:
Already merged in commit 3669c04f

PR #625, opened 2026-01-17
feat (agent): introduce Goose agent

DavideRutigliano:
## Changes

- Add goose agent

## Testing

## Checklist

### DIFF
--- a/internal/cmd/install.go
+++ b/internal/cmd/install.go
@@ -11,9 +11,9 @@ import (
 
 	"github.com/spf13/cobra"
 	"github.com/steveyegge/gastown/internal/beads"
-	"github.com/steveyegge/gastown/internal/constants"
 	"github.com/steveyegge/gastown/internal/claude"
 	"github.com/steveyegge/gastown/internal/config"
+	"github.com/steveyegge/gastown/internal/constants"
 	"github.com/steveyegge/gastown/internal/deps"
 	"github.com/steveyegge/gastown/internal/formula"
 	"github.com/steveyegge/gastown/internal/shell"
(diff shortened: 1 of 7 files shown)


julianknutsen:
Hey @DavideRutigliano — thanks for the contribution and for thinking about Goose integration with Gas Town!

Since this PR was opened, the codebase has added a comprehensive custom agent registration system that makes hardcoded agent presets unnecessary for most use cases. Users can now configure Goose (or any agent) entirely through CLI commands and config files:

**1. Register Goose as a custom agent:**
```bash
gt config agent set goose "goose"
```

**2. Configure it as default or per-role:**
```bash
gt config default-agent goose
```

**3. For the `gt prime` wrapper behavior** (running `gt prime` before launch), create a simple wrapper script:
```bash
#!/bin/bash
# ~/bin/gt-goose
gt prime 2>/dev/null || true
exec goose "$@"
```

Then: `gt config agent set goose "~/bin/gt-goose"`

**4. Fine-grained config** (session resume, env vars, process names) can be set via `settings/config.json`:
```json
{
  "agents": {
    "goose": {
      "command": "goose",
      "args": [],
      "session": { "id_env": "GOOSE_SESSION_ID" },
      "tmux": { "process_names": ["goose"] }
    }
  }
}
```

The only thing that can't be configured this way is orphan process detection in `gt doctor`, which has a hardcoded process name list — but that's informational-only and doesn't affect normal operation. If Goose adoption grows, we can add it to the detection list in a targeted PR.

Given that the existing config system covers the full scope of what this PR does, we're going to close this in favor of the configuration-based approach. This keeps the codebase leaner and means any agent can be added without code changes.

Really appreciate you taking the time — and if you run into any friction getting Goose working via config, please open an issue and we'll make sure the config path is smooth!

PR #627, opened 2026-01-17
fix(web): buffer template output to prevent superfluous WriteHeader

GraemeF:
## Summary

- Buffer template output to `bytes.Buffer` before writing to ResponseWriter
- Ensures proper 500 status on template errors (instead of 200 with partial content)
- Add test verifying template errors return correct status code

## Problem

When `template.ExecuteTemplate` writes directly to `http.ResponseWriter` and fails mid-execution, the implicit `WriteHeader(200)` has already been sent. Calling `http.Error` then triggers:

```
http: superfluous response.WriteHeader call from github.com/steveyegge/gastown/internal/web.(*ConvoyHandler).ServeHTTP (handler.go:63)
```

## Solution

Buffer template output first, only writing to ResponseWriter on success. This is the standard Go pattern for this problem.

## Test plan

- [x] Added `TestConvoyHandler_TemplateErrorReturns500` that verifies template errors return 500
- [x] All existing tests pass: `go test ./internal/web/...`

### DIFF
--- a/internal/web/handler.go
+++ b/internal/web/handler.go
@@ -1,6 +1,7 @@
 package web
 
 import (
+	"bytes"
 	"html/template"
 	"net/http"
 )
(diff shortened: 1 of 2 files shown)


julianknutsen:
Thanks for the excellent contribution, @GraemeF! This is a real bug and your fix using `bytes.Buffer` is the correct standard Go pattern for it.

We've rebased this onto current main (which had some conflicts due to upstream security/logging improvements added to the same error path) and incorporated two small refinements from our automated review:
- Log `WriteTo` errors on the success path for observability
- Added a negative assertion in the test to verify no partial template content leaks on error

Your original authorship is preserved in the rebased PR: #1392

Thanks again — great catch!

PR #628, opened 2026-01-17
fix(costs): avoid daemon autostart in beads queries

dwsmith1983:
## Summary
- Add `--no-daemon` flag to `bd list/show` calls in `querySessionEventsFromLocation` to prevent daemon autostart issues during read-only queries
- Add `--sandbox` flag to `bd create/list` calls in the integration test to isolate from daemon interactions

## Test plan
- [x] `TestQuerySessionEvents_FindsEventsFromAllLocations` now passes
- [x] Lint passes (`go vet`)

Fixes stack overflow in `TestQuerySessionEvents_FindsEventsFromAllLocations` caused by infinite recursion in bd daemon's `acquireStartLock` function.


julianknutsen:
Thanks for this contribution! I'm closing this PR as part of a backlog triage pass. The --no-daemon flag is now deprecated as we are moving to the Dolt backend.

PR #676, opened 2026-01-17
feat(config): add agent-specific environment variables and mayor persistence

### LINKED ISSUE #692: Support per-agent environment variables in runtime config
## Problem

All agents share the same environment configuration. There's no way to:
- Set agent-specific API keys
- Configure per-agent feature flags
- Customize environment per role

Additionally, Mayor agent override (`--agent` flag) is not persisted across attach/respawn cycles.

## Proposed Solution

1. Add `Env` field to `RuntimeConfig` for per-agent environment variables
2. Export agent env vars in startup command builders
3. Persist Mayor agent override to survive attach/respawn

sauerdaniel:
## Summary

Adds support for per-agent environment variables in runtime config and fixes Mayor agent override persistence across attach/respawn cycles.

## Problem

1. **No per-agent env vars**: All agents share the same environment, making it impossible to configure agent-specific settings (API keys, feature flags, etc.)

2. **Mayor agent override not persisted**: Running `gt mayor start --agent claude-org` works initially, but `gt mayor attach` resets to the default agent because the override isn't saved.

3. **False respawn detection**: ExpectedPaneCommands returns wrong values for claude-* wrapper scripts, causing unnecessary respawns.

### Agent-specific environment variables
- Add `Env` field to `RuntimeConfig` for per-agent environment variables
- Update `fillRuntimeDefaults` to preserve Env field
- Export agent env vars in startup command builders

### Mayor agent persistence
- Add `PersistAgentOverride`/`LoadAgentOverride` to mayor manager
- Persist agent override on start, load on attach
- Override survives attach/respawn cycles

### ExpectedPaneCommands fix
- Return `["node", "claude"]` for claude-* wrapper scripts
- Fixes false "runtime exited" detection

## Changes

- `internal/config/runtime.go`: Add Env field and preserve in defaults
- `internal/config/startup.go`: Export env vars in startup commands
- `internal/mayor/manager.go`: Add override persistence

## Test Plan

- [x] Build passes
- [x] `gt mayor start --agent claude-org` persists across attach
- [x] Agent-specific env vars exported correctly

Fixes #692

### DIFF
--- a/internal/cmd/mayor.go
+++ b/internal/cmd/mayor.go
@@ -153,6 +153,11 @@ func runMayorAttach(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	// If no explicit --agent flag, use the persisted override from start
+	if mayorAgentOverride == "" {
+		mayorAgentOverride = mgr.LoadAgentOverride()
+	}
+
 	townRoot, err := workspace.FindFromCwdOrError()
 	if err != nil {
 		return fmt.Errorf("finding workspace: %w", err)
--- a/internal/config/loader.go
+++ b/internal/config/loader.go
@@ -1269,6 +1269,7 @@ func fillRuntimeDefaults(rc *RuntimeConfig) *RuntimeConfig {
 		Command:       rc.Command,
 		InitialPrompt: rc.InitialPrompt,
 		PromptMode:    rc.PromptMode,
+		Env:           rc.Env,
 	}
 
 	// Deep copy Args slice to avoid sharing backing array
@@ -1552,7 +1553,7 @@ func BuildStartupCommand(envVars map[string]string, rigPath, prompt string) stri
 	if rc.Session != nil && rc.Session.SessionIDEnv != "" {
 		resolvedEnv["GT_SESSION_ID_ENV"] = rc.Session.SessionIDEnv
 	}
-	// Merge agent-specific env vars (e.g., OPENCODE_PERMISSION for yolo mode)
+	// Add agent-specific env vars (e.g., ANTHROPIC_BASE_URL, API tokens, OPENCODE_PERMISSION)
 	for k, v := range rc.Env {
 		resolvedEnv[k] = v
 	}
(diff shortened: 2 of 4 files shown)


julianknutsen:
Hey @sauerdaniel — thanks for this one! All three problems you identified were real, and the per-agent env vars piece has already been merged into main (the `Env` field on `RuntimeConfig`, `fillRuntimeDefaults` deep-copy, and the `BuildStartupCommand` merge logic are all in place).

For the remaining two pieces, the codebase has evolved in ways that make the original approach worth reconsidering:

### Mayor agent override persistence

The `.agent-override` file approach works, but it would make Mayor the only role that persists agent overrides via a dot-file — no other role does this. Since the original PR, the codebase has settled on a consistent mechanism for agent overrides:

- **`GT_AGENT` env var** in the tmux session environment, set at startup by `BuildStartupCommandWithAgentOverride`
- **`RoleAgents` map** in `settings/config.json` (both town-level and rig-level), which already supports per-role agent configuration like `{"mayor": "claude-opus", "witness": "claude-haiku"}`
- **Handoff propagation** via `GT_AGENT` re-export in `handoff.go:484-486`

The `gt mayor attach` respawn case (where `GT_AGENT` is lost because the tmux session is gone) is real, but the fix should probably read from `RoleAgents` in settings rather than a separate dot-file. That way it works consistently with how `gt config settings set role_agents.mayor claude-opus` already configures agents, and there's no second source of truth to get out of sync.

### ExpectedPaneCommands for `claude-*` wrappers

This fix is still valid — `claude-org`, `claude-glm`, etc. would cause false "runtime exited" detection. However, the zombie detection path has since been refactored to use `IsAgentAlive` (which reads `GT_AGENT` from the tmux session env and resolves process names via `config.GetProcessNames`), so the impact is narrower than when this was originally written. The `ExpectedPaneCommands` path is still used by `gt start`, `gt crew at`, and other attach commands, so there's still value in the fix — but it could be a standalone one-liner PR.

### Mayor `Start()` changes

The `KillSession` → `KillSessionWithProcesses` and zombie handling changes in `manager.go` conflict with the unified `session.KillExistingSession` refactor that landed since this PR. Those changes are no longer applicable.

---

Closing this since 1/3 is already merged and the other 2/3 need a different approach given codebase evolution. If you're interested in tackling the `ExpectedPaneCommands` fix as a small standalone PR, that would be a clean contribution. And if the Mayor attach override loss is a pain point for you, we'd welcome a PR that reads from `RoleAgents` settings instead of a dot-file. Thanks again for the thorough problem identification!

Issue #663, opened 2026-01-17
BEADS_DIR override breaks prefix-based routing


sauerdaniel:
## Problem

`gt hook` returns "Nothing on hook" even when beads are properly hooked to agents.

The issue is in `beads.run()` which sets `BEADS_DIR` to the resolved beads directory. This bypasses prefix-based routing - beads with prefixes like `gt-` route to different databases, but the forced `BEADS_DIR` makes `bd` search only one database.

## Expected Behavior

When running `gt hook` from a polecat worktree, it should find hooked beads by allowing `bd`'s native prefix routing to work.

## Actual Behavior

`gt hook` returns "Nothing on hook" because all queries go to the wrong database.

## Proposed Solution

Only set `BEADS_DIR` when explicitly provided via `NewWithBeadsDir()`, allowing `bd`'s native prefix routing to work for all other cases.

boshu2:
FYI - I'm tracking this as part of a broader effort to consolidate duplicate issues.

**Related work on beads prefix routing:**
- Issues: #654, #591, #552, #473

See RFC #708 for discussion on preventing duplicates going forward.

Just helping connect the dots for anyone working on this area.

sauerdaniel:
This should be resolved by the same fix that addressed #654 — `beads.run()` now uses `ResolveBeadsDir()` instead of a hardcoded path, which preserves prefix routing. If the issue persists, please reopen.

PR #426, opened 2026-01-13
fix(beads): don't override BEADS_DIR to preserve prefix routing

sauerdaniel:
## Summary
Fix `gt hook` returning "Nothing on hook" even when beads are properly hooked to agents.

## Problem
`beads.run()` was setting `BEADS_DIR` to the resolved beads directory, which bypassed prefix-based routing. Beads with prefixes like `gt-` route to different databases, but the forced `BEADS_DIR` made `bd` search only one database.

## Fix
Only set `BEADS_DIR` when explicitly provided via `NewWithBeadsDir()`, allowing `bd`'s native prefix routing to work.

## Test plan
- [x] Build passes
- [x] Unit tests pass  
- [x] Manual test: `gt hook` from polecat worktree now finds hooked beads

Fixes #663

PR #685, opened 2026-01-17
fix(beads): preserve prefix routing by not overriding BEADS_DIR

### LINKED ISSUE #654: Beads client overrides BEADS_DIR, breaking prefix routing
## Problem

The beads client in `internal/beads/beads.go` overrides the `BEADS_DIR` environment variable when executing `bd` commands. This breaks prefix routing when operating from a context where `BEADS_DIR` is already correctly set.

## Steps to Reproduce

1. Set `BEADS_DIR` to a specific beads directory
2. Use beads client to perform operations
3. Operations go to wrong beads directory because client overrides the env var

## Root Cause

The client unconditionally sets `BEADS_DIR` in the command environment, ignoring any existing value that may have been set for routing purposes.

## Expected Behavior

The beads client should:
1. Respect existing `BEADS_DIR` if set
2. Only set `BEADS_DIR` if not already defined
3. Preserve prefix routing behavior

## Impact

- Bead operations go to wrong directory
- Cross-rig operations fail
- Prefix-based routing is broken

sauerdaniel:
## Summary

Fixes `gt hook` returning "Nothing on hook" even when beads are properly hooked to agents.

## Problem

`beads.run()` unconditionally sets `BEADS_DIR` to the resolved beads directory. This bypasses prefix-based routing:
- Beads with prefixes like `gt-` route to different databases
- Forcing `BEADS_DIR` makes `bd` search only one database
- Hooked beads in other databases become invisible

## Solution

Only set `BEADS_DIR` when explicitly provided via `NewWithBeadsDir()`. Otherwise, let `bd`'s native prefix routing work.

## Changes

- `internal/beads/client.go`: Conditional BEADS_DIR setting

## Test Plan

- [x] Build passes
- [x] Unit tests pass
- [x] `gt hook` from polecat worktree finds hooked beads
- [x] Prefix routing works correctly for gt- prefixed beads

Fixes #654, #663

### DIFF
--- a/internal/beads/beads.go
+++ b/internal/beads/beads.go
@@ -136,14 +136,13 @@ func (b *Beads) run(args ...string) ([]byte, error) {
 	cmd := exec.Command("bd", fullArgs...) //nolint:gosec // G204: bd is a trusted internal tool
 	cmd.Dir = b.workDir
 
-	// Always explicitly set BEADS_DIR to prevent inherited env vars from
-	// causing prefix mismatches. Use explicit beadsDir if set, otherwise
-	// resolve from working directory.
-	beadsDir := b.beadsDir
-	if beadsDir == "" {
-		beadsDir = ResolveBeadsDir(b.workDir)
+	// Only set BEADS_DIR if explicitly provided via NewWithBeadsDir().
+	// Otherwise, let bd use its native prefix-based routing to find the correct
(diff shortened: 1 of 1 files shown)


sauerdaniel:
Closing in favor of PR #426 which has the same fix and is older (preserving review history). 426 may need a rebase.

PR #706, opened 2026-01-17
fix(down): move bd daemon kill after tmux sessions to prevent respawn

dwsmith1983:
## Summary
Regression fix: bd daemons respawn during `gt down --all` because they are killed before tmux sessions that run Claude processes executing bd commands.

## Related Issue
Regression from #457

## Changes
- Move bd daemon killing from Phase 1 to Phase 5 (after all tmux sessions stopped)
- Rename phases for clarity (1a/1b for refineries/witnesses)
- Add comment explaining why bd daemons must be killed after tmux sessions

Before (broken):
```
gt down --all
✓ bd daemon: 4 stopped
...
✓ Daemon: stopped (was PID 40834)

⚠ Warning: Some processes may have respawned:
  • bd daemon (2 running)
```

## Testing
  - `gt up` followed by `gt down --all` completes without respawn warning
  - `pgrep -f "bd daemon"` returns nothing after shutdown

## Checklist

### DIFF
--- a/internal/formula/formulas/mol-deacon-patrol.formula.toml
+++ b/internal/formula/formulas/mol-deacon-patrol.formula.toml
@@ -665,71 +665,46 @@ Skip dispatch - system is healthy.
 
 [[steps]]
 id = "costs-digest"
-title = "Aggregate daily costs [DISABLED]"
+title = "Aggregate daily costs"
 needs = ["session-gc"]
 description = """
-**⚠️ DISABLED** - Skip this step entirely.
+**DAILY DIGEST** - Aggregate yesterday's session cost wisps.
 
-Cost tracking is temporarily disabled because Claude Code does not expose
-session costs in a way that can be captured programmatically.
-
-**Why disabled:**
-- The `gt costs` command uses tmux capture-pane to find costs
-- Claude Code displays costs in the TUI status bar, not in scrollback
-- All sessions show $0.00 because capture-pane can't see TUI chrome
-- The infrastructure is sound but has no data source
-
-**What we need from Claude Code:**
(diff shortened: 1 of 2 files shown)


julianknutsen:
Hi @dwsmith1983, thanks for taking the time to dig into this and put together a well-documented PR — the reproduction steps and testing checklist are really appreciated.

We took a fresh look at this against the current state of `main` and here's where things stand:

**down.go (core fix):** The `gt down` shutdown pipeline has been significantly reworked since this PR was opened. `StopAllBdProcesses` no longer exists in the phase structure — bd process management was removed/rearchitected. The respawn-during-shutdown issue you identified is no longer present in the current code, so this change is no longer applicable.

**mol-deacon-patrol.formula.toml:** The deacon patrol formula has gone through several iterations since January (version bumps, new steps like compact-report, WITNESS_PING removal, dependency chain changes). The costs-digest and patrol-digest changes here wouldn't apply cleanly, and the formula's structure has diverged substantially.

**namepool.go:** We looked closely at this one. The current code `themes[hash%uint32(len(themes))]` is actually correct — since `hash` is `uint32`, the modulo operation on two unsigned values is always non-negative and in bounds. The proposed change to `int(hash)%themeLen` would actually be slightly less safe on 32-bit platforms where `int(hash)` could overflow to negative, making `%` return a negative index. The `//nolint:gosec` annotation is there because the linter flags the `uint32(len(themes))` conversion, but it's fine since themes is a small constant array.

Given all three changes are either already addressed or not applicable, we're going to close this out. Thanks again for the contribution — the bd daemon respawn issue was a real problem at the time and your diagnosis was spot-on.

PR #707, opened 2026-01-17
fix(install): Makefile uses go install to match docs

easel:
Background here is that it's pretty easy to end up with multiple gt binaries, on in $GOPATH, and another in ~/.local/bin. Roles / Claude plugins don't always get the same path (!!), causing weird problems. 

## Summary
- Makefile install target now uses `go install` instead of copying to `~/.local/bin`
- Aligns with documented installation method in INSTALLING.md and README.md
- Follows GOPATH/GOBIN conventions

## Test plan
- [x] Verified `make install` uses `go install`
- [x] Confirmed docs consistently reference `go install` and `~/go/bin`
- [x] Ran `go vet ./...` - passes
- [ ] Pre-existing test failures in `internal/cmd` are unrelated (beads prefix config issue)

Closes: hq-93c


aleiby:
I have definitely hit this and made a similar change locally.

easel:
Superseded by #851 (branch-based update). Fork main was reset to upstream/main; new PR uses easel:pr-707.

PR #709, opened 2026-01-17
feat(config): add role context overrides via settings config

dwsmith1983:
## Summary
- Add ability to customize CLAUDE.md content for each role via `settings/config.json`
- Supports both town-level and rig-level overrides
- Resolution order: Rig settings → Town settings → Default embedded template

## Test plan
- [x] Unit tests for `ResolveRoleContext()` precedence
- [x] Unit tests for `GetRoleContent()` with nil, town, rig, and combined settings
- [x] Build passes
- [x] Lint passes

## Changes
- `internal/config/types.go`: Add `ContextConfig` struct with `Roles` map, add `Context` field to `TownSettings` and `RigSettings`
- `internal/config/loader.go`: Add `ResolveRoleContext()` function for config resolution
- `internal/templates/templates.go`: Add `GetRoleContent()` for config-aware template rendering
- `internal/templates/templates_test.go`: Add comprehensive tests for override behavior
- `docs/reference.md`: Document the feature with examples

## Example Usage

**Town-level** (`settings/config.json`):
```json
{
  "type": "town-settings",
  "version": 1,
  "context": {
    "roles": {
      "polecat": "# Custom Polecat Context\n\nFollow these guidelines..."
    }
  }
}
```

**Rig-level** (`<rig>/settings/config.json`):
```json
{
  "type": "rig-settings", 
  "version": 1,
  "context": {
    "roles": {
      "witness": "# Custom Witness for this rig..."
    }
  }
}
```

### DIFF
--- a/internal/polecat/session_manager.go
+++ b/internal/polecat/session_manager.go
@@ -16,6 +16,7 @@ import (
 	"github.com/steveyegge/gastown/internal/rig"
 	"github.com/steveyegge/gastown/internal/runtime"
 	"github.com/steveyegge/gastown/internal/session"
+	"github.com/steveyegge/gastown/internal/templates"
 	"github.com/steveyegge/gastown/internal/tmux"
 )
 
@@ -178,6 +179,12 @@ func (m *SessionManager) Start(polecat string, opts SessionStartOptions) error {
 	polecatsDir := filepath.Join(m.rig.Path, "polecats")
 	if err := runtime.EnsureSettingsForRole(polecatsDir, "polecat", runtimeConfig); err != nil {
 		return fmt.Errorf("ensuring runtime settings: %w", err)
+	}
+
+	// Regenerate CLAUDE.md with config overrides (supports town/rig context customization).
+	// This writes to polecats/.claude/CLAUDE.md, shared by all polecats.
+	if err := templates.CreatePolecatCLAUDEmd(polecatsDir, m.rig.Path, m.rig.Name); err != nil {
+		return fmt.Errorf("creating polecat CLAUDE.md: %w", err)
 	}
 
 	// Build startup command first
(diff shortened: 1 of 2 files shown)


julianknutsen:
Hey @dwsmith1983 — thanks for putting this together! The problem you're solving here is real and important: town operators absolutely should be able to customize role behavior and selectively disable roles without forking the codebase. Appreciate you thinking through the config resolution chain (rig → town → default) and the role-check formula steps.

After reviewing this against the current architecture, we think the goals are better served by mechanisms that already exist (or can be extended with less risk):

### For customizing role context

Most roles now receive their context dynamically via `gt prime` (a SessionStart hook) rather than through persistent CLAUDE.md files. This matters because the embedded templates contain safety boundaries (role constraints, ZFC compliance, protocol definitions) that would be lost with a full-replacement approach.

The existing `CONTEXT.md` mechanism already injects town-level custom instructions into every role's context at startup — additively, so the safety boundaries are preserved. For per-role customization, we can extend this to support role-specific sections or per-role context files (e.g., `context/witness.md`).

### For disabling roles

The `beads.RoleConfig` structure already provides per-role configuration that's loaded during startup. Adding a `disabled` field there and checking it before spawning the tmux session is cleaner than a formula step that reads file content — it's checked earlier in the lifecycle, fails explicitly, and doesn't couple role availability to file system state.

### Why not persistent CLAUDE.md for all roles

Polecats and crew work in git worktrees of real project repos. Writing a CLAUDE.md there would leak into `git status` or get accidentally committed. The current architecture intentionally avoids this by using dynamic context injection instead.

---

We're going to close this PR, but the use cases you've identified are on our radar. If you're interested in contributing toward the `CONTEXT.md` per-role extension or the `RoleConfig.Disabled` field, we'd love to collaborate on that. Feel free to open an issue to discuss the approach!

PR #711, opened 2026-01-17
feat(config): add refinery disable flag per-rig

dwsmith1983:
## Summary
- Add ability to disable the refinery per-rig via `settings/config.json`
- When disabled, `gt up` shows "disabled" instead of starting the refinery session
- Useful for rigs that don't use the merge queue workflow

## Motivation
Refineries would always come back up when killed, even when not needed for a rig's workflow. This is particularly useful for rigs that use alternative merge workflows or don't need the merge queue.

Inspired by: https://steve-yegge.medium.com/gas-town-emergency-user-manual-cf0e4556d74b

## Test plan
- [x] Unit tests pass
- [x] Build passes
- [x] Lint passes

## Changes
- `internal/config/types.go`: Add `RefineryConfig` struct with `Enabled *bool` field, add to `RigSettings`
- `internal/cmd/up.go`: Check refinery disabled flag before starting
- `docs/reference.md`: Document the feature

## Example Usage

```json
{
  "type": "rig-settings",
  "version": 1,
  "refinery": {
    "enabled": false
  }
}
```

## Future
This pattern could be extended to witness if needed.

### DIFF
--- a/.beads/formulas/mol-refinery-patrol.formula.toml
+++ b/.beads/formulas/mol-refinery-patrol.formula.toml
@@ -33,11 +33,40 @@ Witness                    Refinery                      Git
 After successful merge, Refinery sends MERGED mail back to Witness so it can
 complete cleanup (nuke the polecat worktree)."""
 formula = "mol-refinery-patrol"
-version = 4
+version = 5
+
+[[steps]]
+id = "role-check"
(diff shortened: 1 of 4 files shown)


julianknutsen:
Closing as author is unable to continue contributing.

Issue #702, opened 2026-01-17
CI failing: gosec lint warning and embedded formulas out of sync


sauerdaniel:
## Problem

CI is failing on main branch and all PRs due to two issues:

### 1. Lint failure: gosec G115

```
internal/polecat/namepool.go:381:27: G115: integer overflow conversion int -> uint32 (gosec)
    return themes[hash%uint32(len(themes))]
```

### 2. Embedded formulas out of sync

```
ERROR: Committed formulas are out of sync with .beads/formulas/
Run: go generate ./... && git add -A && git commit
```

The `internal/formula/formulas/` directory is out of sync with `.beads/formulas/`.

## Impact

- All PRs fail CI checks
- Blocks merging of any contributions

## Solution

1. Fix gosec warning with safe conversion pattern
2. Run `go generate ./internal/formula/...` to sync formulas

julianknutsen:
main is green right now

Issue #700, opened 2026-01-17
Orphaned tmux sessions accumulate over time


sauerdaniel:
## Problem

Over time, Gas Town can accumulate "zombie" tmux sessions - sessions where tmux is still alive but the Claude process has died. These orphaned sessions:
- Consume system resources
- Can cause session name conflicts on restart
- Lead to process accumulation that overwhelms the system

## Current Behavior

Orphaned sessions persist indefinitely until manual cleanup.

## Proposed Solution

Add `CleanupOrphanedSessions()` function that runs at `gt start` time:
1. Scan for `gt-*` and `hq-*` prefixed tmux sessions
2. Detect zombie sessions (tmux alive, Claude dead)
3. Kill zombie sessions before starting new agents

PR #739, opened 2026-01-18
fix(refinery): reconcile status with actual tmux session state

joshuavial:
## Summary
- Fix `gt refinery status` reporting stale "running" state when tmux session is dead
- Add unit test for stale state reconciliation  
- Fix gosec G115 lint warning in polecat theme selection

## Problem
`gt refinery status` read from a stale state file and reported "running" even when the tmux session had crashed. Meanwhile `gt rig status` correctly showed "stopped" because it checks the actual tmux session.

## Solution
`Status()` now verifies the tmux session exists before returning "running" state. If the session is gone, it returns "stopped" regardless of what the state file claims.

## Test plan
- [x] `go test ./...` - all pass
- [x] `golangci-lint run ./...` - 0 issues
- [x] Manual: `gt refinery status` now matches `gt rig status`

Partially addresses #702 (fixes gosec G115 warning)
Related to #700 (improves zombie session detection)

### DIFF
--- a/.beads/formulas/mol-deacon-patrol.formula.toml
+++ b/.beads/formulas/mol-deacon-patrol.formula.toml
@@ -665,46 +665,71 @@ Skip dispatch - system is healthy.
 
 [[steps]]
 id = "costs-digest"
-title = "Aggregate daily costs"
+title = "Aggregate daily costs [DISABLED]"
 needs = ["session-gc"]
 description = """
-**DAILY DIGEST** - Aggregate yesterday's session cost wisps.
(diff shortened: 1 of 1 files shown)


joshuavial:
Closing - this PR is obsolete. The codebase moved to ZFC (Zero File Cache) architecture where tmux session existence is the source of truth, eliminating state files entirely. The stale state problem this PR solved no longer exists. The lint fix in commit 3 was also addressed differently upstream with a nolint directive.

PR #746, opened 2026-01-18
Fix critical bug: kill main pane PID to prevent orphaned processes

roland-at-ailtir:
## Problem
`KillSessionWithProcesses()` killed descendant processes but NOT the main pane PID itself. This caused processes to become orphaned (PPID=1) during shutdown operations.

### Evidence
After running `gt shutdown`, 7 orphaned claude processes remained:
```bash
ps -C claude -o pid,ppid,etime,cmd --no-headers
3440879       1       13:33 claude  # PPID=1 (orphaned)
3440891       1       13:33 claude
3453267       1       10:49 claude
# ... 4 more ...
```

## Root Cause
`getAllDescendants(pid)` at tmux.go:185-205 returned only children/grandchildren, not the root PID itself. When sessions were killed:
1. Descendant processes (node) were terminated ✓
2. Main pane process (claude/bash) was NOT killed ✗
3. Main process became orphaned with PPID=1

## Solution
Modified `getAllDescendants()` to include the root PID in the returned list. This ensures `KillSessionWithProcesses()` terminates the main process along with all descendants.

**Change**: Added root PID to end of returned slice (deepest-first order preserved)

## Impact
- **Severity**: HIGH - affects all shutdown operations
- **Affected commands**: `gt down`, `gt shutdown`, `gt mol step done`, agent restarts
- **Fix scope**: Single fix point in `getAllDescendants()` automatically fixes `KillSessionWithProcesses()`

## Testing
- ✅ Verified `gt shutdown` cleanly terminates all processes
- ✅ No orphaned processes remain (verified with `ps -C claude`)
- ✅ Added test case `TestGetAllDescendantsIncludesRoot`
- ✅ Updated existing test `TestGetAllDescendants` to match new behavior
- ✅ Process count remains stable after multiple shutdown/startup cycles

## Files Changed
- `internal/tmux/tmux.go`: Fixed `getAllDescendants()` to include root PID, updated comments
- `internal/tmux/tmux_test.go`: Added test for root PID inclusion, updated existing test

### DIFF
--- a/internal/tmux/tmux.go
+++ b/internal/tmux/tmux.go
@@ -145,13 +145,13 @@ func (t *Tmux) KillSession(name string) error {
 //
 // Process:
 // 1. Get the pane's main process PID
-// 2. Find all descendant processes recursively (not just direct children)
-// 3. Send SIGTERM to all descendants (deepest first)
+// 2. Find all descendant processes recursively (including the main PID)
+// 3. Send SIGTERM to all descendants and the main PID (deepest first)
 // 4. Wait 100ms for graceful shutdown
-// 5. Send SIGKILL to any remaining descendants
+// 5. Send SIGKILL to any remaining descendants and the main PID
 // 6. Kill the tmux session
 //
-// This ensures Claude processes and all their children are properly terminated.
+// This ensures Claude processes, all their children, and the main pane process are properly terminated.
 func (t *Tmux) KillSessionWithProcesses(name string) error {
 	// Get the pane PID
 	pid, err := t.GetPanePID(name)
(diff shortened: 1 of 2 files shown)


julianknutsen:
I think this was already fixed -- https://github.com/steveyegge/gastown/commit/f89ac47ff9db86acd05810c20b8d11d4ed67018c

roland-at-ailtir:
Thanks for catching this! You're absolutely right - this was already fixed in f89ac47ff9db86acd05810c20b8d11d4ed67018c (merged Jan 16, 2 days before I created this PR).

## Analysis

Both fixes address the same problem: pane processes (like `claude`) that call `setsid()` detach from the terminal and survive session kill, becoming orphaned (PPID=1).

### Upstream Solution (f89ac47)
Kills the pane PID explicitly AFTER killing descendants:
```go
// Kill descendants (as before)
for _, dpid := range descendants {
    _ = exec.Command("kill", "-KILL", dpid).Run()
}

// NEW: Kill the pane process itself
_ = exec.Command("kill", "-TERM", pid).Run()
time.Sleep(100 * time.Millisecond)
_ = exec.Command("kill", "-KILL", pid).Run()
```

### My Approach
Modified `getAllDescendants()` to include the root PID, so it gets killed in the existing loop:
```go
func getAllDescendants(pid string) []string {
    // ... find children ...
    
    // Add root PID last (after all descendants)
    result = append(result, pid)
    return result
}
```

## Verdict

The upstream fix is already in place and working. While my approach might be slightly cleaner (single unified loop), the upstream solution is more conservative (doesn't change `getAllDescendants()` semantics) and is already deployed.

Closing this PR as redundant.

roland-at-ailtir:
Already fixed upstream in f89ac47

PR #749, opened 2026-01-18
fix: detect actual rig layout instead of hardcoding mayor/rig path

dwsmith1983:
## Summary
- Doctor checks were hardcoding `mayor/rig` as the route path for all rigs
- Some rigs have `.beads` directly at the rig root (not in `mayor/rig`)
- This caused `routes.jsonl` to be created with incorrect paths during `gt doctor --fix`

## Changes
- Added `determineRigBeadsPath()` helper that checks actual rig layout:
  - If `mayor/rig/.beads` exists → use `<rig>/mayor/rig`
  - Otherwise → use `<rig>` directly
- Updated `routes_check.go` and `beads_check.go` to use this helper
- Updated tests to create the expected directory structure

## Test plan
- [x] `go test ./internal/doctor/...` passes
- [x] `gt doctor` shows routes check passing
- [x] Build verified locally

### DIFF
--- a/internal/doctor/beads_check.go
+++ b/internal/doctor/beads_check.go
@@ -302,7 +302,8 @@ func (c *PrefixMismatchCheck) Run(ctx *CheckContext) *CheckResult {
 		}
 
 		rigsJsonPrefix := rigEntry.BeadsConfig.Prefix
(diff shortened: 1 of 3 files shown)


dwsmith1983:
Closing - will recreate with clean single commit

PR #750, opened 2026-01-18
fix: detect actual rig layout instead of hardcoding mayor/rig path

dwsmith1983:
## Summary
- Routes and beads doctor checks were hardcoding `mayor/rig` as the route path for all rigs
- Some rigs have `.beads` directly at the rig root (direct layout) rather than at `mayor/rig/.beads` (mayor clone layout)
- Adds `determineRigBeadsPath()` helper that checks the actual rig layout, matching the logic in `rig.go`'s `rigInit()`

## Testing
All doctor tests pass.

### DIFF
--- a/internal/doctor/routes_check.go
+++ b/internal/doctor/routes_check.go
@@ -10,6 +10,15 @@ import (
 	"github.com/steveyegge/gastown/internal/config"
 )
 
+// determineRigBeadsPath returns the correct route path for a rig based on its actual layout.
+// Uses ResolveBeadsDir to follow any redirects (e.g., rig/.beads/redirect -> mayor/rig/.beads).
+func determineRigBeadsPath(townRoot, rigName string) string {
+	rigPath := filepath.Join(townRoot, rigName)
+	resolved := beads.ResolveBeadsDir(rigPath)
+	rel, _ := filepath.Rel(townRoot, resolved)
+	return strings.TrimSuffix(rel, "/.beads")
+}
+
 // RoutesCheck verifies that beads routing is properly configured.
 // It checks that routes.jsonl exists, all rigs have routing entries,
 // and all routes point to valid locations.
@@ -111,7 +120,8 @@ func (c *RoutesCheck) Run(ctx *CheckContext) *CheckResult {
 
 	// Check each rig has a route (by path, not just prefix from rigs.json)
 	for rigName, rigEntry := range rigsConfig.Rigs {
-		expectedPath := rigName + "/mayor/rig"
+		// Determine the correct path based on actual rig layout
+		expectedPath := determineRigBeadsPath(ctx.TownRoot, rigName)
 
 		// Check if there's already a route for this rig (by path)
 		if _, hasRoute := routeByPath[expectedPath]; hasRoute {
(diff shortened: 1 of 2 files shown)


internal/doctor/routes_check.go
```diff
@@ -10,6 +10,20 @@ import (
 	"github.com/steveyegge/gastown/internal/config"
 )
 
+// determineRigBeadsPath returns the correct route path for a rig based on its actual layout.
+// Some rigs have .beads at <rig>/mayor/rig/.beads (mayor clone layout),
+// others have .beads directly at <rig>/.beads (direct rig layout).
+// This matches the logic in rig.go's rigInit().
+func determineRigBeadsPath(townRoot, rigName string) string {
```
julianknutsen, line 15: https://github.com/steveyegge/gastown/blob/9cd2696abe68ac0defc612ace5028d327d4f207d/internal/beads/beads_redirect.go#L25

julianknutsen: I believe this already handles the redirect file so we don't have to add more hardcoding

dwsmith1983:
Tested it - ResolveBeadsDir works for both layouts:

Direct layout: returns rig/.beads → route = rigName
Mayor layout: follows redirect → route = rigName/mayor/rig

Will update to use:
resolved := beads.ResolveBeadsDir(filepath.Join(townRoot, rigName))
rel, _ := filepath.Rel(townRoot, resolved)
expectedPath := strings.TrimSuffix(rel, "/.beads")

internal/doctor/routes_check.go
dwsmith1983: changed to use existing method.

dwsmith1983:
Removed unrelated changes (namepool.go lint fix and formula updates) - PR now contains only the rig layout detection fix.
@steveyegge

dwsmith1983:
Rebased on current main - lint should pass now.

julianknutsen:

### Decision: request_changes

Thanks for tackling this, @dwsmith1983 — the hardcoded `mayor/rig` assumption was a real problem for direct-layout rigs, and using `ResolveBeadsDir` to dynamically detect the layout is a clean approach. The main concern is that the fix doesn't yet extend to `PrefixMismatchCheck` (which has the same hardcoded path), and the new behavior lacks direct test coverage. A couple more items below worth considering before merge.

---

### Major: PrefixMismatchCheck still hardcodes `mayor/rig` path

`PrefixMismatchCheck.Run()` and `.Fix()` in `beads_check.go` (lines ~307 and ~375) still use `expectedPath := rigName + "/mayor/rig"`. For direct-layout rigs where the route path is just `rigName`, the mismatch check silently skips them — the exact same class of bug this PR fixes in `RoutesCheck`.

**Suggested fix:** Apply `determineRigBeadsPath` (or a shared resolver) in `PrefixMismatchCheck.Run()` and `.Fix()`.

---

### Major: No RoutesCheck test for direct-layout rigs

The core behavioral change — `determineRigBeadsPath` resolving direct-layout rigs — has zero test coverage in `routes_check_test.go`. The test changes in `beads_check_test.go` add `.beads` directories and redirect files, but `PrefixMismatchCheck` never calls `ResolveBeadsDir`, so those directories are never read. Tests pass for the same reason they passed before.

**Suggested fix:** Add a `RoutesCheck` test with a direct-layout rig (`.beads` at rig root, no redirect) verifying `Run()` matches correctly and `Fix()` writes the right path. Either update `PrefixMismatchCheck` to use the resolver (making the test scaffolding meaningful) or remove the inert scaffolding.

---

### Major: Redirect path could escape town root boundary

If `.beads/redirect` contains a path that resolves outside the town root, `determineRigBeadsPath` would follow it and `gt doctor --fix` could persist an escaped path into `routes.jsonl`. This is a theoretical concern from code reading — no exploit path was demonstrated end-to-end.

**Suggested fix:** Validate that the resolved path stays within the town root boundary in `determineRigBeadsPath`. Reject or fall back to the default when containment fails.

---

### Minor: Silent `filepath.Rel` error discard

`rel, _ := filepath.Rel(townRoot, resolved)` discards the error. If `Rel` fails (unlikely in normal operation since both paths share a common root), `rel` would be `""`, producing an empty route path that could cause silent lookup misses or collide with the town root path.

**Suggested fix:** Handle the error — fall back to `rigName + "/mayor/rig"` or log a diagnostic and skip the rig.

---

### Pre-Merge Checklist
- [ ] Apply `determineRigBeadsPath` to `PrefixMismatchCheck.Run()` and `.Fix()`
- [ ] Add `RoutesCheck` test for direct-layout rig
- [ ] Clean up or exercise the inert test scaffolding in `beads_check_test.go`
- [ ] Handle `filepath.Rel` error in `determineRigBeadsPath`
- [ ] Evaluate redirect containment check

---

dwsmith1983:
@julianknutsen the reply and response time is too slow for me to continue.

PR #765, opened 2026-01-18
fix(config): correct Claude prompt prefix from > to ❯

### LINKED ISSUE #763: fix: Claude prompt detection uses wrong character (> instead of ❯)
## Summary

Refinery startup times out because `defaultReadyPromptPrefix()` returns `"> "` for Claude detection, but Claude Code uses `"❯"` (Unicode U+276F) as its prompt character.

## Symptoms

When running `gt up`, refineries fail to start with:

```
✖ Refinery (rig_name): waiting for refinery to start: timeout waiting for runtime prompt
```

The `WaitForRuntimeReady()` function waits up to 60 seconds looking for `"> "` in the tmux pane, but Claude Code displays `"❯"`, so the prompt is never detected.

## Root Cause

In `internal/config/types.go`:

```go
func defaultReadyPromptPrefix(provider string) string {
    if provider == "claude" {
        return "> "  // Wrong - Claude Code uses ❯
    }
    return ""
}
```

## Suggested Fix

```go
func defaultReadyPromptPrefix(provider string) string {
    if provider == "claude" {
        // Claude Code uses ❯ (U+276F) as the prompt character
        return "❯"
    }
    return ""
}
```

## Verification

Tested locally:
- With `"> "`: Refinery times out after 60 seconds
- With `"❯"`: Refinery starts successfully

## Related Files

- `internal/config/types.go` - Contains `defaultReadyPromptPrefix()`
- `internal/tmux/tmux.go` - Contains `WaitForRuntimeReady()`
- `internal/constants/constants.go` - Defines `ClaudeStartTimeout`

sauerdaniel:
## Summary

Fixes refinery startup timeout by correcting the Claude prompt detection character.

## Problem

When running `gt up`, refineries fail to start with:
```
✖ Refinery (rig_name): waiting for refinery to start: timeout waiting for runtime prompt
```

The `WaitForRuntimeReady()` function looks for `"> "` in the tmux pane to detect when Claude is ready, but Claude Code uses `"❯ "` (U+276F) as its prompt character.

## Solution

Update `defaultReadyPromptPrefix()` to return `"❯ "` instead of `"> "` for the Claude provider.

## Testing

Verified locally:
- With `"> "`: Refinery times out after 60 seconds
- With `"❯ "`: Refinery starts successfully

Fixes #763

### DIFF
--- a/internal/config/types.go
+++ b/internal/config/types.go
@@ -552,7 +552,8 @@ func defaultProcessNames(provider, command string) []string {
 
 func defaultReadyPromptPrefix(provider string) string {
 	if provider == "claude" {
-		return "> "
+		// Claude Code uses ❯ (U+276F) as the prompt character
+		return "❯ "
 	}
 	return ""
 }


aleiby:
On my setup, this change caused the exact issue that it intends to fix.  Maybe a platform difference?  I am running on WSL w/ Claude v2.1.12.

sauerdaniel:
Tested on both macOS and Linux with Claude Code v2.1.12:

**macOS:**
```
❯ Try "refactor index.ts"
```

**Linux (Fedora):**
```
❯ Try "how do I log an error?"
```

Both platforms use the same prompt character: `❯` (U+276F, HEAVY RIGHT-POINTING ANGLE QUOTATION MARK ORNAMENT).

Let's see which character is being used on other people's systems. Could be platform dependent, yes.

aleiby:
> Both platforms use the same prompt character: `❯` (U+276F, HEAVY RIGHT-POINTING ANGLE QUOTATION MARK ORNAMENT).
> 
Interesting.  That is also what I see at the interactive prompt.
When I applied this fix, the refinery hung when running gt up, and when I reverted it did not, so not sure what's going on.  I have not investigated further.

gmorpheme:
Doesn't fix for me. My mayor is claiming that even if the chevron is right, *the space character is wrong* and we need a non-breaking space.

gmorpheme:
Raised as #838

PR #770, opened 2026-01-18
feat(patrol): implement PING/PONG protocol with backoff

dwsmith1983:
## Summary
- Implement request-response PING/PONG protocol between witness and deacon
- Add exponential backoff to prevent mail flood from bricking deacon
- Escalate to Mayor after repeated failures

## Problem
Previously, witnesses would flood the deacon with WITNESS_PING messages. If the deacon was slow to respond (e.g., processing a large patrol cycle), the accumulated pings would interrupt the deacon mid-thought via UserPromptSubmit hook, causing it to get stuck.

## Solution
**PING/PONG Protocol v2:**
1. Request-response model with unique `request_id`
2. One outstanding PING at a time - wait for PONG before sending next
3. Exponential backoff: 30s → 60s → 120s between retries
4. Escalate to Mayor after 3 consecutive failures
5. 5-minute cooldown after escalation

**Witness (mol-witness-patrol.formula.toml):**
- Check for PONG response before sending new PING
- Track pending_ping state with timestamps
- Implement backoff logic

**Deacon (mol-deacon-patrol.formula.toml):**
- Extract request_id from PING subject
- Send PONG response with matching request_id
- Archive processed PING messages

## Test plan
- [x] Witness sends PING with request_id
- [x] Deacon responds with PONG containing request_id
- [x] Witness waits for PONG before next PING
- [x] No mail flood observed during testing

### DIFF
--- a/.beads/formulas/mol-deacon-patrol.formula.toml
+++ b/.beads/formulas/mol-deacon-patrol.formula.toml
@@ -23,11 +23,40 @@ Witnesses detect it and escalate to the Mayor.
 The Deacon's agent bead last_activity timestamp is updated during each patrol
 cycle. Witnesses check this timestamp to verify health."""
 formula = "mol-deacon-patrol"
-version = 8
+version = 9
+
+[[steps]]
+id = "role-check"
+title = "Verify role is enabled"
+description = """
+**STOP. Read your CLAUDE.md role context first.**
+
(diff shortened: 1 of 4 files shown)


julianknutsen:
Closing as author is unable to continue contributing.

PR #777, opened 2026-01-18
Escape backticks and dollar signs in quoteForShell

swhitt:
Mayor startup times out with "still running excluded command".

The startup prompt has markdown with backticks like `` `gt hook` `` and `` `gt mail inbox` ``. When this gets passed through `tmux` in a double-quoted string, the shell interprets those backticks as command substitution. It runs `gt mail inbox` and replaces the backtick expression with its output, which corrupts the command and drops back to shell.

Fix: escape backticks and dollar signs in `quoteForShell()`. Added 8 test cases covering all escaped characters including edge cases.

Mayor startup now works in all cases tested.

### DIFF
--- a/internal/formula/formulas/mol-deacon-patrol.formula.toml
+++ b/internal/formula/formulas/mol-deacon-patrol.formula.toml
@@ -665,71 +665,46 @@ Skip dispatch - system is healthy.
 
 [[steps]]
 id = "costs-digest"
-title = "Aggregate daily costs [DISABLED]"
+title = "Aggregate daily costs"
 needs = ["session-gc"]
 description = """
-**⚠️ DISABLED** - Skip this step entirely.
+**DAILY DIGEST** - Aggregate yesterday's session cost wisps.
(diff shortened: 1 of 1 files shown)


olivierlefloch:
Encountered this same issue and ended up with the same fix.

internal/formula/formulas/mol-deacon-patrol.formula.toml
```diff
@@ -665,71 +665,46 @@ Skip dispatch - system is healthy.
 
 [[steps]]
 id = "costs-digest"
-title = "Aggregate daily costs [DISABLED]"
+title = "Aggregate daily costs"
```
olivierlefloch, line 668: I don't think this should be part of this PR.

swhitt: I didn't either but it looks like it's necessary to pass CI.
It won't pass without a `go generate`.

PR #778, opened 2026-01-19
feat(daemon): add --rig flag for single-rig operation

Atmosphere-Aviation:
## Summary
- Add `--rig` flag to `gt daemon start` to limit witness/refinery management to a single specified rig
- Prevents daemon from auto-spawning agents for all registered rigs (avoids orphan cascade during startup)
- When `--rig` is specified, only that rig's witness/refinery are managed

## Test plan
- [ ] Run `gt daemon start --rig gastown` and verify only gastown witness/refinery are spawned
- [ ] Check daemon logs show "single-rig mode: gastown"
- [ ] Verify process count stays reasonable (≤6 processes)
- [ ] Run `gt daemon start` without `--rig` and verify all rigs are managed (backward compatible)

### DIFF
--- a/internal/cmd/daemon.go
+++ b/internal/cmd/daemon.go
@@ -66,8 +66,9 @@ var daemonRunCmd = &cobra.Command{
 }
 
 var (
-	daemonLogLines int
+	daemonLogLines  int
 	daemonLogFollow bool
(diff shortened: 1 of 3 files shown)


dwsmith1983:
I think we have overlap: https://github.com/steveyegge/gastown/pull/711

PR #785, opened 2026-01-19
fix(crew): use directory name as source of truth for crew worker names

joshuavial:
## Summary
- Fix `gt crew list` showing wrong names when `state.json` contains stale data
- Always use directory name as source of truth in `loadState()` instead of trusting potentially stale `state.json`
- Fix pre-existing gosec G115 lint warning in namepool.go

## Root Cause
`loadState()` only backfilled `Name` when empty, so stale names from copied/renamed/corrupted `state.json` files persisted.

## Test Plan
- Unit tests: Added regression test `TestManagerGetWithStaleStateName`
- Manual: Verified `gt crew list` now shows correct directory-based names


steveyegge:
Merged manually via cherry-pick to main (commit 183a0d7d). Thank you for the contribution!

Issue #10, opened 2026-01-02
Agent Agnostic Engine with Multi-provider support using SRT to safely sandbox headless sessions


kraitsura:
## Problem Statement
Gas Town currently only supports Claude Code as the runtime for agents. The RuntimeConfig struct in internal/config/types.go has the infrastructure for configurable commands, but there's no abstraction layer or implementations for other AI coding assistants. This limits users to a single provider and prevents leveraging different models for different use cases (cost optimization, specialized tasks, redundancy).

## Proposed Solution
 Proposed Solution

  1. Adopt Anthropic's SRT (Sandboxed Runtime) as the execution layer for all coding agents. This provides:
    - Consistent sandboxing across different CLI tools
    - Security isolation for headless agent execution
    - Unified environment management
  2. Create a provider abstraction layer that supports:
    - Claude Code (claude) - current default
    - OpenCode (opencode) - open-source alternative
    - Codex CLI (codex) - OpenAI's coding agent
    - Aider (aider) - popular open-source option
    - Continue (continue) - VS Code-native agent
  3. Extend RuntimeConfig with provider-specific adapters:
  ```
  type RuntimeConfig struct {
        Provider      string            // "claude", "opencode", "codex", "aider"
        Command       string            // CLI command
        Args          []string          // Provider-specific args
        SandboxConfig *SandboxConfig    // SRT configuration
        HookAdapter   string            // How to translate hooks for this provider
  }
```
  4. Hook translation layer - map Gas Town's hook system (SessionStart, UserPromptSubmit, etc.) to each provider's equivalent mechanism.

## Alternatives Considered
  - Docker-based isolation: More overhead, harder to manage tmux integration
  - Direct process spawning without sandbox: Security concerns for headless execution
  - Provider-specific forks: Maintenance nightmare, defeats the purpose of unified orchestration

## Use Case
  - User choice: Let users pick their preferred coding agent - some teams prefer Claude, others are invested in OpenAI's ecosystem, and open-source advocates want self-hosted options. Gas Town should be engine-agnostic.

## Additional Context
Current extensibility point in internal/config/types.go:124-148:
```
  type RuntimeConfig struct {
      Command       string
      Args          []string
      InitialPrompt string
  }
```

The session manager (internal/session/manager.go) already supports environment variable injection and tmux-based execution, making it relatively straightforward to swap the underlying CLI.

  Key challenges:
  - Each CLI has different hook/event systems (or none at all)
[...]

PR #788, opened 2026-01-19
fix(handoff): preserve agent override across session restarts

### LINKED ISSUE #692: Support per-agent environment variables in runtime config

joshuavial:
## Summary
- Adds `GT_AGENT` env var to track agent override when using `--agent` flag
- Handoff reads and preserves `GT_AGENT` so non-default agents persist across restarts

## Problem
When starting a session with `gt sling --agent gemini`, running `gt handoff` would restart with the default agent (claude) instead of preserving gemini.

## Solution
1. `BuildStartupCommandWithAgentOverride` now sets `GT_AGENT=<override>` in exports
2. `buildRestartCommand` in handoff reads `GT_AGENT` and uses it for the restart command
3. `GT_AGENT` is included in exports so new session inherits it

## Test Plan
- Added unit tests for GT_AGENT being set/not set
- All existing tests pass

Fixes #692 (partial - addresses agent override persistence)
Related to #10 (multi-agent support)

### DIFF
--- a/internal/cmd/handoff.go
+++ b/internal/cmd/handoff.go
@@ -383,7 +383,20 @@ func buildRestartCommand(sessionName string) (string, error) {
 	// 3. export Claude-related env vars (not inherited by fresh shell)
 	// 4. run claude with the startup beacon (triggers immediate context loading)
 	// Use exec to ensure clean process replacement.
-	runtimeCmd := config.GetRuntimeCommandWithPrompt("", beacon)
+	//
+	// Check if current session is using a non-default agent (GT_AGENT env var).
+	// If so, preserve it across handoff by using the override variant.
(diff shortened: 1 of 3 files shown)


steveyegge:
Merged manually via cherry-pick to main (commit 48ace2cb). Thank you for the contribution!

PR #797, opened 2026-01-20
fix(seance): support cross-account session access

joshuavial:
## Summary
- Fix `gt seance --talk <session-id>` to work after `gt account switch`
- Sessions from any account are now accessible via temporary symlinks
- All session entry fields preserved when copying to sessions-index.json

## Problem
After `gt account switch`, seance would fail to find sessions from the previous account because it only searched `~/.claude` (which is symlinked to the current account's config dir).

## Solution
- Search all account config directories in `~/gt/mayor/accounts.json` for the session
- Temporarily symlink the session file from source account to current account's project dir
- Update `sessions-index.json` with full session entry (using json.RawMessage to preserve all fields)
- Cleanup removes symlink and index entry when seance exits

## Test Plan
- [x] Tested seance with sessions from different accounts
- [x] Verified symlink creation and index update
- [x] Verified cleanup after seance exits
- [x] All tests pass (`go test ./...`)

### DIFF
--- a/internal/polecat/namepool.go
+++ b/internal/polecat/namepool.go
@@ -378,7 +378,7 @@ func ThemeForRig(rigName string) string {
 	for _, b := range []byte(rigName) {
 		hash = hash*31 + uint32(b)
 	}
-	return themes[hash%uint32(len(themes))]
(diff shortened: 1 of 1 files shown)


steveyegge:
Merged manually with orphan symlink cleanup enhancement. See commit 9f06eb94.

PR #821, opened 2026-01-20
fix(gt done): prevent gt done from killing itself during session cleanup

julianknutsen:
## Summary

When `gt done` runs inside a tmux session (e.g., after polecat task completion), calling `KillSessionWithProcesses` would kill the `gt done` process itself before it could complete cleanup operations like writing handoff state. This causes race conditions where `gt done` terminates prematurely.

This fix adds a new function `KillSessionWithProcessesExcluding` that excludes specified PIDs from the kill sequence, allowing the caller to survive until the session is explicitly destroyed.

## The Problem
  gt done is running INSIDE the tmux session it's trying to kill. When it calls getAllDescendants(pid) on the tmux pane:

  1. The descendant list includes gt done itself (it's a child of the shell, which is a child of the pane)
  2. getAllDescendants returns deepest-first order
  3. gt done gets killed FIRST (exit code 137 = SIGKILL)
  4. The Claude process and the tmux kill-session command NEVER execute because gt done is dead

  Result: gt done commits suicide before finishing the job. The Claude session survives because:
  - The loop that would have killed Claude never completed
  - The tmux kill-session command never ran

  This is a textbook case of "can't pull the trigger on yourself mid-pull."
  
## Changes

- Add `KillSessionWithProcessesExcluding(name string, excludePIDs []string)` function to `internal/tmux/tmux.go`
  - Accepts a list of PIDs to exclude from SIGTERM/SIGKILL before session destruction
  - Uses O(1) map lookup for exclusion checks
  - Exclusion applies to both descendant processes and the pane process itself
- Update `selfKillSession` in `internal/cmd/done.go` to exclude its own PID
  - Uses `strconv.Itoa(os.Getpid())` to get current PID
  - Ensures `gt done` completes handoff state writing before session termination
- Fix both `KillSessionWithProcesses` and `KillSessionWithProcessesExcluding` to ignore "session not found" errors
  - When all processes in a session are killed, tmux may automatically destroy the session
  - Treating this as an error is incorrect since the goal (session gone) was achieved
- Add comprehensive tests for the new functionality:
  - `TestKillSessionWithProcesses` - basic kill functionality
  - `TestKillSessionWithProcesses_NonexistentSession` - handles missing sessions
  - `TestKillSessionWithProcessesExcluding` - empty exclude list behavior
  - `TestKillSessionWithProcessesExcluding_WithExcludePID` - exclusion works
  - `TestKillSessionWithProcessesExcluding_NonexistentSession` - handles missing sessions

## Testing

- [x] New tests added for `KillSessionWithProcessesExcluding`
- [x] Build succeeds (`go build ./...`)

## Checklist

- [x] Tests added for new functionality
- [x] No breaking changes

### DIFF
--- a/internal/cmd/done.go
+++ b/internal/cmd/done.go
@@ -4,6 +4,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strconv"
 	"strings"
 
 	"github.com/spf13/cobra"
@@ -745,9 +746,12 @@ func selfKillSession(townRoot string, roleInfo RoleInfo) error {
 
 	// Kill our own tmux session with proper process cleanup
 	// This will terminate Claude and all child processes, completing the self-cleaning cycle.
-	// We use KillSessionWithProcesses to ensure no orphaned processes are left behind.
+	// We use KillSessionWithProcessesExcluding to ensure no orphaned processes are left behind,
+	// while excluding our own PID to avoid killing ourselves before cleanup completes.
+	// The tmux kill-session at the end will terminate us along with the session.
 	t := tmux.NewTmux()
-	if err := t.KillSessionWithProcesses(sessionName); err != nil {
+	myPID := strconv.Itoa(os.Getpid())
+	if err := t.KillSessionWithProcessesExcluding(sessionName, []string{myPID}); err != nil {
 		return fmt.Errorf("killing session %s: %w", sessionName, err)
 	}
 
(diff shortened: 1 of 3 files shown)


steveyegge (approved):
Code looks good. The fix correctly excludes the calling process from the kill sequence while ensuring it still gets terminated when the session is destroyed. Tests are comprehensive. Windows CI failure is a pre-existing issue unrelated to this PR.

PR #835, opened 2026-01-21
fix(cmd): extract orphan cleanup to platform-specific files

joyshmitz:
## Summary
- Extracts `cleanupOrphanedClaude()` from `start.go` to platform-specific files
- Creates `start_orphan_unix.go` with full SIGTERM/SIGKILL implementation
- Creates `start_orphan_windows.go` with stub (Windows lacks these signals)
- Removes unused `syscall` and `util` imports from `start.go`

## Problem
PR #759 introduced `cleanupOrphanedClaude()` using `syscall.Kill` directly, which is Unix-only. This breaks Windows CI with:
- `internal\cmd\start.go:1031:21: undefined: syscall.Kill`

## Solution
Follow existing codebase pattern (`process_unix.go` / `process_windows.go`) by extracting platform-specific code to separate files with build tags.

## Test plan
- [x] `GOOS=windows GOARCH=amd64 go build -v ./cmd/gt` - passes
- [x] `GOOS=linux GOARCH=amd64 go build -v ./cmd/gt` - passes
- [x] `GOOS=darwin GOARCH=amd64 go build -v ./cmd/gt` - passes
- [x] `go test ./internal/cmd/...` - passes

### DIFF
--- a/internal/cmd/start.go
+++ b/internal/cmd/start.go
@@ -9,7 +9,6 @@ import (
 	"strings"
 	"sync"
 	"sync/atomic"
-	"syscall"
 	"time"
 
 	"github.com/spf13/cobra"
@@ -26,7 +25,6 @@ import (
 	"github.com/steveyegge/gastown/internal/session"
 	"github.com/steveyegge/gastown/internal/style"
 	"github.com/steveyegge/gastown/internal/tmux"
-	"github.com/steveyegge/gastown/internal/util"
 	"github.com/steveyegge/gastown/internal/witness"
 	"github.com/steveyegge/gastown/internal/workspace"
 )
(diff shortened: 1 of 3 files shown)


steveyegge (approved):
Clean fix for the Windows CI issue. The changes:

1. **Follow existing patterns** - Uses the same `//go:build !windows` / `//go:build windows` pattern as `process_unix.go`/`process_windows.go` in this codebase

2. **Unix implementation unchanged** - The `start_orphan_unix.go` is a direct copy of the original code with no functional changes

3. **Windows stub is appropriate** - Since Windows lacks SIGTERM/SIGKILL semantics, a stub that prints a message is the right approach (could be enhanced later with `TerminateProcess` if needed)

4. **Clean import removal** - Correctly removes now-unused `syscall` and `util` imports from `start.go`

LGTM 👍

PR #851, opened 2026-01-21
fix(install): Makefile uses go install to match docs

easel:
Background here is that it's pretty easy to end up with multiple gt binaries, one in $GOPATH and another in ~/.local/bin. Roles / Claude plugins don't always get the same path, causing weird problems.

## Summary
- Makefile install target now uses `go install` instead of copying to `~/.local/bin`
- Ensure `bd config` uses the resolved beads dir when configuring custom types
- Update startup command and bd stub tests for current command behavior

## Test plan
- [x] `go build -o gt ./cmd/gt`
- [x] `go test ./...`

Closes: hq-93c


steveyegge (approved):
Easy-win: Makefile install now uses `go install` to match docs. Also fixes BEADS_DIR env propagation and updates tests.

✅ LGTM - approving and merging.

PR #882, opened 2026-01-23
fix(handoff): prevent self-kill during gt handoff (#881)

### LINKED ISSUE #881: bug(handoff): gt handoff times out due to KillPaneProcesses killing self
## Summary

`gt handoff` times out and fails to respawn the session after commit 405d40ee.

## Root Cause

Commit `405d40ee` ("fix(tmux): KillPaneProcesses must kill pane process itself, not just descendants") added logic to kill the pane's main process in `KillPaneProcesses()`:

```go
// Kill the pane process itself
_ = exec.Command("kill", "-TERM", pid).Run()
time.Sleep(100 * time.Millisecond)
_ = exec.Command("kill", "-KILL", pid).Run()
```

When `gt handoff` runs inside Claude Code and calls `KillPaneProcesses(pane)`, it kills Claude Code (the pane process) - which also kills the `gt handoff` command before it can reach `RespawnPane()`.

## Reproduction

```bash
gt handoff
# Output: 🤝 Handing off gt-gastown-crew-holden...
# Then times out after 2 minutes with no respawn
```

## Expected Behavior

`gt handoff` should successfully respawn the pane with a fresh Claude session.

## Actual Behavior

1. `gt handoff` prints "🤝 Handing off..."
2. Calls `KillPaneProcesses(pane)`
3. `KillPaneProcesses` sends SIGKILL to Claude Code (the pane process)
4. Claude dies, taking `gt handoff` with it
5. `RespawnPane` is never called
6. Pane is left with dead process, no respawn occurs

## Suggested Fix

Option A: Don't kill the pane process in `KillPaneProcesses` - let `respawn-pane -k` handle it (the `-k` flag already sends SIGHUP to the pane process).

Option B: Have `gt handoff` pass a flag to skip killing the pane process itself when doing self-handoff, since `respawn-pane -k` will handle it.

## Workaround

Use tmux directly:
```bash
tmux respawn-pane -k
```

Or start fresh from another terminal:
```bash
gt crew cycle <name>
```

aleiby:
## Summary

Fixes #881. `gt handoff` was calling `KillPaneProcesses` which killed Claude Code (the pane process) before `RespawnPane` could be called, causing handoff to silently fail with no respawn.

**Changes:**
- Add `KillPaneProcessesExcluding` function in `internal/tmux/tmux.go` that allows excluding specific PIDs from being killed
- Modify self-handoff in `internal/cmd/handoff.go` to exclude the current process and its parent (Claude Code)
- Add unit tests for the new function

## Test plan

- [x] All existing tests pass (`go test ./...`)
- [x] New unit tests for `KillPaneProcessesExcluding` pass
- [x] Linters pass (`go vet ./...`, `gofmt`)
- [x] Manual test: run `gt handoff` from a crew session and verify respawn completes

### DIFF
--- a/internal/cmd/handoff.go
+++ b/internal/cmd/handoff.go
@@ -5,6 +5,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strconv"
 	"strings"
 
 	"github.com/spf13/cobra"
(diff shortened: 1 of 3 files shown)


steveyegge:
Approved, but needs rebase to resolve merge conflicts with main.

steveyegge:
This PR has merge conflicts and cannot be merged cleanly. Could you please rebase against main?

aleiby:
Rebased onto main. Resolved conflicts by keeping both:
- `SetRemainOnExit` (added in main for pane survival during handoff)
- `KillPaneProcessesExcluding` with self-PID exclusion (this PR's fix)

Also merged both test sets in tmux_test.go.

Note: CI failures are pre-existing on main (binary signing issue in integration tests).

PR #892, opened 2026-01-23
Draft: Fix SessionManager treating witness/refinery as polecats

Kazanir:
Marking this PR as draft because I'm not sure if this is the right/even-a-good fix or if it reflects a larger incoherency/separation-of-concerns issue with SessionManager.

## Summary

The bug: SessionManager.List() returned ALL sessions matching gt-<rig>-*, including witness, refinery, and crew sessions. Commands like gt rig shutdown would try to sync beads in non-existent directories like <prefix>/polecats/refinery, causing "beads sync failed" warnings.

The fix: Add ListPolecats() method that filters out witness, refinery, and crew-* sessions. Update call sites that should only see polecats (rig shutdown/stop/restart, gt down) to use ListPolecats() instead of List().

Leave List() unchanged so gt session list continues to show all sessions.

## Testing

## Checklist

### DIFF
--- a/internal/cmd/down.go
+++ b/internal/cmd/down.go
@@ -343,7 +343,7 @@ func stopAllPolecats(t *tmux.Tmux, townRoot string, rigNames []string, force boo
 		}
 
 		polecatMgr := polecat.NewSessionManager(t, r)
-		infos, err := polecatMgr.List()
+		infos, err := polecatMgr.ListPolecats()
 		if err != nil {
 			continue
 		}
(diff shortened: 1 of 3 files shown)


julianknutsen:
You've stumbled on an unfortunate overloading of "session" that makes the expected behavior non-obvious. Based on `gt session list -h` it is supposed to just return the polecats as well and likely wasn't changed when witness/refinery/crew were moved from town-level to rig-level agents (hq -> gt prefixes). It also makes sense to me that a `List()` call on a polecat manager just returns the polecats so maybe we can just move the fix to the original method. What do you think? Are there any other stray callers of `List()` that expect all the sessions? If so, we should probably have another component handle that query.

Kazanir:
I guess it is only important if the overloading of `gt session` means that people have come to depend on it returning all sessions for a given rig, which is what the name implies if you don't check the help message. I believe there are no other call sites beyond these 6, of which only the CLI command itself was not converted -- maybe the entire subcommand should just be `polecats` and not `sessions`? It is definitely confusing to new citizens :P

PR #927, opened 2026-01-24
fix(tmux): use syscall.Kill() for process group termination

Kazanir:
fix(tmux): use syscall.Kill() for process group termination to avoid systemd session teardown

ROOT CAUSE: exec.Command("kill", "-TERM", "-"+pgid) invokes /usr/bin/kill, which sometimes requires a '--' prepended before the '-pgid' in order to avoid the '-pgid' being interpreted as a command option. Ultimately, when this occurs, the resulting call to /usr/bin/kill results in the SIGTERM being broadcast to all processes (-1) and having systemd shut down the entire slice.

The fix uses syscall.Kill() directly with a negative PID, which is the POSIX- correct way to signal a process group via the kernel interface. This bypasses the /usr/bin/kill binary entirely.

Fixes orphan cleanup bug where `gt down` would kill the entire user session.

Co-Authored-By: GLM 4.7 (https://z.ai)

## Testing

## Checklist

### DIFF
--- a/internal/tmux/tmux.go
+++ b/internal/tmux/tmux.go
@@ -187,11 +187,8 @@ func (t *Tmux) KillSessionWithProcesses(name string) error {
 		// Note: Processes that called setsid() will have a new PGID and won't be killed here
 		pgid := getProcessGroupID(pid)
 		if pgid != "" && pgid != "0" && pgid != "1" {
-			// Kill process group using syscall.Kill() directly.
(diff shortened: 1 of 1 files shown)


steveyegge:
Applied this fix directly to main in commit 0605db97. Using `syscall.Kill()` is the right approach — it eliminates all shell parsing ambiguity. The previous `"--"` workaround was fragile. Thank you for the fix! Closing.

Issue #27, opened 2026-01-02
Memory pressure from multi-agent workflows causes system OOM


eric-distyl-ai:
## Problem
Running multiple polecats/agents simultaneously leads to system memory exhaustion.
Current system shows ~253 Claude processes consuming ~24.4 GB RSS total.

## Impact
- System becomes unresponsive
- Requires hard restart
- Work in progress may be lost (e.g., stranded uncommitted work in polecat worktrees)

## Proposed Solution
Add a memory monitor that:
- Tracks total RSS across all Claude Code process subtrees
- Kills polecats when threshold exceeded (e.g., 14 GB)
- Notifies mayor via mail before/after action

## Environment
- macOS Darwin 24.6.0
- Multiple rigs with concurrent polecats

eric-distyl-ai:
This was a Claude Code bug (or a bug with my local .zshrc or whatever) - I ran into this issue even without Gas Town, and I fixed it a different way.

luisoala:
ehi @eric-distyl-ai care to share the problem and fix? running into similar issues atm

PR #26, opened 2026-01-02
fix: Enable gt doctor --fix to remediate missing patrol role templates

Avyukth:
## Summary

Enable `gt doctor --fix` to auto-remediate missing patrol role templates by making `PatrolRolesHavePromptsCheck` fixable.

## Problem

Running `gt doctor` shows this warning that `--fix` cannot remediate:

```
⚠ patrol-roles-have-prompts: 3 role prompt template(s) missing
    myrig: deacon.md.tmpl
    myrig: witness.md.tmpl
    myrig: refinery.md.tmpl
    → Role prompt templates should be in the project repository under internal/templates/roles/
```

The check detects missing templates but provides no auto-fix capability.

### `internal/doctor/patrol_check.go`

**Struct changes:**
- Changed `BaseCheck` → `FixableCheck` to enable auto-fix
- Added `missingByRig map[string][]string` to track missing templates per rig

**Run() changes:**
- Now populates `missingByRig` during detection for use by `Fix()`

**New Fix() method:**
```go
func (c *PatrolRolesHavePromptsCheck) Fix(ctx *CheckContext) error {
    allTemplates, err := templates.GetAllRoleTemplates()
    // For each rig with missing templates:
    //   - Create directory at <rig>/mayor/rig/internal/templates/roles/
    //   - Copy embedded templates to disk
}
```

**FixHint change:**
- Before: `"Role prompt templates should be in the project repository under internal/templates/roles/"`
- After: `"Run 'gt doctor --fix' to copy embedded templates to rig repos"`

### `internal/templates/templates.go`

**New function:**
```go
func GetAllRoleTemplates() (map[string][]byte, error)
```
Reads all role templates from embedded FS and returns as filename → content map.

### `internal/doctor/patrol_check_test.go` (new file)

13 test cases covering:
- `TestNewPatrolRolesHavePromptsCheck` - Constructor and `CanFix()` returns true
- `TestPatrolRolesHavePromptsCheck_NoRigs` - Returns OK when no rigs
- `TestPatrolRolesHavePromptsCheck_NoTemplatesDir` - Detects missing dir
- `TestPatrolRolesHavePromptsCheck_SomeTemplatesMissing` - Detects partial
- `TestPatrolRolesHavePromptsCheck_AllTemplatesExist` - Returns OK
- `TestPatrolRolesHavePromptsCheck_Fix` - Creates all templates
- `TestPatrolRolesHavePromptsCheck_FixPartial` - Preserves existing files
[...]

Issue #25, opened 2026-01-02
Harden bd daemon setup to improve gt status reliability


eric-distyl-ai:
## Problem

`gt status` can become very slow (39+ seconds) when the bd daemon is not running properly. This happens because:

1. Each `bd` call falls back to "direct mode" taking 2-5 seconds
2. With 8+ polecats across multiple rigs, the time adds up significantly

## Root Cause

The bd daemon fails to start silently when:
- Legacy database lacks repository fingerprint (pre-0.17.5)
- Database migration needed (`bd migrate --update-repo-id`)

## Solution

1. **`gt status` should check daemon health first**
   - If daemon not running, attempt to start it
   - If daemon fails to start, show warning (not block status display)

2. **`gt install` / `bd init` should ensure daemon can start**
   - Validate database fingerprint during setup
   - Auto-migrate legacy databases

3. **`gt doctor` should check bd daemon status**
   - Add check for daemon running
   - Add check for database fingerprint
   - `--fix` should be able to migrate and restart daemon

4. **Better error messages when daemon fails**
   - Surface the actual error (e.g., "legacy database detected")
   - Provide one-liner fix command

## Fix Applied Today

Manually ran:
```
bd migrate --update-repo-id
bd daemon --start
```

This reduced `gt status` from 39s to 2s.

Related bead: hq-0nu

vessenes:
Related, I just filed https://github.com/steveyegge/beads/issues/863; gastown created the files that triggered this error in my case as well.

johann-taberlet:
I traced one of the daemon startup failures to a timing issue in `gt install`: git initialization happens **after** beads initialization, but `bd` needs a git repo to compute the repository fingerprint.

When `gt install --git` runs:
1. Beads init runs first → `bd migrate --update-repo-id` fails with "not a git repository"
2. Git init runs after → too late, fingerprint already missing
3. Daemon can't start → multiple daemon instances spawn, 5s+ timeouts

**Root cause in `internal/cmd/install.go`:**
- Git init block is at line ~246 (after beads init at ~200)

**Fix:** Move git initialization before beads initialization so `bd` can compute the fingerprint from the existing git repo.

I have a minimal fix ready (9 lines moved, same logic) if you're interested - happy to open a PR or take a different approach if you prefer.

johann-taberlet:
## Related: Git init timing issue during `gt install --git`

[...]

PR #928, opened 2026-01-24
fix(tmux): add -- to kill commands to prevent procps-ng argument parsing bug

groblegark:
## The Murder Investigation

After **26 Claude AI instances were killed** while running `TestCleanupOrphanedSessions`, we finally discovered the root cause.

### The Bug

The procps-ng `/usr/bin/kill` binary (v4.0.4) misparses negative PIDs:

```
/usr/bin/kill -KILL -12345   -> kill(-1, SIGKILL)  # WRONG! Kills ALL processes!
/usr/bin/kill -KILL -- -12345 -> kill(-12345, SIGKILL) # Correct!
```

The negative PGID `-12345` looks like an option (starts with `-`), so procps-ng tries to parse it as a signal, fails, and falls back to `-1` (all processes).

### The Fix

Add `--` before negative PGID arguments:

```go
// Before (FATAL):
exec.Command("kill", "-KILL", "-"+pgid).Run()

// After (SAFE):
exec.Command("kill", "-KILL", "--", "-"+pgid).Run()
```

### Investigation Summary

| Attempt | Status | Finding |
|---------|--------|----------|
| #1-#25 | Killed | Various theories, all died |
| #26 | BREAKTHROUGH | Shell wrapper SAFE, direct kill FATAL |
| #27 | SOLVED | Found the `--` argument parsing bug |

### Full Investigation

See the complete 27-investigator murder mystery:
https://github.com/groblegark/gastown/blob/main/MURDER_INVESTIGATION.md

---
Generated with Claude Code

### DIFF
--- a/internal/tmux/tmux.go
+++ b/internal/tmux/tmux.go
@@ -187,10 +187,13 @@ func (t *Tmux) KillSessionWithProcesses(name string) error {
 		if pgid != "" && pgid != "0" && pgid != "1" {
 			// Kill process group with negative PGID (POSIX convention)
 			// Use SIGTERM first for graceful shutdown
-			_ = exec.Command("kill", "-TERM", "-"+pgid).Run()
+			// NOTE: The "--" is CRITICAL! Without it, procps-ng kill (v4.0.4+)
+			// misparses "-PGID" as an option and ends up calling kill(-1) which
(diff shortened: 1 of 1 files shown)


Kazanir:
I have had a similar past 36 hours over at #927

PR #929, opened 2026-01-24
fix(settings): set editorMode to normal for Claude sessions

btucker:
This ensures Claude Code starts in normal editor mode rather than potentially using vim mode, which can cause issues with automated text input via tmux send-keys.

## Changes
- Added editorMode: normal to settings-autonomous.json
- Added editorMode: normal to settings-interactive.json

## Test plan
- [x] Rebuilt gt binary with changes
- [x] Booted selkie rig polecats
- [x] Verified witness/.claude/settings.json contains editorMode: normal

### DIFF
--- a/internal/claude/config/settings-autonomous.json
+++ b/internal/claude/config/settings-autonomous.json
@@ -1,4 +1,5 @@
 {
+  "editorMode": "normal",
   "enabledPlugins": {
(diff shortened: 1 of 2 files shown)


bobrich:
Thank you!  This took more time than it should have for me to figure out.

PR #952, opened 2026-01-25
fix(sling) race condition in sling polecat spawn and duplicate nudges

PiotrTrzpil:
## Problem

When `gt sling` spawns a fresh polecat for a rig, there were two issues:

1. **Race condition**: The tmux session starts before `attached_molecule` can be set, so `gt prime` (which runs on session start) might not see the molecule.

2. **Duplicate nudges**: Fresh polecats received multiple nudges in rapid succession (StartupNudge, PropulsionNudge 2s later, then injectStartPrompt), causing Claude to interpret them as interruptions ("What should Claude do instead?") instead of working.

### Commit 1: Defer session start until bead setup complete

- `SpawnPolecatForSling` now creates the worktree but defers session start
- Add `StartSession()` method to `SpawnedPolecatInfo`
- Sling calls `StartSession()` after `attached_molecule` is set

### Commit 2: Remove duplicate nudges

- Remove `PropulsionNudge` in `session_manager.go` (redundant with `StartupNudge`)
- Skip `injectStartPrompt` in `sling.go` for freshly spawned polecats

Now fresh polecats receive exactly one nudge (StartupNudge with "Run gt hook" instructions) and start working immediately.

## Changes

- `internal/cmd/polecat_spawn.go` - Defer session start, add `StartSession()` and `SessionStarted()` methods
- `internal/cmd/sling.go` - Call `StartSession()` after bead setup, skip nudge for fresh spawns
- `internal/polecat/manager.go` - Add `SetAgentState()` method
- `internal/polecat/session_manager.go` - Remove redundant PropulsionNudge
- `internal/cmd/polecat_spawn_test.go` - Unit tests for new methods

## Testing

- Unit tests added for `SessionStarted()`, `AgentID()`, and `StartSession()` idempotency
- Manual testing confirmed polecats now start working without interruption
- All tests pass: `go test ./...`

### DIFF
--- a/internal/cmd/sling.go
+++ b/internal/cmd/sling.go
@@ -520,7 +520,8 @@ func runSling(cmd *cobra.Command, args []string) error {
 
 	// Start polecat session now that attached_molecule is set
 	// This ensures polecat sees the molecule when gt prime runs on session start
-	if newPolecatInfo != nil {
+	freshlySpawned := newPolecatInfo != nil
+	if freshlySpawned {
 		pane, err := newPolecatInfo.StartSession()
 		if err != nil {
 			return fmt.Errorf("starting polecat session: %w", err)
@@ -529,7 +530,11 @@ func runSling(cmd *cobra.Command, args []string) error {
 	}
 
 	// Try to inject the "start now" prompt (graceful if no tmux)
-	if targetPane == "" {
+	// Skip for freshly spawned polecats - SessionManager.Start() already sent StartupNudge
+	// with instructions. Sending another nudge causes Claude to interpret it as interruption.
+	if freshlySpawned {
+		// Fresh polecat already got StartupNudge from SessionManager.Start()
+	} else if targetPane == "" {
 		fmt.Printf("%s No pane to nudge (agent will discover work via gt prime)\n", style.Dim.Render("○"))
 	} else {
 		// Ensure agent is ready before nudging (prevents race condition where


boshu2:
## Review: LGTM with minor question ✅

Well-designed solution. Deferring session start until bead setup is complete is the right fix for this race.

**What I like:**
- Clean separation: `SpawnPolecatForSling` creates worktree, `StartSession()` activates it
- Idempotent `StartSession()` - safe to call multiple times
- `SessionStarted()` check is useful for callers
- Good unit test coverage

**Question:** In `StartSession()`, if `SetAgentState()` fails, it logs a warning but continues:
```go
if err := polecatMgr.SetAgentState(s.PolecatName, "working"); err != nil {
    fmt.Printf("Warning: could not update agent state: %v\n", err)
}
```

Is this the right behavior? The session has started at this point, so continuing makes sense, but curious if there's a scenario where we'd want to fail here instead.

Otherwise, looks solid and ready for merge.

boshu2:
**Correction to my previous comment:** I should clarify - the above is my analysis as a contributor, not a merge approval. That's Steve's call.

PiotrTrzpil:
Thanks for the review @boshu2!

Re: the `SetAgentState()` error handling - yes, warn-and-continue is intentional here because:

1. **The session has already started** - at this point `polecatSessMgr.Start()` succeeded. Failing now would leave us with a running session but no pane returned to the caller, which is a worse state.

2. **State is informational** - `SetAgentState("working")` updates the rig's tracking of the polecat. It's useful for monitoring (`gt polecat list`) but not required for the polecat to do its work.

3. **Graceful degradation** - the polecat will function correctly even if state tracking fails. Better to have a working polecat with stale state than to abort the whole sling operation.

The only scenario where failing might make sense is if we had a way to rollback the started session, but that would add complexity for minimal benefit.

PiotrTrzpil:
Added a second commit that removes duplicate nudges during polecat startup. While not directly related to the race condition fix, it addresses another polecat startup issue we discovered during testing: fresh polecats were receiving 3 nudges in quick succession (StartupNudge → PropulsionNudge → injectStartPrompt), which Claude interpreted as interruptions. Now they receive just one clear instruction.

Both commits improve polecat startup reliability, so bundling them seemed reasonable. Happy to split if preferred.

steveyegge:
This PR has been approved but has merge conflicts that need to be resolved. The race condition fix is especially valuable - could you please rebase on main? Thanks!

PiotrTrzpil:
Thanks @steveyegge ! Done - resolved conflicts.

PiotrTrzpil:
It still needs a look (previously it was approved but workflows needed approval to run after fixing conflicts) - would you have a moment @steveyegge ?

steveyegge:
Applied the deferred session start pattern to main in commit 1f293e57. The key changes:\n\n- `SpawnPolecatForSling()` now defers session creation\n- `StartSession()` method starts the session after bead/molecule attachment\n- `sling.go` calls `StartSession()` after setting up the hook\n- Added `SetAgentState()` to polecat manager\n- Preserved worktree verification and runtime ready wait from recent merges\n\nThe rig_lifecycle.go refactoring wasn't included (separate concern). Closing.

PR #963, opened 2026-01-25
Fix gt ready using wrong beads path for rigs

baldvin-kovacs:
## Summary

gt ready was looking for the .beads directory within the mayor's subdir, this change makes it look for it right under the rig.

## Changes
- Use r.BeadsPath() in "gt ready" implementation instead of constants.RigMayorPath(r.Path)

## Testing
  - They didn't even before.
- [X] Manual testing performed

## Checklist
- [X] Code follows project style
- [?] No breaking changes (or documented in summary)

### DIFF
--- a/internal/cmd/ready.go
+++ b/internal/cmd/ready.go
@@ -147,9 +147,9 @@ func runReady(cmd *cobra.Command, args []string) error {
 		wg.Add(1)
 		go func(r *rig.Rig) {
 			defer wg.Done()
-			// Use mayor/rig path where rig-level beads are stored
-			rigBeadsPath := constants.RigMayorPath(r.Path)
-			rigBeads := beads.New(rigBeadsPath)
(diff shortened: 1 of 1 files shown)


boshu2:
## Review: LGTM ✅

Clean, focused fix. Using `r.BeadsPath()` is the right abstraction - the rig knows where its beads live.

**Why this is correct:**
- `constants.RigMayorPath()` was looking in `mayor/rig` subdirectory
- Beads actually live at rig root (routing handles the rest)
- The rig's `BeadsPath()` method should be the source of truth

**One note:** You mentioned unit tests didn't pass before this change. Is that a known issue or something we should track? If it's test debt worth addressing, might be worth filing an issue.

Otherwise, ready to merge.

boshu2:
**Correction to my previous comment:** I should clarify - the above is my analysis as a contributor, not a merge approval. That's Steve's call.

PR #981, opened 2026-01-26
fix(shutdown): prevent self-kill during gt shutdown --all

aleiby:
## Summary

When running `gt shutdown --all` from a crew session, the command would terminate prematurely with 'Terminated' because `killSessionsInOrder()` killed the current session before cleanup completed.

## Root Cause

`killSessionsInOrder()` in start.go doesn't account for the current session. When it kills the session we're running in, the gt process dies before completing:
- Polecat cleanup
- Daemon stop
- Success message

## Fix

1. Detect current tmux session using existing `getCurrentTmuxSession()` from handoff.go
2. Add `skipSession` parameter to `killSessionsInOrder()` 
3. Defer killing own session until after all cleanup completes

This follows the same pattern established in PR #882 for `gt handoff`.

## Test Plan

- [ ] Manual test: `gt shutdown --all` from crew session completes cleanup before terminating
- [ ] Build passes
- [ ] Existing tests pass

---

### DIFF
--- a/internal/cmd/start.go
+++ b/internal/cmd/start.go
@@ -550,6 +550,17 @@ func categorizeSessions(sessions []string, mayorSession, deaconSession string) (
 func runGracefulShutdown(t *tmux.Tmux, gtSessions []string, townRoot string) error {
 	fmt.Printf("Graceful shutdown of Gas Town (waiting up to %ds)...\n\n", shutdownWait)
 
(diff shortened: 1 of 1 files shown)


aleiby:
Work already landed on main via direct push (commit 0e873e96). Closing redundant PR.

PR #977, opened 2026-01-25
fix(startup): unify agent startup with beacon + instructions in CLI prompt

aleiby:
## Summary

All agents now receive their startup beacon + role-specific instructions via the CLI prompt, making sessions identifiable in /resume picker while removing unreliable post-startup nudges.

**Key changes:**
- Rename `FormatStartupNudge` → `FormatStartupBeacon`, `StartupNudgeConfig` → `BeaconConfig`
- Remove `StartupNudge()`, `PropulsionNudge()`, and `PropulsionNudgeForRole()` functions
- Update all role managers (deacon, witness, refinery, polecat, boot) to include beacon in CLI prompt
- Update daemon/lifecycle.go and cmd/deacon.go to include beacon
- Remove redundant post-startup nudge calls from all startup paths

## GUPP Preservation

**This change does NOT remove GUPP (Gas Town Universal Propulsion Principle).** It changes the mechanism:

**Before (post-startup nudge):**
1. Start session
2. Send beacon via `StartupNudge()` (tmux send-keys)
3. Wait 2 seconds for Claude to process
4. Send `PropulsionNudge()` to trigger autonomous execution

**After (pre-startup injection via CLI prompt + SessionStart hook):**
1. CLI prompt includes beacon + instructions (queued before Claude starts)
2. SessionStart hook runs `gt prime --hook` synchronously
3. `gt prime` injects full context including "AUTONOMOUS WORK MODE" instructions
4. Claude sees everything in first turn and executes immediately

The old two-step nudge was a workaround for unreliable context injection. SessionStart hooks now handle propulsion reliably via `gt prime`, making the separate PropulsionNudge redundant.

GUPP is still enforced by:
- `gt prime` injecting "Work is on your hook... Begin IMMEDIATELY" when work is hooked
- Daemon's `checkGUPPViolations()` detecting agents with hooked work that aren't progressing
- Witness remediation of stuck agents

## Test plan
- [x] Unit tests pass for all affected packages
- [x] Build compiles without errors
- [ ] Manual test: Start deacon and verify beacon appears in prompt
- [ ] Manual test: Verify agents with hooked work start immediately

---
🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #984, opened 2026-01-26
fix(version): improve bd version check resilience during multi-agent startup

### LINKED ISSUE #503: bd version check causes contention under high concurrency

aleiby:
## Summary

Fixes "bd version check timed out" errors during multi-agent startup by:
- Using `--no-daemon` flag to avoid daemon contention (version check doesn't need DB access)
- Adding retry with backoff (3 attempts) for resilience during heavy startup load

## Related Issue

Fixes #503 (bd version check causes contention under high concurrency)

Alternative approach to PR #504 (caching) - this PR avoids the contention source entirely rather than caching the result.

## Changes

- `internal/cmd/beads_version.go`: Add retry wrapper and `--no-daemon` flag

## Testing

- Unit tests pass for version parsing/comparison
- Manual verification: `bd version --no-daemon` is same speed as with daemon (~0.15s)
- Under load, `--no-daemon` avoids daemon bottleneck entirely

---

### DIFF
--- a/internal/cmd/beads_version.go
+++ b/internal/cmd/beads_version.go
@@ -94,7 +94,9 @@ func getBeadsVersion() (string, error) {
 	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
 	defer cancel()
 
-	cmd := exec.CommandContext(ctx, "bd", "version")
+	// Use --no-daemon to avoid contention when multiple agents start simultaneously.
+	// Version check doesn't need database access, so direct mode is faster and more reliable.
+	cmd := exec.CommandContext(ctx, "bd", "version", "--no-daemon")
 	output, err := cmd.Output()
 	if err != nil {
 		if ctx.Err() == context.DeadlineExceeded {


aleiby:
Note: This improves robustness of #977 (unify agent startup) - the SessionStart hook calling `gt prime` was failing due to version check timeouts during multi-agent startup.

PR #989, opened 2026-01-26
perf(status): parallelize beads pre-fetching (~2x faster)

aleiby:
## Summary
- Parallelize rig-level beads fetching using goroutines
- Use mutex to protect shared maps during concurrent writes
- Refactor `discoverRigHooks` to use pre-fetched agent beads instead of handoff beads

## Key Refactoring: discoverRigHooks

The old `getAgentHook` function used `FindHandoffBead` to look up pinned "handoff" beads with `AttachedMolecule` in attachment fields. This is the **legacy data model**.

The new `buildHook` helper uses agent beads with the `HookBead` database column - the same model that `discoverRigAgents` already uses. This aligns with commit 278abf15 ("fix(hook): Read hook_bead from database column, not description") which established that:

> "the slot update code writes to the hook_bead database column via 'bd slot set'"

**Why this matters**: The old code could read stale handoff beads instead of the authoritative hook state from agent beads. This refactor makes `discoverRigHooks` consistent with `discoverRigAgents` and the rest of the codebase.

## Performance
- Before: ~7s with `--fast` flag
- After: ~3s with `--fast` flag
(Tested with two rigs.)

## Test Plan
- [x] All existing tests pass
- [x] Manual testing: `time gt status --fast` shows ~2x improvement
- [x] Linter passes with no issues

Fixes: gt-evb125

---

### DIFF
--- a/internal/cmd/status.go
+++ b/internal/cmd/status.go
@@ -380,12 +380,8 @@ func runStatusOnce(_ *cobra.Command, _ []string) error {
 				rs.CrewCount = len(workers)
 			}
 
-			// Discover hooks for all agents in this rig
-			// In --fast mode, skip expensive handoff bead lookups. Hook info comes from
-			// preloaded agent beads via discoverRigAgents instead.
-			if !statusFast {
-				rs.Hooks = discoverRigHooks(r, rs.Crews)
-			}
+			// Discover hooks for all agents in this rig (uses pre-fetched data for O(1) lookups)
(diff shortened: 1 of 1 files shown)


steveyegge:
This PR now conflicts with main after #1065 was merged, which also parallelizes beads pre-fetching in `gt status`. The hook lookup refactoring (using pre-fetched agent beads for `discoverRigHooks`) is still valuable and not covered by #1065, so a rebase would be worthwhile to bring in that improvement.

aleiby:
Already merged in commit d9b86dfa

PR #993, opened 2026-01-26
fix: Correct bootstrap file placement and settings.local.json migration

julianknutsen:
Let's see if work with _PR Sheriff Review_ that looks to use LIFO order when reviewing merge candidates. Move #430 to the top of the queue as it fixes a critical bug and necessary for correct operation without explicit nudges when using Claude agents now that #977 is in.

## Summary

This PR fixes critical issues with Claude Code configuration file placement that caused hooks not to fire and agents to miss project instructions:

1. **settings.json → settings.local.json**: Claude Code does NOT traverse parent directories for settings files (only for CLAUDE.md). Settings must be placed directly in each agent's working directory. We now use `settings.local.json` which is gitignored by default, preventing pollution of user repos.

2. **Bootstrap file placement**: CLAUDE.md and AGENTS.md are now placed at agent directory level (e.g., `<rig>/refinery/CLAUDE.md`) where Claude can find them via upward traversal, rather than inside git worktrees where they would pollute source repos.

3. **Documentation fixes**: Updated stale references to polecat path structure (`polecats/<name>/<rigname>/`) and other directory layouts to match actual code.

## Fixed Issues

- **Hooks not firing for polecats/crew**: Settings placed in parent directories were never found because Claude Code doesn't traverse upward for settings. Now each agent provisions `settings.local.json` in their working directory at startup.

- **Source repo pollution**: Writing to CLAUDE.md or AGENTS.md inside git worktrees would leak Gas Town internals if merged. Bootstrap files are now outside worktrees.

## Related Issues (Not Fixed)

- **#392 - Polecats/crew exclude essential instructions**: Project-specific CLAUDE.md is still excluded by sparse checkout. This is intentional (prevents Gas Town context from conflicting with project instructions), but means agents don't see project-specific instructions. A future solution might merge project instructions into `gt prime` output.

BUT, we now have a path for introducing repo-specific settings and CLAUDE.md files that still allows our hooks and instructions to run.

Fixes: #742 -- settings.json fix
Fixes: #735 

## Breaking Changes

| Before | After |
|--------|-------|
| `<rig>/polecats/.claude/settings.json` | `<rig>/polecats/<name>/<rigname>/.claude/settings.local.json` |
| `<rig>/crew/.claude/settings.json` | `<rig>/crew/<name>/.claude/settings.local.json` |
| `<rig>/refinery/.claude/settings.json` | `<rig>/refinery/rig/.claude/settings.local.json` |
[...]

PR #994, opened 2026-01-26
fix(doctor): update claude_settings_check to delete old/misplaced settings

aleiby:
## Summary

The claude_settings_check doctor check had backwards logic - it considered parent directories as CORRECT locations for settings when Claude Code actually requires settings in the working directory.

Changes:
- Detect settings in both old (settings.json) and new (settings.local.json) filenames
- Mark parent-directory settings as stale (should be deleted)
- Mark old settings.json files as stale (should be settings.local.json)
- Simplify Fix() to only delete - agent startup handles recreation
- Remove content validation (we only care about location/filename now)
- Update tests to match new behavior

### Why delete-only (no recreation)?

We considered having doctor recreate correct settings after deletion, but Claude Code only reads settings at startup - it doesn't hot-reload them during a session (see [anthropics/claude-code#5513](https://github.com/anthropics/claude-code/issues/5513) for a feature request to add `/reloadSettings`).

Recreating files would give a false sense of security since running agents wouldn't pick up the changes anyway. Instead, we:
1. Delete stale files
2. Warn that agents need restart
3. Let agent startup (EnsureSettingsForRole) create correct settings when they restart

This keeps the creation logic in one authoritative place.

## Test plan

- [x] All existing tests pass
- [x] New tests verify correct behavior for working directories vs parent directories
- [x] Manual: run `gt doctor` and verify it correctly identifies stale settings

## Related

- #993 (bootstrap files fix - this PR adds the doctor check portion)
- anthropics/claude-code#12962 (Claude Code doesn't traverse parent directories for settings)
- anthropics/claude-code#5513 (no hot-reload for settings - why we don't recreate)

---
🤖 Tackled with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/internal/doctor/claude_settings_check_test.go
+++ b/internal/doctor/claude_settings_check_test.go
@@ -1,7 +1,6 @@
 package doctor
 
 import (
-	"encoding/json"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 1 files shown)


aleiby:
Closing as obsolete. The core fix (detecting settings in wrong locations) was addressed by upstream PR #239 and subsequent changes. The current upstream `claude_settings_check.go` already has `wrongLocation` logic for parent directories. The remaining changes in this PR are just test simplifications that aren't worth maintaining separately.

PR #1033, opened 2026-01-27
fix(mail): mark messages as read when viewed via gt mail read

### LINKED ISSUE #1031: Mail not marked as read/archived after processing, causing re-processing on restart
## Summary

Agents re-process mail after restart/handoff because mail is never marked as read or archived after being acted upon. The tools exist (`gt mail mark-read`, `gt mail archive`) but nothing in the agent workflow prompts their use.

## The Gap

1. Agent receives mail (e.g., "Tackle gt-9aj5gp")
2. Agent reads it with `gt mail read <id>` — which **intentionally does not mark it as read** (per design in #936, to preserve handoff messages)
3. Agent acts on the mail (starts the work)
4. Agent never runs `gt mail mark-read` or `gt mail archive`
5. Mail stays unread
6. Next session: same mail shows as unread in inbox
7. Agent processes it again → duplicate work, confusion, or worse

## Why It Happens

- `gt mail read` is a **view** command, not an **acknowledge** command — by design
- The mail protocol doc (`docs/design/mail-protocol.md`, line 344) says agents should "Mark mail as read after processing"
- But no agent startup instructions, PRIME.md, or role context actually tells agents to do this
- The system-reminder ("You have N unread messages") keeps surfacing the same mail

## What Exists Today

| Command | Effect |
|---------|--------|
| `gt mail read <id>` | Views message, does NOT mark as read |
| `gt mail mark-read <id>` | Adds "read" label, stays in inbox |
| `gt mail archive <id>` | Closes the message |
| `gt mail delete <id>` | Removes the message |

The tools are there. The workflow just doesn't use them.

## Possible Fixes

1. **Protocol update**: Add explicit "mark-read or archive after acting on mail" step to agent startup instructions and PRIME.md
2. **Tooling**: Add an `--ack` flag to `gt mail read` that marks as read after display
3. **Convention**: Establish that `gt mail archive` is the standard "I handled this" action
4. **Automation**: Have `gt prime` or the startup hook auto-archive stale mail (the `--stale` flag in `mail_inbox.go:345-408` partially does this but isn't widely used)

## Environment

- Observed across crew workers and polecats
- Happens after any session restart, handoff, or compaction
- The `--stale` archiving mechanism exists but isn't integrated into standard startup

---
🤖 Filed with [Claude Code](https://claude.com/claude-code)

aleiby:
## Summary

`gt mail read` previously did not mark messages as read, causing agents to
re-process the same mail after restart or handoff. The original comment
cited preserving handoff messages, but handoff mail is already preserved
via the hook mechanism (`gt hook`), making inbox preservation redundant.

## Changes

- `internal/cmd/mail_inbox.go`: Replace "don't mark read" comment with
  `MarkReadOnly()` call in `runMailRead()`. Adds the "read" label without
  closing or archiving — mail stays in inbox but won't appear as unread.
- `internal/mail/mailbox_test.go`: Add test verifying `MarkReadOnly()`
  excludes messages from `ListUnread()` while keeping them in `List()`.

## Testing

Empirically validated across 3 sessions with `beads/crew/ruby`:

| Session | Without Fix | With Fix |
|---------|------------|----------|
| 1 | Read mail, replied. Mail stayed unread. | Read mail, replied. Auto-marked read. |
| 2 (restart) | Re-read same mail, sent duplicate reply. | 0 unread, no re-processing. |
| 2 (handoff) | N/A | Handoff context found via hook, work continued correctly. |
| 3 (post-handoff) | Would re-process everything. | 0 unread, clean idle. |

Fixes #1031

---

### DIFF
--- a/internal/cmd/mail_inbox.go
+++ b/internal/cmd/mail_inbox.go
@@ -149,9 +149,13 @@ func runMailRead(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("getting message: %w", err)
 	}
 
-	// Note: We intentionally do NOT mark as read/ack on read.
-	// User must explicitly delete/ack the message.
-	// This preserves handoff messages for reference.
+	// Mark as read when viewed (adds "read" label, does not close/archive).
+	// Handoff messages are preserved via the hook mechanism, so marking
+	// read here is safe — hooked mail is found via gt hook, not the inbox.
+	if err := mailbox.MarkReadOnly(msgID); err != nil {
+		// Non-fatal: message was retrieved, just couldn't mark
+		style.PrintWarning("could not mark message as read: %v", err)
+	}
 
 	// JSON output
 	if mailReadJSON {
--- a/internal/mail/mailbox_test.go
+++ b/internal/mail/mailbox_test.go
@@ -264,6 +264,56 @@ func TestMailboxLegacyListUnread(t *testing.T) {
 	}
 }
 
+func TestMailboxMarkReadOnlyExcludesFromUnread(t *testing.T) {
+	tmpDir := t.TempDir()
+	m := NewMailbox(tmpDir)
+
+	msgs := []*Message{
+		{ID: "msg-001", Read: false, Subject: "First"},
+		{ID: "msg-002", Read: false, Subject: "Second"},
+	}
+	for _, msg := range msgs {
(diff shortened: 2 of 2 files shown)


arttttt:
it was partially fixed in the PR #913 for Witness and POLECAT_DONE messages. It looks like that it needed to be taken into account

PR #417, opened 2026-01-12
fix(agents): use CLI prompt for propulsion instead of unreliable nudges

sauerdaniel:
## Summary

Fixes agent startup reliability by passing `"gt prime"` as a CLI argument to Claude instead of using post-startup nudges. The CLI prompt is queued before Claude starts, avoiding timing issues entirely.

## Problem

Post-startup nudges are unreliable in Claude Code v2.1.4+ - text arrives before input is ready, causing only Enter to make it through (empty input). This causes agents to hang or enter endless restart cycles.

## Related

- Related to #410 which uses `WaitForRuntimeReady` + nudges (different approach)
- This PR removes nudges entirely, which is more robust

## Why CLI Prompt is Superior

| Approach | Timing Issue? | Complexity |
|----------|---------------|------------|
| **Post-startup nudges** (#410) | Yes - must wait for prompt | Polling + sleeps |
| **CLI prompt** (this PR) | No - queued before start | Simple, one-line change |

The CLI argument is processed before Claude's TUI initializes, so there's no race condition.

## Changes

Affected agents:
- Deacon (`internal/deacon/manager.go`)
- Witness (`internal/witness/manager.go`)  
- Refinery (`internal/refinery/manager.go`)
- Polecat (`internal/polecat/session_manager.go`)

For each:
1. Pass `"gt prime"` as CLI prompt parameter to `BuildAgentStartupCommand*`
2. Remove post-startup nudges (StartupNudge, PropulsionNudge)
3. Remove timing sleeps that waited for input readiness
4. Keep SessionStart hook as backup propulsion mechanism

## Test Plan

- [x] `go build ./...` passes
- [x] `go test ./...` passes
- [x] Manual testing: agents start and execute gt prime immediately

Issue #866, opened 2026-01-22
Race condition: gt prime queries status=hooked before bead is hooked during spawn


julianknutsen:
## Bug Description

When slinging work to a rig (which spawns a fresh polecat), there's a race condition where `gt prime` queries for `status=hooked` before the bead is actually hooked. This causes the polecat to miss its work on startup.

The race window exists because the hook operation (`bd update --status=hooked`) ran in `sling.go` AFTER the session was started in `SpawnPolecatForSling()`.

## Steps to Reproduce

1. Sling work to a rig (spawns fresh polecat):
   ```bash
   gt sling <bead> <rig>
   ```

2. The polecat spawns and Claude immediately runs `gt prime` to check for work

3. **Race**: `gt prime` queries `bd list --status=hooked --assignee=<agent>` but the hook hasn't happened yet

4. **Bug**: `gt prime` returns no work, polecat sits idle waiting for nudge

## Expected Behavior

When a polecat spawns via `gt sling`, the work should already be hooked BEFORE the session starts. This ensures that when Claude runs `gt prime`, the hooked work is visible.

## Actual Behavior

The sequence was:
1. `SpawnPolecatForSling()` creates polecat and starts tmux session
2. Claude starts in session, runs `gt prime`
3. `gt prime` queries `status=hooked` → NOT FOUND (bead not yet hooked)
4. `sling.go` runs `bd update --status=hooked` (too late)

The polecat misses its work on the initial `gt prime` query. It may eventually discover work via:
- The nudge prompt injected into tmux
- A subsequent `gt prime` call
- Manual intervention

But the race can cause confusion, duplicate work attempts, or silent failures.

## Environment
- **OS**: Linux
- **Go version**: 1.21+
- **Gas Town version**: current main (before fix)

## Logs / Error Output

No explicit error - the bug manifests as:
- Polecat appears to ignore slung work initially
- Work is eventually picked up via nudge or later gt prime
- Possible duplicate processing if nudge and delayed gt prime both trigger

## Root Cause Analysis

The issue is the ordering of operations in the spawn + sling flow:

**Before fix:**
```
SpawnPolecatForSling():
  1. Allocate name
  2. Create/repair polecat
  3. Create agent bead (sets hook_bead field)
  4. Start session ← Claude runs gt prime here
  
sling.go (after SpawnPolecatForSling returns):
  5. bd update --status=hooked ← TOO LATE
```

**After fix:**
```
SpawnPolecatForSling():
  1. Allocate name
[...]

Issue #667, opened 2026-01-17
Boot watchdog lacks context and Deacon startup unreliable


sauerdaniel:
## Problem

Two related issues with daemon-managed agent startup:

### 1. Boot Watchdog Missing Context

When Boot spawns, it doesn't have a CLAUDE.md file with context about its role. This prevents Boot from making intelligent triage decisions when handling agent issues.

### 2. Deacon Startup Auto-execution Failure

In Claude Code v2.1.4+, the Deacon startup nudge text appears in the prompt area but doesn't execute. This causes:
- `gt prime` text visible but not running
- Endless restart loops as daemon detects unhealthy deacon

### For Boot:
- Add `CreateBootCLAUDEmd` function to templates package
- Add `EnsureCLAUDEmd` method to create context before session spawn

### For Deacon:
- Execute `gt prime` directly via SendKeys instead of nudge message
- This ensures the command actually executes rather than just appearing in the prompt

PR #1037, opened 2026-01-27
fix(session): remove redundant gt prime instruction from start/restart beacons

aleiby:
## Summary

- Remove the fallback "Run `gt prime`" instruction from start/restart beacons
- The SessionStart hook now reliably injects context via `gt prime --hook`, making this instruction redundant
- When the hook works (normal case), the beacon text causes agents to run `gt prime` a second time, wasting context window

## Historical context

This fallback was added in 605eeec8 (Jan 12) as a safety net for when the SessionStart hook failed to inject context, causing agents to sit idle. Since then, hook infrastructure has stabilized.

For non-hook agents (Codex, Cursor) that genuinely need "Run gt prime", PR #1029 introduces a capability-aware fallback matrix (`IncludePrimeInstruction` field) that only targets agents without hooks. This change removes the blunt fallback; #1029 adds the surgical replacement. The two are complementary and merge-order independent.

## Test plan

- [x] Updated startup_test.go with negative assertions (start/restart beacons must NOT contain "gt prime")
- [x] All session package tests pass
- [x] Change isolated to 2 files (startup.go, startup_test.go)

---

### DIFF
--- a/internal/session/startup.go
+++ b/internal/session/startup.go
@@ -70,14 +70,6 @@ func FormatStartupBeacon(cfg BeaconConfig) string {
 		beacon += "\n\nWork is on your hook. Run `gt hook` now and begin immediately."
 	}
 
-	// For start/restart, add fallback instructions in case SessionStart hook fails
-	// to inject context via gt prime. This prevents the "No recent activity" state
-	// where agents sit idle because they received only metadata, no instructions.
-	// See: gt-uoc64 (crew workers starting without proper context injection)
-	if cfg.Topic == "start" || cfg.Topic == "restart" {
-		beacon += "\n\nRun `gt prime` now for full context, then check your hook and mail."
-	}
-
 	return beacon
 }
 
--- a/internal/session/startup_test.go
+++ b/internal/session/startup_test.go
@@ -90,7 +90,7 @@ func TestFormatStartupBeacon(t *testing.T) {
 			},
 		},
 		{
-			name: "start includes fallback instructions",
+			name: "start beacon has no prime instruction",
 			cfg: BeaconConfig{
 				Recipient: "beads/crew/fang",
 				Sender:    "human",
@@ -101,11 +101,13 @@ func TestFormatStartupBeacon(t *testing.T) {
 				"beads/crew/fang",
 				"<- human",
 				"start",
-				"gt prime", // fallback instruction for when SessionStart hook fails
+			},
+			wantNot: []string{
+				"gt prime",
 			},
 		},
 		{
-			name: "restart includes fallback instructions",
+			name: "restart beacon has no prime instruction",
 			cfg: BeaconConfig{
 				Recipient: "gastown/crew/george",
 				Sender:    "human",
@@ -115,7 +117,9 @@ func TestFormatStartupBeacon(t *testing.T) {
 				"[GAS TOWN]",
 				"gastown/crew/george",
 				"restart",
-				"gt prime", // fallback instruction for when SessionStart hook fails
+			},
+			wantNot: []string{
+				"gt prime",
 			},
 		},
 	}


arttttt:
Does gt hook inject full context? Personally, I often use not as advanced models as Opus or GPT 5.2, etc(GLM 4.7 for example). The problem is that gt hook is not enough for GLM to start working properly and sometimes I manually call gt prime. I mean that maybe we should extent gt hook if it does not provide full context atm

aleiby:
> Does gt hook inject full context? Personally, I often use not as advanced models as Opus or GPT 5.2, etc(GLM 4.7 for example). The problem is that gt hook is not enough for GLM to start working properly and sometimes I manually call gt prime. I mean that maybe we should extent gt hook if it does not provide full context atm

https://github.com/steveyegge/gastown/pull/977 ensures `gt prime` is injected (already merged but relies on SessionStart hook support and CLI injection).

https://github.com/steveyegge/gastown/pull/1029 (still open) adds support for other agent situations.  See details in https://github.com/steveyegge/gastown/issues/978.

The startup nudge in 1029 says, "Check your hook with `gt hook`. If work is present, begin immediately."

I'm pretty sure the "then check your mail" part is also covered somewhere, but would be good to verify once the dust settles.

arttttt:
I guess that it would be better to create a common startup and context providing system rather than trying to remember what which platform can do or cannot. We can try using initial nudges, because all coding agent apps support queues.

Example:

gt sling -> nudge gt prime, delay, nudge gt hook

aleiby:
> I guess that it would be better to create a common startup and context providing system rather than trying to remember what which platform can do or cannot. We can try using initial nudges, because all coding agent apps support queues.
> 
> Example:
> 
> gt sling -> nudge gt prime, delay, nudge gt hook

I believe that was the original solution and proved unreliable.  So the options are:
1. Make all agents unreliable
2. Make some agents reliable, and have a fallback for agents that do not support the reliable method(s).

Then as agents add features like SessionStart hooks, they can take advantage of the reliable solution.

arttttt:
> I believe that was the original solution and proved unreliable

can not agree, in my opinion the only reason was because claude supports it(originally it was a claude code only solution) and you don't need to compile a whole project in case you need to test smth

aleiby:
Thanks for the perspective. I think the evidence in the issue tracker actually does support the unreliability claim pretty strongly — there were ~48 related issues/PRs filed over a 25-day period (Jan 2-27) documenting systematic startup failures rooted in nudge-based propulsion. Here are the highlights:

**Stampeding nudges:**
- #571: Concurrent `gt nudge deacon session-started` calls during `gt up` interleave in the tmux input buffer, causing garbled input
- #952: Fresh polecats receive multiple nudges in rapid succession (StartupNudge → PropulsionNudge → injectStartPrompt), which Claude interprets as interruptions

**Cancelled tasks:**
- #470: Witness nudges arrive during Claude's API initialization, triggering the "Interrupted" prompt
- #787: `NudgeSession` sends ESC before Enter (for vim mode), interrupting Claude during startup
- #319: Messages arrive in wrong order, burning 48k+ tokens on exploratory work

**Flaky startup / race conditions:**
- #417/#418: Text arrives before input is ready in Claude Code v2.1.4+; only Enter makes it through, causing hangs or restart loops
- #661: Comprehensive filing documenting timing issues, lost nudges, no delivery confirmation, and race conditions
- #667: Deacon nudge text visible in prompt but doesn't execute, causing endless restart loops
- #866: Race condition where `gt prime` queries `status=hooked` before the bead is actually hooked
- #660: Multiple components try to start Deacon simultaneously; session appears running but isn't initialized

The fix attempts escalated through increasingly targeted workarounds — `WaitForRuntimeReady` (#410) → serialized nudges (#571) → ESC key removal (#787) → beacon formatting (#617) → sleep-to-polling (#431) — before #977 addressed the root cause by moving to CLI prompt injection + SessionStart hooks.

You're right that Claude Code hook support was a prerequisite that made the cleaner approach possible, but the motivation was the documented unreliability of the nudge-based approach, not just convenience.

arttttt:
The 48 issues cited document **implementation bugs**, not fundamental architectural flaws. The current `NudgeSession()` implementation in `internal/tmux/tmux.go` already includes fixes for many of them:

- Per-session mutex serialization (fixes #571 interleaving)
- 500ms empirically-tested debounce (fixes paste timing)
- Escape before Enter (fixes #787 vim mode)
- SIGWINCH wake for detached sessions (fixes #417)
- Retry logic with 3 attempts

The hook-based approach creates a maintenance burden: each new agent requires its own capability detection and fallback logic (see `IncludePrimeInstruction` matrix in #1029). As new coding agents emerge (Cursor, Codex, etc.), this means N implementations for N agents.

Nudges work at the tmux transport layer — one implementation can cover all agents that run in tmux sessions. This is a centralized solution vs. per-agent fragmentation.

If there are remaining reliability issues with nudges, they can be fixed in one place. With hooks, every new agent requires separate integration work.

PR #1045, opened 2026-01-27
fix(git): preserve symlinks when copying directories

### LINKED ISSUE #1044: copyDir fails with symlinks pointing to directories
## Description

`copyDir` (used by `moveDir` for cross-filesystem moves during clone operations) fails with the error:

```
copy_file_range: is a directory
```

when the source directory contains symlinks that point to other directories.

## Root Cause

In `copyDir`, the code checks `entry.IsDir()` to decide whether to recurse or call `copyFile`. However, `entry.IsDir()` returns `false` for symlinks, so symlinks fall into the `else` branch and `copyFile` is called.

`copyFile` uses `os.Open(src)` which **follows** the symlink. When the symlink target is a directory, `io.Copy` (which uses `copy_file_range` on Linux) fails because that syscall doesn't work on directory file descriptors.

## Reproduction

Any repository with symlinks pointing to directories will trigger this. For example:

```
rust_crates/proto/proto_arrow_types/proto -> ../proto_definitions/proto
```

When `gt rig add` clones this repo and the clone needs to be moved cross-filesystem (temp dir to final location), `moveDir` falls back to `copyDir` and fails.

## Fix

Check `entry.Type()&os.ModeSymlink` before `entry.IsDir()`, and recreate symlinks using `os.Readlink` + `os.Symlink` rather than following them.

See PR #XXX for the fix with tests.

markjen:
## Summary

Fixes #1044

- Fix `copyDir` to preserve symlinks rather than following them
- Add tests for symlink handling (both symlinks to directories and files)

## Problem

`copyDir` was failing with `copy_file_range: is a directory` when copying directories containing symlinks that point to other directories.

The bug: `entry.IsDir()` returns `false` for symlinks, so they fell into the `else` branch and `copyFile` was called. `copyFile` uses `os.Open` which follows the symlink, and `io.Copy` (which uses `copy_file_range` on Linux) fails when the file descriptor points to a directory.

## Solution

Check `entry.Type()&os.ModeSymlink` before `entry.IsDir()`, and recreate symlinks using `os.Readlink` + `os.Symlink` rather than following them.

## Test plan

- [x] Added `TestCopyDirWithSymlinks` - verifies symlinks to directories are preserved
- [x] Added `TestCopyDirWithSymlinkToFile` - verifies symlinks to files are preserved
- [x] All existing git tests pass

### DIFF
--- a/internal/git/git.go
+++ b/internal/git/git.go
@@ -54,6 +54,7 @@ func moveDir(src, dest string) error {
 }
 
 // copyDir recursively copies a directory from src to dest.
+// Symlinks are preserved (recreated with the same target) rather than followed.
 func copyDir(src, dest string) error {
 	srcInfo, err := os.Stat(src)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


karl82:
👋 I ran into the same issue and have slightly different approach. Check #1057. I think handling all different cases for files, directories and other nuances on file system, let's use `cp` or `robocopy` to do all dirty work.

markjen:
closing PR in favor of #1057

Issue #470, opened 2026-01-13
Bug: Polecat startup race condition causes Interrupted prompt


moudy:
# Bug Report: Polecat startup race condition causes Interrupted prompt

## Summary

Polecats spawned via `gt sling` consistently hit the Claude Code "Interrupted" prompt immediately after startup, preventing autonomous work execution. This is caused by witness nudges arriving during Claude Code's API initialization phase.

## Environment

- **Gas Town version**: Current (as of 2026-01-13)
- **Claude Code version**: 2.1.6
- **OS**: macOS Darwin 25.2.0
- **Rig**: web

## Reproduction Steps

1. Start witness for a rig: `tmux attach -t gt-web-witness`
2. Spawn a polecat with work: `gt sling hq-xyz web`
3. Attach to polecat session: `tmux attach -t gt-web-<polecat>`
4. Observe: Session shows "Interrupted · What should Claude do instead?" prompt
5. Check hook status: `gt hook show web/<polecat>` → empty

### Timeline from Debug Logs

From Claude Code debug log (`~/.claude/debug/*.txt`) for a typical polecat spawn:

```
17:12:34.377 [DEBUG] Watching for changes in setting files
17:12:34.628 [DEBUG] Getting matching hook commands for SessionStart
17:12:34.629 [DEBUG] Found 0 hook matchers in settings
17:12:35.162 [DEBUG] [API:request] Creating client
17:12:46.338 [DEBUG] Stream started - received first chunk
17:12:48.396 [DEBUG] Streaming aborted by user: Request was aborted
17:12:48.396 [ERROR] Error in non-streaming fallback: Request was aborted
```

**Key observations:**
- **17:12:34**: Claude Code starts, begins initialization
- **17:12:46**: API stream starts (SessionStart processing)
- **17:12:48**: Stream aborted "by user" (actually by witness nudge)
- **Duration**: Only **2 seconds** between API stream start and abort

### The Race Condition

When `gt sling` spawns a polecat, this sequence occurs:

1. `gt sling` creates worktree and starts tmux session with Claude Code
2. `gt sling` sends "▶ Start prompt sent" nudge immediately
3. Witness detects new polecat session
4. Witness sends "Work is on your hook. Run `gt hook` now..." nudge
5. **Race**: These nudges arrive while Claude Code is still initializing
6. Nudge triggers `UserPromptSubmit` event → aborts ongoing API stream
7. Result: "Interrupted" prompt instead of autonomous execution

### Why This Happens

[...]

PR #1076, opened 2026-01-29
fix: Write polecat settings to individual home directories

### LINKED ISSUE #742: Polecat sessions fail: settings.json written to wrong directory
## Problem

Polecat sessions die during startup because `.claude/settings.json` is written to the wrong directory.

## Current behavior

In `internal/polecat/session_manager.go`:
```go
// Ensure runtime settings exist in polecats/ (not polecats/<name>/) so we don't
// write into the source repo. Runtime walks up the tree to find settings.
polecatsDir := filepath.Join(m.rig.Path, "polecats")
runtime.EnsureSettingsForRole(polecatsDir, "polecat", runtimeConfig)
```

Settings end up at: `polecats/.claude/settings.json`
Claude starts in: `polecats/<name>/<rigname>/`

## Root cause

The comment says "Runtime walks up the tree to find settings" but Claude Code does NOT walk up parent directories looking for `.claude/settings.json`. It only looks in the current working directory.

## Impact

- Polecat sessions spawn, Claude starts, but no hooks fire
- Session times out waiting for start prompt
- All polecats fail to start

## Suggested fix

Write settings to the worktree directory (clonePath) instead of polecatsDir:

```go
workDir := opts.WorkDir
if workDir == "" {
    workDir = m.clonePath(polecat)
}

// Ensure settings in the actual working directory where Claude will run
if err := runtime.EnsureSettingsForRole(workDir, "polecat", runtimeConfig); err != nil {
    return fmt.Errorf("ensuring runtime settings: %w", err)
}
```

## Workaround

Manually copy/symlink `.claude/` into polecat worktrees, or route work to witnesses instead.

ztbrown:
## Summary
- Settings.json is now written to each polecat's home directory (`polecats/<name>/`) instead of the shared `polecats/` directory
- Provides isolated settings per polecat while keeping them out of the git worktree
- Runtime's parent-directory walk still finds settings when searching up from workDir

## Test plan
- [ ] Create multiple polecats and verify each gets its own settings.json in `polecats/<name>/`
- [ ] Verify settings.json is not written into the git worktree (`polecats/<name>/<rigname>/`)
- [ ] Verify runtime can find settings by walking up from workDir

Fixes #742

### DIFF
--- a/internal/polecat/session_manager.go
+++ b/internal/polecat/session_manager.go
@@ -173,10 +173,12 @@ func (m *SessionManager) Start(polecat string, opts SessionStartOptions) error {
 
 	runtimeConfig := config.LoadRuntimeConfig(m.rig.Path)
 
-	// Ensure runtime settings exist in polecats/ (not polecats/<name>/) so we don't
-	// write into the source repo. Runtime walks up the tree to find settings.
-	polecatsDir := filepath.Join(m.rig.Path, "polecats")
-	if err := runtime.EnsureSettingsForRole(polecatsDir, "polecat", runtimeConfig); err != nil {
+	// Ensure runtime settings exist in polecat's home directory (polecats/<name>/).
+	// This keeps settings out of the git worktree while allowing runtime to find them
+	// when walking up the tree from workDir (polecats/<name>/<rigname>/).
+	// Each polecat gets isolated settings rather than sharing a single settings file.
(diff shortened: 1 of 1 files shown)


ztbrown:
Superseded by #1077 which includes this fix along with the startup race condition fix (#470)

PR #1088, opened 2026-01-29
fix(sling): pass GT_ROOT to bd cook for formula path resolution

jpogah:
## Summary
- Fix formula path resolution when slinging work to polecats
- Pass `GT_ROOT` environment variable to `bd cook` so it can locate town-level formulas at `~/gt/.beads/formulas/`
- Update `CookFormula` function signature to accept `townRoot` parameter

## Problem
The `bd cook` command was looking for formulas in the rig directory instead of the town-level formulas directory, causing errors like:
```
Error: parsing formula: read mol-polecat-work: open /Users/.../gt/myproject/mol-polecat-work: no such file or directory
```

## Test plan
- [x] `TestCookFormula` passes
- [x] `TestInstantiateFormulaOnBead` passes
- [x] All sling-related tests pass

### DIFF
--- a/internal/formula/formulas/mol-convoy-cleanup.formula.toml
+++ b/internal/formula/formulas/mol-convoy-cleanup.formula.toml
@@ -135,7 +135,12 @@ bd show {{convoy}}
 # Status should reflect archived state
 ```
 
-**Exit criteria:** Convoy archived."""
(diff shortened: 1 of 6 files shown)


steveyegge:

PR #1094, opened 2026-01-29
fix(rig): zero struct before unmarshaling in unsetNestedValue

### LINKED ISSUE #1007: Fix unsetNestedValue bug in gt rig settings unset command
## Description

The `unsetNestedValue` function in `internal/cmd/rig_settings.go` has a bug where it deletes keys from the intermediate map but the struct field retains the value after unmarshaling back.

When running `gt rig settings unset <rig> <key-path>`, the key is deleted from the intermediate map but when the map is unmarshaled back into the RigSettings struct, the struct field still contains the old value. This causes the key to persist in the saved JSON file.

## Steps to Reproduce

1. Set a value: `gt rig settings set testrig agent claude`
2. Unset the value: `gt rig settings unset testrig agent`
3. Check settings file: `gt rig settings show testrig`
4. Observe that the `agent` key is still present in the JSON

## Expected Behavior

The key should be removed from both the struct and the JSON file. When a key is unset, it should not appear in the saved settings file.

## Actual Behavior

The key remains in the JSON file after unsetting, indicating the struct field was not properly cleared during the unmarshal operation.

## Location

`internal/cmd/rig_settings.go` - `unsetNestedValue` function (lines 299-351)

## Test Coverage

Tests are written but currently skipped in `rig_settings_test.go`:
- `TestRigSettingsUnset/unsets_top-level_keys`
- `TestRigSettingsUnset/unsets_nested_keys`

These tests can be re-enabled once the bug is fixed.

## Additional Context

The issue appears to be in how the modified map is unmarshaled back into the struct. The deletion from the map works correctly, but the struct field retains its value after the unmarshal operation.

ztbrown:
## Summary
- Fixes bug in `unsetNestedValue` where deleted keys persist after unsetting
- Uses `reflect.Zero()` to clear the struct before `json.Unmarshal`, ensuring deleted keys are removed
- Re-enables previously skipped unset tests with full coverage

## Details
The bug occurred because Go's `json.Unmarshal` only updates struct fields that are present in the JSON data - it doesn't clear fields that are missing. When `unsetNestedValue` deleted a key from the intermediate map and marshaled back to JSON, the subsequent unmarshal into the original struct would preserve the old field value.

The fix adds a reflection-based reset of the struct to its zero value before unmarshaling, ensuring the struct accurately reflects the modified JSON data.

## Test plan
- [x] Run `go test ./internal/cmd/... -run TestRigSettingsUnset` - all tests pass
- [x] Run `go test ./internal/cmd/... -run TestRigSettings` - full test suite passes
- [x] Verify `TestRigSettingsUnset/unsets_top-level_keys` passes (previously skipped)
- [x] Verify `TestRigSettingsUnset/unsets_nested_keys` passes (previously skipped)

Fixes #1007

### DIFF
--- a/internal/cmd/rig_settings.go
+++ b/internal/cmd/rig_settings.go
@@ -1,0 +1,410 @@
+// Package cmd provides CLI commands for the gt tool.
+// This file implements the gt rig settings commands for viewing and manipulating
+// rig settings/config.json files.
+package cmd
+
+import (
+	"encoding/json"
+	"errors"
+	"fmt"
+	"path/filepath"
+	"reflect"
+	"strconv"
+	"strings"
+
+	"github.com/spf13/cobra"
+	"github.com/steveyegge/gastown/internal/config"
+	"github.com/steveyegge/gastown/internal/style"
+)
+
+var rigSettingsCmd = &cobra.Command{
+	Use:   "settings",
+	Short: "View and manage rig settings",
+	Long: `View and manage rig settings (settings/config.json).
+
+Rig settings control behavioral configuration for a rig:
+- Agent selection and overrides
+- Merge queue settings
+- Theme configuration
+- Namepool settings
+- Crew startup settings
+- Workflow settings
+
+Settings are stored in settings/config.json within each rig directory.
+Use dot notation to access nested keys (e.g., role_agents.witness).`,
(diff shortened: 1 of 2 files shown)


ztbrown:
Closing - will resubmit manually after review

PR #1096, opened 2026-01-29
fix(agents): fall back to list view when tmux menu unsupported

ztbrown:
## Summary
- Add terminal capability detection for `gt agents` command
- Automatically fall back to list view when running outside tmux or in macOS Terminal.app
- Show helpful message explaining why list view is being used

## Problem
`gt agents` uses `tmux display-menu` which fails silently on macOS Terminal.app - the command runs but nothing appears on screen. Users see the command return immediately with no output.

## Solution
Added `supportsDisplayMenu()` function that checks:
1. If we're inside tmux (required for display-menu to work)
2. If TERM_PROGRAM is "Apple_Terminal" (known to not support tmux popup/menu overlays)

When display-menu won't work, the command falls back to `gt agents list` with a message explaining why and suggesting alternatives.

## Test plan
- [x] Tests pass: `go test ./...`
- [x] Build succeeds: `go build ./...`
- [ ] Manual test on macOS Terminal.app - should show list with explanation
- [ ] Manual test on iTerm2/other terminals - should show interactive menu

Fixes: gt-ge1

### DIFF
--- a/internal/cmd/agents.go
+++ b/internal/cmd/agents.go
@@ -284,6 +284,25 @@ func shortcutKey(index int) string {
 	return ""
 }
 
+// supportsDisplayMenu checks if the current terminal environment supports tmux display-menu.
+// Returns false for terminals known to not support popup menus (e.g., macOS Terminal.app)
+// or when not running inside a tmux session.
+func supportsDisplayMenu() bool {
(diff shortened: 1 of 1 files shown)


ztbrown:
Closing - will resubmit manually after review

PR #1110, opened 2026-01-30
fix(ci): use make build for integration tests

ztbrown:
## Summary
Fix integration tests failing with:
```
ERROR: This binary was built with 'go build' directly.
Use 'make build' to create a properly signed binary.
```

## Root Cause
The integration tests invoke the `gt` binary, which checks `BuiltProperly != ""` before running. This variable is only set when building via `make build` (through ldflags).

The CI was using `go build -v -o gt ./cmd/gt` which leaves `BuiltProperly` empty.

## Fix
Change from `go build` to `make build` which:
- Sets `BuiltProperly=1` via ldflags
- Sets Version, Commit, BuildTime metadata  
- Signs the binary on macOS

### DIFF
--- a/internal/cmd/test_helpers_test.go
+++ b/internal/cmd/test_helpers_test.go
@@ -50,8 +50,9 @@ func buildGT(t *testing.T) string {
 		binaryName += ".exe"
 	}
 	tmpBinary := filepath.Join(tmpDir, binaryName)
-	ldFlags := "-X github.com/steveyegge/gastown/internal/cmd.BuiltProperly=1"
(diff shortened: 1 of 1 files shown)


ztbrown:
Ok, so integration tests aren't passing but they are no longer failing because of the `go build` vs `make build` reason.

steveyegge:
This PR has merge conflicts with recently merged changes (#1164 which overlaps in test_helpers_test.go and ci.yml). Please rebase against main.

ztbrown:
@steveyegge merge conflicts fixed

PR #1144, opened 2026-01-31
fix(polecat): clean up polecatDir when worktree creation fails

flipch:
## Summary

- Fix partial state bug where `AddWithOptions` creates `polecatDir` but fails to create the worktree
- Added `cleanupOnError` function that removes `polecatDir` when worktree creation fails
- Ensures atomic behavior: polecat is fully created or nothing is left behind

## Problem

`gt sling` spawns polecats but the git worktree is not actually created in some error cases. The session starts but the working directory doesn't exist. This blocks polecat work dispatch.

Root cause: `AddWithOptions` creates `polecatDir` (`polecats/<name>/`) first, then tries to create the worktree (`polecats/<name>/<rigname>/`). If any step fails after creating `polecatDir` but before creating the worktree:
1. `exists(name)` returns `true` (it only checks `polecatDir`)
2. `clonePath(name)` returns the worktree path that doesn't exist
3. Session is started with a non-existent working directory

## Solution

Add a `cleanupOnError` function that removes `polecatDir` when:
- `repoBase()` fails (no `.repo.git` or `mayor/rig`)
- `WorktreeAddFromRef()` fails (git worktree creation error)

This ensures `exists()` returns `false` for incomplete polecats.

## Test plan

- [x] Added `TestAddWithOptions_CleansUpOnRepoBaseError` - verifies polecatDir cleanup
- [x] All existing tests pass (`go test ./internal/polecat/...`)

Fixes: br-w2ee9

### DIFF
--- a/internal/polecat/manager.go
+++ b/internal/polecat/manager.go
@@ -363,9 +363,18 @@ func (m *Manager) AddWithOptions(name string, opts AddOptions) (*Polecat, error)
 		return nil, fmt.Errorf("creating polecat dir: %w", err)
 	}
 
+	// cleanupOnError removes polecatDir if worktree creation fails.
+	// This ensures exists() returns false for incomplete polecats, preventing
+	// a partial state where polecatDir exists but the worktree doesn't.
+	// See: br-w2ee9 (worktrees not being created)
+	cleanupOnError := func() {
+		_ = os.RemoveAll(polecatDir)
+	}
+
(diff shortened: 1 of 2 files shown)


flipch:
Hey @steveyegge just wanted to let you know that this was actually gastown itself making this PR. I was very confused last night when I saw I had made this PR.

It was one of the polecats 😂

PR #1155, opened 2026-02-01
fix(sling): route standalone formula wisps to target rig beads

joshuavial:
## Summary
- Fix `gt sling <formula> <rig>` to create wisps in target rig's beads instead of HQ beads
- Route `bd cook` and `bd mol wisp` commands to the correct working directory

## Problem
When running `gt sling sync-upstream gastown`, the wisp was created in `~/gt/.beads/` (HQ) instead of `~/gt/gastown/.beads/` (target rig).

The root cause was that `runSlingFormula` ran `bd cook` and `bd mol wisp` without setting `cmd.Dir`, so they defaulted to HQ beads via the current working directory.

## Solution
Track the target's working directory in `formulaWorkDir` and set it on both `cookCmd.Dir` and `wispCmd.Dir`. This routes the commands to the correct rig's beads based on the target type:
- Rig target: Use `spawnInfo.ClonePath`
- Existing agent: Use resolved target working directory
- Self: Use self working directory
- Fall back to `townRoot` for HQ beads if no specific directory

## Test Plan
- Unit tests pass (`go test ./internal/cmd/...`)
- Manual test: `gt sling sync-upstream gastown` creates wisp in gastown beads

Note: Two pre-existing test failures unrelated to this change:
- `TestCrossPlatformBuild`: Windows build fails due to Unix-only `syscall.Kill`
- `TestParallelReadySteps`: Formula was simplified to linear but test expects parallel steps

### DIFF
--- a/internal/cmd/sling_formula.go
+++ b/internal/cmd/sling_formula.go
@@ -91,11 +91,12 @@ func runSlingFormula(args []string) error {
 	var targetAgent string
 	var targetPane string
 	var delayedDogInfo *DogDispatchInfo // For delayed session start after hook is set
+	var formulaWorkDir string            // Working directory for bd cook/wisp (routes to correct rig beads)
 
 	if target != "" {
 		// Resolve "." to current agent identity (like git's "." meaning current directory)
 		if target == "." {
-			targetAgent, targetPane, _, err = resolveSelfTarget()
+			targetAgent, targetPane, formulaWorkDir, err = resolveSelfTarget()
 			if err != nil {
 				return fmt.Errorf("resolving self for '.' target: %w", err)
 			}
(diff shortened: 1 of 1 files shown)


joshuavial:
CI failures are pre-existing issues in the codebase, not related to this fix:

**Lint (10 issues):**
- `errcheck` errors in formula.go, dolt.go, krc.go (unchecked return values)
- `gosec` warnings in krc.go (unhandled errors)  
- `unparam` in rig_settings.go, dolt.go, mail/router.go

**Integration Tests:**
- Binary built without signature: "This binary was built with 'go build' directly. Use 'make build'"

**Windows Build:**
- `syscall.Kill` is undefined on Windows (Unix-only syscall)

None of these files were modified by this PR. The fix only touches `internal/cmd/sling_formula.go`.

PR #1157, opened 2026-02-01
fix(tmux): detect Claude when pane command is version string

### LINKED ISSUE #926: gt mayor attach kills live Claude session due to version-number process name mismatch
### Summary

`gt mayor attach` incorrectly detects a running Claude session as dead and restarts it. The native Claude Code installer names binaries by version number (e.g., `2.1.19`), which tmux reports as the pane command. While `IsClaudeRunning` correctly handles version patterns (#273), `runMayorAttach` uses `IsAgentRunning` instead, which only does exact matching against `["node", "claude"]`.

### Environment

- gastown v0.5.0
- Claude Code 2.1.19 (native installer)
- macOS

### Reproduction

1. Start mayor: `gt mayor start`
2. Attach and verify working: `gt mayor attach`, then detach with Ctrl-B D
3. Check tmux reports version as command:
   ```
   $ tmux list-panes -t hq-mayor -F '#{pane_current_command}'
   2.1.19
   ```
4. Run `gt mayor attach` again
5. **Expected:** Attaches to existing session
6. **Actual:** Prints "Runtime exited, restarting with context..." and kills/restarts Claude

### Root Cause

In `internal/cmd/mayor.go`, `runMayorAttach` calls `IsAgentRunning` with `ExpectedPaneCommands`, which returns `["node", "claude"]`. The native installer binary is named `2.1.19` (symlinked from `~/.local/bin/claude`), so tmux reports `2.1.19` as the pane command. `IsAgentRunning` does exact string matching, finds no match, and triggers a restart.

`IsClaudeRunning` (used in `Start()` for zombie detection) correctly handles version patterns. The fix from #273 was applied there but not to the attach path.

### Suggested Fix

Have `IsAgentRunning` delegate to `IsClaudeRunning` when Claude is in the expected commands list. This keeps Claude-specific detection logic in one place and fixes the issue for all callers.

### Workaround

Use direct tmux attach instead of `gt mayor attach`:
```
tmux attach -t hq-mayor
```

joshuavial:
## Summary
- Fix zombie detection killing active Claude sessions
- Handle Claude Code's argv[0] being version number (e.g., "2.1.29")

## Problem
`IsRuntimeRunning()` checks if Claude is running by:
1. Direct pane command match against ["node", "claude"]
2. Child process check only when pane command is a shell

Claude Code's native installer sets argv[0] to the version number, so tmux reports "2.1.29" as the pane command. This doesn't match "node" or "claude", and isn't a shell, so the child check is skipped. Sessions are incorrectly marked as zombies and killed.

## Solution
1. Add `hasDescendantWithNames()` for recursive process tree traversal
2. Check descendants as fallback for any unrecognized pane command

## Test Plan
- Unit tests pass (TestIsRuntimeRunning, TestIsRuntimeRunning_ShellWithNodeChild)
- Manual: Verified Claude sessions no longer killed by gt doctor

Fixes #926

### DIFF
--- a/internal/tmux/tmux.go
+++ b/internal/tmux/tmux.go
@@ -1182,18 +1182,23 @@ func (t *Tmux) IsRuntimeRunning(session string, processNames []string) bool {
 			return true
 		}
 	}
-	// If pane command is a shell, check for child processes.
-	// This handles agents started with "bash -c 'export ... && agent ...'"
+	// Check for child processes if pane command is a shell or unrecognized.
+	// This handles:
+	// - Agents started with "bash -c 'export ... && agent ...'"
+	// - Claude Code showing version as argv[0] (e.g., "2.1.29")
+	pid, err := t.GetPanePID(session)
+	if err != nil || pid == "" {
+		return false
+	}
+	// If pane command is a shell, check descendants
 	for _, shell := range constants.SupportedShells {
 		if cmd == shell {
-			pid, err := t.GetPanePID(session)
-			if err == nil && pid != "" {
-				return hasChildWithNames(pid, processNames)
-			}
-			break
+			return hasDescendantWithNames(pid, processNames, 0)
 		}
 	}
-	return false
+	// If pane command is unrecognized (not in processNames, not a shell),
+	// still check descendants as fallback. This handles version-as-argv[0].
+	return hasDescendantWithNames(pid, processNames, 0)
 }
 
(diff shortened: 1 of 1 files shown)


gmorpheme:
Still broken for me on main.

joshuavial:
> Still broken for me on main.

I've been noticing issues too, not as bad as it was but pretty frustrating when it happens. haven't spotted the root cause yet

PR #1166, opened 2026-02-02
improving agent instructions

MovGP0:
## Summary
Improving agent instructions by the following:
1. instructions to develop code that works on Unix, Linux and Windows
2. instructions to validate work by executing linting and tests
3. instructions for keeping a record of lessons learned

## Changes

- Update of `AGENTS.md` file

## Testing

-✅ Unit tests pass (`go test ./...`)
- [n/a] Manual testing performed

## Checklist
- [n/a] Code follows project style
- ✅ Documentation updated (if applicable)
- ✅ No breaking changes (or documented in summary)

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -7,3 +7,17 @@ This file exists for compatibility with tools that look for AGENTS.md.
 > **Recovery**: Run `gt prime` after compaction, clear, or new session
 
 Full context is injected by `gt prime` at session start.
+
+**Platforms**: Always write code that works on Unix (MacOS), Linux, and Windows.
+
+**Validation**: Before closing the bd ticket, always check that the code works using the following commands:
+
+```bash
+golangci-lint run ./...
+go test -v -timeout=4m ./...
+go test -v -tags=integration -timeout=4m ./internal/cmd/...
+```
+
+When you fix errors, add a note in the "Lessons Learned" section of the `AGENTS.md` file how to prevent the problem, in order to ensure we do not repeat the same mistake
+
+## Lessons Learned


MovGP0:
**Note:**
- the "Lessons Learned" section could also be implemented as a separate file

steveyegge:
Thanks for the contribution. A few concerns with this PR:

1. Typo: "Plattforms" should be "Platforms"
2. The code block syntax uses single backticks instead of triple backticks
3. Gas Town agents get their context via `gt prime`, not AGENTS.md - the file explicitly says "Full context is injected by gt prime at session start." Adding detailed instructions here could cause conflicts with the prime-injected context.
4. The "Lessons Learned" section would grow unbounded and doesn't fit the project's context injection model.

Consider instead contributing these improvements to the role templates in `internal/templates/roles/` where they'd be properly scoped per role.

MovGP0:
> Typo: "Plattforms" should be "Platforms"

✅ Problem when having english as a 3rd language. Have fixed it now.

> single backticks instead of triple backticks

✅ Fixed

> Gas Town agents get their context via `gt prime`

❓ Where would i best place that?

> The "Lessons Learned" section would grow unbounded and doesn't fit the project's context injection model.

At the moment I do not have a better solution 💩

In my projects the `Lessons Learned` section is something that works really well, since most of the errors the Agents make are the same all-over-again ([Groundhog Day](https://www.imdb.com/title/tt0107048/) and [50 first Dates](https://www.imdb.com/title/tt0343660/) issues). My experience is that the section is smaller for the context window than having lots of repeated errors. But this may not apply to other projects.

If you have a better alternative, I'm all ears.

steveyegge:
Thank you for the contribution\! Closing as this conflicts with the existing gt prime startup protocol and adds Windows-specific requirements that are outside our current target platform scope. The session close protocol is already covered by the Landing the Plane section in our worker context. Appreciate the thought though\!

Issue #925, opened 2026-01-24
Mayor ignore --agent flag


lexasub:
## Steps to Reproduce
1. gt config agent set opencode "opencode --model llama.cpp/llama.cpp.109"
Agent 'opencode' set to: opencode --model llama.cpp/llama.cpp.109

2.  gt mayor attach --agent opencode
Mayor session not running, starting...
can't find session: hq-mayor
Error: exit status 1
Usage:
  gt mayor attach [flags]

Aliases:
  attach, at

Flags:
      --agent string   Agent alias to run the Mayor with (overrides town default)
  -h, --help           help for attach

3. tmux at

## Expected Behavior
runned opencode

## Actual Behavior
failed claude

## Environment
- **OS**: archlinux
- **Go version**: go1.25.5
- **Gas Town version**: 0.5.0
- **tmux version** (if applicable):

lexasub:
> [@lexasub](https://github.com/lexasub) try gt install --wrappers Install gt-codex/gt-opencode wrapper scripts to ~/bin

Not working, Please, give plan for using it ~
1)  gt install --wrappers 
2) gt rig add project1 /path/to/your/project1
3) gt crew add dev --rig project1
4) gt config agent set opencode-109 "opencode --model llama.cpp/llama.cpp.109"
5) gt config default-agent opencode-109
6) gt mayor attach

lexasub:
@Avyukth, may be you have ideas?

Avyukth:
This is a valid use case, but I haven’t run Mayor on OpenCode yet. At the moment, Claude is acting as Mayor, while I mainly use OpenCode for Polcats and Refinery. Fully replacing Claude would require significant testing. While OpenCode does provide out-of-the-box agents (Codex, Gemini CLI, Antigravity, GLM), introducing custom agents via OpenCode tends to be more fragile and adds additional testing and maintenance overhead.

lexasub:
> This is a valid use case, but I haven’t run Mayor on OpenCode yet. At the moment, Claude is acting as Mayor, while I mainly use OpenCode for Polcats and Refinery. Fully replacing Claude would require significant testing. While OpenCode does provide out-of-the-box agents (Codex, Gemini CLI, Antigravity, GLM), introducing custom agents via OpenCode tends to be more fragile and adds additional testing and maintenance overhead.

I understand that replacing Claude with OpenCode for Mayor is a serious undertaking that requires testing. However, I have objective constraints that make Claude a non-starter:
[...]

Issue #741, opened 2026-01-18
Fresh install of 4.0 on ubuntu 24.04 won't launch mayor.


perlsaiyan:
## Bug Description
Install 0.4.0

tom@aggro:~/gt$ gt mayor attach
Mayor session not running, starting...
no sessions
Error: exit status 1
Usage:
  gt mayor attach [flags]

Aliases:
  attach, at

Flags:
      --agent string   Agent alias to run the Mayor with (overrides town default)
  -h, --help           help for attach

However, deacon works.

## Steps to Reproduce

gt install . --git --wrappers --shell --force
gt config agent set opencode "gt-opencode --model dallas-ollama/qwen3-coder-32k:latest"
gt config default-agent opencode

Mayor won't start.
gt doctor complains about not being able to create roles do to UNIQUE constraints in sqlite.

remove, install 0.3.1, initializes okay, mayor won't start there either.
Install 0.4.0, doctor says everything good again.

gt doctor thinks it okay: ✓ 56 passed  ⚠ 0 warnings  ✖ 0 failed

## Expected Behavior
Mayor session attach?

## Actual Behavior
No mayor session attach

## Environment
- **OS**: Ubuntu 24.04.3 LTS (fresh install)
- **Go version**: go1.25.6 linux/amd64
- **Gas Town version**: gt version 0.4.0 (dev)
- **tmux version** (if applicable): tom@aggro:~/gt$ tmux -V
tmux 3.4

## Logs / Error Output
```
<!-- Paste relevant logs or error messages -->
```

## Additional Context

Not sure where to get logs from to diagnose

perlsaiyan:
I'm using opencode as well, got any idea for the PR?

TechnicallyShaun:
Have you tried `gt start && gt may at`?

I found on Ubuntu 24 tmux throws a different message when using tmux, which means using attach and start (with crew) wasn't working as expected.

PR here: https://github.com/steveyegge/gastown/pull/755

ppsplus-bradh:
Same errors here.  I'm using Claude.

   ⚠ Could not create role bead hq-mayor-role: bd create --json --id=hq-mayor-role --force --title=Mayor Role --labels=gt:role --description=Role definition for Mayor agents. Global coordinator for cross-rig work.: Error: insert issue: failed to insert issue: sqlite3: constraint failed: UNIQUE constraint failed: issues.id
[...]

PR #1169, opened 2026-02-02
fix(config): set Provider in RuntimeConfigFromPreset for correct prompt mode

scbrown:
## Summary

Fixes OpenCode (and other non-Claude agents) failing to start when used as the default agent for mayor/deacon/witness roles.

## Problem

`RuntimeConfigFromPreset()` was not setting the `Provider` field in the returned `RuntimeConfig`. This caused `normalizeRuntimeConfig()` to default `Provider` to `"claude"`, which made `defaultPromptMode()` return `"arg"` instead of the correct mode for the actual agent.

For OpenCode, this resulted in:
- Startup command: `opencode "prompt text here"`
- OpenCode interprets bare arguments as directory paths
- Crash: `Error: Failed to change directory to /path/to/prompt text here`

## Solution

Set `Provider: string(preset)` in `RuntimeConfigFromPreset()` so each agent uses its correct prompt mode:
- `claude`/`gemini`: `"arg"` (prompt as argument)
- `opencode`/`codex`: `"none"` (no prompt argument)

## Testing

1. Set opencode as default agent: edit `settings/config.json` → `"default_agent": "opencode"`
2. `gt mayor start`
3. Verify mayor session stays running with `gt mayor status`
4. Verify process: `tmux list-panes -t hq-mayor -F '#{pane_current_command}'` shows `opencode`

## Related Issues

- #5 - OpenCode support
- #741 - Fresh install won't launch mayor (likely same root cause)
- #925 - Mayor ignores --agent flag (related)

### DIFF
--- a/internal/config/agents.go
+++ b/internal/config/agents.go
@@ -137,7 +137,7 @@ var builtinPresets = map[AgentPreset]*AgentPresetInfo{
 		Command:             "codex",
 		Args:                []string{"--yolo"},
 		ProcessNames:        []string{"codex"}, // Codex CLI binary
-		SessionIDEnv:        "", // Codex captures from JSONL output
+		SessionIDEnv:        "",                // Codex captures from JSONL output
 		ResumeFlag:          "resume",
 		ResumeStyle:         "subcommand",
 		SupportsHooks:       false, // Use env/files instead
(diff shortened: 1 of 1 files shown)


scbrown:
Closing - PR included unrelated merge commits. Will resubmit with clean fix.

PR #1179, opened 2026-02-03
fix(sling): skip nudge when self-slinging

aleiby:
## Summary

When an agent slings work to itself, the nudge is unnecessary and potentially problematic:

1. **Unreliable delivery**: On some setups, the nudge message appears in the input buffer but Enter may not be processed reliably, leaving the agent stuck waiting for input.

2. **User interruption**: Even when delivery works, nudging yourself while you're actively working can interrupt user prompts, which is annoying and disruptive.

Since the agent already knows about the work it just slung (and will see it on the next turn via `gt hook`), skipping the nudge entirely is the right behavior for self-sling cases.

## Changes

- Add `isSelfSling` flag to `sling.go` and `sling_formula.go`
- Set flag when target is "." or no target (implicit self)
- Skip nudge/injectStartPrompt when `isSelfSling` is true

## Test plan

- [x] Code compiles
- [x] Existing tests pass (one pre-existing failure on upstream)
- [ ] Manual test: `gt sling <bead>` (self-sling) no longer shows stuck input

---
🤖 [Tackled](https://github.com/aleiby/claude-skills/tree/main/tackle) with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/internal/cmd/sling.go
+++ b/internal/cmd/sling.go
@@ -203,6 +203,7 @@ func runSling(cmd *cobra.Command, args []string) error {
 	var hookSetAtomically bool              // True if hook was set during polecat spawn (skip redundant update)
 	var delayedDogInfo *DogDispatchInfo     // For delayed dog session start after hook is set
 	var newPolecatInfo *SpawnedPolecatInfo  // Spawned polecat info (session started after bead setup)
+	var isSelfSling bool                    // True if slinging to self (skip nudge - agent already knows)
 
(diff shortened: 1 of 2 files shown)


steveyegge (approved):
Easy-win: skips nudge when self-slinging. Agent already knows about the work it just slung, and nudging mid-execution is disruptive. Clean implementation in both sling.go and sling_formula.go.

PR #1190, opened 2026-02-03
fix(doctor): correct shell integration install hint

l0g1x:
## Summary
- Fix `gt doctor` global-state check to recommend `gt shell install` instead of `gt install --shell`
- The latter fails on existing HQ setups without `--force`, while the standalone command works correctly

## Test plan
- [x] Run `gt doctor` on existing HQ setup - verify hint shows `gt shell install`
- [x] Run `gt shell install` - verify it works without requiring `--force`


steveyegge (approved):
Easy-win: correct command name in fix hint. Clean 2-line string fix.

PR #1178, opened 2026-02-03
fix(boot): detect stale molecule progress for idle Deacon

aleiby:
## Summary

- Boot's IDLE_CHECK only fired when Deacon had no work on hook
- But Deacon can have work hooked (patrol molecule) while sitting idle at the prompt
- This adds molecule progress tracking to detect that case

## Changes

- Refactor `hasDeaconWork()` → `getDeaconHookBead()` (returns bead ID instead of bool)
- Add `getMoleculeLastActivity()` to check most recent step `closed_at`
- Update triage logic: if hooked work hasn't progressed in >15 min → IDLE_CHECK

## Test plan

- [x] `go build` compiles
- [x] `go test ./internal/cmd/...` passes
- [ ] Manual test with idle Deacon (has patrol molecule hooked but at prompt)

## Related

- #1177 - upstream issue for `bd mol last-activity` command (TODO in code to switch when available)

---
🤖 [Tackled](https://github.com/aleiby/claude-skills/tree/main/tackle) with [Claude Code](https://claude.com/claude-code)

PR #1193, opened 2026-02-03
fix(boot): skip IDLE_CHECK when Deacon is in await-signal backoff

aleiby:
## Summary

When the Deacon enters await-signal mode after completing patrol, it sets an `idle:N` label on hq-deacon to track backoff cycles. Boot's molecule staleness check was incorrectly triggering IDLE_CHECK during this legitimate wait period, causing rapid patrol cycling (observed as patrols 15→16→17→18 in rapid succession).

## Changes

- Add `isDeaconInBackoff()` function that checks for `idle:N` label on hq-deacon
- Call this before the molecule staleness check in `runDegradedTriage()`
- If in backoff mode, return "nothing" - the Deacon is legitimately waiting for signals

## Test plan

- [x] Added tests for `isDeaconInBackoff()` JSON parsing
- [ ] Manual test: Deacon should stay in await-signal without being interrupted by IDLE_CHECK

## Related

- Fixes regression from #1178 which added molecule progress tracking but didn't account for await-signal mode

---
🤖 [Tackled](https://github.com/aleiby/claude-skills/tree/main/tackle) with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/internal/cmd/boot.go
+++ b/internal/cmd/boot.go
@@ -321,6 +321,15 @@ func runDegradedTriage(b *boot.Boot) (action, target string, err error) {
 		} else {
 			// Heartbeat is fresh - but is Deacon actually working?
 			// Check for idle state (no work on hook, or work not progressing)
+
+			// If deacon is in backoff mode (await-signal), skip all idle checks.
+			// The idle:N label indicates the deacon is legitimately waiting for
+			// signals, not stuck. This covers both "no work on hook" and
+			// "work not progressing" paths.
+			if isDeaconInBackoff() {
+				return "nothing", "", nil
+			}
+
 			hookBead := getDeaconHookBead()
 			if hookBead == "" {
 				fmt.Println("Deacon heartbeat fresh but no work on hook - nudging to restart patrol")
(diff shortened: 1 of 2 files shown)


seanbearden:
Found a gap in the backoff check placement while testing this fix in our fork.

`isDeaconInBackoff()` only guards the "hooked work not progressing" branch (line 337 in boot.go), but the "no work on hook" branch (line 325) fires the IDLE_CHECK nudge *before* ever reaching the backoff check. When the deacon is in await-signal with no hook bead (normal idle state), the nudge still interrupts it.

```go
// Current flow (simplified):
hookBead := getDeaconHookBead()
if hookBead == "" {
    // ← IDLE_CHECK fires here, BEFORE backoff is checked
    nudge("IDLE_CHECK: No active work on hook...")
    return
}
// backoff check only reached if hookBead != ""
if isDeaconInBackoff() {
    return "nothing", "", nil
}
```

Fix: move `isDeaconInBackoff()` check above the `hookBead == ""` branch so it covers both paths.

aleiby:
Fixed - moved the `isDeaconInBackoff()` check above the `hookBead == ""` branch so it now covers both the "no work on hook" and "work not progressing" paths.

PR #1221, opened 2026-02-06
Fix DatabasePrefixCheck scoping with dolt backend

olpie101:
## Summary

- `DatabasePrefixCheck.getDBPrefix()` and `Fix()` used `--db beadsDir` flag which doesn't scope correctly with the dolt backend — `bd` falls back to the town root database, causing every rig to report `hq` as its prefix
- Replace `--db` flag with `cmd.Dir = rigPath` so `bd` runs from the rig directory and discovers the correct database naturally
- Prevents `gt doctor --fix` from overwriting the town root `issue_prefix` with the last rig's prefix

## Test plan

- [ ] Run `gt doctor` on a town with multiple rigs using dolt backend — verify no false prefix mismatches
- [ ] Run `gt doctor --fix` and verify it targets the correct rig database, not the town root

### DIFF
--- a/internal/doctor/beads_check.go
+++ b/internal/doctor/beads_check.go
@@ -592,7 +592,6 @@ type DatabasePrefixCheck struct {
 
 type databasePrefixMismatch struct {
 	rigPath      string
-	beadsDir     string
 	routesPrefix string // From routes.jsonl (without trailing hyphen)
 	dbPrefix     string // From database config
 }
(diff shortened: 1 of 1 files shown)


steveyegge (approved):
Easy-win: Fixes DatabasePrefixCheck to use cmd.Dir instead of --db flag, which doesn't scope correctly with dolt backend. Clean, focused fix.

PR #1247, opened 2026-02-07
fix(nudge): target agent pane in multi-pane tmux sessions

joshuavial:
## Summary

- `NudgeSession` now identifies the correct agent pane in multi-pane tmux sessions instead of sending to the focused/active pane
- Adds `FindAgentPane` method that enumerates all panes and matches against agent process names (node, claude, etc.)
- Mail delivery (`notifyRecipient`) automatically benefits since it calls `NudgeSession`

## Problem

When a tmux session has multiple panes (e.g. monitoring split + agent), `send-keys -t <session>` targets the active/focused pane. Messages and nudges silently land in the wrong pane, breaking inter-agent communication.

## Changes

- `internal/tmux/tmux.go`: Add `FindAgentPane(session)` — lists all panes, detects which one runs the agent using the same process detection as `IsRuntimeRunning` (direct command match, shell+descendant check, argv[0] rewrite detection). `NudgeSession` calls this to resolve the target pane before sending.
- `internal/tmux/tmux_test.go`: 4 new tests covering single-pane (no-op), multi-pane with node agent, nonexistent session, and multi-pane with no agent (fallback).

## Test plan

- [x] `TestFindAgentPane_SinglePane` — single pane returns empty (no change to existing behavior)
- [x] `TestFindAgentPane_MultiPaneWithNode` — shell + node split, correctly identifies node pane
- [x] `TestFindAgentPane_NonexistentSession` — returns error
- [x] `TestFindAgentPane_MultiPaneNoAgent` — two shell panes, returns empty (falls back to session targeting)
- [x] All existing tmux and cmd tests pass
- [x] `golangci-lint` clean

### DIFF
--- a/internal/tmux/tmux.go
+++ b/internal/tmux/tmux.go
@@ -734,8 +734,15 @@ func (t *Tmux) NudgeSession(session, message string) error {
 	lock.Lock()
 	defer lock.Unlock()
 
+	// Resolve the correct target: in multi-pane sessions, find the pane
+	// running the agent rather than sending to the focused pane.
+	target := session
+	if agentPane, err := t.FindAgentPane(session); err == nil && agentPane != "" {
+		target = agentPane
+	}
+
 	// 1. Send text in literal mode (handles special characters)
-	if _, err := t.run("send-keys", "-t", session, "-l", message); err != nil {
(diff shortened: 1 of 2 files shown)


steveyegge:

PR #1254, opened 2026-02-08
fix: generate message ID before Validate in sendToSingle

### LINKED ISSUE #1253: gt mail send broken: 'message must have an ID' regression from e3a4b022
## Bug

`gt mail send` fails for all agents across all rigs with:

```
Error: all sends failed: <target>: invalid message: message must have an ID
```

## Root Cause

Commit e3a4b022 (`fix: validate From, Subject, and ID in Message.Validate()`) added an `ID == ""` check to `Message.Validate()`. However, `sendToSingle()` calls `Validate()` **before** `bd create` generates the bead ID. The message never has an ID at validation time because the ID is system-generated downstream.

The CLI send path (`internal/cmd/mail_send.go:75`) builds messages with a struct literal (`&mail.Message{...}`) rather than using `NewMessage()`, so no ID is ever set.

## Impact

All inter-agent mail communication is blocked town-wide. Affects direct sends, fan-out (group/list), and all callers of `Router.Send()` that don't pre-generate an ID.

## Fix

Generate the ID in `sendToSingle()` before calling `Validate()` when it's missing, using the existing `generateID()` function. This preserves the upstream validation guard (messages loaded from storage still get checked) while ensuring the send path works.

```go
func (r *Router) sendToSingle(msg *Message) error {
    if msg.ID == "" {
        msg.ID = generateID()
    }
    // Validate message before sending
    if err := msg.Validate(); err != nil {
```

PR incoming.

seanbearden:
## Summary

- Fixes regression from e3a4b022 where `Message.Validate()` requires an ID but `sendToSingle()` calls it before `bd create` generates one
- Generates the ID via existing `generateID()` in `sendToSingle` when missing, preserving the validation guard for other contexts

Closes #1253

## Test plan

- [x] `TestMessageValidate` passes (ID check still enforced)
- [x] `TestValidateRecipient` passes
- [x] `gt mail send --self` succeeds (was failing with "message must have an ID")
- [x] `gt mail inbox` confirms message delivery
- [ ] Verify fan-out paths (group, list) work (they route through `sendToSingle`)

### DIFF
--- a/internal/mail/router.go
+++ b/internal/mail/router.go
@@ -755,6 +755,11 @@ func (r *Router) validateRecipient(identity string) error {
 
 // sendToSingle sends a message to a single recipient.
 func (r *Router) sendToSingle(msg *Message) error {
+	// Ensure message has an ID (callers may omit it; bd create doesn't generate one)
+	if msg.ID == "" {
+		msg.ID = generateID()
+	}
+
 	// Validate message before sending
 	if err := msg.Validate(); err != nil {
 		return fmt.Errorf("invalid message: %w", err)


steveyegge:
Closing: changes included in #1257

PR #1257, opened 2026-02-08
feat: only interrupt agents for urgent mail, defer normal notifications

### LINKED ISSUE #1256: Mail notifications interrupt agents mid-task, causing Claude Code to treat them as rejections
## Problem

When an agent is executing a multi-step task (tool calls in progress), incoming mail triggers a `<system-reminder>` notification injected into tool results via the UserPromptSubmit hook. Claude Code interprets this mid-execution notification as user input, which can:

1. **Cause the agent to treat it as rejection** of the current approach, abandoning or re-explaining work
2. **Derail multi-tool workflows** where the agent was about to chain dependent operations
3. **Waste tokens and context** on unnecessary apologies, re-explanations, or pivots
4. **Compound with frequency** — in a healthy town with active mail, agents get interrupted constantly

## Reproduction

1. Start an agent on a multi-step task (e.g., investigating a bug across multiple files)
2. Have another agent send mail to the working agent
3. Observe the working agent pause, context-switch, or treat the notification as user rejection

Example from a real session — agent was mid-commit when a mail notification arrived:
```
<system-reminder>
You have 2 unread message(s) in your inbox.
- hq-r0d082 from gastown/sean: installed binary test
...
</system-reminder>
```

The agent may interpret this as the user interrupting their current work.

## Impact

- **Token waste**: Agents burn context on unnecessary pivots
- **Reliability**: Multi-step operations become fragile under mail load
- **Throughput**: The busier the town, the worse agents perform (negative scaling)

## Possible Approaches

1. **Defer notifications**: Queue mail notifications and deliver only when the agent is idle (between tool calls, at prompt boundaries)
2. **Batch notifications**: Aggregate notifications and deliver at defined checkpoints rather than per-message
3. **Annotate notifications clearly**: Add explicit framing like `[NON-INTERRUPTING NOTIFICATION - continue current task]` that Claude Code can learn to ignore
4. **Hook-level gating**: The UserPromptSubmit hook could suppress mail notifications when the agent has active tool calls in flight
5. **Separate channel**: Use a mechanism other than `<system-reminder>` for async notifications (e.g., write to a file the agent checks voluntarily)

## Related

- #584 — Agents act on stale mail after handoff (related mail timing issue)
- #1143 — Learning from agent execution costs (this is a source of thrash)
- #144 — Agents need autonomous patrol/mail-check loops (voluntary checking would avoid this)

seanbearden:
## Summary

- Fixes regression from e3a4b022 where `Message.Validate()` requires an ID but `sendToSingle()` calls it before `bd create` generates one
- Mail notifications now use priority-based framing to prevent agent derailment

## Problem

Mail notifications via `gt mail check --inject` (UserPromptSubmit hook) inject `<system-reminder>` text that Claude Code interprets as user input. This causes agents to pause, pivot, or abandon multi-step tasks every time normal mail arrives. The busier the town, the worse agents perform.

## Design

All mail is still notified — agents don't need to manually check. The difference is in framing:

| Priority | Notification Behavior |
|----------|----------------------|
| Urgent (P0) | Interrupts: "require immediate attention" |
| Normal | Background: "Do NOT stop or interrupt your current task" |

Urgent mail (`--urgent`) breaks through. Normal mail is delivered with explicit non-interrupting framing so Claude Code treats it as background context, not an instruction to act.

Closes #1256

## Test plan

- [x] `gt mail check --inject` with only normal mail → notification delivered with "Do NOT stop or interrupt" framing
- [x] `gt mail check --inject` with urgent mail → interrupting notification with "require immediate attention"
- [x] `gt mail check --inject` with mixed → urgent interrupts, normal count mentioned as "read after current task"
- [x] `gt mail check --inject` with no mail → no output
- [x] `gt mail check` (non-inject) and `--json` modes → unchanged
- [ ] End-to-end: agent working on task while normal mail arrives → no derailment

### DIFF
--- a/internal/cmd/mail_check.go
+++ b/internal/cmd/mail_check.go
@@ -60,24 +60,49 @@ func runMailCheck(cmd *cobra.Command, args []string) error {
 		return enc.Encode(result)
 	}
 
-	// Inject mode: output system-reminder if mail exists
+	// Inject mode: notify agent of mail with priority-appropriate framing.
+	// Urgent mail interrupts (agent should act now). Normal mail is delivered
+	// as background context that does NOT interrupt the current task.
 	if mailCheckInject {
 		if unread > 0 {
-			// Get subjects for context
 			messages, _ := mailbox.ListUnread()
-			var subjects []string
+
+			// Separate urgent from non-urgent
+			var urgent, normal []*mail.Message
 			for _, msg := range messages {
-				subjects = append(subjects, fmt.Sprintf("- %s from %s: %s", msg.ID, msg.From, msg.Subject))
+				if msg.Priority == mail.PriorityUrgent {
+					urgent = append(urgent, msg)
+				} else {
+					normal = append(normal, msg)
+				}
 			}
 
-			fmt.Println("<system-reminder>")
-			fmt.Printf("You have %d unread message(s) in your inbox.\n\n", unread)
-			for _, s := range subjects {
-				fmt.Println(s)
+			if len(urgent) > 0 {
+				// Urgent mail: interrupt — agent should stop and read
+				fmt.Println("<system-reminder>")
+				fmt.Printf("URGENT: %d urgent message(s) require immediate attention.\n\n", len(urgent))
+				for _, msg := range urgent {
+					fmt.Printf("- %s from %s: %s\n", msg.ID, msg.From, msg.Subject)
+				}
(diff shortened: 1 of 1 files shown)


steveyegge:

julianknutsen:
Merged 9a8ae05fea381b46002e7a407598d7c79b69940f

PR #1258, opened 2026-02-08
fix(tmux): target agent pane in multi-pane sessions

joshuavial:
## Summary
- `NudgeSession` now auto-detects the agent pane in multi-pane tmux sessions instead of sending to the focused pane
- Adds `FindAgentPane()` which identifies the non-shell pane in a session
- All callers (nudge, mail, daemon, boot, sling, broadcast) get the fix automatically

## Problem
When a tmux window has multiple panes, `gt nudge` and mail delivery sent input to the currently focused pane rather than the pane running the target agent. Messages silently landed in the wrong place (e.g., a shell split or monitoring pane).

## Solution
Added `FindAgentPane(session)` to `internal/tmux/tmux.go` that lists all panes and finds the one running a non-shell process. Modified `NudgeSession` to delegate to `NudgePane(agentPaneID)` when multiple panes exist. Single-pane sessions retain original behavior. Falls back gracefully if no agent pane is detected.

## Test Plan
- Unit tests: `TestFindAgentPane_SinglePane`, `TestFindAgentPane_MultiPane`, `TestFindAgentPane_NonexistentSession`, `TestIsShellCommand`
- Manual: verified with Claude and Codex agents in multi-pane sessions — nudge correctly targets agent pane while shell pane is focused
- Full test suite passes (`go test ./...`), 1 pre-existing failure in daemon unrelated

### DIFF
--- a/internal/tmux/tmux.go
+++ b/internal/tmux/tmux.go
@@ -718,17 +718,65 @@ func (t *Tmux) WakePaneIfDetached(target string) {
 	t.WakePane(target)
 }
 
+// FindAgentPane finds the pane running an agent process in a multi-pane session.
+// Returns the pane ID (e.g., "%9") of the pane running a non-shell process,
+// or empty string if no agent pane is found or the session has only one pane.
+// In multi-pane sessions, the agent pane is identified as the one NOT running
+// a known shell (bash, zsh, fish, etc.).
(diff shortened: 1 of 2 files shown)


steveyegge:
Closing: superseded by #1247 which has more robust agent pane detection.

PR #1260, opened 2026-02-08
fix(convoy): auto-convoy tracking link broken for cross-rig beads

olpie101:
## Summary

- Auto-convoys created via `gt sling` were missing their `tracks` dependency link to the dispatched bead, causing `gt convoy status` to show 0/0 progress
- Root cause: `formatTrackBeadID()` incorrectly assumed rig prefixes are always two hyphen-separated segments (e.g., `gt-mol`), but single-segment prefixes like `nx-` and `rpk-` caused it to produce bogus external refs (e.g., `external:nx-dxvmk:nx-dxvmk` instead of `external:nexus:nx-dxvmk`)
- `bd dep add` failed silently on the malformed ref, and the warning was only printed to stderr

## Fix

- Remove `formatTrackBeadID()` and pass the raw bead ID directly to `bd dep add`, letting `bd` handle cross-rig resolution via `routes.jsonl`
- This matches what `gt convoy create` and `gt convoy add` already do (they pass raw IDs and work correctly)
- Change `depCmd.Dir` from `townBeads` to `townRoot` so `bd` can resolve routes properly

## Scope

- Only the auto-convoy path (`gt sling`) was affected
- Manual convoy commands (`gt convoy create`, `gt convoy add`) were not affected — they already pass raw IDs

## Test plan

- [x] Build compiles clean
- [x] All sling and convoy tests pass
- [x] Manual verification: `bd dep add` with raw cross-rig ID resolves correctly via `routes.jsonl`
- [x] Manual verification: `gt convoy status` displays tracked issues after fix


steveyegge:

PR #1261, opened 2026-02-08
fix(config): inject --settings flag for roles with mismatched session directories

olpie101:
## Issue

Claude Code resolves project-level settings (`settings.json`) from its working directory. Roles like crew, witness, refinery, and polecat run in subdirectories (e.g., `polecats/<name>/`, `gastown/crew/`) that differ from where their settings are located. This means these roles fail to pick up the correct project settings.

## Fix

Add `withRoleSettingsFlag()` to `ResolveRoleAgentConfig()` in `internal/config/loader.go`. For roles whose session working directory differs from their settings location, it appends `--settings <path>` to the Claude args, pointing at the correct `settings.json`.

This approach is chosen so that we don't update the `.claude/settings.json` file in a project that may already have `.claude/` checked in. In such projects, even with sparse-checkout it avoids any problems. Also avoids the need to rely on using `*.local.json` files.

This could be potential allow the ability to actually keep the original `.claude/` directory that may already be checked into projects.

Role-to-settings mapping:
- `crew`, `witness`, `refinery` → `<rigPath>/<role>/.claude/settings.json`
- `polecat` → `<rigPath>/polecats/.claude/settings.json`
- `mayor`, `deacon` → no flag needed (settings and session dir are the same)

## Test plan

- [x] Build compiles clean
- [x] Verified role settings resolution logic matches directory structure

### DIFF
--- a/internal/config/loader.go
+++ b/internal/config/loader.go
@@ -1054,6 +1054,62 @@ func lookupAgentConfigIfExists(name string, townSettings *TownSettings, rigSetti
 // townRoot is the path to the town directory (e.g., ~/gt).
 // rigPath is the path to the rig directory (e.g., ~/gt/gastown), or empty for town-level roles.
 func ResolveRoleAgentConfig(role, townRoot, rigPath string) *RuntimeConfig {
+	rc := resolveRoleAgentConfigCore(role, townRoot, rigPath)
+	return withRoleSettingsFlag(rc, role, rigPath)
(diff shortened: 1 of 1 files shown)


julianknutsen:
@olpie101 - as of 096c96836f2033e10717a337ffc554d3d20f1573 we now keep the user's .claude/ files. 🚀

PR #1279, opened 2026-02-10
fix: remove deprecated --no-daemon flag from all bd command invocations

Xexr:
## Summary

- Removes all `--no-daemon` flag references from production code (19 files) and test code (13 files) after bd daemon mode was removed upstream
- Adds nil guard for `runtimeConfig` and `runtimeConfig.Tmux` in `crew_at.go` to prevent panic when config resolution returns nil

The `--no-daemon` flag was removed from the `bd` CLI but gt still passed it in ~50 call sites, causing `bd version` checks to fail and a nil pointer dereference in `crew_at.go` when attaching to crew members.

## Test plan

- [x] `go build ./...` passes cleanly
- [x] Zero remaining `--no-daemon` references in codebase (verified via grep)
- [x] `bd version` no longer errors on unknown flag
- [x] `gt crew at <name>` attaches without panic

### DIFF
--- a/internal/beads/beads.go
+++ b/internal/beads/beads.go
@@ -190,11 +190,9 @@ func (b *Beads) Init(prefix string) error {
 
 // run executes a bd command and returns stdout.
 func (b *Beads) run(args ...string) ([]byte, error) {
-	// Use --no-daemon for faster read operations (avoids daemon IPC overhead)
-	// The daemon is primarily useful for write coalescing, not reads.
 	// Use --allow-stale to prevent failures when db is out of sync with JSONL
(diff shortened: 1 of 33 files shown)


julianknutsen:
Thank you @Xexr for being the first to identify and fix this issue! Your commit was cherry-picked (with authorship preserved) as the foundation for #1301, which builds on your work with additional cleanup of the `BEADS_NO_DAEMON` env var and related docs. Really appreciate you jumping on this so quickly — your fix covered the vast majority of the codebase and made it straightforward to carry forward. 🙏

Closing in favor of #1301.

PR #1292, opened 2026-02-10
fix: crew shell inherits parent GT_ROLE instead of session-level env

seanbearden:
## Summary

- Add `NewSessionWithCommandAndEnv` to tmux package that passes env vars via `-e` flags on `tmux new-session`, setting session environment **before** the shell starts
- Update `crew/manager.go:Start()` to compute env vars before session creation and pass them via `-e` flags, preventing parent env (e.g., `GT_ROLE=mayor`) from leaking into crew sessions
- Consolidate `GT_AGENT` override into the same env map instead of a separate `SetEnvironment` call

The existing `exec env` in the startup command is preserved for `WaitForCommand` detection compatibility — this is belt-and-suspenders.

Fixes https://github.com/steveyegge/gastown/issues/1289

## Test plan

- [ ] New unit tests: `TestNewSessionWithCommandAndEnv` and `TestNewSessionWithCommandAndEnvEmpty` verify session creation with env vars and that session env is correctly set
- [ ] Manual: Set `GT_ROLE=mayor`, run `gt crew start <rig> <name>`, verify the crew session has the correct `GT_ROLE` (not mayor)
- [ ] Verify `gt hook` in the crew session shows the correct role
- [ ] Existing tmux and crew tests still pass (`go test ./internal/tmux/ ./internal/crew/`)

### DIFF
--- a/internal/crew/manager.go
+++ b/internal/crew/manager.go
@@ -629,7 +629,6 @@ func (m *Manager) Start(name string, opts StartOptions) error {
 		AgentName:        name,
 		TownRoot:         townRoot,
 		RuntimeConfigDir: opts.ClaudeConfigDir,
-		BeadsNoDaemon:    true,
 	})
 	if opts.AgentOverride != "" {
 		envVars["GT_AGENT"] = opts.AgentOverride


steveyegge (approved):
Good fix -- using tmux -e flags to set session env before the shell starts is the right approach for preventing parent env leakage. Code is clean, tests are adequate, and the belt-and-suspenders with exec env is reasonable.

One regression to fix: the moved config.AgentEnv() call is missing BeadsNoDaemon: true that exists in the current code on main. I will push a follow-up fix after merge.

Windows CI failure (TestExpandHomePath_RootUser) is unrelated to this PR.

PR #1300, opened 2026-02-11
Fix crew --agent override not provisioning correct hooks/plugin

sredfern:
## Summary

- When using `gt crew start --agent opencode`, `EnsureSettingsForRole` was resolving the default role agent config (claude) instead of the override. This meant the gastown.js plugin was never deployed to crew workspaces for non-default agents like opencode.
- Now uses `ResolveAgentConfigWithOverride` when `crewAgentOverride` is set, so the correct provider's hooks/plugins get provisioned.

## Test plan

- [x] Verified with `gt crew at doc_reviewer --rig mikrotik_docs --agent opencode --debug` — debug output confirms `provider="opencode"` and `hooks.dir=".opencode/plugin"`
- [x] Confirmed `.opencode/plugin/gastown.js` is auto-deployed to crew workspace
- [x] Tested cartographer and doc_reviewer both launch with OpenCode + MiniMax M2.1
- [x] Verified gt prime context injection, bd list, and gt mail inbox all work in session
- [ ] Run `go test ./internal/cmd/...` to confirm no regressions

### DIFF
--- a/internal/cmd/crew_at.go
+++ b/internal/cmd/crew_at.go
@@ -105,7 +105,18 @@ func runCrewAt(cmd *cobra.Command, args []string) error {
 		fmt.Printf("Using account: %s\n", accountHandle)
 	}
 
-	runtimeConfig := config.ResolveRoleAgentConfig("crew", townRoot, r.Path)
+	var runtimeConfig *config.RuntimeConfig
+	if crewAgentOverride != "" {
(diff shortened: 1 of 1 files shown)


steveyegge:
Closing — both changes in this PR are already on main. The sling_helpers ensureAgentReady change was merged via another PR, and the crew_at.go agent override fix was also landed independently. Thank you for the contribution!

PR #1310, opened 2026-02-11
fix(refinery): serialize main pushes to preserve merge determinism

### LINKED ISSUE #594: Race condition: Refinery and conflict-resolution polecat can push to main concurrently
Race condition between refinery's normal merge path and conflict-resolution polecat when pushing to `main`. Both code paths can attempt to push to `origin/main` concurrently, causing one push to fail with a non-fast-forward error.

  The merge slot mechanism successfully prevents multiple conflict-resolution polecats from racing each other, but does **not** prevent a conflict-resolution polecat from racing with the refinery's normal merge path.

  ## Steps to Reproduce

  1. MR-1 has conflicts → Refinery acquires merge slot, creates conflict-task-1
  2. Polecat-A starts resolving MR-1 (slot held)
  3. Refinery continues processing queue (MR-2, MR-3, MR-4 merge successfully)
  4. Refinery starts processing MR-5 (no conflict):
     - `git pull origin/main` → Gets SHA-X
     - `git merge MR-5 --no-ff` → Creates SHA-Y locally
  5. **RACE CONDITION:**
     - Refinery: `git push origin main` → Pushing SHA-Y (based on SHA-X)
     - Polecat-A: `git push origin temp-resolve:main` → Pushing SHA-Z (based on SHA-X)

  ## Expected Behavior

  All pushes to `main` should be serialized, regardless of whether they come from the normal merge queue or conflict resolution. Only one process should be able to push to `main` at a time.

  ## Actual Behavior

  The refinery can push to `main` while a conflict-resolution polecat is also pushing, causing one to fail with:
  ! [rejected] main -> main (non-fast-forward)

  The losing process has no explicit retry/recovery logic.

  ## Environment

  - **OS**:
  - **Go version**:
  - **Gas Town version**:
  - **tmux version** (if applicable):

  ## Logs / Error Output
  ! [rejected] main -> main (non-fast-forward)
  error: failed to push some refs to 'origin'

  ## Additional Context

  ### Root Cause

  Two independent code paths push to main:

  1. **Normal merge path** (`internal/refinery/engineer.go:257-364`, `doMerge()`):
     - Line 356: `git.Push("origin", target, false)`
     - **No merge slot check**

  2. **Conflict resolution path** (`internal/formula/formulas/mol-polecat-conflict-resolve.formula.toml:266-277`):
     - Line 276: `git push origin temp-resolve:main`
     - Protected by merge slot (prevents multiple polecats)

  The merge slot is acquired during conflict task **creation** (`engineer.go:654`), which prevents multiple conflict tasks from being created concurrently. However, it does **not** serialize the actual push operations in `doMerge()`.

  ### Impact

  - **Frequency**: Low (requires conflict resolution and normal merge to complete within same time window)
  - **Severity**: Medium-High (causes merge failure, unclear recovery path for failed push)

  ### Suggested Fix

  Acquire the merge slot in `doMerge()` before pushing (around line 355 in `internal/refinery/engineer.go`):

  ```go
  // Acquire merge slot before pushing to main
  holder := e.rig.Name + "/refinery"
  status, err := e.beads.MergeSlotAcquire(holder, false)
  if err != nil {
      return ProcessResult{
          Success: false,
          Error:   fmt.Sprintf("could not acquire merge slot: %v", err),
      }
  }
  defer e.beads.MergeSlotRelease(holder)

  // Now safe to push
  if err := e.git.Push("origin", target, false); err != nil {
      return ProcessResult{
          Success: false,
          Error:   fmt.Sprintf("failed to push to origin: %v", err),
      }
  }

  This ensures:
  - Normal merges acquire slot before pushing
  - Conflict resolutions wait for normal merges to complete
  - Only one process can push to main at a time (serialized at push boundary)

  Files to Review

  - internal/refinery/engineer.go:255-364 (doMerge - normal path)
  - internal/refinery/engineer.go:644-667 (createConflictResolutionTaskForMR - slot acquisition)
  - internal/formula/formulas/mol-polecat-conflict-resolve.formula.toml:253-284 (conflict resolution push)
  - internal/beads/beads_merge_slot.go (merge slot implementation)

davidahmann:
## Problem
Concurrent main pushes can race and violate deterministic merge behavior (issue #594).

## What changed
- Added merge-slot hooks on `Engineer` to allow deterministic acquisition/release before push.
- Wired default merge-slot operations through the beads client in `NewEngineer`.
- Acquired slot immediately before `git push` and deferred release with warning on release failure.
- Added concurrency test validating single-writer acquisition semantics under parallel contenders.

## Validation
- `go test ./internal/refinery`
- Result: pass

Fixes #594

### DIFF
--- a/internal/git/git.go
+++ b/internal/git/git.go
@@ -813,8 +813,17 @@ func (g *Git) ListBranches(pattern string) ([]string, error) {
 
 // ResetBranch force-updates a branch to point to a ref.
 // This is useful for resetting stale polecat branches to main.
+// NOTE: This uses `git branch -f` which fails on the currently checked-out branch.
+// Use ResetHard instead when the target branch is checked out.
 func (g *Git) ResetBranch(name, ref string) error {
 	_, err := g.run("branch", "-f", name, ref)
+	return err
+}
+
+// ResetHard resets the current working tree and index to the given ref.
+// Unlike ResetBranch, this works on the currently checked-out branch.
+func (g *Git) ResetHard(ref string) error {
+	_, err := g.run("reset", "--hard", ref)
 	return err
 }
 
--- a/internal/refinery/engineer.go
+++ b/internal/refinery/engineer.go
@@ -5,12 +5,14 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"strings"
+	"sync/atomic"
 	"time"
 
 	"github.com/steveyegge/gastown/internal/beads"
@@ -109,6 +111,17 @@ const (
 	staleClaimCriticalAfter = 6 * time.Hour
 )
 
+// errMergeSlotTimeout is returned by acquireMainPushSlot when retries are
+// exhausted due to slot contention. Infrastructure errors (beads down,
+// permission errors) return a different error so callers can distinguish
+// transient contention from real failures that need operator attention.
+var errMergeSlotTimeout = errors.New("merge slot contention timeout")
+
+// mergeSlotSeq is a package-level counter for unique merge slot holder IDs.
+// Using time.Now().UnixNano() alone is insufficient on Windows where timer
+// resolution can cause identical timestamps across concurrent goroutines.
+var mergeSlotSeq uint64
+
 // Engineer is the merge queue processor that polls for ready merge-requests
 // and processes them according to the merge queue design.
 type Engineer struct {
@@ -122,6 +135,8 @@ type Engineer struct {
 	mergeSlotEnsureExists func() (string, error)
 	mergeSlotAcquire      func(holder string, addWaiter bool) (*beads.MergeSlotStatus, error)
 	mergeSlotRelease      func(holder string) error
+	mergeSlotMaxRetries   int           // Max retries for slot acquisition (0 = no retry)
+	mergeSlotRetryBackoff time.Duration // Initial backoff between retries
 }
 
 // NewEngineer creates a new Engineer for the given rig.
@@ -154,6 +169,8 @@ func NewEngineer(r *rig.Rig) *Engineer {
 		mergeSlotRelease: func(holder string) error {
 			return beadsClient.MergeSlotRelease(holder)
 		},
+		mergeSlotMaxRetries:   10,
+		mergeSlotRetryBackoff: 500 * time.Millisecond,
 	}
 }
 
@@ -250,30 +267,10 @@ type ProcessResult struct {
 	Error       string
 	Conflict    bool
 	TestsFailed bool
-}
-
-// ProcessMR processes a single merge request from a beads issue.
-func (e *Engineer) ProcessMR(ctx context.Context, mr *beads.Issue) ProcessResult {
-	// Parse MR fields from description
-	mrFields := beads.ParseMRFields(mr)
-	if mrFields == nil {
-		return ProcessResult{
-			Success: false,
-			Error:   "no MR fields found in description",
-		}
-	}
-
-	// Log what we're processing
-	_, _ = fmt.Fprintln(e.output, "[Engineer] Processing MR:")
-	_, _ = fmt.Fprintf(e.output, "  Branch: %s\n", mrFields.Branch)
-	_, _ = fmt.Fprintf(e.output, "  Target: %s\n", mrFields.Target)
-	_, _ = fmt.Fprintf(e.output, "  Worker: %s\n", mrFields.Worker)
-
-	return e.doMerge(ctx, mrFields.Branch, mrFields.Target, mrFields.SourceIssue)
+	SlotTimeout bool // Merge slot contention timeout (distinct from build/test failure)
 }
 
 // doMerge performs the actual git merge operation.
-// This is the core merge logic shared by ProcessMR and ProcessMRFromQueue.
 func (e *Engineer) doMerge(ctx context.Context, branch, target, sourceIssue string) ProcessResult {
 	// Step 1: Verify source branch exists locally (shared .repo.git with polecats)
 	_, _ = fmt.Fprintf(e.output, "[Engineer] Checking local branch %s...\n", branch)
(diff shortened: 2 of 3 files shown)


davidahmann:
Context link: broader stale-state/determinism discussion for this area is here:
- https://github.com/steveyegge/gastown/discussions/1312

I opened this thread to capture the wider stale-state recovery contract beyond this scoped merge-slot fix.

julianknutsen:

Thanks for tackling the concurrent push race — this is a real problem and the right area to fix. The approach of using merge-slot acquisition before push is solid in concept. Both reviewers converged on a few design concerns that should be addressed before merge, but the core idea is sound.

Please respond with your intention for each item below (fix, defer, or disagree) — not all may require changes depending on your read of the severity.

---

### Blocker: Shared slot deadlock between push and conflict resolution
**Source:** both reviewers, high confidence

The new push code acquires the singleton merge slot with holder `<rig>/refinery/push/<nanos>`, but the existing conflict-resolution path (`createConflictResolutionTaskForMR`) holds the same slot with holder `<rig>/refinery`. These are different holder strings competing for one lock. A conflict-resolution polecat holding the slot will cause all normal merges to fail at push time, and vice versa.

**Suggested fix:** Introduce a separate slot for push serialization, or unify the holder identity so both paths coordinate intentionally.

### Major: Local git state corruption on slot failure

The slot is acquired *after* the local `git merge` commit (Steps 5-6) but *before* push. If acquisition fails, `doMerge` returns failure but the local branch now has an unpushed merge commit — corrupting the worktree for subsequent operations.

**Suggested fix:** Acquire the slot before the local merge (before Step 5), or add rollback logic (`git reset --hard origin/<target>`) on acquisition failure.

### Major: Contention treated as hard merge failure
**Source:** both reviewers, high confidence

`MergeSlotAcquire` is non-blocking — contention returns immediately with `Available: false`. The code treats this as a hard error, flowing into `MERGE_FAILED` notifications and potentially spawning spurious conflict-resolution tasks for what was just transient lock contention.

**Suggested fix:** Implement a bounded retry loop with backoff, or introduce a distinct failure type so the caller re-enqueues rather than treating contention as merge failure.

### Major: Verify `stopCh` field preserved

The diff rewrites the `Engineer` struct with aligned fields plus three new function fields. It's unclear from the diff alone whether `stopCh` was preserved. If dropped, nil-channel operations panic at runtime.

**Action:** Please confirm all existing fields are intact on your branch.

### Major: Test coverage gap
**Source:** both reviewers, high confidence

The test validates mock single-writer semantics (good!) but doesn't exercise the critical interaction between the push slot and the existing conflict-resolution slot, or worktree state after acquisition failure.

**Suggested fix:** Add tests covering: (1) slot pre-held by conflict resolver, (2) local state after mid-merge acquisition failure, (3) contention path doesn't emit spurious `MERGE_FAILED`.

### Minor: Holder uniqueness via UnixNano

`time.Now().UnixNano()` can theoretically produce duplicates under clock skew. Consider UUID or atomic counter.

---

### Pre-Merge Checklist
- [ ] Resolve shared-slot deadlock (separate slot or unified holder)
- [ ] Move slot acquisition before local merge, or add rollback
- [ ] Add retry/backoff or distinct re-enqueue failure type
- [ ] Confirm `stopCh` and all existing struct fields preserved
- [ ] Add integration tests for cross-path slot contention

---

julianknutsen:
### Suggestion from closed PR #639

PR #639 (now closed as superseded by this one) had one feature worth considering here: **blocking wait/retry when the slot is already held**.

Currently, `acquireMainPushSlot()` fails immediately if the lock file exists. In practice, two polecats finishing conflict resolution at nearly the same time is a realistic scenario — the second one will fail its merge entirely even though it just needs to wait a few seconds for the first push to complete.

PR #639 handled this with a retry loop that polls for the slot with backoff. Something like:

```go
// In acquireMainPushSlot(), before returning error:
// Retry with backoff if slot is held by another process
for attempt := 0; attempt < maxRetries; attempt++ {
    time.Sleep(backoff)
    // try again...
}
```

This would make the merge slot mechanism more resilient without sacrificing the fail-closed semantics — if retries are exhausted, it still fails the merge cleanly.

julianknutsen:
## Maintainer Adoption Review

### Original PR Review

Initial dual-model review (Claude + Codex) found **1 blocker and 3 majors**:
1. **Blocker:** `ResetBranch` (`git branch -f`) fails on the currently checked-out target branch, leaving stale local state after slot timeout.
2. **Major:** `SlotTimeout` misclassified as "build" failure — slot contention routed through polecat notification instead of silent retry.
3. **Major:** PR branch diverged from current main — `e.config.TargetBranch` was removed upstream, causing a compile failure after rebase.
4. **Major:** Push failure path didn't reset local branch state, leaving stale squash commit for next retry.

The core serialization approach, retry/backoff logic, Windows holder-ID fix, and function-field testability pattern were all validated as correct.

### Maintainer Changes

Rebased onto current `upstream/main` preserving original contributor authorship, then added a fixup commit (+413 -159 across 3 files):
- Added `ResetHard` method to git package (works on checked-out branches unlike `ResetBranch`)
- Added `SlotTimeout` early-return in `HandleMRInfoFailure` (no polecat notification for transient contention)
- Added `ResetHard` on push failure path to prevent stale local state
- Replaced `e.config.TargetBranch` (removed upstream) with `e.rig.DefaultBranch()`
- Introduced sentinel `errMergeSlotTimeout` so infrastructure errors (beads down) route through normal failure path for operator visibility
- Added 3 tests verifying infrastructure errors are NOT classified as `SlotTimeout`
- Removed dead code: `ProcessMR`, `handleFailure`, `handleSuccess` (replaced by `ProcessMRInfo`/`HandleMRInfoFailure`/`HandleMRInfoSuccess`)

### Final Review Status

Final review decision: **approve**. All original blocker and major findings resolved. Remaining findings are medium-confidence edge cases:
- **Major (medium confidence):** Slot release failure has no retry/escalation — could stall queue if release fails.
- **Major (medium confidence):** Self-conflict bypass keyed on `rigName/refinery` not per-process — mitigated by single-instance-per-rig invariant.
- **Minor:** No integration test for `doMerge` wiring with slot acquisition path (P2).

None are blocking — all depend on conditions that don't currently exist in the architecture.

### Review Iterations

3 review passes performed (1 initial PR-mode + 2 local branch iterations).

---

PR #1325, opened 2026-02-11
feat(dolt sync): add --gc flag to purge closed ephemeral beads

seanbearden:
## Summary

Adds `--gc` flag to `gt dolt sync` that purges closed ephemeral beads (wisps, convoys) before pushing databases to DoltHub remotes.

Closed ephemeral beads accumulate rapidly as transient workflow state. In one database, 7,384 closed wisps made up 89% of all issues (618MB → 3MB after purge). The `--gc` flag keeps DoltHub remotes lean by cleaning up before each push.

## Dependencies

- **Requires PR #1324** (`gt dolt sync` base command) to merge first
- **Requires steveyegge/beads#1693** (`bd purge` command) to merge first

This PR is opened as a **draft** because both dependencies must land first.

## Changes

- `internal/doltserver/sync.go`: Add `GC`/`TownRoot` fields to `SyncOptions`, `Purged`/`PurgeError` to `SyncResult`, new `purgeClosedEphemerals()` function
- `internal/cmd/dolt.go`: Add `--gc` flag, updated help text, purge stats in output

## Usage

```bash
gt dolt sync --gc           # Purge closed ephemeral beads, then push
gt dolt sync --gc --dry-run # Preview purge + push without changes
```

## How it works

For each database being synced:
1. Resolves the rig's `.beads` directory
2. Runs `bd purge --json` (or `bd purge --json --dry-run`)
3. Reports purge count per database
4. Proceeds with commit + push

Purge errors are non-fatal — logged as warnings, sync continues.

## Test plan

- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] `gt dolt sync --help` shows `--gc` flag
- [ ] `gt dolt sync --gc --dry-run` previews purge + push
- [ ] `gt dolt sync --gc` purges and pushes

Ref: steveyegge/beads#1692, gt-o97y

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1329, opened 2026-02-12
fix(dolt): verify server accepts connections before reporting restart success

seanbearden:
## Summary
- `Start()` previously only checked PID alive (signal 0) after spawning the Dolt server
- Now retries `CheckServerReachable()` (TCP dial) up to 10 times (500ms intervals, 5s total) to confirm the port is actually listening
- Prevents false "Dolt server restarted" messages that caused agents to discover a dead server and fall back to embedded mode

## Test plan
- [x] `go test ./internal/doltserver/...` passes
- [ ] Manual: `gt dolt stop && gt dolt start` — verify server is accepting connections before success message
- [ ] Manual: `gt dolt sync --gc` — verify restart reports "(accepting connections)"

### DIFF
--- a/internal/doltserver/doltserver.go
+++ b/internal/doltserver/doltserver.go
@@ -426,18 +426,29 @@ func Start(townRoot string) error {
 		fmt.Fprintf(os.Stderr, "Warning: failed to save state: %v\n", err)
 	}
 
-	// Wait briefly and verify it started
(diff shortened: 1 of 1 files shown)


seanbearden:
Closing — this fix belongs on #1325 (feat/sync-gc draft) instead.

PR #1343, opened 2026-02-12
fix(rig add): include dolt_mode and dolt_database in .beads/metadata.json

seanbearden:
## Summary
- `gt rig add` creates `.beads/metadata.json` via `bd init --backend dolt` but omits `--server`, so the metadata is missing `dolt_mode=server` — causing `bd` to fall back to embedded mode instead of connecting to the centralized Dolt sql-server
- Adds `--server` flag to both `bd init` calls in `AddRig()` (rig-level and mayor/rig tracked beads)
- Calls `doltserver.EnsureMetadata()` after `initBeads()` to set `dolt_database` to the rig name, which `bd init --server` alone doesn't provide

## Context
After `gt rig add`, agents in the new rig would find `metadata.json` with `backend: dolt` but no `dolt_mode` or `dolt_database`. This caused `bd` to use embedded mode, which can't acquire the lock held by the running Dolt server — breaking all beads operations until manually fixed.

The `--adopt` path (`runRigAdopt`) already used `--backend dolt --server` correctly (line 731). This aligns the normal `AddRig` path to match.

## Test plan
- [x] `go build ./...` passes
- [ ] Manual: `gt rig add testrig <url>` → verify `.beads/metadata.json` contains `dolt_mode: "server"` and `dolt_database: "testrig"`

### DIFF
--- a/internal/rig/manager.go
+++ b/internal/rig/manager.go
@@ -15,6 +15,7 @@ import (
 	"github.com/steveyegge/gastown/internal/beads"
 	"github.com/steveyegge/gastown/internal/config"
 	"github.com/steveyegge/gastown/internal/constants"
+	"github.com/steveyegge/gastown/internal/doltserver"
 	"github.com/steveyegge/gastown/internal/git"
 	"github.com/steveyegge/gastown/internal/runtime"
 )
@@ -424,7 +425,7 @@ func (m *Manager) AddRig(opts AddRigOptions) (*Rig, error) {
 		// DB files are gitignored so they won't exist after clone — bd init creates them.
 		// bd init --prefix will create the database and auto-import from issues.jsonl.
 		if !bdDatabaseExists(sourceBeadsDir) {
-			cmd := exec.Command("bd", "init", "--prefix", opts.BeadsPrefix, "--backend", "dolt") // opts.BeadsPrefix validated earlier
+			cmd := exec.Command("bd", "init", "--prefix", opts.BeadsPrefix, "--backend", "dolt", "--server") // opts.BeadsPrefix validated earlier
 			cmd.Dir = mayorRigPath
 			if output, err := cmd.CombinedOutput(); err != nil {
 				fmt.Printf("  Warning: Could not init bd database: %v (%s)\n", err, strings.TrimSpace(string(output)))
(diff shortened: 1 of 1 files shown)


julianknutsen:

Thanks for fixing this, @seanbearden! The embedded-mode fallback when `--server` is missing is a real pain — agents silently fail to acquire the Dolt lock, and the root cause isn't obvious from the error. Your fix is clean: adding `--server` to both `bd init` calls and calling `EnsureMetadata` to set `dolt_database` covers both the tracked and untracked `.beads` paths. Good catch that `runRigAdopt` already had this right.

---

### Major: Rig named "hq" could mutate town-level metadata

`EnsureMetadata` special-cases the name `"hq"` as a town-level alias and writes to town-level `.beads/` instead of the rig's `.beads/`. If someone were to create a rig named `hq`, the `EnsureMetadata` call would update the wrong metadata target. This is unlikely in practice (and may already fail other validation), but worth adding `"hq"` to the reserved-name list as a follow-up.

**Suggested fix:** Follow-up: add `"hq"` to reserved rig names in `AddRig`/`RegisterRig`. Not blocking this PR.

---

### Minor: No automated test asserts metadata content after AddRig

Existing rig-add tests validate the prefix/config flow but don't check that `metadata.json` actually contains `dolt_mode=server` and `dolt_database=<rigName>`. The exact regression this PR fixes has no automated guard against recurrence.

**Suggested fix:** Follow-up: extend `AddRig` integration tests to assert metadata content for both tracked and untracked `.beads` paths.

---

### Nit: Warning message could suggest recovery path

If `EnsureMetadata` fails, the rig is left in the broken state this PR fixes. The non-fatal approach is fine given the daemon's `EnsureAllMetadata` self-healing on startup, but the warning could help users self-serve: `"Warning: Could not set Dolt server metadata: %v (run 'gt doctor --fix' to repair)"`.

**Suggested fix:** Optional: amend the warning message to suggest `gt doctor --fix`.

---

### Pre-Merge Checklist
- [x] Manual test passes (`metadata.json` has correct fields)
- [x] `go build` and `go test` pass
- [ ] Follow-up: add `"hq"` to reserved rig names
- [ ] Follow-up: add integration test asserting metadata content

---

seanbearden:
Thanks for the thorough review!

**Major: Rig named "hq" could mutate town-level metadata** — **defer**. Valid observation. `EnsureMetadata` does special-case `"hq"` as the town-level alias. In practice `AddRig` would likely fail before reaching `EnsureMetadata` since `"hq"` collides with the existing `mayor/` structure, but adding it to a reserved-names list is cheap insurance. Happy to file a follow-up.

**Minor: No automated test asserts metadata content after AddRig** — **defer**. Agreed this would guard against recurrence. The existing integration tests cover the prefix/config path but not metadata content assertions. Follow-up makes sense.

**Nit: Warning message could suggest recovery path** — **agree**. Good suggestion. The daemon's `EnsureAllMetadata` self-heals on next startup, but pointing users to `gt doctor --fix` in the warning is a quick improvement for the manual-recovery case. Can add in a follow-up.

PR #1358, opened 2026-02-12
fix(convoy): use issue-level blocked state for stranded detection

corymhall:
## Motivation

`gt convoy stranded` can misclassify tracked issues when blocked-state is derived from a separate/global blocked query instead of each issue’s current details. In mixed/cross-rig scenarios, this can surface work as "ready" when it is actually blocked.

## What this changes

- Uses issue-level blocked state (`blocked_by_count`) when evaluating tracked issues.
- Removes dependency on a separate global blocked-issue lookup for stranded classification.
- Threads blocked-state through tracked issue/detail structs and readiness checks.
- Updates `isReadyIssue` to rely on per-issue `Blocked` status.

## Why this is safe

- Uses canonical issue detail fields already returned by `bd show`.
- Reduces ambiguity from out-of-band blocked lookups.
- Keeps existing readiness behavior for status/assignee checks.

## Tests

- Adds `internal/cmd/convoy_stranded_test.go` coverage for:
  - closed issues not ready
  - blocked open issues not ready
  - open unblocked issues ready
  - non-open unblocked recovery-ready path

### DIFF
--- a/internal/formula/formulas/mol-convoy-feed.formula.toml
+++ b/internal/formula/formulas/mol-convoy-feed.formula.toml
@@ -72,16 +72,13 @@ bd show <issue-id> --json
 
 An issue is "ready" if ALL of these are true:
 - status = "open" (NOT in_progress, closed, or hooked)
-- not blocked from issue details (`blocked_by_count > 0`, `blocked_by`, or open `dependencies` edges with `dependency_type="blocks"`)
+- not in blocked list (check: bd blocked --json)
 - assignee is empty OR assignee session is dead
 
 Check blocked status:
 ```bash
-bd show <issue-id> --json
-# Blocked when any of these are true:
-# - blocked_by_count > 0
-# - blocked_by has entries
-# - dependencies has an entry with dependency_type="blocks" and status not in {"closed","tombstone"}
+bd blocked --json
+# If issue ID appears here, it's blocked (skip it)
 ```
 
 Check assignee session if set:


julianknutsen:

### Decision: request_changes

Thanks for tackling cross-rig blocked-state detection, @corymhall — this is a real pain point and moving to per-issue `blocked_by_count` is the right architectural direction. The removal of the global `bd blocked` query eliminates a potential TOCTOU race and simplifies the data flow. Nice work adding unit tests for `isReadyIssue` too.

That said, both reviewers flagged a potential silent regression: the codebase consistently uses a dual-check pattern for blocked detection (`blocked_by_count > 0 || len(blocked_by) > 0`), and `blocked_by_count` is annotated as "Counts from list output" in `beads.go:52` — meaning `bd show` may only return the `blocked_by` array, not the count. If so, blocked issues would silently appear unblocked.

---

### Major: `blocked_by_count` may not be populated by `bd show`
**Source:** both, medium confidence

The `beads.Issue` struct documents `BlockedByCount` as "Counts from list output" (`beads.go:52-55`), and every other call site in the codebase uses a dual-check: `len(blocked_by) > 0 || blocked_by_count > 0` (see `mq_next.go:80`, `mq_list.go:164`, `status.go:1200`). This PR only checks `blocked_by_count > 0`. If `bd show --json` doesn't return that field, every blocked issue will appear unblocked and get dispatched as "stranded ready" work.

**Suggested fix:** Decode both `blocked_by` (array) and `blocked_by_count` (int) from `bd show` output, and use the same dual-check pattern. Verify with a live `bd show --json` against a known-blocked issue.

---

### Major: Test coverage doesn't reach the `BlockedByCount → Blocked` mapping

The new tests exercise `isReadyIssue` with pre-filled `trackedIssueInfo` structs, which is good for the function's logic. But the critical new behavior — converting `details.BlockedByCount > 0` to `dep.Blocked = true` inside `getTrackedIssues` — has no test coverage. If that threading is wrong or `bd show` returns unexpected data, the tests won't catch it.

**Suggested fix:** Add a test that constructs `issueDetails{BlockedByCount: 1}` and verifies the resulting `trackedIssueInfo.Blocked == true`. Ideally a command-stubbed integration test for `findStrandedConvoys` covering both the happy path and the missing-details fallback.

---

### Minor: `Blocked` field on intermediate `deps` struct is misleading

The `deps` struct (parsed from `bd dep list`) now has `Blocked bool` tagged `json:"blocked"`, but `bd dep list` doesn't return a `"blocked"` boolean. The field always deserializes as `false` and only gets set later by `freshDetails`. If someone reads the code, they might think `bd dep list` provides this field.

**Suggested fix:** Remove `Blocked` from the `deps` struct. Set it only on `trackedIssueInfo` during the result-building loop.

---

### Minor: Formula template still references `bd blocked --json`

`mol-convoy-feed.formula.toml` still instructs AI agents to check `bd blocked --json`. After this PR, the Go code uses per-issue state instead. This could cause agent/code divergence in readiness classification.

**Suggested fix:** Update the formula to reference per-issue blocked state. Can be a follow-up PR.

---

### Pre-Merge Checklist
- [ ] Verify `bd show --json` returns `blocked_by_count` for a known-blocked issue
- [ ] Adopt dual-check: `blocked_by_count > 0 || len(blocked_by) > 0`
- [ ] Remove `Blocked` from intermediate `deps` struct
- [ ] Add test for `BlockedByCount → Blocked` mapping
- [ ] Track formula template update (this PR or follow-up)

---

corymhall:
@julianknutsen thanks for the review!

Reviewed and addressed all four items.

Intention/status:
- Major (`blocked_by_count` from `bd show`): **fix**
- Major (test coverage for `BlockedByCount -> Blocked` mapping): **fix**
- Minor (`Blocked` on intermediate deps struct): **fix**
- Minor (formula still references `bd blocked --json`): **fix**

What changed:

1) Blocked detection hardening
- `internal/cmd/convoy.go`
- Added `issueDetails.IsBlocked()` to treat an issue as blocked when:
  - `blocked_by_count > 0`, or
  - `blocked_by` has entries, or
  - `dependencies` contains an open `dependency_type="blocks"` edge.
- `getTrackedIssues` now maps blocked state via enriched details (`applyFreshIssueDetails`) before readiness checks.

2) Mapping-path test coverage
- `internal/cmd/convoy_stranded_test.go`
- Added:
  - `TestApplyFreshIssueDetails_SetsBlockedFlag`
  - `TestIssueDetailsIsBlocked`

3) Intermediate deps struct cleanup
- `internal/cmd/convoy.go`
- Replaced ad-hoc dep decode with `trackedDependency`.
- `Blocked` is now derived state (`json:"-"`), not decoded from `bd dep list`.

4) Formula alignment
- `internal/formula/formulas/mol-convoy-feed.formula.toml`
- Updated blocked-state guidance to per-issue `bd show` checks (`blocked_by_count`, `blocked_by`, dependency-edge fallback).
- Removed instruction to use `bd blocked --json`.

julianknutsen:
## Cross-rig blocking: `bd show --json` silently drops external dependencies

Hey @corymhall — appreciate the work here, and the per-issue approach is architecturally cleaner than the global `bd blocked` call. However, I think there's a fundamental problem with using `bd show --json` as the source of blocking state for cross-rig scenarios — which is the exact case this PR aims to fix.

### The issue

`bd show --json` gets its dependency list from `GetDependenciesWithMetadata`, which runs:

```sql
SELECT i.id, ... , d.type
FROM issues i
JOIN dependencies d ON i.id = d.depends_on_id
WHERE d.issue_id = ?
```

Cross-rig dependencies are stored as `external:<project>:<issue-id>` strings in `depends_on_id`. These don't exist in the local `issues` table, so the JOIN silently drops them. The `dependencies` array in `bd show --json` output will be **empty** for any cross-rig blockers.

This means all three checks in `IsBlocked()` fail for cross-rig deps:

| Check | Works cross-rig? | Why not |
|---|---|---|
| `BlockedByCount > 0` | No | Field lives on `BlockedIssue` type, not `IssueDetails` — `bd show` never returns it |
| `len(BlockedBy) > 0` | No | Same — not part of `bd show` output |
| dependency edge scan | No | JOIN drops `external:*` refs |

### What does handle cross-rig correctly

- **`bd blocked --json`** — the thing this PR removed. It explicitly includes `external:%` refs in its SQL and post-filters them via `filterBlockedByExternalDeps`. It was the right mechanism; the bug may have been calling it from the wrong rig context.
- **`bd dep list --direction=down`** — resolves external refs by opening the target rig's database via `resolveExternalDependencies`. This is the only query path that actually fetches live cross-rig status.

### Suggested investigation

1. Verify this by running `bd show --json <issue-id>` against a known cross-rig-blocked issue and checking whether `dependencies`, `blocked_by`, or `blocked_by_count` contain anything.
2. Consider whether `bd dep list` (which does resolve cross-rig) could be used instead of `bd show` for blocking detection.
3. Alternatively, revisit the `bd blocked` approach but ensure it's called from the correct rig context (or from HQ).

steveyegge:
## Review: Needs Changes - Cross-Rig Regression

The refactoring work is solid (named structs, toIssueDetails(), applyFreshIssueDetails, IsBlocked()), but there is a blocking regression:

**Critical: bd show --json cannot see cross-rig blockers.** The underlying SQL JOINs dependencies against the local issues table. Cross-rig deps stored as external:project:issue-id have no matching local row, so the JOIN silently drops them. All three legs of IsBlocked() fail for cross-rig scenarios - the exact case this PR is trying to fix.

The old `bd blocked --json` explicitly handled cross-rig deps. This PR removes that mechanism.

**Also missing:** Formula file update referenced in the description but not in the diff.

**Suggested path forward:**
1. Verify empirically: bd show --json against a known cross-rig-blocked issue
2. Consider hybrid approach: per-issue bd show for same-rig blocking + bd blocked or bd dep list for cross-rig
3. Push the missing formula update
4. Add a test for the cross-rig blocked detection path

The structural refactoring is worth keeping - the architecture direction is correct. But the implementation needs to account for the cross-rig data gap.

PR #1373, opened 2026-02-12
fix(daemon): Fix convoy watcher and polecat spawn issues

athosmartins:
This PR includes 3 bug fixes for Gas Town discovered during production deployment:

## 1. Fix convoy watcher daemon dependency (fixes hq-s004g)
**Commit:** 20cee311
**Author:** whatsapp_automation/polecats/guzzle  
**Issue:** Convoy watcher was calling `bd activity` without `--no-daemon` flag, causing failures when no daemon is running at HQ level.  
**Fix:** Add `--no-daemon` flag to use direct storage mode.  
**Impact:** Eliminates daemon dependency for convoy activity monitoring.

**File:** `internal/daemon/convoy_watcher.go` (line 88)

## 2. Fix polecat spawn failure from NODE_OPTIONS inheritance  
**Commit:** 16162bf9  
**Issue:** Polecats were dying immediately on spawn with "Debugger attached" error when NODE_OPTIONS env var contained debugger flags from parent environment (VSCode, tmux).  
**Fix:** Explicitly clear NODE_OPTIONS in BuildStartupCommand to prevent debugger flags from breaking agents.  
**Impact:** Polecats spawn successfully regardless of parent shell environment.

**File:** `internal/config/loader.go` (line 1379)

## 3. Fix hook building for RoleBoot agents
**Commit:** bd508c82  
**Fix:** Add RoleBoot case to buildAgentIdentity() in hook builder.  
**Impact:** Deacon boot polecats can properly use hook system.

---

All fixes tested in production Gas Town deployment at `/Users/athos/gt/`.

**Testing:**
- ✅ Convoy watcher no longer fails with daemon errors
- ✅ Polecats spawn successfully from VSCode/tmux environments  
- ✅ Boot polecats can use hook system

Co-authored-by: whatsapp_automation/polecats/guzzle <athosmartins@gmail.com>

PR #1375, opened 2026-02-12
fix(spawn): clear NODE_OPTIONS to prevent polecat debugger crashes

julianknutsen:
## Summary

Cherry-pick of @athosmartins' fix from PR #1373 (commit 16162bf9). Polecats die immediately on spawn when `NODE_OPTIONS` contains debugger flags inherited from the parent environment (VSCode, tmux).

## Related Issue

Extracted from PR #1373. Credit to @athosmartins for identifying and fixing this.

## Changes

- Explicitly clear `NODE_OPTIONS=""` in `BuildStartupCommand` environment before spawning agents
- Prevents Node.js debugger flags (e.g., `--inspect` from VSCode) from being inherited through tmux into Claude's Node.js runtime, which causes "Debugger attached" errors and immediate process death

## Testing

- [x] Tested in production by @athosmartins — polecats spawn successfully from VSCode/tmux environments

## Checklist

- [x] No breaking changes
- [x] Original authorship preserved via cherry-pick

### DIFF
--- a/internal/cmd/handoff.go
+++ b/internal/cmd/handoff.go
@@ -454,7 +454,20 @@ func buildRestartCommand(sessionName string) (string, error) {
 	var agentEnv map[string]string // agent config Env (rc.toml [agents.X.env])
 	if gtRole != "" {
 		simpleRole := config.ExtractSimpleRole(gtRole)
-		runtimeConfig := config.ResolveRoleAgentConfig(simpleRole, townRoot, "")
+		// When GT_AGENT is set, resolve config with the override so we pick up
+		// the active agent's env (e.g., NODE_OPTIONS from [agents.X.env]).
+		// Otherwise, fall back to role-based resolution.
+		var runtimeConfig *config.RuntimeConfig
+		if currentAgent != "" {
+			rc, _, err := config.ResolveAgentConfigWithOverride(townRoot, "", currentAgent)
+			if err == nil {
+				runtimeConfig = rc
+			} else {
+				runtimeConfig = config.ResolveRoleAgentConfig(simpleRole, townRoot, "")
+			}
+		} else {
+			runtimeConfig = config.ResolveRoleAgentConfig(simpleRole, townRoot, "")
+		}
 		agentEnv = runtimeConfig.Env
 		exports = append(exports, "GT_ROLE="+gtRole)
 		exports = append(exports, "BD_ACTOR="+gtRole)


julianknutsen:

### Decision: request_changes

Thanks for cherry-picking this fix from @athosmartins' PR #1373 — the NODE_OPTIONS debugger crash is a real and frustrating production issue, especially for developers running VSCode. The approach of clearing the variable at the spawn boundary is the right call. However, both reviewers independently found that the fix is incomplete: it only covers one of two parallel code paths, leaving most agent types still vulnerable. A small adjustment will make this solid.

---

### Major: Fix only covers daemon restart path, not initial agent spawns

The `NODE_OPTIONS` clear is added to `BuildStartupCommand`, but nearly all agent spawns (witness, refinery, mayor, deacon, crew, dog, boot, polecat CLI) route through the parallel function `BuildStartupCommandWithAgentOverride`, which has an identical env-building block without the fix. Only daemon-triggered polecat restarts (`internal/daemon/daemon.go:1621`) use the patched path.

**Suggested fix:** Add the same `resolvedEnv["NODE_OPTIONS"] = ""` to `BuildStartupCommandWithAgentOverride`, or extract a shared `sanitizeEnv()` helper called by both functions.

---

### Minor: Unconditional clear clobbers intentional NODE_OPTIONS

The hardcoded `resolvedEnv["NODE_OPTIONS"] = ""` discards any value intentionally provided via `envVars` or `RuntimeConfig.Env` (e.g., `--max-old-space-size` for memory tuning). No configs use this today, but the contract is silently broken for future use.

**Suggested fix:** Only clear when not explicitly provided:
```go
if _, ok := envVars["NODE_OPTIONS"]; !ok {
    resolvedEnv["NODE_OPTIONS"] = ""
}
```

---

### Pre-Merge Checklist
- [ ] Apply NODE_OPTIONS clearing to `BuildStartupCommandWithAgentOverride`
- [ ] Guard against clobbering intentional NODE_OPTIONS values
- [ ] Optional: use unset semantics instead of empty-string export

---

julianknutsen:

### Decision: request_changes

Thanks for tackling the NODE_OPTIONS inheritance problem — this is a real and painful issue that kills polecats silently in VSCode/tmux environments, and credit to @athosmartins for the original diagnosis. The `SanitizeAgentEnv` helper with its two-layer guard (preserving intentional overrides from callerEnv and rc.Env) is a thoughtful design. The current iteration covers the main `BuildStartupCommand` paths well — there are a couple of consistency gaps worth addressing to avoid "works on first spawn, breaks on handoff" surprises.

---

### Major: handoff.go unconditionally clears NODE_OPTIONS, bypassing SanitizeAgentEnv guard

`handoff.go:482` appends `NODE_OPTIONS=` with no guard for caller/rc.Env overrides, while `SanitizeAgentEnv` has a two-layer guard that preserves intentionally-set values. This means a user who legitimately sets `NODE_OPTIONS` via rc.Env (e.g., `--max-old-space-size=4096` for memory tuning) will have it preserved on initial spawn but wiped on every handoff/respawn — a confusing "works on first start, breaks on restart" behavior.

**Suggested fix:** Route `buildRestartCommand` through `SanitizeAgentEnv` over a map that feeds into the exports list, or extract a shared "should we clear NODE_OPTIONS?" predicate both paths can use.

### Major: tmux session-level environment not sanitized

The PR sanitizes NODE_OPTIONS in the startup *command string* paths, but every role also calls `tmux.SetEnvironment` to set session-level env vars — and none of these clear NODE_OPTIONS. The startup `export NODE_OPTIONS=` only affects the agent's own process. Any subprocess spawned later in that tmux session (e.g., by Claude's internal tooling) may still inherit the poisoned value from the tmux session environment. Additionally, `boot.go`'s degraded-mode path uses `os.Environ()` directly with no filtering.

**Suggested fix:** The simplest fix would be to add NODE_OPTIONS clearing to `AgentEnv()` itself (the single source of truth for agent env), which would protect all paths automatically. Alternatively, add `tmux.SetEnvironment(session, "NODE_OPTIONS", "")` in each manager's session setup.

### Minor: SanitizeAgentEnv(x, x) same-map pattern collapses two-layer semantics

In `lifecycle.go`, `SanitizeAgentEnv` is called with the same map for both `resolvedEnv` and `callerEnv` (lines 489, 501, 517). This collapses the distinction between "what the caller explicitly asked for" and "what was computed from AgentEnv + rc.Env". Safe today since `AgentEnv()` never sets NODE_OPTIONS, but a latent correctness risk.

**Suggested fix:** Pass an empty map as `callerEnv` for the computed-env cases to preserve the intended two-layer semantics.

### Minor: Missing test coverage for handoff and same-map patterns

The new `TestSanitizeAgentEnv` covers core scenarios well, but there's no test for the `SanitizeAgentEnv(x, x)` same-map pattern used in lifecycle.go, no integration test verifying `BuildStartupCommand` output contains `NODE_OPTIONS=`, and the handoff.go path is completely untested for this behavior.

**Suggested fix:** Add a same-map test case, a `BuildStartupCommand` integration test, and a `buildRestartCommand` test.

---

### Pre-Merge Checklist
- [ ] Route handoff.go through `SanitizeAgentEnv` instead of unconditional append
- [ ] Address tmux session-level NODE_OPTIONS inheritance (verify in VSCode + tmux, or add clearing to `AgentEnv()`)
- [ ] Fix same-map calls in lifecycle.go to pass distinct maps
- [ ] Add test coverage for handoff path and same-map pattern

---

julianknutsen:

### Decision: request_changes

Thanks for iterating on this — the two-layer guard pattern is really well-structured now, and the defense-in-depth across AgentEnv, SanitizeAgentEnv, and the handoff path is solid. The NODE_OPTIONS clearing covers all spawn paths comprehensively. One remaining issue around shell quoting in the handoff export path needs attention before merge.

---

### Major: Shell quoting gap in handoff NODE_OPTIONS export

When `runtimeConfig.Env["NODE_OPTIONS"]` contains multiple flags (e.g., `--max-old-space-size=4096 --expose-gc`), the handoff export at `handoff.go:489-491` uses raw concatenation (`"NODE_OPTIONS="+val`) without shell quoting. This produces `export NODE_OPTIONS=--max-old-space-size=4096 --expose-gc` where bash interprets `--expose-gc` as a separate argument to `export`, not part of the value. The PR documents rc.toml NODE_OPTIONS tuning as a supported use case.

**Suggested fix:** Use `fmt.Sprintf("NODE_OPTIONS=%q", val)` to match the quoting pattern already used for `claudeEnvVars` at handoff.go:478.

---

### Minor: Tmux SetEnvironment blast radius

`AgentEnv()` is called ~19 times across 6+ tmux `SetEnvironment` callers (witness, refinery, polecat, mayor, deacon, boot managers). All now set `NODE_OPTIONS=""` in tmux session environments. This is net-positive but unmentioned in the PR description. If any agent needs NODE_OPTIONS via tmux environment (rather than command-line export), it gets silently overridden with no override mechanism (unlike the command-line path which has rc.Env merge).

**Suggested fix:** Verify `tmux set-environment -t <session> NODE_OPTIONS ""` produces `NODE_OPTIONS=` (not unset). Document in PR description that tmux environment paths are also affected. No code change required.

---

### Minor: Lifecycle default path format change

The default path in `getStartCommand` (witness, refinery, mayor, deacon, boot, dog) changes command format from `exec cmd` to `export NODE_OPTIONS= && exec cmd`. The `exec` still replaces the shell so `pane_current_command` detection should be unaffected, but the format change is worth confirming against `WaitForCommand`.

**Suggested fix:** Confirm `WaitForCommand` detection is unaffected by the `export ... &&` prefix. Already documented with inline comment.

---

### Pre-Merge Checklist
- [ ] Fix shell quoting in handoff.go — use `fmt.Sprintf("NODE_OPTIONS=%q", val)` at line 489
- [ ] Verify tmux `set-environment` with empty value produces `NODE_OPTIONS=` (not unset)
- [ ] Confirm `WaitForCommand` compatibility with `export ... && exec cmd` format
- [ ] Update PR description to note tmux `SetEnvironment` blast radius

---

PR #1376, opened 2026-02-12
fix: molecule step ordering - dep type filter + sequence sort

Henry-E:
## Summary

- **Bug 1 (dep type filter)**: The dependency filter only treated `"blocks"` as blocking. Dependencies with empty string, `"needs"`, or other unknown types were silently skipped, causing all steps to appear ready even when they had unsatisfied dependencies. Fix: `isNonBlockingDepType()` allowlist that only skips known non-blocking types (`parent-child`, `tracks`, `related`, etc.). Unknown types default to blocking (safe default).
- **Bug 2 (step ordering)**: `readySteps` inherited reverse creation order from `bd list`, so `readySteps[0]` was the last formula step. `gt hook` would suggest "Start next ready step: bd update gt-xxx.5 --status=in_progress" (Finalize) instead of step 1. Fix: `sortStepsBySequence()` sorts by the `.N` suffix in step IDs.
- Updated 6 filter locations across `molecule_status.go`, `molecule_step.go`, `molecule_dag.go` and added sorting in 4 locations where readySteps is built.

**Before fix**: `Ready: 5 (gt-xxx.5, gt-xxx.4, gt-xxx.3, gt-xxx.2, gt-xxx.1)` — all steps ready, last step first
**After fix**: `Ready: 1 (gt-xxx.1)` — only step 1 ready, correct order

## Test plan

- [x] 3 previously-failing tests now pass (TestNonBlocksDepTypeIgnored, TestReadyStepOrderReversed, TestGetMoleculeProgressInfoDepTypeFilter)
- [x] All existing tests pass (`go test ./internal/cmd/`)
- [x] Integration tested: built fixed binary, created 5-step molecule with linear deps, verified `gt mol progress` shows correct ready/blocked counts and ordering
- [x] Compared old vs fixed binary output side-by-side

### DIFF
--- a/internal/cmd/molecule_dag.go
+++ b/internal/cmd/molecule_dag.go
@@ -159,7 +159,7 @@ func buildDAG(b *beads.Beads, root *beads.Issue, children []*beads.Issue) (*DAGI
 
 		// Extract dependencies (all blocking types)
 		for _, dep := range step.Dependencies {
-			if !isNonBlockingDepType(dep.DependencyType) {
+			if isBlockingDepType(dep.DependencyType) {
 				node.Dependencies = append(node.Dependencies, dep.ID)
 			}
 		}
--- a/internal/cmd/molecule_dep.go
+++ b/internal/cmd/molecule_dep.go
@@ -8,15 +8,16 @@ import (
 	"github.com/steveyegge/gastown/internal/beads"
 )
 
-// isNonBlockingDepType returns true for dependency types that should NOT
-// block step progress. Unknown types default to BLOCKING (safe default).
-func isNonBlockingDepType(depType string) bool {
+// isBlockingDepType returns true for dependency types that block molecule step
+// progress. Matches beads' canonical blocking types (AffectsReadyWork) except
+// parent-child, which represents molecule→step hierarchy in this context.
+// Unknown/custom types are non-blocking, matching beads' default behavior.
+func isBlockingDepType(depType string) bool {
 	switch depType {
-	case "parent-child", "tracks", "related", "discovered-from",
-		"caused-by", "validates", "relates-to", "supersedes":
+	case "blocks", "conditional-blocks", "waits-for":
 		return true
 	default:
-		return false // "blocks", "", "needs", unknown -> blocking
+		return false
 	}
 }
 
--- a/internal/cmd/molecule_status.go
+++ b/internal/cmd/molecule_status.go
@@ -228,7 +228,7 @@ func runMoleculeProgress(cmd *cobra.Command, args []string) error {
 				deps = step.Dependencies
 			}
 			for _, dep := range deps {
-				if isNonBlockingDepType(dep.DependencyType) {
+				if !isBlockingDepType(dep.DependencyType) {
 					continue // Skip parent-child and other non-blocking relationships
 				}
 				hasBlockingDeps = true
@@ -604,7 +604,7 @@ func getMoleculeProgressInfo(b *beads.Beads, moleculeRootID string) (*MoleculePr
 				deps = step.Dependencies
 			}
 			for _, dep := range deps {
-				if isNonBlockingDepType(dep.DependencyType) {
+				if !isBlockingDepType(dep.DependencyType) {
 					continue // Skip parent-child and other non-blocking relationships
 				}
 				hasBlockingDeps = true
@@ -900,7 +900,7 @@ func runMoleculeCurrent(cmd *cobra.Command, args []string) error {
 		allDepsClosed := true
 		hasBlockingDeps := false
 		for _, dep := range step.Dependencies {
-			if isNonBlockingDepType(dep.DependencyType) {
+			if !isBlockingDepType(dep.DependencyType) {
 				continue // Skip parent-child and other non-blocking relationships
 			}
 			hasBlockingDeps = true
--- a/internal/cmd/molecule_step.go
+++ b/internal/cmd/molecule_step.go
@@ -199,93 +199,6 @@ func extractMoleculeIDFromStep(stepID string) string {
 	return stepID[:lastDot]
 }
 
-// findNextReadyStep finds the next ready step in a molecule.
-// Returns (nextStep, allComplete, error).
-// If all steps are complete, returns (nil, true, nil).
-// If no steps are ready but some are blocked/in_progress, returns (nil, false, nil).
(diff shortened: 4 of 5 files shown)


Henry-E:
This was really messing with my custom formulas when running them in codex polecats

Henry-E:
## Reproduction: both bugs demonstrated with a simple 5-step linear molecule

Created a 5-step molecule (`gt-90dy`) with strict linear dependencies (1→2→3→4→5) using `--type "needs"` for the dependency type:

```
bd create -t epic "Manual ordering demo"  # → gt-90dy
bd create -t task "Step 1: Gather requirements" --parent gt-90dy  # → gt-90dy.1
bd create -t task "Step 2: Write design doc" --parent gt-90dy     # → gt-90dy.2
bd create -t task "Step 3: Implement feature" --parent gt-90dy    # → gt-90dy.3
bd create -t task "Step 4: Write tests" --parent gt-90dy          # → gt-90dy.4
bd create -t task "Step 5: Ship it" --parent gt-90dy              # → gt-90dy.5

bd dep add gt-90dy.2 gt-90dy.1 --type "needs"
bd dep add gt-90dy.3 gt-90dy.2 --type "needs"
bd dep add gt-90dy.4 gt-90dy.3 --type "needs"
bd dep add gt-90dy.5 gt-90dy.4 --type "needs"
```

### Before this PR (commit `307d6f73`)

```
🧬 Molecule Progress: Manual ordering demo
  Root: gt-90dy
  [░░░░░░░░░░░░░░░░░░░░] 0% (0/5)

  Done:        0
  In Progress: 0
  Ready:       5 (gt-90dy.5, gt-90dy.4, gt-90dy.3, gt-90dy.2, gt-90dy.1)
  Blocked:     0
```

**Bug 1 (dep type filter):** All 5 steps show as "Ready" because the old code only treated `"blocks"` as blocking. The `"needs"` dep type was silently skipped, so the engine thinks no step has any dependencies.

**Bug 2 (step ordering):** Steps are listed in reverse creation order (`.5` first, `.1` last). If an agent runs `readySteps[0]`, it starts from **step 5 ("Ship it")** instead of step 1.

### After this PR (tip of `main`)

```
🧬 Molecule Progress: Manual ordering demo
  Root: gt-90dy
  [░░░░░░░░░░░░░░░░░░░░] 0% (0/5)

  Done:        0
  In Progress: 0
  Ready:       1 (gt-90dy.1)
  Blocked:     4
```

Only step 1 is ready. The 4 remaining steps are correctly blocked by their unsatisfied `"needs"` dependencies.

### Why this matters

This is a real-world scenario: `bd dep add` supports `--type "needs"` as a valid dependency type, and formulas use `needs = [...]` in their TOML definitions. The old code's allowlist of only `"blocks"` meant any molecule using non-standard dep types would have completely broken ordering — all steps appear ready and in reverse order, causing agents to execute work out of sequence.

julianknutsen:
## Maintainer Adoption Review

Thanks for tackling this, @Henry-E! The molecule step ordering bug was real and impactful — steps appearing in reverse creation order is exactly the kind of subtle issue that causes confusion. Your approach of centralizing the dep type filter and adding sequence-based sorting is the right direction. I've pushed a fixup commit on top of your work to tighten the beads alignment, and ran a dual-model review (Claude + Codex) across two passes.

### Fixes Validated

Both reviewers confirmed all three fixes are correct:

- **Ready-step ordering:** All 4 accumulation sites now sort by `.N` sequence suffix before first-item selection. This covers `CurrentStepID`, `NextStepID`, `handleStepContinue`, and `determineNextAction` consumers. ✅
- **Dep type filter centralization:** All 6 filter sites consistently routed through the shared `isBlockingDepType` helper. ✅
- **DAG dependency extraction:** `buildDAG` now uses `isBlockingDepType`, aligning tier computation with ready-step determination. ✅

### Maintainer Changes

The fixup commit (`dc957bcd`) makes these adjustments:

**1. Replaced non-blocking allowlist with blocking allowlist.** Your original `isNonBlockingDepType()` used a non-blocking allowlist where unknown/custom dep types would default to *blocking*. After checking the beads source, this diverges from beads' canonical behavior — `AffectsReadyWork()` in `beads/internal/types/types.go` defines an explicit set of four blocking types (`blocks`, `parent-child`, `conditional-blocks`, `waits-for`), and everything else is non-blocking by default. The new `isBlockingDepType()` matches this: it returns `true` for `blocks`, `conditional-blocks`, and `waits-for`, and `false` for everything else.

**Why `parent-child` is excluded:** In the molecule context, `parent-child` represents the molecule→step hierarchy relationship, not a step-ordering dependency. Including it would incorrectly block steps based on their structural relationship to the molecule root.

**Why unknown types are non-blocking:** Beads' `IsValid()` accepts any non-empty string up to 50 chars as a dependency type — there's no validation against known types. If we defaulted unknowns to blocking, any custom or future dep type would silently break step progress. Matching beads' default (non-blocking) is the safe choice.

**2. Research note — `needs` normalizes to `blocks`.** In case you were wondering about the `needs` field in formula TOML: it's compiled to `blocks` type during cooking (`beads/cmd/bd/cook.go:896-908`). It never appears as `"needs"` in the dependency database, so the filter doesn't need to handle it.

**3. Dead code removal.** `findNextReadyStep` had zero production callers (superseded by `findAllReadySteps`). Removed along with its tests.

**4. Test consistency.** Fixed remaining inline dep filters in tests to use `isBlockingDepType()` instead of hardcoded `!= "blocks"`.

### Final Review Status

The final review passed with no blockers. One major finding was flagged (test coverage gap — the new tests re-implement readiness logic rather than calling production paths like `getMoleculeProgressInfo`), but both reviewers agreed this requires interface extraction from `*beads.Beads` which is a separate refactoring effort. This is tracked in #1420.

Three minor awareness items (none requiring changes in this PR):
- `swarm/manager.go` uses a different dep filter (different context: epic→task, not step ordering)
- `extractStepSequence` could be fragile with dotted molecule IDs (low risk given current `.N` convention)
- `conditional-blocks` treated as hard block in DAG (safer direction: over-blocking vs under-blocking)

### Related Work

- **#1420** — Molecule step readiness should delegate to `bd ready --mol` instead of reimplementing blocking logic. This is the longer-term fix: gastown shouldn't need `isBlockingDepType` at all.
- **steveyegge/beads#1726** — Tightens beads' `IsValid()` to reject arbitrary dep type strings, which would catch mismatches like this at the source.

### Review Iterations

2 review passes performed (initial PR-mode + post-fixup local branch mode).

---

PR #1384, opened 2026-02-13
fix(daemon): Add --no-daemon flag to convoy watcher bd activity call

athosmartins:
Fixes convoy watcher daemon failures when beads daemon is not running.

## Problem
Convoy watcher calls `bd activity --follow --town --json` at line 88 of `internal/daemon/convoy_watcher.go`. The `bd activity` command requires a beads daemon to be running, but the convoy watcher runs at HQ level where no daemon is guaranteed.

This causes repeated errors:
```
convoy watcher: bd activity error: exit status 1, restarting in 5s
```

## Solution
Add `--no-daemon` flag to use direct storage mode:
```go
cmd := exec.CommandContext(w.ctx, "bd", "activity", "--follow", "--town", "--json", "--no-daemon")
```

## Testing
Verified `bd activity --follow --town --json --no-daemon` works without daemon.

Resolves: hq-s004g

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -29,6 +29,11 @@ install: build
 	@rm -f $(INSTALL_DIR)/$(BINARY)
 	@cp $(BUILD_DIR)/$(BINARY) $(INSTALL_DIR)/$(BINARY)
 	@echo "Installed $(BINARY) to $(INSTALL_DIR)/$(BINARY)"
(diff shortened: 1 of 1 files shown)


julianknutsen:
--no-daemon is no longer a valid beads option and has been removed while we are moving to Dolt. Please ensure you have the latest version of beads in your environment.

PR #1385, opened 2026-02-13
fix(mail): resolve cross-rig agent addresses when description is missing

seanbearden:
## Summary
- `agentBeadToAddress()` returns empty string for non-`gt-` prefixed agent beads when their description metadata is missing (`role_type:`, `rig:` fields)
- Adds `parseRigAgentAddressFromID()` fallback that extracts rig/role/name from the bead ID by scanning for known role markers (`crew`, `polecat`, `witness`, `refinery`)
- Makes cross-rig mail resilient to agent beads created without proper description fields (e.g. via manual `bd create` instead of `gt crew add`)

## Test plan
- [x] Updated `TestAgentBeadToAddress` with 6 new cases covering non-gt prefixes with and without descriptions
- [x] All existing tests pass unchanged
- [x] `go test ./internal/mail/` passes
- [x] `go build ./...` clean
- [x] Verified end-to-end: `gt mail send beads/beavis` resolves correctly

### DIFF
--- a/internal/mail/router.go
+++ b/internal/mail/router.go
@@ -370,7 +370,7 @@ func parseRigAgentAddress(bead *agentBead) string {
 	if rig == "" || rig == "null" || roleType == "" || roleType == "null" {
 		// Fallback: parse from bead ID by scanning for known role markers.
 		// ID format: <prefix>-<rig>-<role>[-<name>]
-		// Known roles: witness, refinery, crew, polecat, mayor, deacon
+		// Known rig-level roles: crew, polecat, witness, refinery
 		return parseRigAgentAddressFromID(bead.ID)
 	}
 
@@ -398,12 +398,20 @@ func parseRigAgentAddress(bead *agentBead) string {
 // parseRigAgentAddressFromID extracts a mail address from a rig-prefixed bead ID
 // when the description metadata is missing. Scans for known role markers in the ID
 // to determine the rig name and agent name.
+//
 // ID format: <prefix>-<rig>-<role>[-<name>]
+//
+// Singleton roles (witness, refinery) must NOT have a name segment — IDs like
+// "bd-beads-witness-extra" are malformed and return "".
+//
+// Keep role lists in sync with beads.RigLevelRoles and beads.NamedRoles.
 func parseRigAgentAddressFromID(id string) string {
-	// Known agent role markers that appear in bead IDs
-	knownRoles := []string{"crew", "polecat", "witness", "refinery"}
+	// Singleton roles: no name segment allowed
+	singletonRoles := []string{"witness", "refinery"}
+	// Named roles: require a name segment
+	namedRoles := []string{"crew", "polecat"}
 
-	for _, role := range knownRoles {
+	for _, role := range namedRoles {
 		marker := "-" + role + "-"
 		if idx := strings.Index(id, marker); idx >= 0 {
 			// Everything between prefix- and -role- is the rig name.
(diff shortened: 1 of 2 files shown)


julianknutsen:

### Decision: request_changes

Thanks for this fix, @seanbearden — making cross-rig mail work for agent beads created without proper description metadata is a real gap, and the ID-based fallback approach is a clean solution. The test coverage for the happy paths is solid too.

Both reviewers flagged one issue worth addressing before merge: the new parser doesn't enforce the singleton constraint for `witness`/`refinery` roles, which could silently misroute mail for malformed IDs. The other items are minor cleanup.

---

### Major: Singleton roles with name segments silently misrouted

The new `parseRigAgentAddressFromID` scans for `-witness-` and `-refinery-` as markers, but these are singleton roles — they should never have a name suffix. An ID like `bd-beads-witness-extra` would resolve to `beads/extra` instead of being rejected. This could misroute or blackhole mail for malformed bead IDs.

**Suggested fix:** After matching `witness` or `refinery` via the `-role-` marker path, check that `name == ""`. If a name is present, skip the match (return `""`). Only `crew` and `polecat` should allow name segments.

---

### Minor: Comment/code mismatch on known roles

The comment at line 371 says `// Known roles: witness, refinery, crew, polecat, mayor, deacon` but the code only includes `{"crew", "polecat", "witness", "refinery"}`. The omission of `mayor`/`deacon` is correct (they're town-level, not rig-level), but the comment could mislead a future maintainer into adding them.

**Suggested fix:** Update the comment to match the code: `// Known rig-level roles: crew, polecat, witness, refinery`

---

### Minor: Missing negative and edge-case tests

The new test cases all cover happy paths. There's no coverage for malformed singleton IDs (`bd-beads-witness-extra` should return `""`) or hyphenated agent names via the fallback path (`bd-beads-crew-my-agent` should return `beads/my-agent`).

**Suggested fix:** Add test cases for: `bd-beads-witness-extra` → `""`, `bd-beads-refinery-extra` → `""`, `bd-beads-crew-my-agent` → `beads/my-agent`

---

### Nit: Hardcoded role list duplicates canonical source

`internal/beads/agent_ids.go` defines `ValidAgentRoles`, `RigLevelRoles`, and `NamedRoles` as the canonical source of truth. The new function hardcodes its own list, which could drift if new roles are added.

**Suggested fix:** Either import the canonical lists or add a `// Keep in sync with beads.ValidAgentRoles` comment.

---

### Pre-Merge Checklist
- [ ] Enforce singleton constraint for `witness`/`refinery` (reject if name segment present)
- [ ] Fix comment to remove `mayor, deacon`
- [ ] Add negative tests for malformed singleton IDs
- [ ] Add test for hyphenated agent name via fallback

---

seanbearden:
Thanks for the thorough review. All four items addressed in the latest push:

| Item | Action | Details |
|------|--------|---------|
| **Major: Singleton roles with name segments** | **fix** | `parseRigAgentAddressFromID` now separates named roles (`crew`, `polecat`) from singleton roles (`witness`, `refinery`) into distinct loops. Singletons only match at end-of-ID via the `-role` suffix path; if a `-role-` marker is found (indicating a name segment), the match is skipped and returns `""`. |
| **Minor: Comment/code mismatch** | **fix** | Updated to `// Known rig-level roles: crew, polecat, witness, refinery` |
| **Minor: Missing negative/edge-case tests** | **fix** | Added 3 new test cases: `bd-beads-witness-extra` → `""`, `bd-beads-refinery-extra` → `""`, `bd-beads-crew-my-agent` → `beads/my-agent` |
| **Nit: Hardcoded role list** | **fix** | Added `// Keep role lists in sync with beads.RigLevelRoles and beads.NamedRoles` comment. Chose not to import directly since `internal/mail` doesn't currently depend on `internal/beads` and adding a cross-package dependency for a 4-element list felt heavier than the sync comment. |

All 19 tests pass (including the 3 new ones). Also rebased onto latest `upstream/main` (1112c906).

seanbearden:
@julianknutsen All four items addressed — ready for re-review when you get a chance.

PR #1392, opened 2026-02-13
fix(web): buffer template output to prevent superfluous WriteHeader

julianknutsen:
## Summary

Buffer template output to `bytes.Buffer` before writing to `ResponseWriter`, ensuring proper 500 status on template errors instead of 200 with partial content.

## Related Issue

Supersedes #627 (rebased onto current main, resolved conflicts with upstream security/logging improvements)

## Changes

- Buffer `ExecuteTemplate` output to `bytes.Buffer` instead of writing directly to `ResponseWriter`
- Only write to `ResponseWriter` on success — prevents implicit `WriteHeader(200)` on partial template execution
- Preserve upstream's server-side error logging and security comment on the error path
- Add `TestConvoyHandler_TemplateErrorReturns500` verifying template errors return 500

## Testing

- [x] Unit tests pass (`go test ./internal/web/...`)
- [x] Build succeeds (`go build ./...`)

## Checklist

- [x] No breaking changes
- [x] Original authorship preserved (Graeme Foster)

### DIFF
--- a/internal/web/handler.go
+++ b/internal/web/handler.go
@@ -1,6 +1,7 @@
 package web
 
 import (
+	"bytes"
 	"context"
 	"embed"
 	"html/template"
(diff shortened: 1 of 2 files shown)


julianknutsen:

### Decision: approve

Thanks for the fix, @GraemeF! The superfluous `WriteHeader` bug is real and your buffered template approach is the standard Go pattern — it matches what the codebase already does in `internal/templates/templates.go`. Both reviewers confirmed the change is correct, tightly scoped, and includes a good test. Two minor hardening improvements were incorporated during rebase.

All findings below have already been addressed in the rebased version — listing them for transparency.

---

### Minor: WriteTo error now logged

`buf.WriteTo(w)` previously discarded write errors silently (e.g., client disconnect). While the original code also had this gap, we added error logging to match the method's existing error-logging pattern: `log.Printf("dashboard: response write failed: %v", err)`.

**Status:** Fixed in rebase.

### Minor: Test now asserts no partial template content on error

The test verified 500 status and error message text, but didn't assert that partial template HTML (like `<!DOCTYPE`) was absent from the error response — which is the core guarantee of the buffering fix. Added a negative assertion.

**Status:** Fixed in rebase.

---

### Pre-Merge Checklist
- [x] Core behavior correct: template errors produce 500, not 200-with-partial-content
- [x] Aligns with existing `internal/templates/templates.go` buffering pattern
- [x] `WriteTo` errors logged
- [x] Test asserts absence of partial template content on error
- [x] All existing tests pass
- [x] Original authorship preserved

---

PR #1395, opened 2026-02-13
fix(startup): replace fixed sleeps with prompt-detection polling

julianknutsen:
## Summary

- Replace the hard-coded `time.Sleep(8 * time.Second)` in `ensureAgentReady()` with `WaitForRuntimeReady()` polling, which detects the agent's prompt prefix (e.g. `❯ ` for Claude) every 200ms instead of blindly sleeping
- Remove redundant 5-second sleep in `swarm.go` after `polecatSessMgr.Start()`, which already handles readiness internally via `WaitForCommand()`, `AcceptBypassPermissionsWarning()`, and `runtime.SleepForReadyDelay()` (fixed delay)
- For known agent presets: uses the preset's `ReadyPromptPrefix` and `ReadyDelayMs` for prompt-detection polling
- For unknown/custom agents: falls back to a 1-second fixed delay with no prompt detection (mirrors old behavior)

Cherry-picked from #431 by @ekoziol and adapted to use existing `WaitForRuntimeReady()` infrastructure.

## Test plan

- [x] `go vet ./...` — no warnings
- [x] `go test ./internal/cmd/... -count=1` — all cmd tests pass
- [x] `go test ./internal/config/... -count=1` — config tests pass
- [ ] Manual: `gt start` on fast machine — agent ready in <1s instead of 8s
- [ ] Manual: `gt start` on slow machine — agent waits up to 60s timeout, then proceeds gracefully
- [ ] Manual: `gt swarm create` — workers start without extra 5s delay
- [ ] Manual: custom/unknown GT_AGENT — falls back to 1s delay (no 60s regression)

### DIFF
--- a/internal/cmd/sling_helpers.go
+++ b/internal/cmd/sling_helpers.go
@@ -276,9 +276,15 @@ func ensureAgentReady(sessionName string) error {
 	// Use prompt-detection polling instead of fixed sleep.
 	// For known presets: uses ReadyPromptPrefix (e.g. "❯ " for Claude) polled every 200ms.
 	// For unknown/custom agents: falls back to a 1s fixed delay (mirrors old behavior).
+	// Note: uses preset-only resolution (not ResolveRoleAgentConfig) because
+	// ensureAgentReady lacks rig/town context — only has the session name.
+	effectiveName := agentName
+	if effectiveName == "" {
+		effectiveName = "claude" // Default sessions without GT_AGENT are Claude
+	}
 	var rc *config.RuntimeConfig
-	if preset := config.GetAgentPreset(config.AgentPreset(agentName)); preset != nil {
-		rc = config.RuntimeConfigFromPreset(config.AgentPreset(agentName))
+	if preset := config.GetAgentPreset(config.AgentPreset(effectiveName)); preset != nil {
+		rc = config.RuntimeConfigFromPreset(config.AgentPreset(effectiveName))
 	} else {
 		// Unknown agent — use minimal config: no prompt detection, short fixed delay.
 		rc = &config.RuntimeConfig{


julianknutsen:

### Decision: request_changes

Thanks for picking up @ekoziol's original work and adapting it to use the existing `WaitForRuntimeReady()` infrastructure — this is a real improvement over the fixed 8-second sleep, and the prompt-detection approach is the right direction. The cherry-pick with conflict resolution was cleanly done.

The main concern is a latency regression for unknown/custom agent presets. Please respond with your intention for each item below (**fix**, **defer**, or **disagree**) — not all may require changes depending on your read of the severity.

---

### Major: Unknown agent presets block for 60s instead of old 1s sleep

`RuntimeConfigFromPreset()` never returns nil — when `GT_AGENT` isn't a known preset, it falls back to `DefaultRuntimeConfig()` which returns Claude defaults (`ReadyPromptPrefix: "❯ "`, `ReadyDelayMs: 10000`). This means `WaitForRuntimeReady()` polls for a Claude-specific prompt that will never appear, blocking for the full `ClaudeStartTimeout` (60s) before the graceful-degradation warning fires. The old code slept just 1 second for non-Claude agents.

**Suggested fix:** Check `config.GetAgentPreset(agentName)` first. If nil (unknown preset), use a minimal `RuntimeConfig` with no prompt prefix and a short `ReadyDelayMs` (~1000ms to match old behavior), or skip `WaitForRuntimeReady` entirely.

---

### Minor: PR description inaccurately describes swarm.go readiness mechanism

The description states `polecatSessMgr.Start()` handles readiness via `WaitForRuntimeReady()`. It actually uses `runtime.SleepForReadyDelay()` (fixed 10s sleep), not prompt-detection polling. The code change itself is correct — removing the redundant 5s post-`Start()` sleep is safe — but the inaccurate description will mislead future readers.

**Suggested fix:** Update PR description to say `Start()` uses `SleepForReadyDelay()` (fixed delay).

---

### Nit: Dead nil-check on RuntimeConfigFromPreset return value

The `if rc == nil { rc = config.DefaultRuntimeConfig() }` guard can never execute because `RuntimeConfigFromPreset()` already contains this exact fallback internally (`agents.go:352-354`). This misleads future readers into thinking a nil path exists.

**Suggested fix:** Remove the dead nil check. If the major finding is addressed by restructuring to check `GetAgentPreset` first, this becomes moot.

---

### Pre-Merge Checklist
- [ ] Handle unknown/custom agent presets explicitly to avoid 60s timeout regression
- [ ] Remove dead `if rc == nil` guard (or restructure)
- [ ] Correct PR description re: `Start()` using `SleepForReadyDelay`

---
_Generated by automated dual-model review pipeline (Claude Opus 4.6 + GPT-5.3 Codex)_

julianknutsen:
Closing — reopening from fork (julianknutsen/gastown) per CONTRIBUTING.md workflow.

PR #1397, opened 2026-02-13
fix(startup): replace fixed sleeps with prompt-detection polling

julianknutsen:
## Summary

- Replace the hard-coded `time.Sleep(8 * time.Second)` in `ensureAgentReady()` with `WaitForRuntimeReady()` polling, which detects the agent's prompt prefix (e.g. `❯ ` for Claude) every 200ms instead of blindly sleeping
- Remove redundant 5-second sleep in `swarm.go` after `polecatSessMgr.Start()`, which already handles readiness internally via `WaitForCommand()`, `AcceptBypassPermissionsWarning()`, and `runtime.SleepForReadyDelay()` (fixed delay)
- For known agent presets: uses the preset's `ReadyPromptPrefix` and `ReadyDelayMs` for prompt-detection polling
- For unknown/custom agents: falls back to a 1-second fixed delay with no prompt detection (mirrors old behavior)
- Empty `GT_AGENT` normalized to "claude" before preset lookup (default sessions are Claude)
- Minimum 1s readiness delay floor for presets without prompt detection (prevents early-input races on gemini, cursor, auggie, amp)

Cherry-picked from #431 by @ekoziol and adapted to use existing `WaitForRuntimeReady()` infrastructure.

Supersedes #1395 (closed — was pushed directly to upstream instead of from fork).

## Test plan

- [x] `go vet ./...` — no warnings
- [x] `go test ./internal/cmd/... -count=1` — all cmd tests pass
- [x] `go test ./internal/config/... -count=1` — config tests pass
- [ ] Manual: `gt start` on fast machine — agent ready in <1s instead of 8s
- [ ] Manual: `gt start` on slow machine — agent waits up to 60s timeout, then proceeds gracefully
- [ ] Manual: `gt swarm create` — workers start with minimum 1s readiness guard
- [ ] Manual: custom/unknown GT_AGENT — falls back to 1s delay (no 60s regression)
- [ ] Manual: zero-delay preset (e.g. gemini) — gets 1s minimum floor, not 0s

### DIFF
--- a/internal/cmd/sling_helpers.go
+++ b/internal/cmd/sling_helpers.go
@@ -293,6 +293,13 @@ func ensureAgentReady(sessionName string) error {
 			},
 		}
 	}
+	// Ensure a minimum 1s readiness delay for presets without prompt detection.
+	// Without this, agents with ReadyPromptPrefix="" and ReadyDelayMs=0
+	// (e.g. gemini, cursor) would skip the readiness guard entirely,
+	// reintroducing early-input races that this function exists to prevent.
+	if rc.Tmux != nil && rc.Tmux.ReadyPromptPrefix == "" && rc.Tmux.ReadyDelayMs < 1000 {
+		rc.Tmux.ReadyDelayMs = 1000
+	}
 	if err := t.WaitForRuntimeReady(sessionName, rc, constants.ClaudeStartTimeout); err != nil {
 		// Graceful degradation: warn but proceed (matches original behavior of always continuing)
 		fmt.Fprintf(os.Stderr, "Warning: agent readiness detection timed out for %s: %v\n", sessionName, err)
--- a/internal/cmd/swarm.go
+++ b/internal/cmd/swarm.go
@@ -8,6 +8,8 @@ import (
 	"os/exec"
 	"path/filepath"
 	"strings"
+	"time"
+
 	"github.com/spf13/cobra"
 	"github.com/steveyegge/gastown/internal/config"
 	"github.com/steveyegge/gastown/internal/git"
@@ -527,6 +529,11 @@ func spawnSwarmWorkersFromBeads(r *rig.Rig, townRoot string, swarmID string, wor
 				style.PrintWarning("  couldn't start %s: %v", worker, err)
 				continue
 			}
+			// Minimum readiness guard before injection. Start() handles
+			// runtime-config-aware delays internally, but presets with
+			// ReadyDelayMs=0 (e.g. gemini, cursor) skip the delay entirely.
+			// This floor prevents early-input races on those agents.
+			time.Sleep(1 * time.Second)
 		}
 
 		// Inject work assignment


julianknutsen:
> _Copied from closed PR #1395 (v1 review, before fixes applied):_

### Decision: request_changes

Thanks for picking up @ekoziol's original work and adapting it to use the existing `WaitForRuntimeReady()` infrastructure — this is a real improvement over the fixed 8-second sleep, and the prompt-detection approach is the right direction. The cherry-pick with conflict resolution was cleanly done.

The main concern is a latency regression for unknown/custom agent presets. Please respond with your intention for each item below (**fix**, **defer**, or **disagree**) — not all may require changes depending on your read of the severity.

---

### Major: Unknown agent presets block for 60s instead of old 1s sleep

`RuntimeConfigFromPreset()` never returns nil — when `GT_AGENT` isn't a known preset, it falls back to `DefaultRuntimeConfig()` which returns Claude defaults (`ReadyPromptPrefix: "❯ "`, `ReadyDelayMs: 10000`). This means `WaitForRuntimeReady()` polls for a Claude-specific prompt that will never appear, blocking for the full `ClaudeStartTimeout` (60s) before the graceful-degradation warning fires. The old code slept just 1 second for non-Claude agents.

**Suggested fix:** Check `config.GetAgentPreset(agentName)` first. If nil (unknown preset), use a minimal `RuntimeConfig` with no prompt prefix and a short `ReadyDelayMs` (~1000ms to match old behavior), or skip `WaitForRuntimeReady` entirely.

> **Status: Fixed** in commit ce49e68c — uses `GetAgentPreset()` first, unknown presets get minimal 1s config.

---

### Minor: PR description inaccurately describes swarm.go readiness mechanism

The description states `polecatSessMgr.Start()` handles readiness via `WaitForRuntimeReady()`. It actually uses `runtime.SleepForReadyDelay()` (fixed 10s sleep), not prompt-detection polling. The code change itself is correct — removing the redundant 5s post-`Start()` sleep is safe — but the inaccurate description will mislead future readers.

**Suggested fix:** Update PR description to say `Start()` uses `SleepForReadyDelay()` (fixed delay).

> **Status: Fixed** — PR description updated.

---

### Nit: Dead nil-check on RuntimeConfigFromPreset return value

The `if rc == nil { rc = config.DefaultRuntimeConfig() }` guard can never execute because `RuntimeConfigFromPreset()` already contains this exact fallback internally (`agents.go:352-354`). This misleads future readers into thinking a nil path exists.

**Suggested fix:** Remove the dead nil check. If the major finding is addressed by restructuring to check `GetAgentPreset` first, this becomes moot.

> **Status: Fixed** in commit ce49e68c — restructured to check `GetAgentPreset` first, dead nil-check removed.

---

### Pre-Merge Checklist
- [x] Handle unknown/custom agent presets explicitly to avoid 60s timeout regression
- [x] Remove dead `if rc == nil` guard (or restructure)
- [x] Correct PR description re: `Start()` using `SleepForReadyDelay`

---
_Generated by automated dual-model review pipeline (Claude Opus 4.6 + GPT-5.3 Codex)_

julianknutsen:

### Decision: approve

Thanks for tackling this, @ekoziol! The 8-second fixed sleep has been a pain point — too slow on fast machines, too short on slow ones, and completely unaware of whether the agent is actually ready. The prompt-detection polling approach is the right fix: it returns faster when the agent is ready, waits longer when it isn't, and never hard-fails. All three fix targets validated as correct across both reviewers. The remaining observations are minor and well-handled by the graceful degradation design.

---

### Fixes Validated

- **Hard-coded 8s sleep replaced with prompt-detection polling** — `ensureAgentReady()` now uses `WaitForRuntimeReady()` which polls for the `❯ ` prompt prefix every 200ms for Claude, and uses configurable `ReadyDelayMs` for other agents. Graceful degradation on timeout preserves the never-fail invariant.
- **Zero-delay presets get minimum readiness guard** — The `>=1000ms` floor when no prompt prefix is present prevents gemini/cursor/auggie/amp from skipping the readiness guard entirely.
- **Swarm 5s redundant sleep reduced to 1s floor** — `Start()` already handles runtime-config-aware delays internally (10s for Claude via `SleepForReadyDelay`). The 1s floor prevents early-input races on zero-delay presets.

---

### Minor: GT_AGENT fallback defaults to Claude prompt polling

If `GT_AGENT` is unset on a non-Claude session, the code defaults to Claude prompt detection and could poll for up to 60s before gracefully proceeding. In practice, `GT_AGENT` is set by gastown during session creation, and the 60s timeout degrades gracefully (warn + continue) — so this is very unlikely to bite. A follow-up to log a warning when `GT_AGENT` is unexpectedly empty would improve debuggability.

**Suggested fix:** Optional follow-up — add a warning log when `GT_AGENT` is empty for debuggability.

### Minor: Preset-only config may diverge from resolved role/town config
**Source:** codex, low confidence

`ensureAgentReady()` resolves runtime config via `RuntimeConfigFromPreset()` rather than the full `ResolveRoleAgentConfig()` chain, so user-specified config overrides (e.g., custom `ReadyDelayMs` in rig config) won't be respected. This is an acceptable tradeoff — `ensureAgentReady` lacks rig/town context, and the pattern matches how `refinery/manager.go` and `polecat_spawn.go` already work.

**Suggested fix:** No action needed. Consider persisting resolved readiness fields into tmux env as a future improvement.

### Minor: Timing profile change — 8s fixed sleep to 60s polling

The old behavior was a fixed 8-second sleep regardless of agent readiness. The new behavior returns faster when the agent is ready (prompt detected early) and waits up to 60s if the prompt never appears, then gracefully proceeds. This is strictly better — faster fast path, more patient slow path, same never-fail guarantee.

**Suggested fix:** None — this is an improvement.

---

### Pre-Merge Checklist
- [x] No blockers identified
- [x] No required fixes for merge
- [ ] **Optional follow-up:** Add warning log when `GT_AGENT` is unexpectedly empty
- [ ] **Optional follow-up:** Consider persisting resolved readiness fields into tmux env

---

PR #1402, opened 2026-02-13
fix(gt done): prevent incorrect main/master detection when cwd is unavailable

thanakijwanavit:
## Summary

When a polecat's working directory is deleted (nuked), gt done falls back to using the rig's mayor clone for git operations. However, the mayor clone is always on main/master, not the polecat's feature branch.

## Bug

When cwdAvailable=false and GT_BRANCH env var is not set, the code would call g.CurrentBranch() on the mayor clone, incorrectly returning main/master and causing 'cannot submit main/master branch to merge queue' error.

## Fix

Check if cwdAvailable is false before calling g.CurrentBranch(). If we're using the mayor clone fallback and don't have GT_BRANCH, we can't determine the actual branch and should return an error rather than guessing incorrectly.

## Testing

- [x] Build passes: go build ./...
- [x] Tests pass: go test ./internal/cmd/... -v -run Done

Fixes: gm-ae9ud

### DIFF
--- a/internal/cmd/done.go
+++ b/internal/cmd/done.go
@@ -196,7 +196,17 @@ func runDone(cmd *cobra.Command, args []string) error {
 	// g.CurrentBranch() in that case would incorrectly return main/master.
 	if branch == "" {
 		if !cwdAvailable {
-			// We don't have GT_BRANCH and we're using mayor clone - can't determine branch
+			// We don't have GT_BRANCH and we're using mayor clone - can't determine branch.
+			// Arm session cleanup before returning so the polecat doesn't get stranded.
+			if polecatEnv := os.Getenv("GT_POLECAT"); polecatEnv != "" {
+				sessionCleanupNeeded = true
+				deferredTownRoot = townRoot
+				deferredRoleInfo = RoleInfo{
+					Role:    RolePolecat,
+					Rig:     rigName,
+					Polecat: polecatEnv,
+				}
+			}
 			return fmt.Errorf("cannot determine branch: GT_BRANCH not set and working directory unavailable")
 		}
 		var err error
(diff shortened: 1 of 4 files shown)


julianknutsen:
## Maintainer Adoption Review

Thanks for this fix, @thanakijwanavit! You correctly identified a nasty failure mode — when a polecat's worktree gets nuked before `gt done` finishes, it falls back to the mayor clone and silently returns the wrong branch. Your guard against calling `CurrentBranch()` in that case is the right fix, and the env var injection for Kimi follows the established `PrependEnv` pattern nicely.

I've reviewed, tested, and pushed a fixup commit on top of your work to address a few things the review caught. Your original commits are fully preserved.

### Original PR Review

The initial review validated both fixes (branch detection guard and Kimi env var inheritance) as correct. It flagged 4 items:

- `GT_BRANCH` was referenced in `done.go` but never set anywhere — the fallback path would always error
- The early return happened before the session cleanup backstop was armed, risking stranded polecats
- `GT_POLECAT_PATH` had the same "referenced but never set" gap
- No regression tests for the new guard logic

### Maintainer Changes

Fixup commit `96d2bc21` (+191 -6 across 4 files):

1. **Populate `GT_BRANCH` at session start** — reads the branch from the worktree and injects it via `PrependEnv` + `tmux.SetEnvironment` so the nuked-worktree fallback actually works end-to-end
2. **Arm cleanup backstop before early return** — constructs a minimal `RoleInfo` from env vars so the deferred session kill fires on the error path
3. **Populate `GT_POLECAT_PATH`** — added alongside `GT_BRANCH` for worktree path fallback
4. **Add regression tests** — `TestBranchDetectionGuard`, `TestBranchDetectionCleanupOnError`, `TestPolecatStartInjectsFallbackEnvVars`

### Final Review Status

All clear — no remaining blockers or majors. Two minor notes: (a) if `CurrentBranch()` fails at session start, `GT_BRANCH` is silently unset, but a broken worktree would fail the polecat long before `gt done` anyway; (b) the new tests simulate the logic rather than calling production code directly — a known gap since `runDone` is hard to unit test.

Build passes, all tests pass. Ready for CI and merge.

---
_Adopted via `/adopt-pr`. Original contributor commits preserved. Merge with merge commit to preserve attribution._

PR #1403, opened 2026-02-13
fix(config): remove incorrect --settings flag for crew/polecat roles

seanbearden:
## Summary

- `roleSettingsDir()` incorrectly returns the parent directory (`<rig>/crew/` or `<rig>/polecats/`) for crew and polecat roles, causing `--settings` to point to a non-existent `settings.local.json`
- Crew and polecat agents each have per-agent settings in their own working directory (`<rig>/crew/<name>/.claude/` and `<rig>/polecats/<name>/.claude/`), so Claude discovers them from CWD naturally — no `--settings` override is needed
- Only single-instance roles (`witness`, `refinery`) need the `--settings` flag since their session directory differs from their settings location

**Symptom**: `gt crew at <name>` sessions crash immediately with `[exited]` because Claude treats a missing `--settings` path as fatal.

**Root cause**: Introduced in #1261 which added `roleSettingsDir` — the mapping assumed crew/polecat settings live at the role directory level, but they live at the per-agent level.

## Changes

- `internal/config/loader.go` — Remove `"crew"` and `"polecat"` cases from `roleSettingsDir()` switch, so they fall through to the default (empty string = no `--settings` flag)
- `internal/config/loader_test.go` — Update test to use `witness` as the positive case; add negative assertions that `crew` and `polecat` do NOT receive `--settings`

## Test plan

- [x] `go build ./internal/config/` compiles clean
- [x] `go test ./internal/config/ -run "TestRoleSettingsDir|TestWithRoleSettingsFlag"` passes
- [x] Cherry-picked cleanly onto latest `upstream/main` (1112c906)

### DIFF
--- a/internal/config/loader.go
+++ b/internal/config/loader.go
@@ -1092,14 +1092,16 @@ func withRoleSettingsFlag(rc *RuntimeConfig, role, rigPath string) *RuntimeConfi
 
 // roleSettingsDir returns the shared settings directory for roles whose session
 // working directory differs from their settings location. Returns empty for
-// roles where settings and session directory are the same (mayor, deacon).
+// roles where settings and session directory are the same (mayor, deacon), or
+// where each agent has its own settings in its working directory (crew, polecat).
 func roleSettingsDir(role, rigPath string) string {
 	switch role {
-	case "crew", "witness", "refinery":
+	case "witness", "refinery":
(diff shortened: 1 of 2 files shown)


julianknutsen:
Superseded by #1418

PR #1406, opened 2026-02-13
fix(config): remove --settings flag for polecat role

olpie101:
## Bug

Polecat sessions crash immediately on startup because Claude Code receives a `--settings` flag pointing to a non-existent file:

```
--settings /Users/.../gt/gastown/polecats/.claude/settings.local.json
```

This path (`<rig>/polecats/.claude/settings.local.json`) is never created by any current code path:

- `gt hooks sync` writes settings to individual polecat worktrees (`polecats/<name>/<rig>/.claude/settings.local.json`), not the parent `polecats/` directory
- `EnsureSettingsForRole` writes settings into the polecat's worktree (cwd), not the parent
- `gt doctor` explicitly marks `polecats/.claude/settings.{json,local.json}` as **stale** and flags them as `wrongLocation`

The file only existed previously as a legacy artifact. After recent cleanup (stale settings removal, hooks sync improvements), the file no longer exists, causing the crash.

## Root Cause

`roleSettingsDir()` in `internal/config/loader.go` returns `filepath.Join(rigPath, "polecats")` for the polecat role. This causes `withRoleSettingsFlag()` to inject `--settings <rig>/polecats/.claude/settings.local.json` into the startup command. This path is inconsistent with the actual settings layout used by `gt hooks sync`, `EnsureSettingsForRole`, and `gt doctor`.

## Fix

Return `""` from `roleSettingsDir()` for the polecat role, so no `--settings` flag is injected. Polecats already get settings via `EnsureSettingsForRole` written directly into their worktree, and Claude Code resolves settings from its working directory.

Updated the corresponding test to verify `--settings` is **not** present for polecats.

## Test plan

- [x] `go test ./internal/config/ -run "TestWithRoleSettingsFlag|TestBuildPolecatStartupCommand"` — all pass
- [x] `make install` — clean build
- [x] `gt sling` spawns polecat successfully (manually verified)

### DIFF
--- a/internal/config/loader.go
+++ b/internal/config/loader.go
@@ -1098,7 +1098,11 @@ func roleSettingsDir(role, rigPath string) string {
 	case "crew", "witness", "refinery":
 		return filepath.Join(rigPath, role)
 	case "polecat":
-		return filepath.Join(rigPath, "polecats")
+		// Polecats get settings via EnsureSettingsForRole into their worktree
+		// (polecats/<name>/<rig>/). Claude Code resolves settings from its cwd,
+		// so no --settings flag is needed. The parent polecats/ directory is not
+		// a valid settings location.
+		return ""
 	default:
 		return ""
 	}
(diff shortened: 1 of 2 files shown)


olpie101:
Superseded by #1418

PR #1410, opened 2026-02-13
fix(tmux): target pane 0 explicitly to prevent false zombie detection on split

### LINKED ISSUE #1409: bug: tmux split pane causes false zombie detection and session restart
## Summary

When a user creates a horizontal split (e.g., right-click → 'h') in any agent's tmux session, the agent's Claude Code session is killed and restarted. This happens either ~30 seconds after the split (on the next health check), or immediately if the user runs a command in the new pane that races with a health check cycle.

## Root Cause

`GetPaneCommand()` in `internal/tmux/tmux.go` uses `list-panes -t session -F #{pane_current_command}`, which returns **all panes** in the session. With a single pane this works fine, but after a split the output becomes multi-line (e.g., `"claude\nzsh"`). This value doesn't match any expected process name, so `IsRuntimeRunning()` returns false → the daemon concludes the agent is dead → kills the healthy session and restarts it.

`GetPanePID()` has the same issue — the multi-line PID output fails `strconv.Atoi`.

### Kill chain

1. Daemon/Deacon heartbeat calls health check
2. Health check calls `IsAgentAlive()` → `IsRuntimeRunning()` → `GetPaneCommand()`
3. `GetPaneCommand` returns multi-line output (all panes) instead of pane 0's command
4. No process name matches → agent declared dead
5. `KillExistingSession` kills the entire tmux session (including pane 0 where agent IS running)
6. `Start()` creates a fresh session → agent resets

### Verification

```bash
# Returns all panes (broken after split):
tmux list-panes -t hq-mayor -F '#{pane_current_command}'
# Returns: "claude\nzsh" (two lines)

# Returns pane 0 specifically (correct):
tmux display-message -t hq-mayor:0.0 -p '#{pane_current_command}'
# Returns: "claude"
```

## Fix

Switch `GetPaneCommand()`, `GetPanePID()`, and `GetPaneWorkDir()` from `list-panes` to `display-message -t session:0.0 -p #{format}`, which explicitly targets pane 0 where agent processes always run. For `GetPanePID`, pane ID targets (e.g., `%5`) are preserved for callers like `KillPaneProcesses`.

PR with fix and test incoming.

## Reproduction

1. Start any agent session normally (e.g., mayor)
2. In the agent's tmux session, create a split pane
3. Wait ~30 seconds, or run any command in the new pane
4. Observe: agent's Claude Code session in pane 0 gets killed and restarted

## Impact

Affects any Gas Town agent session where additional tmux panes exist. The mayor is the most visible case since users interact with it directly.

Jonathan-A-White:
## Summary

Fixes #1409 — tmux split panes cause false zombie detection, killing healthy agent sessions.

`GetPaneCommand()`, `GetPanePID()`, and `GetPaneWorkDir()` used `list-panes` which returns all panes in a session. After a user splits the window, the multi-line output (e.g., `"claude\nzsh"`) doesn't match any expected process name, so `IsRuntimeRunning()` returns false and the daemon kills the healthy session.

## Changes

Switch all three functions from `list-panes -t session -F #{format}` to `display-message -t session:0.0 -p #{format}`, which explicitly targets pane 0 where agent processes always run.

For `GetPanePID`, pane ID targets (e.g., `%5`) are detected and used directly, preserving behavior for callers like `KillPaneProcesses`.

## Test plan

- [x] Added `TestGetPaneCommand_MultiPane` — creates a session with `sleep`, splits the window, verifies `GetPaneCommand` and `GetPanePID` still return pane 0's values
- [x] All existing tmux tests pass (pre-existing flaky `TestIsAgentRunning` unrelated)
- [x] `go build ./...` clean
- [x] Manual verification: `display-message -t session:0.0` returns correct pane 0 command after split

### DIFF
--- a/internal/tmux/tmux.go
+++ b/internal/tmux/tmux.go
@@ -1011,7 +1011,11 @@ func (t *Tmux) GetPaneCommand(session string) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	return strings.TrimSpace(out), nil
+	result := strings.TrimSpace(out)
+	if result == "" {
+		return "", fmt.Errorf("empty command for session %s (session may not exist)", session)
+	}
+	return result, nil
 }
 
 // FindAgentPane finds the pane running an agent process within a session.
@@ -1077,16 +1081,18 @@ func (t *Tmux) FindAgentPane(session string) (string, error) {
 
 // GetPaneID returns the pane identifier for a session's first pane.
 // Returns a pane ID like "%0" that can be used with RespawnPane.
+// Targets pane 0 explicitly to be consistent with GetPaneCommand,
+// GetPanePID, and GetPaneWorkDir.
 func (t *Tmux) GetPaneID(session string) (string, error) {
-	out, err := t.run("list-panes", "-t", session, "-F", "#{pane_id}")
+	out, err := t.run("display-message", "-t", session+":0.0", "-p", "#{pane_id}")
 	if err != nil {
 		return "", err
 	}
-	lines := strings.Split(out, "\n")
-	if len(lines) == 0 || lines[0] == "" {
+	result := strings.TrimSpace(out)
+	if result == "" {
 		return "", fmt.Errorf("no panes found in session %s", session)
 	}
-	return lines[0], nil
+	return result, nil
 }
 
 // GetPaneWorkDir returns the current working directory of a pane.
@@ -1097,7 +1103,11 @@ func (t *Tmux) GetPaneWorkDir(session string) (string, error) {
 	if err != nil {
 		return "", err
 	}
-	return strings.TrimSpace(out), nil
+	result := strings.TrimSpace(out)
+	if result == "" {
+		return "", fmt.Errorf("empty working directory for session %s (session may not exist)", session)
+	}
+	return result, nil
 }
 
 // GetPanePID returns the PID of the pane's main process.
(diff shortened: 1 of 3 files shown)


julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @Jonathan-A-White! This is a great catch — false zombie detection from split panes was silently killing healthy agent sessions, and your fix correctly identifies the root cause and applies a clean, targeted solution.

### Original PR Review
Initial dual-model review (Claude + Codex) found the core fix correct — switching from `list-panes` to `display-message -t session:0.0` properly eliminates false zombie detection when users split tmux panes. Four issues were flagged: (1) `web/api.go:1472` had the same `list-panes` vulnerability, (2) `display-message` on invalid sessions could return empty output silently, (3) `GetPaneID` was inconsistent with the other migrated functions, and (4) test PID assertions were weak with no `GetPaneWorkDir` multi-pane coverage.

### Maintainer Changes
All four review findings addressed in a single fixup commit (+52 -22 across 3 files):
- **web/api.go:** Migrated `isClaudeRunningInSession` from `list-panes` to `display-message -t :0.0` with empty-output guard
- **tmux.go:** Hardened `GetPaneCommand`, `GetPanePID`, `GetPaneWorkDir` to return errors on empty `display-message` output (catches invalid/destroyed sessions)
- **tmux.go:** Migrated `GetPaneID` from `list-panes` to `display-message` for consistency with the other three pane-query functions
- **tmux_test.go:** Strengthened multi-pane test with pre/post-split PID and WorkDir stability assertions

### Final Review Status
Second review pass: **approve**. All original findings resolved. Remaining items are non-blocking follow-ups:
- `:0.0` assumes default `base-index` (accepted — gastown controls session creation)
- `hasQuestionInPane` in api.go not pane-targeted (cosmetic, follow-up)
- `getSessionPane` in handoff.go still uses old `list-panes` pattern (cleanup follow-up)

### Review Iterations
2 review passes performed.

---

PR #1414, opened 2026-02-13
fix(mail): soften inject wording so agents process mail before going idle

seanbearden:
## Summary

- Softens the non-urgent mail inject notification from "Do NOT stop or interrupt" to "Continue your current task. When it completes, check these messages before going idle"
- Adds a high-priority tier that prompts agents to process mail promptly at task boundaries
- Keeps urgent tier unchanged (immediate interrupt)

Closes gt-9xfn

## Problem

`gt mail check --inject` (fired by `UserPromptSubmit` hook) tells agents about new mail. Non-urgent mail was framed as "Do NOT stop or interrupt your current task" — agents took this literally and never processed mail, even after finishing work. Discovered by laser/crew/kelcy when PR notifications from 5 crew members sat unread.

### `internal/cmd/mail_check.go`

**Before (2 tiers):**
| Priority | Behavior |
|----------|----------|
| urgent | "URGENT: require immediate attention" (interrupt) |
| high/normal/low | "Do NOT stop or interrupt your current task" (ignored forever) |

**After (3 tiers):**
| Priority | Behavior |
|----------|----------|
| urgent | "URGENT: require immediate attention" (interrupt) — unchanged |
| high | "Continue your current task. When it completes, process these messages before going idle" |
| normal/low | "Continue your current task. When it completes, check these messages before going idle" |

The key behavioral shift: agents still don't interrupt current work, but they now process mail at task boundaries instead of ignoring it entirely.

## Test plan

- [x] `go test ./internal/cmd/...` — all pass
- [ ] Verify agent processes normal mail after finishing a task (instead of going idle)

### DIFF
--- a/internal/cmd/mail_check.go
+++ b/internal/cmd/mail_check.go
@@ -63,8 +63,9 @@ func runMailCheck(cmd *cobra.Command, args []string) error {
 	}
 
 	// Inject mode: notify agent of mail with priority-appropriate framing.
-	// Urgent mail interrupts (agent should act now). Normal mail is delivered
-	// as background context that does NOT interrupt the current task.
+	// Three tiers: urgent interrupts immediately, high-priority is processed
+	// at the next task boundary, normal/low is informational but still
+	// checked before going idle (prevents mail from sitting unread).
 	if mailCheckInject {
 		if unread > 0 {
 			messages, listErr := mailbox.ListUnread()
@@ -73,12 +74,15 @@ func runMailCheck(cmd *cobra.Command, args []string) error {
 				return nil
 			}
 
-			// Separate urgent from non-urgent
-			var urgent, normal []*mail.Message
+			// Separate by priority: urgent interrupts, high is actionable, rest is informational.
+			var urgent, high, normal []*mail.Message
 			for _, msg := range messages {
-				if msg.Priority == mail.PriorityUrgent {
+				switch msg.Priority {
+				case mail.PriorityUrgent:
 					urgent = append(urgent, msg)
-				} else {
+				case mail.PriorityHigh:
+					high = append(high, msg)
+				default:
 					normal = append(normal, msg)
 				}
 			}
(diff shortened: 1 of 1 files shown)


julianknutsen:

### Decision: approve

Thanks for tackling this, @seanbearden — the "agents never process mail" problem was a real pain point (5 crew members' PR notifications sitting unread!), and this is a clean, well-scoped fix. The three-tier priority approach is a nice improvement over the binary urgent/everything-else split, and it aligns well with the existing `PriorityHigh` semantics already used throughout the codebase. All three fix validations confirmed correct.

Only minor/nit findings below — none blocking. Respond with **fix**, **defer**, or **disagree** for each if you'd like, but these are all non-blocking.

---

### Fixes Validated
- **Non-urgent mail never processed by agents:** Confirmed correct. The old "Do NOT stop or interrupt" phrasing caused agents to literally never process mail after task completion; new wording correctly conveys "don't interrupt, but do process afterward."
- **High-priority mail had no distinct treatment:** Confirmed correct. `mail_check.go` previously collapsed high into the same bucket as normal/low, despite `PriorityHigh` being distinguished in 23+ other callsites. New tier correctly separates it.
- **Urgent behavior preserved:** Confirmed correct. Urgent tier still emits immediate-interrupt language; contract unchanged.

---

### Minor: "normal-priority" label misrepresents low/backlog messages

The `normal` bucket includes low/backlog priorities via the `default` case, but when the high-tier block shows a suffix count, it labels them "normal-priority message(s)." This is technically inaccurate — they could be low or backlog priority.

**Suggested fix:** Change to neutral wording like "non-urgent message(s)" or "additional message(s)."

### Minor: High-priority messages hidden behind urgent when both coexist

When urgent + high messages arrive together, high-priority messages get folded into the `other` count as "non-urgent message(s) — read after current task," losing their distinct "process these messages before going idle" framing. This is the same behavior as before (not a regression), but partially defeats the new tier's intent in that edge case.

**Suggested fix:** Non-blocking. If urgent+high coexistence is common in practice, a follow-up could list high-priority messages separately in the urgent block.

### Minor: No direct tests for inject output branching

The three-tier branching logic is unverified by direct tests — nearest tests validate command presence only. This is a pre-existing gap, not introduced by this PR.

**Suggested fix:** Follow-up tests covering urgent+others, high+normal, and normal-only inject cases would be valuable but aren't blocking.

### Nit: "process" vs "check" verb distinction

The sole mechanism differentiating high from normal priority in inject output is "process these messages" vs "check these messages." Whether LLM agents reliably distinguish these verbs is an empirical question.

**Suggested fix:** None needed now — reasonable starting point. Can be tuned based on observed agent behavior.

---

### Pre-Merge Checklist
- [x] All findings are minor/nit — no blockers or majors
- [ ] Consider neutral wording for "normal-priority" label in high-tier suffix (non-blocking)
- [ ] Consider follow-up issue for high-priority framing when urgent messages coexist (non-blocking)

---

PR #1421, opened 2026-02-14
fix: use town name for tmux session names (hq-gfq)

kaushikNaarayan:
## Summary
- `gt mayor start` hardcoded tmux session name to `hq-mayor`, preventing multiple towns from running concurrent mayor sessions
- Session names now use town name from `town.json` (e.g., `redos-mayor`, `nyx-mayor`) to support multi-town installations
- Also fixes session naming for deacon, daemon, and other town-level agents

## Test plan
- [x] Unit tests for session name generation (`session/names_test.go`)
- [x] Identity tests for town name resolution (`session/identity_test.go`)
- [x] Orphan check tests updated for new naming pattern
- [ ] Manual: start mayor in two different towns concurrently

### DIFF
--- a/internal/beads/fields.go
+++ b/internal/beads/fields.go
@@ -705,6 +705,8 @@ func FormatRoleConfig(config *RoleConfig) string {
 
 // ExpandRolePattern expands placeholders in a pattern string.
 // Supported placeholders: {town}, {rig}, {name}, {role}
+// Note: {town_name} (the town identifier from town.json) should be pre-expanded
+// by callers before passing to this function, as it requires session package access.
 func ExpandRolePattern(pattern, townRoot, rig, name, role string) string {
 	result := pattern
 	result = strings.ReplaceAll(result, "{town}", townRoot)
--- a/internal/cmd/root.go
+++ b/internal/cmd/root.go
@@ -10,6 +10,7 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/steveyegge/gastown/internal/cli"
 	"github.com/steveyegge/gastown/internal/config"
+	"github.com/steveyegge/gastown/internal/session"
 	"github.com/steveyegge/gastown/internal/style"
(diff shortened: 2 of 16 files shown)


steveyegge:
## Review: Needs Changes - Safety Regression

The multi-tenancy problem being solved is legitimate. However, the suffix-only session matching introduces a **safety regression**.

**Critical issue:** `isGasTownSession()` uses `strings.HasSuffix(sess, "-mayor")` etc., meaning ANY tmux session ending in `-mayor`, `-deacon`, `-boot`, or `-overseer` will be treated as Gas Town. A user session named `project-boot` could be **killed by `gt doctor --fix`**. The previous `^(gt|hq)-` prefix match was safe; the new suffix-only approach is not.

**Other issues:**
- `ParseSessionName("unrelated-project-mayor")` would succeed - same false-positive problem
- Boot session round-trip is broken: parse("redos-boot") produces {RoleDeacon, "boot"}, but .SessionName() returns "redos-deacon" (wrong)
- Duplicate `isGasTownSession` in both tmux.go and orphan_check.go - should be shared
- `currentTownName` global has no synchronization (data race risk)

**Recommended fixes:**
1. Use the known town name to build precise matching (e.g., sess == session.MayorSessionName())
2. Extract `isGasTownSession` into the session package
3. Add input validation to SetTownName (regex for alphanumeric+hyphens)
4. Fix doc comment examples that use impossible values like "gt-deacon"

steveyegge:
Closing — session naming has been substantially reworked on main since this PR was opened (ParseSessionName for hyphenated rigs, session name validation, prefix corrections). This PR would conflict with and partially duplicate that work. Thanks for the contribution!

steveyegge:
To clarify — the core idea here (dynamic town-name prefixes instead of hardcoded `hq-`) was solid and addressed a real problem. The work just happened to land through a different path while this PR was open. We appreciate you identifying the issue and putting together a thorough fix. Hope to see more contributions!

PR #1440, opened 2026-02-15
fix(tmux): tolerate 'server exited unexpectedly' error

sauerdaniel:
## Summary
- Treat `server exited unexpectedly` stderr like `no server running` (ErrNoServer)
- Make `KillSessionWithProcesses` and `KillSessionWithProcessesExcluding` tolerate ErrNoServer in addition to ErrSessionNotFound

## Why
The tmux server can exit unexpectedly during session cleanup, causing flaky test failures like:
- `TestEnsureSessionFresh_ZombieSession`
- `TestKillSessionWithProcesses`

These tests were failing with "tmux kill-session: server exited unexpectedly" when the server died mid-cleanup.

## Validation
```
go test ./internal/tmux -run 'TestEnsureSessionFresh_ZombieSession|TestKillSessionWithProcesses'
ok  github.com/steveyegge/gastown/internal/tmux  29.371s
```

### DIFF
--- a/internal/tmux/tmux.go
+++ b/internal/tmux/tmux.go
@@ -87,7 +87,8 @@ func (t *Tmux) wrapError(err error, stderr string, args []string) error {
 	// Detect specific error types
 	if strings.Contains(stderr, "no server running") ||
 		strings.Contains(stderr, "error connecting to") ||
-		strings.Contains(stderr, "no current target") {
+		strings.Contains(stderr, "no current target") ||
+		strings.Contains(stderr, "server exited unexpectedly") {
 		return ErrNoServer
 	}
 	if strings.Contains(stderr, "duplicate session") {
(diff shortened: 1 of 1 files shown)


julianknutsen:

### Decision: approve

Thanks for tracking this down, Daniel! The "server exited unexpectedly" race during cleanup is a real and annoying source of flaky test failures, and this is a clean, well-scoped fix. The approach of extending the existing `ErrNoServer` tolerance pattern at both the early and final kill paths is exactly right — both reviewers confirmed all four fix sites are correct with no regressions.

---

### Fixes Validated
- **`server exited unexpectedly` not mapped to `ErrNoServer`:** Confirmed correct. New stderr string correctly added to the existing server-unavailability detection block in `wrapError`.
- **`KillSessionWithProcesses*` not tolerating `ErrNoServer`:** Confirmed correct at all 4 sites (2 functions × 2 error paths each). Cleanup is now idempotent under server-race conditions.

---

### Minor: Missing `TestWrapError` case for new error string

The existing `TestWrapError` table test doesn't include `"server exited unexpectedly"` → `ErrNoServer`. While the fix is clearly correct, adding this case would prevent a future refactor from silently breaking the classification.

**Suggested fix:** Add `{"server exited unexpectedly", ErrNoServer}` to the test table in `tmux_test.go:196`.

---

### Pre-Merge Checklist
- [ ] Add `TestWrapError` case for `"server exited unexpectedly"` (non-blocking)

---

PR #1441, opened 2026-02-15
fix(doctor): detect and fix suboptimal routes that rely on beads redirect

julianknutsen:
## Summary

Doctor's routes-config check now detects and rewrites legacy rig routes that rely on `.beads/redirect` resolution instead of pointing directly to the canonical beads path. This works around the beads resolver bug (beads#1749) where `resolveRedirect` in routing resolves relative to `.beads/` instead of its parent.

## Related Issue

Workaround for beads#1749 — redirect resolution bug in `resolveRedirect` (routes.go) uses wrong base directory.

## Changes

- **Detect suboptimal routes in `Run`:** When a route prefix exists but points to a rig root (e.g., `"crom"`) that has a `.beads/redirect` instead of the canonical path (e.g., `"crom/mayor/rig"`), flag it as suboptimal with an actionable detail message
- **Rewrite suboptimal routes in `Fix`:** `gt doctor --fix` rewrites redirect-dependent routes to the canonical path, bypassing the resolver bug entirely. Gated by `isRedirectDependent` (only rewrites legacy redirect patterns) and `hasRealBeadsDir` (only rewrites to paths with real `.beads` directories)
- **Duplicate prefix guard:** Skip auto-fix for prefixes shared by multiple rigs in `rigs.json` to avoid non-deterministic behavior from map iteration order
- **Double-counting prevention:** Routes flagged as suboptimal are excluded from the validity check loop to prevent appearing in both diagnostic lists
- **Stderr warnings:** Emit operator-facing warnings when a suboptimal route can't be auto-fixed (canonical target has no `.beads` directory) or when duplicate prefixes are detected
- **`routeMap` type change:** Changed from `map[string]bool` to `map[string]int` (prefix → slice index) to support in-place route mutation

## Testing

- [x] Unit tests pass (`go test ./internal/doctor/`)
- [x] Three new test cases in `TestRoutesCheck_SuboptimalRoutes`:
  - `Run detects legacy rig-root route as suboptimal` — verifies StatusWarning with correct detail message
  - `Fix rewrites suboptimal route to canonical path` — verifies route rewrite and StatusOK on re-run
  - `Fix skips rewrite when canonical target has no .beads directory` — verifies route unchanged and stderr warning emitted
- [x] Existing tests unchanged and passing

## Checklist
- [x] No breaking changes — additive detection and fix behavior only
- [x] New rigs via `gt rig add` are unaffected (already route to canonical path)

### DIFF
--- a/internal/doctor/routes_check.go
+++ b/internal/doctor/routes_check.go
@@ -154,11 +154,11 @@ func (c *RoutesCheck) Run(ctx *CheckContext) *CheckResult {
 		if prefix != "" {
 			if existingPath, found := routeByPrefix[prefix]; found {
 				// Route exists but points to a different path than expected.
-				// This happens with legacy rigs where the route points to the rig
-				// root (e.g., "crom") instead of the canonical beads location
-				// (e.g., "crom/mayor/rig"). The rig-root path relies on a .beads/redirect
-				// file, which breaks with bd's routing resolver (beads#1749).
-				if existingPath != expectedPath {
+				// Only flag as suboptimal if the existing path relies on a
+				// .beads/redirect file — this is the specific legacy pattern
+				// broken by beads#1749. Intentional non-canonical routes
+				// (without redirect) are left alone.
+				if existingPath != expectedPath && isRedirectDependent(ctx.TownRoot, existingPath) {
 					suboptimalRoutes = append(suboptimalRoutes, prefix)
 					details = append(details, fmt.Sprintf("Route %s -> %s should be %s -> %s (avoids redirect resolution bug)", prefix, existingPath, prefix, expectedPath))
 				}
@@ -169,6 +169,12 @@ func (c *RoutesCheck) Run(ctx *CheckContext) *CheckResult {
 		}
 	}
 
+	// Build set of suboptimal prefixes to avoid double-counting in validity check
+	suboptimalSet := make(map[string]bool, len(suboptimalRoutes))
+	for _, p := range suboptimalRoutes {
+		suboptimalSet[p] = true
+	}
+
 	// Check each route points to a valid location
 	for _, r := range routes {
 		rigPath := filepath.Join(ctx.TownRoot, r.Path)
@@ -176,6 +182,11 @@ func (c *RoutesCheck) Run(ctx *CheckContext) *CheckResult {
 
 		// Special case: "." path is town root, already checked
 		if r.Path == "." {
+			continue
+		}
+
+		// Skip routes already flagged as suboptimal to avoid double-counting
+		if suboptimalSet[r.Prefix] {
 			continue
 		}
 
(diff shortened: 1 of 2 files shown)


julianknutsen:

### Decision: approve

This is a well-targeted workaround for the beads resolver bug (beads#1749), with careful gating to avoid clobbering intentional non-canonical routes. The detection logic in `Run` correctly scopes to redirect-dependent routes only, and `Fix` has appropriate guards (`hasRealBeadsDir`, `isRedirectDependent`, duplicate prefix detection). Three new test cases cover the detect/fix/skip-fix paths, including stderr assertion for the skip case. No blockers or majors — the remaining findings are all minor/nit and relate to pre-existing edge cases around duplicate prefixes in `routes.jsonl`.

---

### Fixes Validated
- **Legacy redirect-dependent routes invisible to doctor:** Confirmed correct. `Run` now flags prefix routes pointing at rig-root redirect paths with actionable detail messages.
- **Doctor Fix could only add routes, never rewrite stale ones:** Confirmed correct. `routeMap` changed to `map[string]int` to support in-place mutation via slice index.
- **Unsafe rewrite when canonical target has no .beads:** Confirmed correct. `hasRealBeadsDir` guard prevents writing broken routes; stderr warning emitted.

---

### Minor: Duplicate prefix in routes.jsonl — Fix operates on last map entry while runtime uses first match

If `routes.jsonl` somehow contains duplicate prefixes, `routeMap` (built with last-write-wins) would index a different entry than `GetRigPathForPrefix` (first-match iteration). A rewrite could modify the wrong row. However, this requires an already-invalid state that existing checks (`FindConflictingPrefixes`) should catch. The PR's `prefixCount` guard handles the `rigs.json` side of duplicates. Not introduced by this PR.

**Suggested fix:** Follow-up ticket at most. No action needed for merge.

---

### Nit: Missing route + invalid canonical target silently no-ops in Fix
**Source:** both, medium confidence

When a rig has no route AND the canonical path lacks `.beads`, Fix silently does nothing. The rewrite path correctly emits a stderr warning in the equivalent case. The asymmetry could confuse operators, though a rig without canonical `.beads` is likely not fully set up.

**Suggested fix:** Consider adding a warning in the `else` clause for consistency with the rewrite path. Optional.

---

### Nit: `hasRealBeadsDir` name implies redirect exclusion but only checks existence

The function name and comment say "not just a redirect" but it only does `os.Stat(path/.beads)`. Correctness is maintained because callers only pass canonical paths (which by definition have real `.beads` dirs), but the name is mildly misleading.

**Suggested fix:** Consider renaming to `hasBeadsDir` or adding a caller-contract comment. Cosmetic only.

---

### Pre-Merge Checklist
- [x] All new findings are minor/nit severity — no blockers or majors
- [x] Three new code paths have dedicated test coverage
- [x] Fix validations confirmed correct
- [x] `prefixCount` guard handles non-deterministic map iteration

---

PR #1467, opened 2026-02-15
fix(role): add RoleBoot case to ActorString and fix isTownLevelRole

### LINKED ISSUE #1383: Boot role: ActorString() missing RoleBoot case and bare "boot" match in isTownLevelRole
## Summary

Two related identity-mapping gaps for the Boot role introduced in #1370 (GUPP integration). Both stem from Boot's dual identity: `RoleBoot` constant vs `"boot"` string vs `"deacon-boot"` in `BD_ACTOR`.

## Bug 1: `ActorString()` has no `RoleBoot` case

**File:** `internal/cmd/role.go:364-393`

`ActorString()` has no `case RoleBoot`, so it falls through to the `default` branch returning `"boot"`. Meanwhile `BD_ACTOR` is set to `"deacon-boot"` (`internal/config/env.go:57`). If any code path compares `ActorString()` output against `BD_ACTOR` for identity matching (e.g. beads attribution via `sling_helpers.go:313`), it will silently diverge.

**Expected:** `ActorString()` returns a value consistent with the `BD_ACTOR` convention for Boot.

**Fix:** Add explicit `case RoleBoot` to `ActorString()` returning `"deacon-boot"` (or document why the `"boot"` default is intentional and the mismatch with `BD_ACTOR` is acceptable).

## Bug 2: Bare `"boot"` match in `cmd.isTownLevelRole` may be unreachable

**File:** `internal/cmd/molecule_status.go` — `isTownLevelRole()`

`isTownLevelRole` in `cmd` matches bare `"boot"`, but `buildAgentIdentity` only produces `"deacon/boot"` as the agent ID. If no code path ever produces bare `"boot"` as an agentID, this match is dead code.

**Expected:** The match should use the same identifier that `buildAgentIdentity` produces, or be removed if unreachable.

**Fix:** Verify whether any code path produces bare `"boot"` as an agentID. If not, switch the match to `"deacon/boot"` or remove it.

## Context

Found during dual-model review of #1370. Both are minor severity — no known runtime breakage today, but the inconsistency creates a maintenance hazard as more code relies on Boot identity strings.

seanbearden:
## Summary
- Add explicit `case RoleBoot` to `ActorString()` returning `"deacon-boot"` to match `BD_ACTOR` convention set in `config/env.go`
- Fix `isTownLevelRole`: replace unreachable bare `"boot"` match with `"deacon-boot"` (matching `ActorString` output); `buildAgentIdentity` only produces `"deacon/boot"`, never bare `"boot"`
- Add test coverage for `ActorString()` with `RoleBoot` and consistency check against `BD_ACTOR`

Closes #1383

## Test plan
- [x] `go build ./...` passes
- [x] All boot-related tests pass (`TestActorStringBoot`, `TestActorStringConsistentWithBDActorBoot`, `TestIsTownLevelRoleBoot`, `TestParseRoleStringBoot`, `TestGetRoleHomeBoot`)
- [x] Full `go test ./internal/cmd/` passes with no regressions

### DIFF
--- a/internal/cmd/molecule_status.go
+++ b/internal/cmd/molecule_status.go
@@ -993,7 +993,7 @@ func outputMoleculeCurrent(info MoleculeCurrentInfo) error {
 func isTownLevelRole(agentID string) bool {
 	return agentID == "mayor" || agentID == "mayor/" ||
 		agentID == "deacon" || agentID == "deacon/" ||
-		agentID == "boot" || agentID == "deacon/boot"
+		agentID == "deacon/boot" || agentID == "deacon-boot"
 }
 
 // extractMailSender extracts the sender from mail bead labels.
--- a/internal/cmd/role.go
+++ b/internal/cmd/role.go
@@ -405,6 +405,8 @@ func (info RoleInfo) ActorString() string {
 			return fmt.Sprintf("%s/crew/%s", info.Rig, info.Polecat)
 		}
 		return "crew"
+	case RoleBoot:
+		return "deacon-boot"
 	default:
 		return string(info.Role)
 	}
--- a/internal/cmd/role_boot_test.go
+++ b/internal/cmd/role_boot_test.go
@@ -50,14 +50,15 @@ func TestIsTownLevelRoleBoot(t *testing.T) {
 		agentID string
 		want    bool
 	}{
-		{"boot", true},
 		{"deacon/boot", true},
+		{"deacon-boot", true},
 		{"mayor", true},
 		{"mayor/", true},
 		{"deacon", true},
 		{"deacon/", true},
 		{"gastown/witness", false},
 		{"west/boot", false},
+		{"boot", false}, // bare "boot" is not a valid agentID
 	}
 
 	for _, tt := range tests {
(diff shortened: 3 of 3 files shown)


julianknutsen:

### Decision: approve

Thanks for fixing this, @seanbearden! The boot role identity mismatch was a real problem — `ActorString()` returning bare `"boot"` while `BD_ACTOR` expected `"deacon-boot"` meant event attribution was silently wrong for boot sessions. Your fix is clean, correctly scoped, and well-tested. Both reviewers independently verified that removing bare `"boot"` from `isTownLevelRole` is safe (callers never produce it) and that the new `ActorString` case matches the `BD_ACTOR` convention.

A couple of minor items below — none are blocking.

---

### Major: `buildAgentBeadID` missing `RoleBoot` case (pre-existing, not blocking)

Both reviewers flagged that `buildAgentBeadID` in `molecule_status.go:65-96` has no `RoleBoot` case — it falls through to `default:` returning `""`. This means `checkStopSlungWork()` in `signal_stop.go:146` silently skips slung-work detection for boot sessions. This bug predates your PR and isn't worsened by it, but it's the most significant remaining gap in boot role support.

**Suggested fix:** File a follow-up issue to add `case RoleBoot` to `buildAgentBeadID`. Not needed in this PR.

---

### Minor: `ActorString()` docstring doesn't mention `"deacon-boot"` convention

The docstring above `ActorString()` describes town-level agents as simple names (`"mayor"`, `"deacon"`), but boot now returns the hyphenated `"deacon-boot"` form. Minor documentation drift.

**Suggested fix:** Add a note to the docstring that boot uses `"deacon-boot"` (matching `BD_ACTOR`).

---

### Nit: Consistency test hardcodes `BD_ACTOR` value

`TestActorStringConsistentWithBDActorBoot` claims to verify consistency with `BD_ACTOR` in `config/env.go` but hardcodes `"deacon-boot"` instead of reading the actual value. If someone later changes `BD_ACTOR`, the test would still pass while consistency is broken.

**Suggested fix:** Import the actual `BD_ACTOR` value from config, or add a comment acknowledging the snapshot.

---

### Pre-Merge Checklist
- [x] No new blockers — all fixes validated as correct
- [ ] (Optional) Update `ActorString()` docstring
- [ ] (Follow-up) File issue for `buildAgentBeadID` missing `RoleBoot` case

---

PR #1469, opened 2026-02-15
fix(rig): init .beads/ on adopt when no existing database found

### LINKED ISSUE #1355: gt rig add --adopt doesn't initialize .beads/ when no existing beads database found
## Bug Description

`gt rig add <name> --adopt` does not create a `.beads/` directory at the rig level when neither `<rig>/.beads` nor `<rig>/mayor/rig/.beads` already exists. The normal `gt rig add` (non-adopt) path calls `mgr.initBeads()` explicitly, but `--adopt` assumes an existing beads database and only re-initializes if `metadata.json` is missing inside an existing `.beads/`.

This leaves the rig in a degraded state where multiple subsystems silently fail.

## Steps to Reproduce
1. Create a local git repo with no `.beads/` directory
2. `gt install ~/gt`
3. `mkdir -p ~/gt/testrig/refinery/rig && git clone <repo> ~/gt/testrig/refinery/rig`
4. `cd ~/gt && gt rig add testrig --adopt --url <url>`
5. `ls ~/gt/testrig/.beads` → No such file or directory

## Expected Behavior

`gt rig add --adopt` should initialize `.beads/` (via `bd init`) when no existing beads database is found, matching the behavior of the normal `gt rig add` path.

## Actual Behavior

`.beads/` is never created. This causes:

1. **Rig not recognized as valid** — `.beads/` existence is the heuristic for rig detection in `up.go:507`, `costs.go:446`, doctor `priming_check.go:212`, and `stale_beads_redirect_check.go:340`. All silently skip the rig.
2. **No beads database** — `runRigAdopt()` (rig.go:654-740) only enters the `bd init` path if it finds an existing `.beads/` directory. If neither candidate exists, the loop body is never entered.
3. **Worktree redirects fail** — `beads_redirect.go:182-208`: crew/polecat/refinery worktrees can't compute redirect targets → ERROR.
4. **Formulas not provisioned** — Embedded formulas for agents are missing.
5. **PRIME.md not provisioned** — Agents lack Gas Town context fallback.
6. **Routes may point to non-existent directory** — Cross-rig routing breaks.
7. **Doctor checks are silent** — No warnings about missing `.beads/` because the rig isn't recognized as valid in the first place.

A subsequent `gt doctor --fix` self-heals via the `beads-redirect` check, but the gap between `--adopt` and the first `doctor --fix` leaves the rig broken.

## Root Cause

In `internal/cmd/rig.go`, `runRigAdopt()` lines 655-662:

```go
candidates := []string{
    filepath.Join(rigPath, ".beads"),
    filepath.Join(rigPath, "mayor", "rig", ".beads"),
}
for _, candidate := range candidates {
    if !dirExists(candidate) {
        continue  // ← skips both candidates when neither exists
    }
    // ... bd init logic only runs if candidate exists
}
```

The normal `runRigAdd()` path explicitly calls `mgr.initBeads()` at line 380, which always creates the database.

## Environment
- Discovered during manual testing of PR #1082

## Additional Context

Workaround: run `gt doctor --fix` after `gt rig add --adopt` to bootstrap `.beads/`.

seanbearden:
## Summary
- Add fallback `InitBeads()` call in `runRigAdopt()` when no existing `.beads/` candidate is found, matching the behavior of the normal `gt rig add` path
- Export `InitBeads()` from the `rig` package so the `cmd` layer can call it directly
- Add test coverage for candidate detection logic and fallback init decision

Closes #1355

## Test plan
- [x] `go build ./...` passes
- [x] `go test ./internal/cmd/ -run TestRigAdopt` — all candidate detection and fallback tests pass
- [x] `go test ./internal/cmd/` — full cmd package tests pass
- [x] `go test ./internal/rig/` — full rig package tests pass (InitBeads rename)
- [ ] Manual test: `gt rig add <name> --adopt` on a repo with no `.beads/` → verify `.beads/` is created

### DIFF
--- a/internal/cmd/rig.go
+++ b/internal/cmd/rig.go
@@ -723,10 +723,12 @@ func runRigAdopt(_ *cobra.Command, args []string) error {
 		filepath.Join(rigPath, ".beads"),
 		filepath.Join(rigPath, "mayor", "rig", ".beads"),
 	}
+	foundBeadsCandidate := false
 	for _, beadsDir := range beadsDirCandidates {
 		if _, err := os.Stat(beadsDir); err != nil {
 			continue
 		}
+		foundBeadsCandidate = true
 
 		// Detect prefix: try dolt backend first, fall back to metadata.json, then issues.jsonl.
 		// With dolt, metadata.json survives clone (dolt/ is gitignored since bd v0.50+).
@@ -845,6 +847,16 @@ func runRigAdopt(_ *cobra.Command, args []string) error {
 			}
 		}
 		break
+	}
+
+	// If no existing .beads/ candidate was found, initialize a fresh database
+	// to match the behavior of the normal (non-adopt) gt rig add path.
+	if !foundBeadsCandidate && result.BeadsPrefix != "" {
+		if err := mgr.InitBeads(rigPath, result.BeadsPrefix); err != nil {
+			fmt.Printf("  %s Could not init beads database: %v\n", style.Warning.Render("!"), err)
+		} else {
+			fmt.Printf("  %s Initialized beads database\n", style.Success.Render("✓"))
+		}
 	}
 
 	// Print results
--- a/internal/cmd/rig_adopt_beads_test.go
+++ b/internal/cmd/rig_adopt_beads_test.go
@@ -1,0 +1,121 @@
+package cmd
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+// TestRigAdoptBeadsCandidateDetection verifies the .beads/ candidate detection
+// logic used by runRigAdopt to decide whether to initialize a fresh database.
+func TestRigAdoptBeadsCandidateDetection(t *testing.T) {
+	tests := []struct {
+		name           string
+		setupDirs      []string // directories to create under rigPath
+		wantFoundBeads bool     // whether any candidate should be found
+	}{
+		{
+			name:           "no beads directory exists",
+			setupDirs:      nil,
+			wantFoundBeads: false,
+		},
+		{
+			name:           "rig-level .beads exists",
+			setupDirs:      []string{".beads"},
+			wantFoundBeads: true,
+		},
+		{
+			name:           "mayor/rig/.beads exists (tracked beads)",
+			setupDirs:      []string{"mayor/rig/.beads"},
+			wantFoundBeads: true,
+		},
+		{
+			name:           "both candidates exist",
+			setupDirs:      []string{".beads", "mayor/rig/.beads"},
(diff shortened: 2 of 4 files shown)


julianknutsen:

### Decision: approve

Good catch — `gt rig add --adopt` on a repo with no `.beads/` directory was silently leaving the rig without a beads database, causing confusing downstream failures. The fix correctly delegates to `InitBeads()` which handles prefix validation, BEADS_DIR env isolation, tracked-beads redirect, custom types configuration, and fallback config creation. Clean and well-scoped.

No blocking or major issues found. Three minor observations for awareness:

---

### Minor: Two divergent init paths in `runRigAdopt`

The inline init (lines 870-887, for existing `.beads/` needing re-init after clone) runs raw `bd init` directly, while the new fallback uses `mgr.InitBeads()`. The `InitBeads` path is more thorough — it validates the prefix format, filters `BEADS_DIR` from the environment to prevent parent-directory discovery, configures custom types, and creates a fallback `config.yaml` if `bd` isn't installed. The inline path skips all of these. Pre-existing divergence, not introduced by this PR, but the two paths could drift further over time.

**Suggested fix:** Consider consolidating the inline init to also use `mgr.InitBeads()` in a follow-up.

---

### Minor: Tests validate logic copies, not actual code paths

`TestRigAdoptBeadsCandidateDetection` replicates the `os.Stat` candidate loop inline rather than calling `runRigAdopt`. `TestRigAdoptFallbackInitNeeded` replicates the boolean condition `!foundBeadsCandidate && beadsPrefix != ""` rather than testing through the real function. If the logic in `runRigAdopt` changes, the tests could pass while the production code regresses.

**Suggested fix:** Consider integration-style tests that invoke the actual adopt path with a mock filesystem.

---

### Nit: Non-fatal error handling differs from `AddRig`
**Source:** both, low confidence

`AddRig` treats `InitBeads` failure as fatal (`return nil, fmt.Errorf(...)`). The adopt path prints a warning and continues. This is likely intentional — adopt is about registering an existing rig, and beads failure shouldn't block that — but the asymmetry could surprise users who expect adopt to produce a fully functional rig.

**Suggested fix:** No change needed if intentional. Consider a follow-up note in `gt rig add --adopt --help` that beads initialization is best-effort.

---

### Pre-Merge Checklist
- [x] No blockers or majors
- [x] Fallback correctly uses `InitBeads()` (not raw `bd init`)
- [x] `InitBeads` export is clean (no API contract changes)
- [ ] *Optional:* Consolidate inline init to use `InitBeads()` too
- [ ] *Optional:* Add integration-level adopt test

---

Issue #1230, opened 2026-02-07
Race conditions in TUI feed, web options cache, and feed curator


steveyegge:
Eight race conditions identified across subsystems:

1. TUI `feed/model.go:81` - `closeOnce`/`done` channel accessed without sync
2. TUI `feed/model.go:188-189` - `m.events` slice mutated during `addEvent` without guard
3. TUI `convoy/model.go:256-271` - `cursorToConvoyIndex` iterates `m.convoys` which could be modified concurrently
4. Web `api.go:466-475` - options cache RWMutex releases read lock before cache is fully consumed
5. Web `api.go:483-555` - multiple goroutines call `runGtCommand` without synchronization
6. Feed `curator.go:105-110` - `bufio.Reader` without locks, concurrent `Start()` calls race on file pointer
7. Feed `curator.go:191-217` - `readRecentFeedEvents` reads while `writeFeedEvent` appends, partial line read possible
8. Witness `manager.go:123-127` - role config loading not thread-safe across rigs

Priority: P2
Found by: gastown/crew/max code review

seanbearden:
## All 8 race conditions addressed

Five PRs collectively fix every item in this issue. Each PR includes race-detector tests (`go test -race`) validating the fix.

### PR mapping

| # | Race condition | Fix | PR |
|---|---------------|-----|-----|
| 1 | `feed/model.go:81` — `closeOnce`/`done` channel accessed without sync | Added `sync.RWMutex` to feed `Model`, protected `View()` with RLock | #1474 |
| 2 | `feed/model.go:188-189` — `m.events` slice mutated during `addEvent` without guard | Extracted `addEventLocked` with clean lock boundaries, mutations under write lock | #1474 |
| 3 | `convoy/model.go:256-271` — `cursorToConvoyIndex` iterates `m.convoys` concurrently | Added `sync.RWMutex`, converted to pointer receivers (copying mutex is UB), protected reads/writes | #1475 |
| 4 | `api.go:466-475` — options cache RWMutex released before cache fully consumed | Hold RLock through `json.Encode` so concurrent writer can't replace cached pointer mid-serialization | #1472 |
| 5 | `api.go:483-555` — `runGtCommand` called without synchronization | Added command execution semaphore (max 12 concurrent) preventing resource exhaustion | #1472 |
[...]

PR #1473, opened 2026-02-15
fix(feed): prevent concurrent Start() and guard feed file access

seanbearden:
## Summary
- Wrap `Start()` in `sync.Once` so concurrent callers don't spawn multiple `run()` goroutines that would produce duplicate feed events (race 6)
- Add in-process `sync.Mutex` (`feedMu`) to coordinate `readRecentFeedEvents` and `writeFeedEvent` across goroutines within the same process — the existing `flock` only coordinates across processes (race 7)

Part of #1230 (items 6-7)

## Test plan
- [x] `go build ./...` passes
- [x] `TestCurator_ConcurrentStartIsIdempotent` — 10 concurrent Start() → exactly 1 feed event
- [x] `TestCurator_ConcurrentFeedReadWrite` — 20 concurrent reader/writer goroutines → valid JSONL, no data races
- [x] All existing `curator_test.go` tests pass with `-race` (16/16)

### DIFF
--- a/internal/feed/curator.go
+++ b/internal/feed/curator.go
@@ -47,7 +47,8 @@ type Curator struct {
 	ctx             context.Context
 	cancel          context.CancelFunc
 	wg              sync.WaitGroup
-	startOnce       sync.Once // prevents concurrent Start() calls from spawning multiple goroutines
+	startOnce sync.Once // prevents concurrent Start() calls from spawning multiple goroutines
+	startErr  error     // result of the one-shot Start; visible to all callers via sync.Once happens-before
 
 	// feedMu guards in-process access to the feed file. The flock in
 	// readRecentFeedEvents/writeFeedEvent coordinates across processes;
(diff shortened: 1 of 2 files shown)


julianknutsen:
## Maintainer Adoption Review

### Original PR Review
Both Claude and Codex reviewers independently identified one major issue: `startErr` was a function-local variable inside `Start()`, meaning concurrent or subsequent callers would silently receive `nil` after a failed initialization — violating the doc comment's promise of concurrent-call safety. The `sync.Once` guard and `feedMu` mutex additions were confirmed correct. Tests passed including `-race`.

### Maintainer Changes
Promoted `startErr` from function-local to a `Curator` struct field (`c.startErr`) so all `Start()` callers observe the persisted initialization error via `sync.Once`'s happens-before guarantee. Added `TestCurator_StartErrorPersistsAcrossCalls` to verify that a second `Start()` call returns the same error after failure. Fixed misleading test comment about sync.Once semantics. (+32 -6 across 2 files)

### Final Review Status
All clear — no remaining blockers or majors. Both reviewers confirmed the `startErr` fix is correct. Only low/nit observations remain: timing-fragile test pattern (pre-existing in other curator tests, not a regression) and a suggestion to document the non-reentrant `feedMu` contract.

### Review Iterations
2 review passes performed (initial PR-mode review + local branch re-review after maintainer fix).

---

PR #1477, opened 2026-02-15
fix(witness): serialize config resolution to prevent registry corruption

seanbearden:
## Summary
- Add `configMu sync.Mutex` at package level to serialize `ResolveRoleAgentConfig` + `EnsureSettingsForRole` + `roleConfig()` in witness `Start()`
- Concurrent witness starts for different rigs would load rig-specific agents into the global registry, corrupting each other's config lookups
- Moved `roleConfig()` call inside the locked section and `EnsureGitignorePatterns` outside it (doesn't need config lock)

## Context
Part of #1230 race condition audit (item 8: witness manager config resolution race).

Epic: gt-83ug | Bead: gt-6o0s

## Test plan
- [x] All witness package tests pass with `-race`
- [x] Full `go build ./internal/witness/...` succeeds
- [ ] Verify concurrent `gt rig start` for multiple rigs doesn't produce config corruption

### DIFF
--- a/internal/config/loader.go
+++ b/internal/config/loader.go
@@ -9,10 +9,17 @@ import (
 	"path/filepath"
 	"sort"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/steveyegge/gastown/internal/constants"
 )
+
+// resolveConfigMu serializes agent config resolution across all callers.
+// ResolveRoleAgentConfig and ResolveAgentConfig load rig-specific agents
+// into a global registry; concurrent calls for different rigs would corrupt
+// each other's lookups.
+var resolveConfigMu sync.Mutex
 
 var (
 	// ErrNotFound indicates the config file does not exist.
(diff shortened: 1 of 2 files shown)


julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @seanbearden! This PR addresses a real concurrency hazard in the witness config resolution path — concurrent rig starts corrupting each other's agent lookups via the global registry is exactly the kind of subtle race that's hard to catch without deliberate auditing. The mail check inject refactoring with the comprehensive test suite is a nice bonus.

### Original PR Review
The initial dual-model review (Claude + Codex) returned **request_changes** with 1 blocker and 1 major:

1. **Blocker (codex):** `buildWitnessStartCommand` called `BuildAgentStartupCommandWithAgentOverride` which re-entered `ResolveRoleAgentConfig` after the witness mutex was released — the race was still reachable for default witness starts.
2. **Major (claude):** The `configMu` mutex was scoped to `package witness`, but the same race exists in every other caller of `ResolveRoleAgentConfig` (polecat, refinery, deacon, session lifecycle). The mutex belongs in `package config`.
3. **Minor (claude):** `roleConfig()` reads from beads, not the agent registry — didn't need to be inside the critical section.

### Maintainer Changes
Commit `9e49ccd2` (+35 -19 across 2 files):

- Moved serialization mutex (`resolveConfigMu`) into `package config` so all 20+ callers of `ResolveRoleAgentConfig`, `ResolveAgentConfig`, and `ResolveAgentConfigWithOverride` are protected
- Extracted internal lock-free versions (`resolveAgentConfigInternal`, `resolveAgentConfigWithOverrideInternal`) to handle the recursive call chain (`resolveRoleAgentConfigCore` → `ResolveAgentConfig`) without deadlock
- This also fixes the blocker: `buildWitnessStartCommand`'s second resolve call is now independently serialized by the config-level mutex
- Moved `roleConfig()` back outside the critical section
- Removed witness-level `configMu` and `sync` import

### Final Review Status
Iteration review returned **approve**. All original findings resolved. Two minor non-blocking observations:

- **Minor (pre-existing):** Two-resolve window in witness `Start()` means cross-rig agent name collisions are theoretically possible — this is a pre-existing architectural limitation of the global registry, not introduced by this PR. Tracked as follow-up under #1230.
- **Minor:** `formatInjectOutput` empty-slice path returns empty string, but guarded by call site.

All tests pass including `-race`.

### Review Iterations
2 review passes performed.

---

PR #1491, opened 2026-02-15
fix(deacon): surface heartbeat in gt deacon status

### LINKED ISSUE #1265: Witness false-alarm: deacon reported stopped during backoff wait
## Bug

Witness repeatedly sends false "Deacon appears stopped" alerts while deacon is alive and healthy in its backoff wait between patrol cycles.

## Root Cause

Witness checks raw tmux session state which appears inactive during deacon's `await-signal` backoff. Should instead check deacon's heartbeat/last_patrol timestamp to determine health.

## Evidence (2026-02-08 session)

- Multiple false alarms from lowendinsight_get/witness
- Deacon confirmed healthy with 12 patrol cycles completed
- Deacon self-reported: "Deacon is running. Heartbeat just updated."

## Suggested Fix

Witness health check should use `last_patrol` timestamp (or heartbeat) rather than tmux session activity state. If `last_patrol` is within expected patrol interval + backoff max, deacon is healthy.

seanbearden:
## Summary
- Add `--json` flag to `gt deacon status` with heartbeat data (timestamp, age, cycle, freshness)
- Add heartbeat info to human-readable output (age, last action, health status)
- Enables witness to programmatically check deacon health via `gt deacon status --json` instead of relying on raw tmux pane activity

Closes #1265

## Test plan
- [x] `go build ./cmd/gt/` — compiles
- [x] `go test ./internal/cmd/ -run Deacon -v` — existing tests pass
- [ ] Manual: `gt deacon status` shows heartbeat info in human output
- [ ] Manual: `gt deacon status --json` outputs parseable JSON with heartbeat fields

### DIFF
--- a/internal/cmd/deacon_status_test.go
+++ b/internal/cmd/deacon_status_test.go
@@ -1,0 +1,228 @@
+package cmd
+
+import (
+	"encoding/json"
+	"testing"
+	"time"
+)
+
+// TestDeaconStatusJSON_Schema verifies the JSON output schema for gt deacon status --json.
+// This catches schema changes that would break witness parsing.
+func TestDeaconStatusJSON_Schema(t *testing.T) {
+	now := time.Now().UTC()
+
+	out := DeaconStatusOutput{
+		Running: true,
+		Paused:  false,
+		Session: "gt-deacon",
+		Heartbeat: &HeartbeatStatus{
+			Timestamp:  now,
+			AgeSec:     42.5,
+			Cycle:      12,
+			LastAction: "patrol complete",
+			Fresh:      true,
+			Stale:      false,
+			VeryStale:  false,
+		},
+	}
+
+	data, err := json.Marshal(out)
+	if err != nil {
+		t.Fatalf("Marshal DeaconStatusOutput: %v", err)
+	}
+
+	// Parse back as generic map to verify field names
+	var m map[string]interface{}
+	if err := json.Unmarshal(data, &m); err != nil {
+		t.Fatalf("Unmarshal to map: %v", err)
+	}
+
+	// Top-level fields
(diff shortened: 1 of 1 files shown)


julianknutsen:

### Decision: approve

Clean implementation — the refactoring that separates state gathering from rendering (JSON vs human) is a nice structural improvement over the original inline print approach. The JSON schema is well-designed with `omitempty` on optional fields (`heartbeat`, `last_action`), and the test suite provides real contract validation (schema field names, roundtrip, omitempty behavior) rather than trivial struct assertions.

The heartbeat surfacing gives the witness a reliable programmatic health check for deacon monitoring, which is a meaningful operational improvement.

No blocking or major issues found. One minor observation:

---

### Minor: Three independent freshness booleans vs. a single status field

`HeartbeatStatus` exposes `fresh`, `stale`, and `very_stale` as independent booleans, which allows structurally inconsistent states (e.g., all three true). The human output handles this correctly via a priority cascade (`VeryStale > Stale > Fresh`), but JSON consumers need to implement the same cascade. A single `"health": "fresh|stale|very_stale"` string field would be simpler to consume.

The test `TestDeaconStatusJSON_FreshnessStates` documents the expected mutual exclusivity, which helps. This is a minor API design preference, not a correctness issue.

**Suggested fix:** Optional — if the witness parses this field, consider adding a `health` string field alongside (or instead of) the three booleans. Low priority since the current approach works.

---

### Pre-Merge Checklist
- [x] No blockers or majors
- [x] JSON schema tests validate field names and omitempty
- [x] Roundtrip serialization verified
- [x] Heartbeat nil → `heartbeat` field omitted (correct)
- [x] Human output refactored without behavioral changes
- [x] "Attach with" hint moved to end of output (better UX)

---

PR #1492, opened 2026-02-15
fix(deacon): scope zombie/orphan detection to Gas Town workspace

Rome-1:
## Summary

Deacon's zombie and orphan detection kills any Claude Code process it doesn't track—including instances running in repos outside the Gas Town workspace (e.g., `~/GithubRepos/some-project/`). This actively destroys user work.

This PR adds a workspace check before classifying a Claude process as orphaned or zombie. Only processes whose working directory is at or under a Gas Town workspace root (`mayor/town.json` marker) are candidates for cleanup. Everything else is left alone.

Also fixes `null` vs `[]` inconsistency in JSON output for `gt polecat list` and `gt mail inbox`.

## Changes

- Add `getProcessCwd()`: resolves process working directory via `/proc/<pid>/cwd` (Linux) with `lsof` fallback (macOS). Fails safe—returns `""` if cwd can't be determined.
- Add `isInGasTownWorkspace()`: walks up from cwd looking for `mayor/town.json`. Returns `false` when cwd is unknown (don't kill what you can't identify).
- Gate both `FindOrphanedClaudeProcesses()` and `FindZombieClaudeProcesses()` on the workspace check.
- Initialize nil slices to `[]T{}` before JSON marshaling in `polecat list`, `mail inbox`, and `mailbox.go` so empty results serialize as `[]` instead of `null`.
- Handle Linux `/proc/<pid>/cwd` returning path with `" (deleted)"` suffix when worktree is removed.

## Testing

- [x] Unit tests pass (`go test ./internal/util/... ./internal/mail/... ./internal/cmd/...`)
- [x] `go vet` clean
- [x] `TestGetProcessCwd`: verifies cwd resolution against `os.Getwd()` for the current process
- [x] `TestIsInGasTownWorkspace`: creates temp workspace with `mayor/town.json`, tests detection from root and subdirectory, plus negative case outside workspace
- [x] Manual testing: Claude Code instance outside `~/gt/` survives Deacon cycle

## Checklist
- [x] No breaking changes
- [x] Tests included for new functionality

### DIFF
--- a/internal/mail/mailbox.go
+++ b/internal/mail/mailbox.go
@@ -210,13 +210,13 @@ func (m *Mailbox) listLegacy() ([]*Message, error) {
 	file, err := os.Open(m.path)
 	if err != nil {
 		if os.IsNotExist(err) {
-			return nil, nil
+			return make([]*Message, 0), nil
 		}
 		return nil, err
 	}
 	defer func() { _ = file.Close() }() // non-fatal: OS will close on exit
 
-	var messages []*Message
+	messages := make([]*Message, 0)
 	scanner := bufio.NewScanner(file)
 	lineNum := 0
 	for scanner.Scan() {
--- a/internal/util/orphan.go
+++ b/internal/util/orphan.go
@@ -218,18 +218,31 @@ func processExists(pid int) bool {
 }
 
 // getProcessCwd returns the current working directory of a process.
-// On macOS, uses lsof; on Linux, reads /proc/<pid>/cwd.
+// On Linux, reads /proc/<pid>/cwd. On macOS and other Unix, uses lsof.
 // Returns empty string if the cwd cannot be determined.
+//
+// On hardened Linux kernels (Ubuntu default: kernel.yama.ptrace_scope=1),
+// readlink(/proc/<pid>/cwd) fails with EACCES for non-descendant same-user
+// processes. The lsof fallback handles this when lsof is installed (it may
+// be setuid or hold CAP_SYS_PTRACE). If neither method works, "" is returned
+// and the caller fails safe by not killing the process.
 func getProcessCwd(pid int) string {
 	pidStr := strconv.Itoa(pid)
 
-	// Try /proc/<pid>/cwd first (Linux)
+	// Try /proc/<pid>/cwd first (Linux).
+	// Fails on hardened kernels (ptrace_scope>=1) for non-descendant processes.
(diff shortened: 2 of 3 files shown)


julianknutsen:

### Decision: approve

Important safety fix — thank you for catching this. Deacon killing user Claude Code instances outside the workspace was a real data-loss risk. The fail-safe design is exactly right: when cwd can't be determined, the process is left alone. The `getProcessCwd` function handles platform differences cleanly (Linux `/proc`, macOS `lsof`, hardened kernels via `lsof` fallback), and the `" (deleted)"` suffix handling for nuked worktrees is a nice detail.

The JSON `null` → `[]` fixes are a good hygiene improvement — downstream consumers shouldn't need to handle both.

No blocking or major issues found. Two minor observations:

---

### Minor: Workspace check matches ANY Gas Town workspace

`isInGasTownWorkspace` walks up from the process cwd looking for `mayor/town.json` in any ancestor — so it identifies whether the process is in *any* Gas Town workspace, not specifically the workspace whose deacon is running. In a multi-workspace setup (`~/gt/` and `~/gt-staging/`), deacon in workspace A could classify a non-tmux process in workspace B as zombie.

This is largely mitigated by `getTmuxSessionPIDs()` which collects ALL tmux sessions (not just `gt-*`), so a process in workspace B's tmux session would be protected. The unmitigated case — process in workspace B, not in any tmux session — is genuinely orphaned, so cleanup is arguably correct.

**Suggested fix:** Optional follow-up: pass the current workspace root to `isInGasTownWorkspace` and check `strings.HasPrefix(cwd, workspaceRoot)` for exact scoping. Low priority given the tmux session cross-reference.

---

### Nit: No diagnostic logging when workspace check skips a process
**Source:** codex, low confidence

When a Claude process is skipped because it's not in any Gas Town workspace, there's no log output. If an operator wonders "why didn't deacon clean up PID X?", there's no visibility into the workspace check decision.

**Suggested fix:** Add a debug log line: `log.Printf("skipping PID %d: cwd %q not in any Gas Town workspace", pid, cwd)` — only useful if deacon has a verbose/debug mode.

---

### Pre-Merge Checklist
- [x] No blockers or majors
- [x] Fail-safe semantics correct (unknown cwd → don't kill)
- [x] Both orphan and zombie paths gated on workspace check
- [x] `" (deleted)"` suffix handling correct for nuked worktrees
- [x] JSON null → [] fixes consistent across all sites
- [ ] *Optional:* Scope workspace check to current workspace
- [ ] *Optional:* Add debug logging for workspace check decisions

---

PR #1493, opened 2026-02-15
fix(rig): kill tmux sessions on rig remove (#1223)

### LINKED ISSUE #1223: gt rig remove leaves orphaned tmux sessions running
## Summary

`gt rig remove` unregisters the rig from `rigs.json` but doesn't kill associated tmux sessions, leaving orphaned processes.

## Steps to Reproduce

1. `gt rig add echo <url>`
2. `gt rig boot echo`
3. `gt rig remove echo`
4. `tmux list-sessions | grep echo` → sessions still running

## Expected

`gt rig remove` should kill (or offer to kill) the rig's tmux sessions.

seanbearden:
## Summary

- `gt rig remove` now detects running `gt-{rig}-*` tmux sessions before removing a rig
- Without `--force`: lists running sessions and errors with guidance to run `gt rig shutdown` first
- With `--force`: kills all rig sessions via `tmux.KillSessionWithProcesses()` then proceeds with removal
- Adds `findRigSessions()` helper that matches by the `gt-{rig}-` prefix, catching witness, refinery, polecat, and crew sessions in one pass

Closes #1223

## Test plan

- [x] `go build ./...` passes
- [x] `go vet ./...` passes  
- [x] `TestFindRigSessions` — creates real tmux sessions with matching/non-matching names, verifies correct filtering
- [x] `TestFindRigSessions_NoSessions` — verifies empty result for nonexistent rig
- [x] Race detection passes (`go test -race`)

### DIFF
--- a/internal/cmd/rig.go
+++ b/internal/cmd/rig.go
@@ -21,6 +21,7 @@ import (
 	"github.com/steveyegge/gastown/internal/polecat"
 	"github.com/steveyegge/gastown/internal/refinery"
 	"github.com/steveyegge/gastown/internal/rig"
+	"github.com/steveyegge/gastown/internal/session"
 	"github.com/steveyegge/gastown/internal/style"
 	"github.com/steveyegge/gastown/internal/tmux"
 	"github.com/steveyegge/gastown/internal/wisp"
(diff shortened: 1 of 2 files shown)


julianknutsen:
## Post-Merge Review — Dual-Model (Claude + Codex)

### Decision: request_changes (addressed in follow-up)

Nice feature, @seanbearden — detecting running sessions before rig removal is exactly the right guard. The prefix-matching approach with `session.Prefix` is clean, and the tests with real tmux sessions are solid. Two safety issues with the `--force` path need fixing.

**Reviewer disagreement resolved:** Codex flagged prefix collision (rig `foo` matching `foo-bar`) as a blocker. Claude verified that rig names cannot contain hyphens (`rig/manager.go:275`), making this collision impossible. Codex finding dismissed.

---

### Major: --force continues removal after session kill failures
**Source:** both reviewers, high confidence

If `KillSessionWithProcesses` fails for some sessions, the rig is still deregistered from `rigs.json`. Those sessions become invisible to all `gt rig` commands.

### Major: --force bypasses uncommitted-work safety check

`gt rig shutdown` has a two-tier model: `--force` checks uncommitted work, `--nuclear` skips everything. `gt rig remove --force` goes straight to `KillSessionWithProcesses` without checking — equivalent to `--nuclear` without the scary naming.

### Minor: findRigSessions swallows ListSessions errors

Returns empty slice on error, making the safety guard unable to distinguish "no sessions" from "check failed."

---

### Follow-Up

All findings addressed in PR #1510:
- `findRigSessions` now returns `([]string, error)` — tmux errors abort removal in non-force mode
- `--force` path accumulates kill errors and aborts removal if any session survives
- Updated `--force` flag description to warn about potential data loss

---

PR #1513, opened 2026-02-15
fix(tmux): add -u flag to remaining client-side tmux callsites

### LINKED ISSUE #1219: fix: tmux attach renders UTF-8 as underscores (missing -u flag)
## Bug Description

`gt mayor attach` (and all other attach/switch paths) renders UTF-8 characters — box drawing, emojis, Unicode symbols — as underscores. Running `tmux -u new` directly works fine.

## Steps to Reproduce

1. Start a Gastown container
2. `gt mayor start`
3. `gt mayor attach`
4. Observe: box-drawing characters, emojis, and other Unicode render as underscores

## Expected Behavior

UTF-8 characters render correctly, matching the behavior of `tmux -u new`.

## Actual Behavior

All non-ASCII characters display as underscores in the attached session.

## Root Cause

Two issues in the Go source:

1. **`internal/tmux/tmux.go`** — The `run()` method calls `exec.Command("tmux", args...)` without the `-u` flag. This means the tmux server may start without UTF-8 awareness.

2. **`internal/cmd/crew_helpers.go`** — `attachToTmuxSession()` calls tmux without `-u` and uses `cmd.Run()` (child process) instead of `syscall.Exec` (process replacement). This means the attaching tmux client doesn't get the `-u` flag, and doesn't get direct terminal control.

## Proposed Fix

**`internal/tmux/tmux.go`** — Prepend `-u` to all tmux invocations:
```go
func (t *Tmux) run(args ...string) (string, error) {
    cmd := exec.Command("tmux", append([]string{"-u"}, args...)...)
```

**`internal/cmd/crew_helpers.go`** — Add `-u` and use `syscall.Exec`:
```go
func attachToTmuxSession(sessionID string) error {
    tmuxPath, err := exec.LookPath("tmux")
    if err != nil {
        return fmt.Errorf("tmux not found: %w", err)
    }
    var args []string
    if os.Getenv("TMUX") != "" {
        args = []string{"tmux", "-u", "switch-client", "-t", sessionID}
    } else {
        args = []string{"tmux", "-u", "attach-session", "-t", sessionID}
    }
    return syscall.Exec(tmuxPath, args, os.Environ())
}
```

## Environment

- OS: Linux (Docker container, arm64)
- Go: 1.24.6
- Gas Town: 0.5.0
- tmux: 3.4

## Logs / Error Output

No error output — tmux silently renders UTF-8 characters as underscores when `-u` is not passed.

## Additional Context

- The `-u` flag forces tmux to assume UTF-8 support regardless of locale detection
- `syscall.Exec` replaces the Go process with tmux (matching how `tmux -u new` works from a shell), while `cmd.Run()` keeps Go as the parent process
- Dockerfile/entrypoint locale fixes (`LANG=C.utf8`, `set-environment -g`) help but are not sufficient without the binary changes

seanbearden:
## Summary

- Add `-u` flag to 6 bare `exec.Command("tmux", ...)` callsites that bypass `Tmux.run()` and lack UTF-8 support
- Switch `attachToTmuxSession()` from `cmd.Run()` to `syscall.Exec` for direct terminal control
- Completes the fix started in 2c3d6ba (server-side `-u` via `Tmux.run()`)

### Callsites fixed

| File | Function | Command |
|------|----------|---------|
| `helpers.go` | `attachToTmuxSession` | `switch-client` / `attach-session` + `syscall.Exec` |
| `cycle.go` | `cycleSessions` | `switch-client` |
| `polecat_cycle.go` | `cyclePolecatSessions` | `switch-client` |
| `town_cycle.go` | `cycleTownSessions` | `switch-client` |
| `crew_cycle.go` | `cycleCrewSessions` | `switch-client` |
| `handoff.go` | `runHandoff` | `switch-client` |

Note: `witness.go:336` from the maintainer comment no longer has a bare tmux callsite (already fixed or removed).

Per maintainer comment: https://github.com/steveyegge/gastown/issues/1219#issuecomment-3893790505

Closes #1219

## Test plan

- [x] `go build ./...` compiles cleanly
- [x] `go test -short ./internal/cmd/` passes
- [x] 6 files, net +2 lines

### DIFF
--- a/internal/cmd/crew_cycle.go
+++ b/internal/cmd/crew_cycle.go
@@ -79,7 +79,7 @@ func cycleCrewSession(direction int, sessionOverride string) error {
 	targetSession := sessions[targetIdx]
 
 	// Switch to target session
-	cmd := exec.Command("tmux", "switch-client", "-t", targetSession)
+	cmd := exec.Command("tmux", "-u", "switch-client", "-t", targetSession)
 	if err := cmd.Run(); err != nil {
 		return fmt.Errorf("switching to %s: %w", targetSession, err)
 	}
--- a/internal/cmd/cycle.go
+++ b/internal/cmd/cycle.go
@@ -182,7 +182,7 @@ func cycleRigInfraSession(direction int, currentSession, rig string) error {
 	}
 
 	// Switch to target session
-	cmd := exec.Command("tmux", "switch-client", "-t", sessions[targetIdx])
+	cmd := exec.Command("tmux", "-u", "switch-client", "-t", sessions[targetIdx])
 	return cmd.Run()
 }
 
--- a/internal/cmd/handoff.go
+++ b/internal/cmd/handoff.go
@@ -751,7 +751,7 @@ func handoffRemoteSession(t *tmux.Tmux, targetSession, restartCmd string) error 
 	if handoffWatch {
 		fmt.Printf("Switching to %s...\n", targetSession)
 		// Use tmux switch-client to move our view to the target session
-		if err := exec.Command("tmux", "switch-client", "-t", targetSession).Run(); err != nil {
+		if err := exec.Command("tmux", "-u", "switch-client", "-t", targetSession).Run(); err != nil {
 			// Non-fatal - they can manually switch
 			fmt.Printf("Note: Could not auto-switch (use: tmux switch-client -t %s)\n", targetSession)
 		}
--- a/internal/cmd/helpers.go
+++ b/internal/cmd/helpers.go
@@ -76,25 +76,27 @@ func isInTmuxSession(targetSession string) bool {
 
 // attachToTmuxSession attaches to a tmux session.
 // If already inside tmux, uses switch-client instead of attach-session.
+// Uses syscall.Exec to replace the Go process with tmux for direct terminal
+// control, and passes -u for UTF-8 support regardless of locale settings.
+// See: https://github.com/steveyegge/gastown/issues/1219
 func attachToTmuxSession(sessionID string) error {
 	tmuxPath, err := exec.LookPath("tmux")
 	if err != nil {
 		return fmt.Errorf("tmux not found: %w", err)
 	}
 
-	// Check if we're already inside a tmux session
-	var cmd *exec.Cmd
+	// Build args with -u for UTF-8 support
+	var args []string
 	if os.Getenv("TMUX") != "" {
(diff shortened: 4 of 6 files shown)


julianknutsen:

### Decision: approve

Thanks for this cleanup, @seanbearden! The `-u` flag gap on client-side tmux calls was a real source of UTF-8 rendering issues, and this PR methodically closes it across all 6 remaining bare `exec.Command("tmux", ...)` callsites. The `syscall.Exec` upgrade in `attachToTmuxSession` is a nice bonus that aligns with the existing `execAgent`/`execRuntime` pattern. Both reviewers agree this is clean and safe to merge.

Two minor observations below — neither is blocking.

---

### Fixes Validated
- **`-u` on all 5 switch-client callsites:** Confirmed correct. Flag positioned before subcommand, matching `Tmux.run()` convention.
- **`-u` on both branches of `attachToTmuxSession`:** Confirmed correct. Both `switch-client` and `attach-session` paths now include `-u`.
- **`syscall.Exec` in `attachToTmuxSession`:** Confirmed safe. All 5 callers tail-return, no deferred cleanup is skipped, and no `PersistentPostRun` hooks exist.

---

### Minor: `syscall.Exec` on `switch-client` branch is unnecessary process replacement

`switch-client` is a transient, non-interactive command that completes instantly — `syscall.Exec` replaces the entire Go process for it, which is overkill compared to the `cmd.Run()` pattern used by the 5 other `switch-client` sites in this same PR. In practice this is harmless since all callers tail-return `attachToTmuxSession(...)`, but a clarifying comment would help future maintainers understand the intent.

**Suggested fix:** Consider adding a brief comment explaining that both branches intentionally use `syscall.Exec` for simplicity, or split the execution strategy per-branch in a follow-up.

### Minor: Pre-existing `witness.go:336` still lacks `-u`

Both reviewers independently found one remaining bare `attach-session` call at `internal/cmd/witness.go:336` that doesn't include `-u`. This is pre-existing and outside this PR's scope, but worth tracking as a follow-up.

**Suggested fix:** Add `-u` to `witness.go:336` in a follow-up.

---

### Pre-Merge Checklist
- [x] All 5 `switch-client` callsites include `-u`
- [x] `attachToTmuxSession` includes `-u` on both branches
- [x] `syscall.Exec` callers verified: all tail-return, no skipped cleanup
- [ ] (Follow-up) Add `-u` to `witness.go:336`

---

PR #1521, opened 2026-02-16
fix(install): preserve existing configs on re-install

### LINKED ISSUE #1252: bug: gt install unconditionally overwrites rigs.json, wiping rig registrations
## Summary

`gt install` unconditionally overwrites `mayor/rigs.json` with an empty registry (`internal/cmd/install.go:185-194`). There is no check for an existing file. This means any re-run of install (or anything that triggers the install path) nukes all rig registrations, requiring manual `gt rig add --adopt` for each rig.

## Impact

- Rigs silently disappear from `gt rig list`
- `gt sling` to a rig fails because the rig is unknown
- Users must manually re-register every rig after each install
- No warning is given that existing registrations are being destroyed

## Root Cause

```go
// internal/cmd/install.go:185-194
rigsConfig := &config.RigsConfig{
    Version: config.CurrentRigsVersion,
    Rigs:    make(map[string]config.RigEntry),  // Always empty
}
rigsPath := filepath.Join(mayorDir, "rigs.json")
if err := config.SaveRigsConfig(rigsPath, rigsConfig); err != nil {
    return fmt.Errorf("writing rigs.json: %w", err)
}
```

No check for existing `rigs.json` before overwriting.

## Suggested Fix

Before writing, check if `rigs.json` already exists. If it does, either:
1. Skip the write entirely (preferred — don't clobber existing config)
2. Load existing entries and merge, only adding defaults if missing

Same pattern should be applied to `town.json` and any other config files created during install.

## Reproduction

1. `gt install` (sets up town with rigs registered)
2. Register rigs: `gt rig add foo --adopt`, `gt rig add bar --adopt`
3. `gt rig list` → shows foo, bar
4. `gt install` again (or anything triggering install path)
5. `gt rig list` → empty, all registrations gone

seanbearden:
## Summary
- Wraps `town.json` and `rigs.json` creation in `os.Stat` existence checks
- Re-running `gt install` no longer clobbers existing rig registrations or town metadata
- Existing files are preserved with a `• already exists, preserving` message

## Test plan
- [ ] Manual: run `gt install`, verify configs created; run again, verify configs preserved

Closes #1252

### DIFF
--- a/internal/cmd/install.go
+++ b/internal/cmd/install.go
@@ -72,7 +72,7 @@ Examples:
 }
 
 func init() {
-	installCmd.Flags().BoolVarP(&installForce, "force", "f", false, "Overwrite existing HQ")
+	installCmd.Flags().BoolVarP(&installForce, "force", "f", false, "Re-run install in existing HQ (preserves town.json and rigs.json)")
 	installCmd.Flags().StringVarP(&installName, "name", "n", "", "Town name (defaults to directory name)")
 	installCmd.Flags().StringVar(&installOwner, "owner", "", "Owner email for entity identity (defaults to git config user.email)")
 	installCmd.Flags().StringVar(&installPublicName, "public-name", "", "Public display name (defaults to town name)")
@@ -184,7 +184,7 @@ func runInstall(cmd *cobra.Command, args []string) error {
 
 	// Create town.json in mayor/ (only if it doesn't already exist).
 	townPath := filepath.Join(mayorDir, "town.json")
-	if _, err := os.Stat(townPath); os.IsNotExist(err) {
+	if townInfo, err := os.Stat(townPath); os.IsNotExist(err) {
 		townConfig := &config.TownConfig{
 			Type:       "town",
 			Version:    config.CurrentTownVersion,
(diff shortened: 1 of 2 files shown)


julianknutsen:
## Maintainer Adoption Review

### Original PR Review
Initial dual-model review (Claude + Codex) found the core preserve logic correct but identified three issues: (1) `os.Stat` errors other than `ENOENT` were silently swallowed in the `else` branch, masking real filesystem faults (major, both reviewers); (2) `--force` flag help text still said "Overwrite existing HQ" despite new preserve semantics (minor); (3) no test coverage for the preserve behavior (minor).

### Maintainer Changes
Added three-way `os.Stat` error handling (`IsNotExist` → create, `err != nil` → return error, `IsRegular()` check → reject non-file paths, `else` → preserve), updated `--force` help text, and added two integration tests: `TestInstallForcePreservesConfigs` (sentinel injection proving data survives `--force`) and `TestInstallForceRejectsNonRegularConfigs` (directory-as-file edge case).

### Final Review Status
All clear — no remaining issues found. Three review passes performed. Both reviewers approve. The `IsRegular()` guard and error propagation match existing codebase patterns (`createTownRootCLAUDEmd`, `EnsureDaemonPatrolConfig`).

### Review Iterations
3 review passes performed.

---

PR #1523, opened 2026-02-16
feat(doltserver): configurable remote Dolt server via env vars

pae23:
## Summary

- Add environment variable overrides (`GT_DOLT_HOST`, `GT_DOLT_PORT`, `GT_DOLT_USER`, `GT_DOLT_PASSWORD`) for Dolt server connection configuration
- When the server is remote (not localhost), `dolt sql` commands use explicit `--host/--port/--user/--password` flags instead of relying on data-directory auto-detection
- Fully backward compatible: without env vars, behavior is unchanged

## Motivation

Enables containerized deployments where the Dolt server runs in a separate container (e.g. Docker Compose), or any setup where the Dolt server is on a different host than the Gas Town process.

## Changes

| File | What |
|---|---|
| `internal/doltserver/doltserver.go` | Add `Host`, `Password` fields to `Config`; read `GT_DOLT_*` env vars in `DefaultConfig()`; add `IsRemote()`, `SQLArgs()`, `userDSN()` methods; update `doltSQL`, `doltSQLScript`, `GetActiveConnectionCount` to use connection flags when remote |
| `internal/doltserver/doltserver_test.go` | Add tests for env var overrides, password in DSN, `IsRemote()`, `SQLArgs()` |

## Test plan

- [x] All existing tests pass unchanged (backward compatible)
- [x] New tests cover: env var overrides, password in connection strings, `IsRemote()` for local/remote hosts, `SQLArgs()` with and without password

### DIFF
--- a/internal/doltserver/doltserver.go
+++ b/internal/doltserver/doltserver.go
@@ -122,6 +122,7 @@ func setDoltGlobalConfig(key, value string) error {
 
 // Default configuration
 const (
+	DefaultHost           = "127.0.0.1"
 	DefaultPort           = 3307
 	DefaultUser           = "root" // Default Dolt user (no password for local access)
 	DefaultMaxConnections = 50     // Conservative default to prevent connection storms
@@ -143,11 +144,17 @@ type Config struct {
 	// TownRoot is the Gas Town workspace root.
 	TownRoot string
 
-	// Port is the MySQL protocol port.
+	// Host is the Dolt server hostname or IP address.
+	Host string
+
+	// Port is the Dolt SQL server port.
 	Port int
 
-	// User is the MySQL user name.
+	// User is the Dolt authentication user.
 	User string
+
+	// Password is the Dolt authentication password (empty for no password).
+	Password string
 
 	// DataDir is the root directory containing all rig databases.
 	// Each subdirectory is a separate database that will be served.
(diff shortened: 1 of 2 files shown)


julianknutsen:

### Decision: request_changes

Thanks for putting together remote Dolt server support, @pae23 — this is a real need for containerized deployments and the env-var-based approach is clean and backward-compatible. The `Config.IsRemote()` / `SQLArgs()` abstraction is a solid foundation. However, both reviewers found that the implementation is currently partial: only about 3 of ~10 `dolt sql` exec.Command call sites were updated, and several core lifecycle functions still assume localhost. Remote mode won't work end-to-end until these are completed.

---

### Blocker: Multiple `dolt sql` call sites not updated for remote mode

The PR updates `doltSQL`, `doltSQLScript`, and `GetActiveConnectionCount` to use `SQLArgs()` and conditional `cmd.Dir`, but several other `dolt sql` exec.Command call sites were not updated: `verifyDatabasesWithRetry`, `CheckReadOnly`, `MeasureQueryLatency`, `serverExecSQL`, and several in `cmd/dolt.go`. These still rely on `cmd.Dir = config.DataDir` for auto-detection, which won't work when there's no local data directory.

**Suggested fix:** Apply the same `args := append([]string{"sql"}, config.SQLArgs()...) / if !config.IsRemote() { cmd.Dir = config.DataDir }` pattern to all remaining call sites. Consider centralizing into a `buildDoltSQLCmd()` helper to prevent future drift.

---

### Major: `CheckServerReachable` hardcodes `127.0.0.1`

`CheckServerReachable` dials `127.0.0.1:port` regardless of `config.Host`. A correctly-running remote server will always appear unreachable, triggering unnecessary startup attempts or error states.

**Suggested fix:** Change to `addr := fmt.Sprintf("%s:%d", config.Host, config.Port)`.

---

### Major: `IsRunning` assumes local server

`IsRunning` checks PID files and runs `lsof`/`ps` — purely local checks. When `config.IsRemote()`, these always return false, causing `up`/`start` to attempt launching a local Dolt server even though the user configured a remote one. This creates a split-brain scenario.

**Suggested fix:** When `config.IsRemote()`, skip PID/lsof checks and use a TCP reachability check. Guard `Start()` to return early or error when remote.

---

### Major: `gt dolt sql` interactive shell hardcodes localhost

The interactive SQL shell at `cmd/dolt.go:481-488` passes `--host 127.0.0.1` and `--password ""` regardless of config, making it unusable with a remote server.

**Suggested fix:** Use `config.Host`, `config.Port`, `config.User`, `config.Password` from `DefaultConfig`.

---

### Minor: Password exposed in terminal output

`GetConnectionString` includes the plaintext password in the DSN, which is printed to stdout by `gt dolt start` and `gt dolt status`.

**Suggested fix:** Add a `GetConnectionStringRedacted()` helper that masks the password for user-facing output.

---

### Minor: DSN special characters not escaped

If a password contains `@`, `:`, or `/`, the DSN string will be parsed incorrectly by go-sql-driver.

**Suggested fix:** Use `url.PathEscape()` on the password field or go-sql-driver's `mysql.Config{}.FormatDSN()`.

---

### Minor: Silent port parse failure

Invalid `GT_DOLT_PORT` values silently fall back to the default. Typos like `GT_DOLT_PORT=33o7` are invisible.

**Suggested fix:** Log a warning when the env var is present but cannot be parsed.

---

### Minor: Tests don't clear `GT_DOLT_PASSWORD` and lack remote coverage

Default-config tests don't clear `GT_DOLT_PASSWORD`, making them environment-sensitive. No tests exercise remote-mode behavior of runtime call sites.

**Suggested fix:** Add `t.Setenv("GT_DOLT_PASSWORD", "")` in default tests. Add remote-mode tests for `IsRunning`/`CheckServerReachable`.

---

### Pre-Merge Checklist
- [ ] Update ALL `dolt sql` exec.Command call sites to use `SQLArgs()` + conditional `cmd.Dir`
- [ ] Fix `CheckServerReachable` to use `config.Host`
- [ ] Make `IsRunning`/`Start`/`Stop` remote-aware
- [ ] Fix `gt dolt sql` interactive shell to use config values
- [ ] Add redacted connection string helper
- [ ] Escape DSN special characters in password
- [ ] Log warning on invalid `GT_DOLT_PORT`
- [ ] Add `t.Setenv` guards and remote-mode test coverage

---

steveyegge:
Closing per author's earlier note about consolidating PRs.

**Review notes for the next iteration:**

Good foundation -- IsRemote(), SQLArgs(), userDSN() are clean abstractions. Three dolt sql callsites were missed though, which would break remote mode for health checks and diagnostics:

- CheckServerReachable() (hardcoded 127.0.0.1)
- serverExecSQL() (no remote flags)
- MeasureQueryLatency() (no remote flags)

Minor: IsRemote() does not handle IPv6 loopback (::1).

Happy to review the consolidated PR when it is ready. Thanks for the contribution!

PR #1524, opened 2026-02-16
fix(guard): block Task tool for Mayor to enforce polecat dispatch

### LINKED ISSUE #904: Mayor reverts to using user defined agents to do work itself
## Bug Description
When talking to the mayor, occasionally they will spin up a claude agent to do the work that I have asked them to sling out instead of starting a convoy or slinging to a polecat. 

This blocks the mayor agent until the agent responds back, essentially invalidating GUZZ. 

## Steps to Reproduce
1. Create a heavily opinionated CLAUDE.md that explicitly states that agents are required to be used
2. Create a "developer" agent that is referenced in the CLAUDE.md
3. Ask the mayor to plan out fixing a bug on a rig. 

## Expected Behavior
The mayor slings the work to a polecat

## Actual Behavior
The mayor utilizes the claude subtask functionality to start fixing the bug. 

## Environment
- **OS**: Mac OS Sequoia 15.6.1
- **Go version**: go version go1.25.6 darwin/arm64
- **Gas Town version**: (`gt version`) gt version 0.5.0 (Homebrew: v0.5.0@Homebrew)
- **tmux version** (if applicable): 

## Logs / Error Output
```
<!-- Paste relevant logs or error messages -->
```

seanbearden:
## Summary

- Adds `gt tap guard task-dispatch` PreToolUse guard that blocks the Claude Code Task tool when `GT_MAYOR` is set, directing the Mayor to use `bd create` + `gt sling` instead
- Wires the guard into `DefaultOverrides()` so `gt hooks sync` auto-deploys it to mayor sessions — no manual configuration needed
- Prevents Mayor from spawning subagents that block its context and break parallel polecat execution (GUPP violation)

Fixes #904

## How it works

The guard leverages the existing `gt tap guard` infrastructure (same pattern as `pr-workflow`):

1. **Guard command** (`gt tap guard task-dispatch`): Checks `GT_MAYOR` env var. If set, exits code 2 (BLOCK). Otherwise returns 0 (ALLOW).
2. **Built-in hook override** (`DefaultOverrides()`): Registers a PreToolUse hook matching `"Task"` that runs the guard. Applied automatically when no on-disk mayor override exists.
3. **Auto-deployment**: `gt hooks sync` picks up the default override and writes it to the mayor's `.claude/settings.json`.

## Test plan

- [x] `TestTapGuardTaskDispatch_BlocksWhenMayor` — GT_MAYOR set → SilentExitError code 2
- [x] `TestTapGuardTaskDispatch_AllowsWhenNotMayor` — no env → nil (allowed)
- [x] `TestTapGuardTaskDispatch_AllowsForCrew` — GT_CREW set, no GT_MAYOR → nil
- [x] `TestTapGuardTaskDispatch_AllowsForPolecat` — GT_POLECAT set, no GT_MAYOR → nil
- [x] `TestComputeExpectedNoBase` — mayor gets DefaultBase + mayor override; crew gets DefaultBase only
- [x] `go test ./internal/cmd/... ./internal/hooks/...` — all pass
- [x] `golangci-lint run ./internal/cmd/... ./internal/hooks/...` — 0 issues

### DIFF
--- a/internal/cmd/tap_guard.go
+++ b/internal/cmd/tap_guard.go
@@ -18,7 +18,8 @@ is violated. They're called before the tool runs, preventing the
 forbidden operation entirely.
 
 Available guards:
-  pr-workflow   - Block PR creation and feature branches
+  pr-workflow      - Block PR creation and feature branches
+  task-dispatch    - Block Task tool for Mayor (use gt sling instead)
 
 Example hook configuration:
   {
@@ -51,9 +52,33 @@ witness, etc.). Humans running outside Gas Town can still use PRs.`,
 	RunE: runTapGuardPRWorkflow,
 }
 
+var tapGuardTaskDispatchCmd = &cobra.Command{
+	Use:   "task-dispatch",
+	Short: "Block Task tool for Mayor (use gt sling instead)",
+	Long: `Block the Claude Code Task tool when running as Mayor.
+
+The Mayor should dispatch work via gt sling, not by spawning Claude Code
+subagents. Subagents block the Mayor's context until they finish, breaking
+GUPP (the propulsion principle) and preventing parallel polecat execution.
+
+This guard blocks:
+  - Task tool invocations (Claude Code subagent spawning)
+
+Exit codes:
+  0 - Operation allowed (not Mayor, or --force override)
+  2 - Operation BLOCKED (Mayor context detected)
+
+The guard only blocks when GT_MAYOR is set. Crew members and polecats
+can still use the Task tool for parallel research.
+
+See: https://github.com/steveyegge/gastown/issues/904`,
+	RunE: runTapGuardTaskDispatch,
+}
+
 func init() {
 	tapCmd.AddCommand(tapGuardCmd)
 	tapGuardCmd.AddCommand(tapGuardPRWorkflowCmd)
+	tapGuardCmd.AddCommand(tapGuardTaskDispatchCmd)
 }
 
 func runTapGuardPRWorkflow(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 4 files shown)


julianknutsen:

### Decision: request_changes

Thanks for tackling #904, Sean — the Task tool guard for Mayor is exactly the right approach, and the implementation follows the existing `pr-workflow` guard pattern cleanly. The test coverage and `DefaultOverrides()` extension point are well-designed. Two issues need fixing before this can merge, both caught independently by both reviewers.

---

### Blocker: `GT_MAYOR` is never set — guard is a no-op in production

The guard checks `os.Getenv("GT_MAYOR") == ""`, but the mayor's environment setup in `config/env.go` only sets `GT_ROLE=mayor`. Compare with polecat (`GT_POLECAT=name`) and crew (`GT_CREW=name`) which do set role-specific vars — mayor, witness, and deacon only get `GT_ROLE`.

This means the guard will never fire. Mayor sessions will sail past it unchecked.

**Suggested fix:** Either change the guard to check `os.Getenv("GT_ROLE") == "mayor"`, or add `env["GT_MAYOR"] = "1"` to the mayor case in `config/env.go`. The former is simpler and avoids coordinating env changes. Update the tests to match.

---

### Blocker: On-disk override silently drops built-in default

`ComputeExpected()` only applies the built-in default when no on-disk override file exists. If someone creates `~/.gt/hooks-overrides/mayor.json` (e.g., to add a custom `SessionStart` hook), the task-dispatch guard vanishes without warning.

**Suggested fix:** Always merge built-in defaults first, then layer on-disk overrides on top:

```go
if def, ok := defaults[overrideKey]; ok {
    result = Merge(result, def)
}
override, err := LoadOverride(overrideKey)
if err != nil {
    if os.IsNotExist(err) {
        continue
    }
    return nil, fmt.Errorf(...)
}
result = Merge(result, override)
```

This preserves the ability for on-disk overrides to explicitly disable the guard (empty Hooks list for the "Task" matcher) while keeping the guard present by default. Add a test covering the coexistence case.

---

### Minor: Help text references non-existent `--force` flag

The help text at `tap_guard.go:68` says `"0 - Operation allowed (not Mayor, or --force override)"` but no `--force` flag is registered. Running `gt tap guard task-dispatch --force` returns an unknown flag error.

**Suggested fix:** Remove the `--force override` text from the help string, or implement the flag if an escape hatch is intended.

---

### Pre-Merge Checklist
- [ ] Guard checks `GT_ROLE` (or `GT_MAYOR` is added to env setup)
- [ ] `ComputeExpected()` merges built-in defaults before on-disk overrides, with coexistence test
- [ ] Help text exit code 0 description no longer references `--force`

---

steveyegge:
Closing — the concept is right but Julian's review identified two blockers that need addressing:

1. **GT_MAYOR is never set** — the guard checks an env var that doesn't exist in production. Mayor gets GT_ROLE=mayor, not GT_MAYOR. The guard never fires.

2. **Override precedence bug** — on-disk overrides silently drop the built-in default, so custom mayor hooks would disable the guard.

Happy to review a v2 that fixes these. The guard approach itself is solid — just needs the env check fixed to GT_ROLE and the ComputeExpected merge order adjusted so built-in defaults layer under on-disk overrides. Thanks!

PR #1526, opened 2026-02-16
fix(install): detect and stop orphaned dolt server during gt install

julianknutsen:
## Summary

Fix `gt install` failure when a user deletes `~/gt` and re-installs while an orphaned dolt sql-server is still running on port 3307.

**Reported symptom:**
```
Could not init HQ database: creating database on running server: chdir /home/ubuntu/gt/.dolt-data
```

**Root cause:** `InitRig("hq")` detects the orphaned server as "running" via `findDoltServerOnPort(3307)` and tries `CREATE DATABASE` via `serverExecSQL`, which sets `cmd.Dir = config.DataDir` (`~/gt/.dolt-data/`). Since the user deleted `~/gt`, the directory doesn't exist and the `chdir` fails.

## Changes

- Add orphaned server detection to `InitRig`: if a dolt server is running but `config.DataDir` doesn't exist, stop the orphan and fall through to the offline `dolt init` path
- Add the same orphaned server detection to `Start`: if a server is "running" but the data directory is gone, kill the orphan and proceed to start a fresh server instead of returning "already running"
- Both paths emit a warning to stderr so the user knows what happened
- Force-kill fallback if graceful `Stop()` fails (orphaned server may lack PID file)

## Testing

- [x] Unit tests pass (`go test ./internal/doltserver/`)
- [x] Manual verification: the fix correctly handles the reported scenario

## Checklist

- [x] No breaking changes — backward compatible (no behavior change when data dir exists)

### DIFF
--- a/internal/doltserver/doltserver.go
+++ b/internal/doltserver/doltserver.go
@@ -415,27 +415,42 @@ func Start(townRoot string) error {
 		return fmt.Errorf("checking server status: %w", err)
 	}
 	if running {
-		// Server is running - verify PID file is correct (gm-ouur fix)
-		// If PID file is stale/missing but server is on port, update it
-		pidFromFile := 0
-		if data, err := os.ReadFile(config.PidFile); err == nil {
-			pidFromFile, _ = strconv.Atoi(strings.TrimSpace(string(data)))
-		}
-		if pidFromFile != pid {
-			// PID file is stale/wrong - update it
-			fmt.Printf("Updating stale PID file (was %d, actual %d)\n", pidFromFile, pid)
-			if err := os.WriteFile(config.PidFile, []byte(strconv.Itoa(pid)), 0644); err != nil {
-				fmt.Fprintf(os.Stderr, "Warning: could not update PID file: %v\n", err)
+		// If data directory doesn't exist, this is an orphaned server (e.g., user
+		// deleted ~/gt and re-ran gt install). Kill it so we can start fresh.
+		if _, statErr := os.Stat(config.DataDir); os.IsNotExist(statErr) {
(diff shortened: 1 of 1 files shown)


julianknutsen:

### Decision: approve

Clean, focused bugfix that correctly addresses the reported install failure. Both reviewers agree the core fix is correct. Findings below are about edge-case hardening — none are blocking.

---

### Major: Port reuse timing after orphan kill in `Start()`

In `Start()`, after force-killing the orphan with only 100ms sleep, the new server start may fail with EADDRINUSE. Neither `Start()` nor `InitRig()` re-checks `IsRunning()` after the kill to confirm the orphan is actually dead.

**Assessment:** In the install flow, `InitRig` kills the orphan first, then `Start()` is called separately — the orphan is already dead by then. The standalone `gt dolt start` path is the only place this matters, and `Start()` already has retry/verification logic after launching the new server. Low risk in practice. Can harden in a follow-up if needed.

---

### Minor: Foreign dolt server on same port could be killed

If a different dolt server occupies port 3307 while the data dir is missing, the orphan detection could kill the wrong server.

**Assessment:** `findDoltServerOnPort` already verifies the process is `dolt` via `isDoltProcess`. The scenario requires: (a) user deleted `~/gt`, (b) a *different* dolt server occupies port 3307, (c) that server's data dir also doesn't exist. Extremely unlikely. No change needed.

---

### Minor: No test coverage for orphan paths

The orphan detection involves process killing and filesystem checks with no unit tests.

**Assessment:** Testing process lifecycle in unit tests is fragile. The fix is straightforward (stat + kill). Acceptable for this bugfix scope.

---

### Pre-Merge Checklist
- [x] Core fix correctly addresses the reported bug
- [x] Backward compatible (no behavior change when data dir exists)
- [x] Full test suite passes
- [ ] Optional follow-up: add post-kill `IsRunning()` re-check in `Start()` for robustness

---

PR #1539, opened 2026-02-16
fix(feed): fall back to .events.jsonl when bd activity unavailable

### LINKED ISSUE #1535: gt feed broken: depends on non-existent bd activity command
## Bug

`gt feed` fails with `Error: unknown command "activity" for "bd"`.

Both TUI mode and plain mode depend on `bd activity` which was never implemented in the beads CLI.

**Reproduction:**
```bash
gt feed          # TUI mode — crashes creating bd activity source  
gt feed --plain  # Plain mode — exec fails immediately
gt feed --window # Window mode — runs bd activity in tmux, fails
```

**Root cause:**
- `internal/cmd/feed.go` calls `bd activity` in all three modes
- `internal/tui/feed/events.go` spawns `bd activity --follow` as a subprocess
- The `bd` CLI has no `activity` subcommand

**Impact:** The entire feed system is non-functional. No agent activity visibility.

**Fix approach:** Make `BdActivitySource` optional in TUI mode (like MQ and GT sources already are). For plain mode, fall back to reading `.events.jsonl` directly or show a clear error.

ekoziol:
## Summary
- `gt feed` depends on `bd activity` which doesn't exist in the current bd CLI, causing all feed modes to fail
- Makes `BdActivitySource` optional in TUI mode — the feed dashboard still works with GT events and MQ sources
- Plain mode (`--plain`) falls back to reading `.events.jsonl` directly via new `PrintGtEvents()` function
- Window mode (`--window`) uses `gt feed --plain --follow` instead of `bd activity --follow`

Closes #1535

## Test plan
- [ ] Run `gt feed` — TUI dashboard should launch with GT events and convoy data
- [ ] Run `gt feed --plain` — should print recent events from `.events.jsonl`
- [ ] Run `gt feed --window` — should open tmux window with plain feed
- [ ] Verify no regression if `bd activity` becomes available in future

### DIFF
--- a/internal/cmd/feed.go
+++ b/internal/cmd/feed.go
@@ -4,7 +4,6 @@ import (
 	"fmt"
 	"os"
 	"os/exec"
-	"path/filepath"
 	"strings"
 
 	tea "github.com/charmbracelet/bubbletea"
@@ -36,15 +35,15 @@ func init() {
 	feedCmd.Flags().StringVar(&feedSince, "since", "", "Show events since duration (e.g., 5m, 1h, 30s)")
 	feedCmd.Flags().StringVar(&feedMol, "mol", "", "Filter by molecule/issue ID prefix")
 	feedCmd.Flags().StringVar(&feedType, "type", "", "Filter by event type (create, update, delete, comment)")
-	feedCmd.Flags().StringVar(&feedRig, "rig", "", "Run from specific rig's beads directory")
+	feedCmd.Flags().StringVar(&feedRig, "rig", "", "Filter events by rig name")
 	feedCmd.Flags().BoolVarP(&feedWindow, "window", "w", false, "Open in dedicated tmux window (creates 'feed' window)")
-	feedCmd.Flags().BoolVar(&feedPlain, "plain", false, "Use plain text output (bd activity) instead of TUI")
+	feedCmd.Flags().BoolVar(&feedPlain, "plain", false, "Use plain text output instead of TUI")
 }
 
 var feedCmd = &cobra.Command{
 	Use:     "feed",
 	GroupID: GroupDiag,
-	Short:   "Show real-time activity feed from beads and gt events",
+	Short:   "Show real-time activity feed of gt events",
 	Long: `Display a real-time feed of issue changes and agent activity.
 
 By default, launches an interactive TUI dashboard with:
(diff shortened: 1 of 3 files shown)


julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @ekoziol! Removing the hard `bd activity` dependency from `gt feed` is a great improvement — having the feed silently break when `bd` isn't available was a real pain point, and switching to native `.events.jsonl` reading is the right long-term direction. The dashboard Escape key fix is also a nice quality-of-life improvement.

### Original PR Review
Initial dual-model review (Claude + Codex) found 5 issues: `--follow` flag silently broken in plain mode (blocker), `--since`/`--mol`/`--type` flags silently discarded (major), `filepath.Abs(os.Args[0])` wrong binary path in window mode (major), zero test coverage for new `PrintGtEvents` (major), and dead `printGtEventLine` function (nit).

### Maintainer Changes
+547 -77 across 3 files:
- Rewrote `PrintGtEvents` with proper `PrintOptions` struct wiring all CLI flags
- Fixed `bufio.Scanner` reuse-after-EOF bug in tail loop (fresh scanner per tick)
- Added TTY detection for follow-mode default (prevents non-TTY blocking)
- Replaced `filepath.Abs(os.Args[0])` with `os.Executable()`
- Added `--rig` event filtering via `matchesFilters` (rig field from event payload)
- Forwarded `--rig` through `buildFeedArgs()` for window mode
- Reconciled `--rig` semantics: pure event filter for plain/window, beads dir resolution for TUI
- Updated CLI help text to reflect `.events.jsonl`-based behavior
- Quoted shell-interpolated paths in `runFeedInWindow`
- Added `context.Context` support for testable follow-mode lifecycle
- Added 22 tests: formatting, limits, since/type/rig/visibility filters, follow streaming, matchesFilters unit tests

### Final Review Status
All clear — no remaining blockers or majors. Final dual-model review returned **approve** with only minor/nit findings (all deferrable):
- Shell arg quoting in tmux command (pre-existing pattern, follow-up candidate)
- Missing `.events.jsonl` hard-errors in follow mode on fresh workspaces (follow-up)
- Follow test assertion could be tighter (nit)

### Review Iterations
5 review passes performed (1 initial PR-mode, 4 local-branch iterations).

---

PR #1550, opened 2026-02-16
fix(doltserver): create rig-root .beads instead of mayor path for new rigs

julianknutsen:
## Summary

When adding a new rig for an untracked repo while the dolt server is running, `FindOrCreateRigBeadsDir` created `mayor/rig/.beads/` instead of the rig-root `.beads/`. This caused `InitBeads` to misdetect the repo as having tracked beads, take the redirect early-return, and skip creating `config.yaml` and `issues.jsonl`. The result: a degraded rig where all `bd` commands fail until `gt doctor --fix` repairs it.

## Changes

- **`FindOrCreateRigBeadsDir`**: Changed fallback (when neither beads path exists) to create rig-root `.beads/` instead of `mayor/rig/.beads/`. The mayor path should only be used when it already exists (i.e., the source repo has tracked beads checked out via git clone).
- **`FindRigBeadsDir`**: Aligned read-only fallback to also return rig-root path, for consistency.
- **`EnsureMetadata` docstring**: Updated to reflect actual resolution order (mayor if present, else rig-root).
- **Tests**: Updated existing tests to expect rig-root path; added regression test verifying mayor path is NOT created for untracked repos.

### Why this happens

`AddRig` calls `InitRig` before `InitBeads`. Inside `InitRig`, when the dolt server `IsRunning()`, `EnsureMetadata` calls `FindOrCreateRigBeadsDir` — which was eagerly creating `mayor/rig/.beads/`. Then when `InitBeads` runs, it sees that directory, assumes the repo has tracked beads, and takes the redirect path instead of initializing the local beads database.

The mayor/rig/.beads path is only correct when the source repo actually has tracked beads (checked out via git clone). For new rigs with untracked repos, the rig-root `.beads/` is the right location.

## Testing

- [x] Unit tests pass (`go test ./internal/doltserver/ -v`)
- [x] All 8 `TestFindOrCreateRigBeadsDir` subtests pass including new regression test
- [x] `TestFindRigBeadsDir` passes with updated assertion
- [x] All rig, doctor, and cmd tests pass
- [x] Dual-model review (Claude + Codex) passed with `approve` decision

## Checklist

- [x] Documentation updated (docstring corrections)
- [x] No breaking changes

### DIFF
--- a/internal/doltserver/doltserver.go
+++ b/internal/doltserver/doltserver.go
@@ -1352,7 +1352,8 @@ func RepairWorkspace(townRoot string, ws BrokenWorkspace) (string, error) {
 // centralized Dolt server.
 //
 // For the "hq" rig, it writes to <townRoot>/.beads/metadata.json.
-// For other rigs, it writes to <townRoot>/<rigName>/mayor/rig/.beads/metadata.json.
+// For other rigs, it writes to mayor/rig/.beads/metadata.json if that path exists,
+// otherwise to <townRoot>/<rigName>/.beads/metadata.json.
 func EnsureMetadata(townRoot, rigName string) error {
 	// Use FindOrCreateRigBeadsDir to atomically resolve and create the directory,
 	// avoiding the TOCTOU race where the directory state changes between
@@ -1453,8 +1454,8 @@ func FindRigBeadsDir(townRoot, rigName string) string {
 		return rigBeads
 	}
 
-	// Neither exists; return mayor path (caller will create it)
-	return mayorBeads
+	// Neither exists; return rig-root path (consistent with FindOrCreateRigBeadsDir)
+	return rigBeads
 }
 
 // FindOrCreateRigBeadsDir atomically resolves and ensures the .beads directory
--- a/internal/doltserver/doltserver_test.go
+++ b/internal/doltserver/doltserver_test.go
@@ -503,11 +503,11 @@ func TestFindRigBeadsDir(t *testing.T) {
 		t.Errorf("otherrig beads dir = %q, want %q", dir, rigBeads)
 	}
 
-	// Test rig with neither directory existing — should return mayor path for creation
+	// Test rig with neither directory existing — should return rig-root path
 	neitherRig := "newrig"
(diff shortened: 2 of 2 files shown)


julianknutsen:

### Decision: approve

Nice targeted fix for a real production issue — users adding rigs for untracked repos while the dolt server is running would end up with a degraded rig where `bd` commands fail. The one-line change in `FindOrCreateRigBeadsDir` is the right fix: the mayor path should only be used when it already exists from tracked beads, not eagerly created. Both reviewers independently confirmed the fix is correct and the blast radius is minimal (2 callers, both safe).

---

### Fixes Validated
- **Untracked rigs misclassified as tracked:** Confirmed correct. `FindOrCreateRigBeadsDir` now creates `rig/.beads` instead of `mayor/rig/.beads`, so `InitBeads` no longer takes the redirect early-return and correctly creates `config.yaml` + `issues.jsonl`.
- **Tracked rigs still resolve to mayor path:** Confirmed correct. Existing `mayor/rig/.beads` remains preferred via the priority check order.
- **Concurrent creation stable:** Confirmed correct. `MkdirAll` convergence behavior unchanged.

---

### Minor: `FindRigBeadsDir` fallback inconsistency

The read-only `FindRigBeadsDir` still defaulted to `mayorBeads` when neither path exists, disagreeing with the updated `FindOrCreateRigBeadsDir`. Current callers tolerate this (they handle read errors gracefully), but the inconsistency is a maintenance hazard.

**Status:** Resolved in second commit (b7c3344).

### Minor: `EnsureMetadata` docstring stale

The docstring said non-`hq` always writes to `mayor/rig/.beads`, which no longer reflects actual behavior.

**Status:** Resolved in second commit (b7c3344).

### Minor: No full-path integration test for server-running AddRig path

The unit-level regression test covers the root cause directly. A full-path integration test exercising `AddRig` with `IsRunning()==true` would be defense-in-depth but is not required — the fix eliminates the problematic directory creation regardless of code path.

**Status:** Accepted as-is. Unit tests cover the root cause.

---

### Pre-Merge Checklist
- [x] All 3 minor findings addressed (2 fixed, 1 accepted)
- [x] No blockers or majors
- [x] All tests pass

---

PR #1554, opened 2026-02-16
fix(await-signal): add --since 1s to prevent stale event replay

scott-pallas:
## Summary\n\n- `waitForActivitySignal()` now passes `--since 1s` to `bd activity --follow`, ensuring only NEW events are received\n- Without this flag, stale events from recent beads operations cause await-signal to fire immediately (~90ms) instead of waiting for new activity\n- This was causing the deacon patrol loop to busy-loop, consuming context and resources\n\n## Root Cause\n\n`bd activity --follow` replays recent history by default. The `--since` flag exists but `await-signal` wasn't using it. One-line fix.\n\n## Test plan\n\n- [x] Existing tests pass (`go test ./internal/cmd/ -run TestCalculateEffectiveTimeout`)\n- [x] Built and installed locally — deacon patrol loop now sleeps correctly between cycles\n- [ ] Manual verification: run `gt mol step await-signal --timeout 30s` after some bd operations, confirm it waits instead of firing immediately\n\n🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/internal/cmd/molecule_await_signal.go
+++ b/internal/cmd/molecule_await_signal.go
@@ -284,17 +284,23 @@ func calculateEffectiveTimeout(idleCycles int) (time.Duration, error) {
 	return time.ParseDuration(awaitSignalTimeout)
 }
 
-// waitForActivitySignal starts bd activity --follow and waits for any output.
-// Returns immediately when a line is received, or when context is canceled.
-func waitForActivitySignal(ctx context.Context, workDir string) (*AwaitSignalResult, error) {
-	// Start bd activity --follow --since 0s
-	// The --since 0s flag ensures we only receive events that occur AFTER the
-	// subscription starts. Without it, stale events (including self-generated
-	// mutations from setAgentBackoffUntil) cause await-signal to fire immediately.
-	// Using 0s instead of 1s avoids the race where our own label update happens
-	// within the 1-second window.
+// activityFollowCmd builds the exec.Cmd for subscribing to new beads activity.
+// Uses --since 0s to scope the feed to post-subscription events only,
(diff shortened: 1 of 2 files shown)


julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @scott-pallas! This is a great catch — the stale event replay causing the deacon patrol busy-loop is exactly the kind of subtle operational issue that's hard to spot but has real resource impact. Clean one-line fix with a clear root cause analysis.

### Original PR Review
Initial dual-model review (Claude + Codex) found the fix itself correct — both reviewers independently confirmed that `--since 0s` correctly scopes `bd activity --follow` to post-subscription events, eliminating the stale replay that caused immediate spurious wakes. One **major** finding: no regression test existed to protect the `--since 0s` argv from being silently removed during future refactors. Two **nits**: function godoc didn't mention `--since 0s`, and the PR title said `--since 1s` while the code correctly uses `--since 0s`.

### Maintainer Changes
- Extracted `activityFollowCmd()` helper for testability (+11 lines in production code)
- Added `TestActivityFollowCmd_IncludesSinceFlag` regression test asserting `--since 0s` is present in command args (+18 lines in test)
- Updated function-level godoc to document `--since 0s` behavior

### Final Review Status
Second review pass returned **APPROVE** with only minor/nit items remaining:
- Cobra long description at line 32 still says `'bd activity --follow'` (cosmetic)
- No `bd` version gate for `--since` support (tracked as follow-up — failure mode is visible and non-destructive)

No blockers or majors. All clear for merge.

### Review Iterations
2 review passes performed (initial PR mode + iteration after maintainer fixup).

---

Issue #1555, opened 2026-02-16
sling: make repeated same-target assignment idempotent (no duplicate dispatch)


davidahmann:
## Problem
Running `gt sling` repeatedly for a bead that is already `pinned`/`hooked` to the same effective target can trigger unnecessary dispatch behavior (or force users into error/retry loops) instead of behaving as a safe no-op.

## Expected behavior
If a bead is already assigned to the same effective target, `gt sling` should return success with no new dispatch side effects.

## Scope (Phase 1)
- Add explicit same-target idempotency guard in single-bead sling path.
- Add focused tests for exact-match and namespace-prefix target matching.

## Acceptance criteria
- Repeated sling to the same target exits successfully without re-dispatch.
- Non-matching targets still follow existing guarded behavior.
- New tests cover idempotent and non-idempotent branches.

julianknutsen:
Closed via #1556

PR #1556, opened 2026-02-16
sling: make same-target assignment idempotent

davidahmann:
## Problem
Repeated `gt sling` on a bead already assigned to the same effective target could error/re-dispatch instead of acting as a no-op.

## What changed
- Added `matchesSlingTarget` matching logic for exact and namespace-prefix equivalence.
- Wired idempotency check into `runSling` before re-sling error path.
- Added focused unit coverage for exact match, shorthand, namespace-prefix, and non-match cases.

## Validation
- `go test ./internal/cmd -run TestMatchesSlingTarget -count=1`
- `go test ./internal/cmd -run TestSling -count=1`

Refs #1555

### DIFF
--- a/internal/cmd/sling.go
+++ b/internal/cmd/sling.go
@@ -287,34 +287,62 @@ func runSling(cmd *cobra.Command, args []string) error {
 	originalAssignee := info.Assignee
 	force := slingForce // local copy to avoid mutating package-level flag
 	if (info.Status == "pinned" || info.Status == "hooked") && !force {
-		target := ""
-		if len(args) > 1 {
-			target = args[1]
-		}
-		selfAgent, _, _, _ := resolveSelfTarget()
-		if matchesSlingTarget(target, info.Assignee, selfAgent) {
-			fmt.Printf("%s Bead %s is already %s to %s, no-op\n",
-				style.Dim.Render("○"), beadID, info.Status, info.Assignee)
-			return nil
-		}
-
 		// Auto-force when hooked agent's session is confirmed dead (gt-pqf9x).
 		// This eliminates the #1 friction in convoy feeding: stale hooks from
 		// dead polecats blocking re-sling without --force.
-		if info.Status == "hooked" && info.Assignee != "" && isHookedAgentDead(info.Assignee) {
+		// IMPORTANT: Stale-hook check must run BEFORE idempotency check so that
+		// a dead polecat with a matching target triggers re-sling, not a no-op.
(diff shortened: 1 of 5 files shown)


davidahmann:
Implemented same-target sling idempotency to avoid duplicate dispatch side effects.

Scope in this PR:
- Added exact and namespace-prefix target matching.
- Wired no-op success path into `runSling` for already-matched assignments.
- Added focused coverage for idempotent and non-idempotent cases.

This contribution was informed by patterns from Gait: https://github.com/davidahmann/gait

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @davidahmann! Making `gt sling` idempotent on same-target assignment is a really valuable improvement — it eliminates a common friction point where repeating a sling command would error out instead of gracefully no-oping. The approach of checking target equivalence before the existing error path is clean and well-structured.

### Original PR Review
Initial dual-model review (Claude + Codex) found 2 major and 4 minor issues: the idempotency no-op could mask stale-hook recovery for dead polecats, rig-only target matching excluded crew (surprising asymmetry needed confirmation), `resolveSelfTarget` error was silently discarded, `args[1]` vs `args[len(args)-1]` fragility, dogs missing from role exclusion list, and no integration tests for the `runSling` idempotency path.

### Maintainer Changes
All 6 findings from the initial review were addressed in a single fixup commit (+625 -55 across 5 files):
- Stale-hook liveness check moved before idempotency so dead polecats trigger auto-force re-sling
- Error guard added on `resolveSelfTarget` (skips idempotency on failure instead of silent discard)
- `args[len(args)-1]` used consistently with explanatory comment
- Dogs added to role exclusion list; rig→crew asymmetry documented as intentional
- `isHookedAgentDeadFn` test seam extracted for integration testing
- Formula-on-bead with matching target now correctly falls through to formula instantiation
- 6 integration tests added: idempotent no-op (hooked + pinned), dead-agent bypass, --force bypass, formula-on-bead bypass, pinned self-dot resolution

### Final Review Status
5 review passes performed. Final decision: **approve**. All blockers and majors resolved. Two minor/nit items remain as follow-up opportunities (not blocking):
1. **Minor:** `args[len(args)-1]` vs `args[1]` divergence between idempotency check and `resolveTarget` — only manifests in contrived >2 args non-batch invocations, worst case is a missed early-return with no data corruption.
2. **Nit:** Redundant `hookBeadWithRetry` call in formula-on-bead path when bead is already hooked to the correct agent — functionally harmless, optimization opportunity.

### Review Iterations
5 review passes performed (1 initial PR-mode + 4 local branch iterations).

---

PR #1565, opened 2026-02-16
fix(doctor): don't write role templates into external repo worktrees

### LINKED ISSUE #1549: gt doctor --fix writes role templates into external repo worktrees
## Bug

`gt doctor --fix` via the `patrol-roles-have-prompts` check writes gastown role template files into external git worktrees (`mayor/rig/`), polluting customer repos.

## Reproduction

1. `gt rig add` an external Go repo that has an `internal/templates/` directory (common in Go projects)
2. Run `gt doctor --fix`
3. Observe gastown-specific files created inside the worktree:

```
<rig>/mayor/rig/internal/templates/roles/deacon.md.tmpl    ← CREATED
<rig>/mayor/rig/internal/templates/roles/witness.md.tmpl   ← CREATED
<rig>/mayor/rig/internal/templates/roles/refinery.md.tmpl  ← CREATED
```

These are untracked files in the external repo and will leak to remote if accidentally committed with `git add .`.

## Root Cause

`PatrolRolesHavePromptsCheck` in `internal/doctor/patrol_check.go`:

- **Run** (line 439): checks if `<rig>/mayor/rig/internal/templates/` exists. If the external repo happens to have `internal/templates/` (common for Go projects), the check misidentifies it as gastown custom template overrides and reports missing role files.
- **Fix** (lines 482-510): writes gastown embedded templates into `<rig>/mayor/rig/internal/templates/roles/`, which is inside the external repo worktree.

```go
// patrol_check.go:489 — writes directly into worktree
mayorRig := filepath.Join(ctx.TownRoot, rigName, "mayor", "rig")
templatesDir := filepath.Join(mayorRig, "internal", "templates", "roles")
```

## Expected Behavior

`gt doctor --fix` should never write files into worktrees that connect to external repositories. Role templates are embedded in the gastown binary — the check should recognize that `internal/templates/` inside a worktree belongs to the customer repo, not gastown.

## Suggested Fix

The check should distinguish between gastown's own `internal/templates/` and an external repo's `internal/templates/`. Options:

1. **Skip external repos entirely**: if `mayor/rig/` is a worktree of an external remote, don't check for role templates there. Templates are embedded in the binary and don't need to exist on disk.
2. **Look for a gastown marker**: only treat `internal/templates/` as gastown overrides if a gastown-specific marker file exists (e.g., check for other gastown files like `internal/templates/roles/` specifically, not just `internal/templates/`).
3. **Write to rig level instead**: if templates must exist on disk, place them at `<rig>/internal/templates/roles/` rather than inside the worktree.

## Related

- #885 (parent issue covering all worktree artifact leaks)
- PR #430 (fixed the CLAUDE.md and settings.json variants of this class of bug)

Avyukth:
## Problem

Fixes #1549

"$(gt doctor --fix)" via "$(PatrolRolesHavePromptsCheck)" writes gastown role template files (deacon.md.tmpl, witness.md.tmpl, refinery.md.tmpl) into external git worktrees at "$(<rig>/mayor/rig/internal/templates/roles/)". If the external repo happens to have "$(internal/templates/)" (common in Go projects), the check misidentifies it as gastown custom template overrides and creates gastown files inside the customer repo.

## Root Cause

"$(PatrolRolesHavePromptsCheck.Run())" (patrol_check.go:439) checks for "$(internal/templates/)" to detect gastown custom template overrides. External Go repos commonly have this directory for their own templates, causing a false match. The "$(Fix())" method then writes gastown-specific "$(.md.tmpl)" files into the external repo's worktree.

## Solution

Two-layer guard to prevent false matches on external repos:

1. **Check for "$(internal/templates/roles)"** instead of "$(internal/templates)". Gastown role templates live specifically in the "$(roles/)" subdirectory — generic Go projects won't have this.

2. **go.mod heuristic** ("$(isLikelyGastownRepo)"): Even if "$(roles/)" exists, only proceed if the repo already contains at least one canonical role template OR its "$(go.mod)" module path contains "gastown". This prevents writes into external repos that coincidentally have a "$(roles/)" directory.

Additionally:
- Added "$(hasAnyRequiredRolePrompt())" helper to detect existing role templates
- Improved error handling for all "$(os.Stat)" calls (handle non-NotExist errors, check IsDir)
- No changes to "$(Fix())" — it only processes rigs already in "$(missingByRig)"

## Test plan

- [x] "$(go build ./...)" passes
- [x] "$(go vet ./...)" passes
- [x] "$(go test ./internal/doctor/ -run TestPatrolRolesHavePrompts -v)" — 16 tests pass (12 existing + 4 new)
- [x] New: external repo with "$(internal/templates/)" but no "$(roles/)" → skipped
- [x] New: external repo with "$(internal/templates/roles/)" but non-gastown go.mod → skipped, Fix() doesn't pollute
- [x] New: mixed rigs (external + gastown) → only gastown rig gets templates
- [x] New: "$(roles)" path is a file, not directory → returns StatusError


julianknutsen:
## Maintainer Review — PR #1565

**Reviewers**: Claude (Opus 4.6) + Codex (GPT-5.3) via dual-model pipeline
**Iterations**: 2 (initial review → maintainer fixup → re-review)

### Summary

The original PR added a two-layer guard to `PatrolRolesHavePromptsCheck.Run()` to prevent writing role templates into external repo worktrees.

During review, we traced the template loading path and discovered the **entire check is based on a false premise**: role templates are `//go:embed`'d into the gastown binary (`templates.go:41`) and loaded at runtime via `template.ParseFS` (`templates.go:123`). On-disk copies in rig worktrees are never read by any code path. The check reports "missing" files that aren't supposed to exist, and `Fix()` writes files nothing reads. Additionally, `Run()` always returns `StatusOK` (since commit `e2bd5ef7`), so `Fix()` is never auto-called by the doctor framework.

### Maintainer Fixup

Rather than patching a dead check, the fixup commit **deletes the entire `PatrolRolesHavePromptsCheck`**:

- Removed struct, constructor, `Run()`, `Fix()`, and all helpers (`requiredRolePrompts`, `hasAnyRequiredRolePrompt`, `isLikelyGastownRepo`)
- Removed registration in `doctor.go` and stale help text
- Removed all 11+ related tests and test helpers (`setupRigConfig`, `setupRigTemplatesDir`)
- Removed `GetAllRoleTemplates()` from `templates.go` (only caller was `Fix()`)
- Removed `TestGetAllRoleTemplates` tests from `templates_test.go`

**Net change from base**: -573 lines (pure deletion) across 5 files.

### Iteration 2 Verification (both reviewers)

- No dangling references to any deleted symbols
- `discoverRigs()` preserved — still used by 5 callers across other checks
- `//go:embed` + `ParseFS` confirms templates are embedded; on-disk copies never read
- All tests pass: `go test ./internal/doctor/ ./internal/templates/ -count=1` OK
- No collateral damage to other checks or tests

### Decision: **APPROVE** (after fixup)

The original issue #1549 is resolved by removing the code that caused it entirely, rather than adding guardrails to dead code.

---
*Reviewed with [aimux](https://github.com/julianknutsen/aimux) dual-model pipeline*

PR #1567, opened 2026-02-16
fix: buffer watch mode output to prevent blank screen

### LINKED ISSUE #1566: gt status --watch shows blank screen until Ctrl+C
## Bug

`gt status --watch` displays a blank screen during watch mode. No status output is visible until the user presses Ctrl+C to exit, at which point the status appears.

## Expected behavior

The status should be visible immediately and refresh every 2-5 seconds (per the `--interval` flag), equivalent to running `gt status` repeatedly.

## Root cause

`runStatusWatch` clears the screen with `\033[H\033[2J` (a single `write()` call), then renders the status through dozens of individual `fmt.Printf` calls in `outputStatusText` and its callees (`renderAgentCompact`, `renderAgentDetails`, etc.). Each `fmt.Printf` is a separate `write()` syscall to the terminal.

The terminal processes the clear-screen escape sequence immediately, producing a blank frame. The subsequent content arrives across many small writes over ~1 second (due to tmux/beads/mail lookups). In a real TTY, this creates a visible blank screen between each refresh cycle. On Ctrl+C the program exits without clearing, so the last rendered frame remains visible.

## Fix

Buffer all output for each watch frame into a `bytes.Buffer`, then write the entire frame (clear + header + status) to stdout in a single `Write()` call. This eliminates the blank-screen window between clear and content.

The fix also extracts a `gatherStatus()` function from `runStatusOnce`, separating data gathering from rendering. The render functions (`outputStatusText`, `renderAgentDetails`, `renderAgentCompact`, `renderAgentCompactWithSuffix`) now accept an `io.Writer` parameter instead of writing to `os.Stdout` directly.

ekoziol:
## Summary

- Buffer the entire watch frame into `bytes.Buffer` and write atomically to stdout, preventing the terminal from rendering a blank screen between the clear-screen escape and the status content
- Extract `gatherStatus()` from `runStatusOnce` to separate data collection from rendering
- Thread `io.Writer` through `outputStatusText`, `renderAgentDetails`, `renderAgentCompact`, and `renderAgentCompactWithSuffix` so callers control the output destination

Fixes #1566

## Test plan

- [x] `gt status` works unchanged (writes to `os.Stdout` directly)
- [x] `gt status --watch` renders status immediately on each refresh cycle
- [x] `gt status --watch --interval 3` respects custom interval
- [x] `gt status --watch --json` still rejects the combination
- [x] `gt status --watch` with `--interval 0` and `--interval -5` still error
- [x] All existing tests pass (`go test ./internal/cmd/...`)
- [x] Full test suite passes (only pre-existing `/tmp` symlink test fails)

### DIFF
--- a/internal/cmd/status.go
+++ b/internal/cmd/status.go
@@ -175,13 +175,13 @@ func runStatusWatch(_ *cobra.Command, _ []string) error {
 		status, err := gatherStatus()
 		if err != nil {
 			fmt.Fprintf(&buf, "Error: %v\n", err)
-		} else {
-			outputStatusText(&buf, status)
+		} else if err := outputStatusText(&buf, status); err != nil {
+			fmt.Fprintf(&buf, "Error: %v\n", err)
 		}
 
 		// Write the entire frame atomically to prevent the terminal from
 		// rendering a blank screen between the clear and the content.
-		os.Stdout.Write(buf.Bytes())
+		_, _ = os.Stdout.Write(buf.Bytes())
 
 		select {
 		case <-sigChan:


julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @ekoziol! This is a great quality-of-life fix — the blank screen flicker during `gt status --watch` refresh cycles was a real usability pain point, and your approach of buffering the entire frame into `bytes.Buffer` before a single `os.Stdout.Write` is the right solution. The `gatherStatus()` extraction and `io.Writer` threading are clean and idiomatic Go.

This PR was reviewed and adopted via dual-model review (Claude + Codex). No maintainer fixup commits were needed.

### Original PR Review

**Decision: approve.** Both reviewers independently confirmed the fix is correct and the `io.Writer` migration is internally complete with no stray direct stdout writes. The PR is well-scoped and backward compatible.

Top findings:
1. **Major — Test gap (codex, high confidence):** No test locks in the new buffering contract. Existing watch-mode tests validate flag guards but don't assert that a rendered frame contains both the ANSI clear-sequence and status payload in a single buffer. This is a test gap, not a behavioral defect — the fix itself is verified correct.
2. **Minor — Silent write discard (both, medium confidence):** `_, _ = os.Stdout.Write(buf.Bytes())` silently discards errors. Pragmatic for a terminal UI but could mask `EPIPE` when piped to a dead process.
3. **Minor — Theoretical short write (claude, low confidence):** Buffers exceeding `PIPE_BUF` (~4KB) aren't guaranteed atomic by POSIX. Only relevant for extremely large towns (100+ agents). Informational only.

### Maintainer Changes

No maintainer changes applied. The PR is clean as submitted.

### Final Review Status

Approve — all clear. The behavioral fix is correct, the refactoring is sound, and the findings are minor/additive (test gap + informational items). No blockers or regressions found.

### Review Iterations
1 review pass performed.

---

PR #1569, opened 2026-02-16
fix: load agent registry in persistentPreRun so IsAgentAlive respects settings/agents.json

GraemeF:
## Summary

Load `settings/agents.json` in `persistentPreRun` so that every `gt` subcommand sees user-configured agent overrides before any liveness check runs.

## Changes

- Add `config.LoadAgentRegistry(config.DefaultAgentRegistryPath(townRoot))` call in `persistentPreRun` (root.go), immediately after `session.InitRegistry`
- Add regression test `TestPersistentPreRunLoadsAgentRegistry` in root_test.go — exercises the full `persistentPreRun` → `GetProcessNames` path with a fake town root and `settings/agents.json` override
- Add unit test `TestGetProcessNamesRespectsRegistryOverride` in agents_test.go — verifies `LoadAgentRegistry` correctly overrides builtin preset `process_names`

## Problem

`settings/agents.json` is the intended mechanism for users to override agent presets (including `process_names` used for liveness detection). However, the file is only loaded as a side effect of `ResolveAgentConfig` / `ResolveRoleAgentConfig` during agent startup. Many code paths call `IsAgentAlive` → `GetProcessNames` → `GetAgentPresetByName` without ever triggering config resolution, so they only see the builtin presets from `ensureRegistry()`.

On NixOS/Nix-installed Claude Code, the binary is `.claude-unwrapped` (a Nix wrapper convention). The builtin Claude preset only lists `["node", "claude"]` as `process_names`. Users can add `.claude-unwrapped` to `settings/agents.json`, but because the registry isn't loaded before liveness checks, this override is never seen. The daemon's heartbeat, zombie cleanup, and dog patrols all think the running Claude session is dead, kill it, and restart it — every 3 minutes, indefinitely.

`LoadAgentRegistry` is idempotent (caches loaded paths), so this adds no overhead to code paths that already load it during config resolution.

## Testing

- [x] Unit tests pass (`go test ./internal/cmd/... ./internal/config/...`)
- [x] New test confirmed to **fail without the fix** and **pass with it**
- [x] `go vet` clean on affected packages

## Checklist

- [x] Documentation updated (if applicable) — N/A, no user-facing docs change needed

### DIFF
--- a/internal/cmd/root.go
+++ b/internal/cmd/root.go
@@ -100,7 +100,10 @@ func persistentPreRun(cmd *cobra.Command, args []string) error {
 	// Best-effort: if town root not found, the default "gt" prefix is used.
 	if townRoot, err := workspace.FindFromCwd(); err == nil && townRoot != "" {
 		_ = session.InitRegistry(townRoot)
-		_ = config.LoadAgentRegistry(config.DefaultAgentRegistryPath(townRoot))
+		if err := config.LoadAgentRegistry(config.DefaultAgentRegistryPath(townRoot)); err != nil {
+			fmt.Fprintf(os.Stderr, "WARNING: failed to load agent registry %s: %v\n",
+				config.DefaultAgentRegistryPath(townRoot), err)
+		}
 	}
 
 	// Get the root command name being run
--- a/internal/cmd/root_test.go
+++ b/internal/cmd/root_test.go
@@ -138,6 +138,8 @@ func TestPersistentPreRunLoadsAgentRegistry(t *testing.T) {
 	// Without this, NixOS users whose Claude binary is ".claude-unwrapped" get
 	// their sessions killed every 3 minutes because the builtin preset only
 	// lists ["node", "claude"].
+	//
+	// NOTE: cannot use t.Parallel() — mutates cwd and global agent registry.
 	config.ResetRegistryForTesting()
 	t.Cleanup(config.ResetRegistryForTesting)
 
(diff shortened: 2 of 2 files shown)


GraemeF:
The CI failure is unrelated to this change — it's `TestRoutesJSONLCorruption/TownLevelRoutesNotCorrupted` failing with `issues.jsonl should exist after gt install`. The same test fails identically on the `pr/push-url-config` branch ([run 22071774920](https://github.com/steveyegge/gastown/actions/runs/22071774920)). The new test added by this PR (`TestPersistentPreRunLoadsAgentRegistry`) passes on CI.

julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @GraemeF! This is a really well-diagnosed fix for a painful NixOS-specific issue — the `.claude-unwrapped` binary name causing daemon heartbeat, zombie cleanup, and dog patrols to falsely detect running Claude sessions as dead and kill them every 3 minutes. The fix is minimal, correctly placed, and idempotent. Great work.

### Original PR Review
Initial dual-model review (Claude + Codex) found the core fix **correct** — `LoadAgentRegistry` is called at the right time in `persistentPreRun`, after `session.InitRegistry` and before any command handler. Both reviewers independently verified idempotency via `loadedPaths` caching and confirmed all 22+ `IsAgentAlive` callers benefit. One **major** finding: the error from `LoadAgentRegistry` was silently discarded with `_`, meaning users with malformed `settings/agents.json` would get zero diagnostic signal — the exact same false-dead behavior this PR fixes.

### Maintainer Changes
- Changed `_ = config.LoadAgentRegistry(...)` to check the error and emit a `WARNING` to stderr with the file path and error message (follows existing warning pattern in `persistentPreRun`)
- Added `TestPersistentPreRunMalformedAgentRegistry` covering the error/fallback path (malformed JSON → warning emitted, builtin defaults preserved)
- Added `// NOTE: cannot use t.Parallel()` annotations to tests that mutate global state

Stats: +51 -1 across 2 files (`root.go`, `root_test.go`)

### Final Review Status
All clear — no remaining issues found. Both reviewers approved on re-review. Two minor nits noted (stderr assertion in test is nice-to-have; `loadedPaths` cache-until-restart is pre-existing behavior) — neither is blocking.

### Review Iterations
2 review passes performed.

---

PR #1575, opened 2026-02-16
fix(nudge): prevent nudge loss from transient Windows file locks

julianknutsen:
## Summary
On Windows, transient sharing violations (antivirus, search indexer) can cause `os.Rename` to fail for all concurrent drainers on a nudge file, silently dropping it. A secondary bug also deleted claimed files when `os.ReadFile` hit a transient error, permanently losing the nudge.

## Related Issue
Fixes nondeterministic Windows CI failure in `TestConcurrentDrainNoDoubleDeli`:
https://github.com/steveyegge/gastown/actions/runs/22066376229/job/63759176127?pr=1559
```
--- FAIL: TestConcurrentDrainNoDoubleDeli (0.08s)
    queue_test.go:545: concurrent Drains delivered 9 total nudges, want exactly 10
```

## Changes
- **queue.go**: On transient `ReadFile` failure (non-ENOENT) after claiming a nudge, unclaim the file (rename back to `.json`) instead of deleting it — nudge is retried on next `Drain` call; orphan sweep remains as safety net
- **queue_test.go**: Add straggler sweep (up to 3 retries, 50ms apart) after concurrent drains, mirroring real-world hook behavior where `Drain` is called repeatedly

## Testing
- [x] Unit tests pass (`go test ./internal/nudge/...`)
- [x] 50-iteration stress test passes (`go test -run TestConcurrentDrain -count=50`)
- [x] Race detector clean (`go test -race ./internal/nudge/...`)
- [ ] Windows CI passes (this PR)

## Checklist

### DIFF
--- a/internal/nudge/queue.go
+++ b/internal/nudge/queue.go
@@ -208,8 +208,10 @@ func Drain(townRoot, session string) ([]QueuedNudge, error) {
 				// File vanished between rename and read — treat as lost race
 				continue
 			}
-			// Genuinely unreadable — clean up so it doesn't become a ghost entry
-			_ = os.Remove(claimPath)
+			// Transient read error (e.g., Windows AV/indexer holding a share
+			// lock) — unclaim so the nudge can be retried on a future Drain
+			// call rather than permanently lost.
+			_ = os.Rename(claimPath, path) // best-effort unclaim; orphan sweep catches failures
 			continue
 		}
 
--- a/internal/nudge/queue_test.go
+++ b/internal/nudge/queue_test.go
@@ -541,7 +541,26 @@ func TestConcurrentDrainNoDoubleDeli(t *testing.T) {
 		total += len(nudges)
 	}
 
+	// On Windows, transient sharing violations (antivirus, search indexer)
+	// can prevent all concurrent drainers from claiming a file.  The nudge
+	// stays as .json and is picked up on the next Drain — mirror that here
+	// with a straggler sweep so the test validates no-loss, not one-shot
+	// completeness.
+	for retries := 0; retries < 3 && total < count; retries++ {
+		time.Sleep(50 * time.Millisecond)
+		stragglers, err := Drain(townRoot, session)
+		if err != nil {
+			t.Fatalf("straggler Drain: %v", err)
+		}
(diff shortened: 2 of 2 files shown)


julianknutsen:
## Self-Review

### Root Cause Analysis

The failure occurs in `TestConcurrentDrainNoDoubleDeli` — 5 goroutines concurrently call `Drain` on a queue of 10 nudges. On Linux this reliably delivers all 10; on Windows it intermittently delivers 9.

**Primary mechanism (rename blocked by sharing violation):**
On Windows, antivirus (Defender) and the search indexer open newly-created `.json` files for scanning, holding a sharing violation. When all 5 concurrent drainers attempt `os.Rename` on a locked file, *all* fail. The nudge remains as `.json` in the directory but no drainer returns it — it appears lost.

This is not a bug in the Drain logic itself. In production, the hook calls `Drain` repeatedly, so the nudge is picked up on the next invocation. The test just didn't account for this.

**Secondary mechanism (ReadFile failure deletes nudge):**
If `os.Rename` succeeds (claiming the file) but `os.ReadFile` subsequently fails with a transient error (sharing violation, not ENOENT), the old code deleted the `.claimed` file — *permanently* losing the nudge. This is a real production bug, not just a test issue.

### What the fix does

1. **`queue.go:211-214`** — On transient `ReadFile` failure, rename the claimed file back to `.json` (unclaim it) so a future `Drain` retries it. Falls back to leaving the `.claimed` file for orphan sweep if the unclaim rename also fails. This matches the existing pattern where the orphan sweep is the safety net for stuck claims.

2. **`queue_test.go:544-556`** — After the concurrent drain burst, if `total < count`, do up to 3 sequential `Drain` calls (50ms apart) to sweep stragglers. This mirrors real-world usage without weakening the assertion: total must still equal exactly `count`.

### Risk assessment

- **Low risk.** The production change is 1 line: `os.Remove` → `os.Rename` (unclaim). The only behavior change is that a transiently unreadable nudge is retried instead of discarded.
- The test change only adds retries *after* the concurrent phase — the concurrent race logic is unchanged.
- No new dependencies, no API changes, no breaking changes.

PR #1576, opened 2026-02-16
RFC: gt patrol step-drift — detect polecats not following molecule steps

chrismilleruk:
## Problem

When polecats work on molecule-driven tasks, they're supposed to close each step as they complete it (`bd close <step-id>`). There are 9 steps in the mol-polecat-work formula, so effectively each polecat has a progress bar — is it on step 3 of 9, or stuck on step 1?

In practice, we found polecats often skip closing steps as they go. They freestyle through the work and either batch-close at the end or forget entirely. The progress stays at ○○○○○○○○○ for the entire session, making it impossible to distinguish a stuck polecat from a productive one.

We think that step drift — working ahead without closing steps — is a useful signal that the polecat isn't following the molecule formula. A nudge to get back on track is more useful than waiting for it to finish and discovering it skipped half the steps.

## Root cause

There's no automated way to detect this during a patrol cycle. The deacon's patrol formula describes the *intent* ("check if polecats are closing steps") but leaves the deacon to figure out the mechanics each time — parsing `bd show` output, finding the right Dolt branch, comparing step counts against session age. This is fragile and expensive in context.

There's also a subtlety with Dolt branch isolation: polecats write to isolated branches (`polecat-<name>-<timestamp>`), so step closures aren't visible on `main`. Any tool checking step status needs to find and read from the polecat's branch, not the default branch.

## What this PR does

Adds `gt patrol step-drift` — a subcommand the deacon can call during patrol to detect and nudge polecats with unclosed steps.

For each working polecat it:
1. Finds the polecat's isolated Dolt branch (most recent `polecat-<name>-*` in the rig's data dir)
2. Reads molecule step status from that branch via `BD_BRANCH` env var
3. Checks session age via tmux
4. Flags drift if the polecat has been working 5+ minutes with zero steps closed
5. Optionally sends a nudge (`--nudge`) reminding the polecat to close steps incrementally

### Example: human-readable output (default)

```
$ gt patrol step-drift
patrol-step-drift
================================================================================
  ▶ obsidian   sa-zdcd      ●●●●○○○○○  Consolidate screenshot pipeline  8m
    │ ⏺ Bash(flutter test test/screenshots/)
    │   ⎿  All 12 tests passed!
    │ ⏺ Good, tests pass. Let me close step 6 and move to cleanup.
    │ ⏺ Bash(bd close sa-wisp-9p3n)
    │   ⎿  ✓ Closed sa-wisp-9p3n: Closed

  ▶ jasper     sa-if74i     ○○○○○○○○○  Fix broken samples + promo CTA  7m
    │ ⏺ Bash(git diff --stat)
    │   ⎿  3 files changed, 142 insertions(+)
    ⚡ Step drift detected (7m, 0 steps closed)

  Steps: ①load ②branch ③preflight ④implement ⑤review ⑥test ⑦cleanup ⑧prepare ⑨submit
  ● = done  ○ = pending  ⚡ = drifting
```

### Example: agent/JSON output (`--agent`)

```json
$ gt patrol step-drift --agent
[
  {
    "rig": "sailingnav",
    "name": "obsidian",
    "bead": "sa-zdcd",
    "title": "Consolidate screenshot pipeline",
    "state": "working",
    "age_min": 8.2,
    "closed": 4,
    "total": 9,
    "drifting": false,
    "nudged": false,
    "branch": "polecat-obsidian-1771244354"
  },
  {
    "rig": "sailingnav",
    "name": "jasper",
    "bead": "sa-if74i",
    "title": "Fix broken samples + promo CTA",
    "state": "working",
    "age_min": 7.1,
    "closed": 0,
    "total": 9,
    "drifting": true,
    "nudged": false,
    "branch": "polecat-jasper-1771195104"
  }
]
```

Two output modes:
- **Default (human-readable):** progress bars, peek at polecat output, session age — useful for humans monitoring polecats
- **`--agent`:** JSON array — useful for the deacon or other automation

Also supports `--watch` for a live refreshing display and `--threshold` to tune the drift detection window.

The formula change updates the `step-drift-nudge` patrol step to call this subcommand instead of describing the detection logic in prose.

## Status

**Not fully proven yet.** We've verified that `BD_BRANCH` correctly reads step closures from a polecat's isolated Dolt branch, but haven't observed the progress bar updating live during an active session — polecats keep finishing or dying before we catch them mid-work. The examples above are illustrative, not captured from a live run.

This is an early take on the problem. Putting it up for discussion on the approach before investing more in hardening it.

## Open questions

- Is this the right home for this? It touches patrol, polecat, and Dolt concerns. Happy to restructure.
- The Dolt branch lookup scans all databases (not just the rig's) because bead prefix routing can place branches in a different DB than the rig name. This feels like it could be fragile — open to suggestions.
- The 5-minute threshold is a guess. Might need tuning based on real usage.
- Should the nudge message be configurable or is the hardcoded one sufficient?

### DIFF
--- a/.beads/formulas/mol-deacon-patrol.formula.toml
+++ b/.beads/formulas/mol-deacon-patrol.formula.toml
@@ -549,9 +549,43 @@ No action needed - Witness is doing its job.
 investigate why the Witness isn't cleaning up properly."""
 
 [[steps]]
+id = "step-drift-nudge"
+title = "Detect and nudge step-drifting polecats"
+needs = ["zombie-scan"]
+description = """
+Detect polecats with unclosed molecule steps and nudge them.
+
+Step drift occurs when a polecat has been working for several minutes without
+closing any molecule steps. This indicates the polecat may not be following the
+formula or is stuck without signaling.
+
+**Run the step-drift scan with nudge:**
+```bash
+gt patrol step-drift --agent --nudge
+```
+
+This command:
+1. Lists all working polecats across all rigs
+2. Reads step status from each polecat's isolated Dolt branch
+3. Detects drift (default: 5 min with 0 steps closed)
+4. Sends a nudge to drifting polecats reminding them to close steps
+5. Returns JSON report for logging
+
+**Interpreting results:**
+- `drifting: true` — polecat exceeded threshold with 0 steps closed
+- `nudged: true` — nudge message was sent
+- `closed: N` — number of canonical steps completed
+
+**If many polecats are drifting:**
+This may indicate a systemic issue (formula too complex, tooling broken).
+Escalate to Mayor if >50% of active polecats show drift.
+
+**Exit criteria:** Step-drift scan completed, drifting polecats nudged."""
+
+[[steps]]
 id = "plugin-run"
 title = "Execute registered plugins"
-needs = ["zombie-scan"]
+needs = ["step-drift-nudge"]
 description = """
 Execute registered plugins.
 
--- a/internal/cmd/patrol.go
+++ b/internal/cmd/patrol.go
@@ -58,6 +58,7 @@ Examples:
 func init() {
 	patrolCmd.AddCommand(patrolDigestCmd)
 	patrolCmd.AddCommand(patrolNewCmd)
+	patrolCmd.AddCommand(patrolStepDriftCmd)
 	rootCmd.AddCommand(patrolCmd)
 
 	// Patrol digest flags
--- a/internal/cmd/patrol_step_drift.go
+++ b/internal/cmd/patrol_step_drift.go
@@ -1,0 +1,496 @@
+package cmd
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"regexp"
+	"strconv"
+	"strings"
+	"time"
+
+	"github.com/spf13/cobra"
+	"github.com/steveyegge/gastown/internal/style"
+	"github.com/steveyegge/gastown/internal/workspace"
+)
+
+var (
+	stepDriftAgent     bool
+	stepDriftNudge     bool
+	stepDriftThreshold int
+	stepDriftWatch     bool
+)
+
+// stepsOrder defines the canonical molecule step names in execution order.
+var stepsOrder = []string{
+	"Load context",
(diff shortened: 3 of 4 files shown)


steveyegge:
## Review: ZFC Compliance & Architecture Fit

Thanks for identifying a real problem — polecats freestyling through steps and not closing them is a genuine observability gap. The diagnosis is solid. But the proposed solution moves in the wrong direction architecturally.

### ZFC Compliance: Poor

ZFC = Zero Framework Cognition — "Agent decides. Go transports."

This PR puts cognition into Go code. It hardcodes:
- A 5-minute threshold for "drift"
- "Zero steps closed" = drifting (a judgment call baked into code)
- A nudge message template
- Branch-scanning logic to find and interpret step status

The patrol formula's own health-scan step demonstrates the correct pattern: *"ZFC Principle: You (Claude) make the judgment call about what is 'stuck' or 'unresponsive' — there are no hardcoded thresholds in Go."* This PR does the opposite.

### Bitter Lesson Positioning: Wrong Side

The bitter lesson says: bet on compute and learning, not hand-crafted heuristics. This PR hand-crafts a heuristic (5min + 0 steps = drift) in Go, when the Deacon is literally an AI agent that can already observe bd show output, check session age via tmux, and *reason* about whether a polecat is drifting. The AI can understand nuance — maybe a polecat is on a complex step 1 that legitimately takes 10 minutes. A hardcoded heuristic can't.

### Should Be a Molecule Step

The Deacon patrol formula already has 18 steps. This should be a patrol step along the lines of:

> "For each working polecat, check bd show on the polecat's branch. Compare closed step count against session age. If the polecat has been working 5+ minutes with no progress, nudge it to close steps incrementally. Use your judgment — some steps are legitimately long."

The Deacon already has bd, tmux, and gt nudge access. No new Go code needed.

### What's Worth Keeping

The BD_BRANCH env trick for reading step closures from polecat isolated Dolt branches is the genuinely useful bit. That could be exposed as a transport primitive (e.g., bd show --branch=polecat-*) rather than embedded in a detection subcommand. Transport, not cognition.

### Recommendation

Close or rework. Extract the branch-reading transport primitive, and express the detection/nudging logic as a molecule step where the AI does the reasoning. That's ZFC-compliant and bitter-lesson-aligned.

PR #1579, opened 2026-02-16
fix(beads): use runWithRouting for agent state updates

athosmartins:
Fixes #gt-wk46l

## Problem

`bd agent state` fails when agent bead ID doesn't match configured prefix:

```
Error: failed to auto-create agent bead wa-whatsapp_automation-polecat-rust: validate issue ID prefix: issue ID 'wa-whatsapp_automation-polecat-rust' does not match configured prefix 'hq'
```

## Solution

Changed `UpdateAgentState` to use `runWithRouting` instead of `run` for bd agent state calls. This allows bd to resolve agent beads with rig-specific prefixes (e.g., `wa-whatsapp_automation-polecat-rust`) when run from a different prefix context (e.g., `hq-`).

The fix follows the same pattern already used for slot operations in the same function, which already use `runWithRouting` for cross-prefix operations.

## Testing

- ✅ Code compiles successfully
- ✅ Follows existing pattern for cross-prefix operations

### DIFF
--- a/internal/beads/beads_agent.go
+++ b/internal/beads/beads_agent.go
@@ -370,7 +370,9 @@ func (b *Beads) ResetAgentBeadForReuse(id, reason string) error {
 // which caused inconsistencies with bd slot commands (see GH #gt-9v52).
 func (b *Beads) UpdateAgentState(id string, state string, hookBead *string) error {
 	// Update agent state using bd agent state command
-	_, err := b.run("agent", "state", id, state)
+	// Use runWithRouting so bd can resolve cross-prefix agent beads (e.g., wa-*
+	// agent beads from hq context) via routes.jsonl instead of BEADS_DIR.
+	_, err := b.runWithRouting("agent", "state", id, state)
 	if err != nil {
 		return fmt.Errorf("updating agent state: %w", err)
 	}


julianknutsen:
## Maintainer Adoption Review

Thanks for the contribution, @athosmartins! This is a clean, well-targeted fix — the `bd agent state` call was the one remaining operation in `UpdateAgentState` that used `run()` instead of `runWithRouting()`, breaking cross-prefix agent bead resolution. Good catch on aligning it with the slot operations pattern.

This PR was reviewed via dual-model pipeline (Claude Opus 4.6 + GPT-5.3 Codex). No maintainer changes needed.

### Original PR Review

**Decision: approve.** Both reviewers validated the fix as correct — `run()` forces `BEADS_DIR` to local prefix, while `runWithRouting()` lets `bd` discover routes via `routes.jsonl` for cross-prefix resolution. Two non-blocking minor observations:

1. **Other `bd agent state` paths** in `done.go` and `deacon.go` may bypass `UpdateAgentState` — flagged by Codex but with low confidence. Claude searched and found those sites don't encounter the cross-prefix scenario. Worth auditing as follow-up.
2. **No regression test** for cross-prefix `UpdateAgentState` — acceptable given this is a one-line method swap matching the established pattern used by 12+ existing `runWithRouting` calls in the same file.

### Maintainer Changes

None — PR is clean as submitted.

### Final Review Status

All clear — no remaining issues found. Approve with zero blockers.

### Review Iterations

1 review pass performed.

---

PR #1580, opened 2026-02-16
feat(mail): auto-nudge idle recipients on gt mail send

seanbearden:
## Summary
- When `gt mail send` delivers a message, check if the recipient's tmux session is idle and send an immediate nudge. If busy, fall back to queued nudge for cooperative delivery at next turn boundary.
- Adds `SuppressNotify` field to `Message` so the CLI can suppress the router's built-in queue-only notification and handle delivery itself.
- Exports `AddressToSessionIDs` and adds `IsRecipientMuted` public wrapper for use from the CLI layer.
- Adds `--no-notify` flag to suppress auto-nudge when desired.

Closes #1578 / gt-qjko

## Test plan
- [x] `go build ./...` — clean compile
- [x] `go test ./internal/mail/...` — passes (exported function names updated)
- [x] `go test ./internal/cmd/...` — passes
- [x] `go vet ./internal/cmd/... ./internal/mail/...` — 0 issues
- [ ] Manual: `gt mail send gastown/crew/X -s "test" -m "hello"` with X idle → should see `↳ Nudged idle recipient`
- [ ] Manual: same with `--no-notify` → no nudge output
- [ ] Manual: same with X busy → should see `↳ Queued notification for`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1586, opened 2026-02-16
fix: regenerate embedded formulas to remove stale mol-wisp-compact

seanbearden:
## Summary
- `mol-wisp-compact.formula.toml` was removed from `.beads/formulas/` but the embedded copy in `internal/formula/formulas/` was never regenerated
- This caused the "Check embedded formulas" CI check to fail on main after merging PR #1580
- Fix: `go generate ./internal/formula/...` to resync

## Test plan
- [x] `go generate ./internal/formula/...` produces no further diff
- [ ] CI "Check embedded formulas" step passes


julianknutsen:
Closing — this PR removed the embedded copy but the correct fix is to add the formula to the canonical location (`.beads/formulas/`), which is the source of truth for `go generate`. See #1593 for the proper fix.

PR #1603, opened 2026-02-17
fix(doctor): prevent killing non-gastown tmux sessions

olpie101:
## Summary

- Remove fallback in `matchPrefix()` that incorrectly matched any session with a dash as a gastown session
- Sessions like `gs-1923` or `dotfiles-main` were being parsed as polecats and killed by `gt doctor --fix`
- Now only sessions with registered prefixes (gt-, bd-, hq-, etc.) are recognized as gastown sessions

## Problem

The `matchPrefix()` function in `registry.go` had a fallback that split ANY session name on the first dash. This meant personal tmux sessions like `gs-1923`, `dotfiles-main`, or any `<name>-<suffix>` pattern would be:

1. Parsed as a gastown session (prefix = first segment)
2. Flagged as "orphan" (since the prefix wasn't a registered rig)
3. **Killed** by `gt doctor --fix`

## Solution

Remove the fallback logic so only sessions with registered prefixes or `hq-*` patterns are recognized as gastown sessions. Unknown sessions are now ignored entirely.

## Testing

- Updated existing tests to expect unknown prefixes to fail parsing
- Verified `gt doctor` no longer flags personal sessions
- All `./internal/session/...` and `./internal/doctor/...` tests pass

### DIFF
--- a/internal/doctor/orphan_check_test.go
+++ b/internal/doctor/orphan_check_test.go
@@ -105,12 +105,12 @@ func TestIsCrewSession(t *testing.T) {
 		session string
 		want    bool
 	}{
-		{"gt-crew-joe", true},   // gastown crew (prefix: gt)
-		{"bd-crew-max", true},   // beads crew (prefix: bd)
-		{"nif-crew-a", true},    // niflheim crew (prefix: nif)
-		{"gt-witness", false},   // witness, not crew
-		{"gt-refinery", false},  // refinery, not crew
-		{"gt-polecat1", false},  // polecat, not crew
+		{"gt-crew-joe", true},  // gastown crew (prefix: gt)
+		{"bd-crew-max", true},  // beads crew (prefix: bd)
+		{"nif-crew-a", true},   // niflheim crew (prefix: nif)
+		{"gt-witness", false},  // witness, not crew
+		{"gt-refinery", false}, // refinery, not crew
+		{"gt-polecat1", false}, // polecat, not crew
 		{"hq-deacon", false},
 		{"hq-mayor", false},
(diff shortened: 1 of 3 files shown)


julianknutsen:
## Adoption Review

Thanks for the contribution, @olpie101! This is a great catch — the fallback dash-splitting in `matchPrefix()` was silently claiming ownership of any tmux session with a hyphen, which meant `gt doctor --fix` could destroy users' personal sessions. The fix is clean and well-scoped: removing the fallback so only registered prefixes are recognized.

### Review Summary
Triple-model review (Claude + Codex + Gemini) unanimously approved. All three reviewers independently confirmed the fix is correct and complete. The 6-line removal in `registry.go` is the right approach, and tests are properly updated to reflect the new strict-matching behavior. No blockers or regressions found.

### Follow-Up Opportunities (non-blocking)

These are minor items worth knowing about — none block this merge:

- **Empty/stale prefix registry (medium confidence):** If `rigs.json` is missing or corrupt, the registry would be empty and all sessions would be skipped by the orphan check. In practice `InitRegistry` runs in `persistentPreRun` so this is unlikely, but a defensive warning when the registry is empty could help future debugging.
- **Deleted-rig session visibility:** If a rig is removed from config while its tmux sessions are still running, those sessions become unparseable and invisible to orphan detection. A debug log line when parsing fails for dash-containing names would aid diagnosis.
- **Stale comments in `identity.go`:** The doc comments on `ParseSessionName`/`ParseSessionNameWithRegistry` still describe the old fallback behavior. Updating them would prevent confusion for future contributors.

1 review pass performed.

---

PR #1604, opened 2026-02-17
fix(done): handle already-deleted molecules gracefully

olpie101:
## Summary

- Handle `ErrNotFound` when closing attached molecules in `gt done`
- Treat already-deleted/burned molecules as successfully closed instead of retrying

## Problem

When a molecule is burned or deleted before `gt done` runs, the close attempt fails with `ErrNotFound`. The code would:

1. Retry 3 times unnecessarily
2. Print a warning suggesting the close failed
3. Skip closing the hooked bead (treating it as blocked)

## Solution

Check for `ErrNotFound` in the retry loop and treat it as success - if the molecule doesn't exist, it's effectively already closed. This avoids unnecessary retries and misleading warnings.

### DIFF
--- a/internal/cmd/done.go
+++ b/internal/cmd/done.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"os/signal"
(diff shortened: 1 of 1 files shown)


julianknutsen:
## Adoption Review

Thanks for the contribution, @olpie101! This is a clean, well-targeted fix for a real problem — when a molecule gets burned or deleted before `gt done` runs, the retry loop was wasting time and then orphaning state by skipping the hooked bead close. Your fix correctly short-circuits the retry and lets cleanup proceed normally.

### Review Summary
Triple-model review (Claude + Codex + Gemini) all confirmed the fix is correct. The `errors.Is(err, beads.ErrNotFound)` check properly handles the TOCTOU window, the error propagation chain through `wrapError` was verified, and the downstream effects (hooked bead close + agent hook clear) are strictly positive. No blockers or majors found.

Two minor items flagged as FYI — neither is blocking:

- **Test coverage (minor):** The existing `TestDoneClosesAttachedMolecule` only covers the happy path. A test variant where `bd close` returns "not found" for the molecule would exercise the new branch and verify the hooked bead still gets closed.
- **Broad `ErrNotFound` mapping (minor, pre-existing):** `wrapError` in `beads.go` maps any stderr containing "not found" to `ErrNotFound`, which is broader than strictly "issue not found." This isn't introduced by your PR — it's the established codebase pattern — but worth noting for future tightening.

1 review pass performed.

---

PR #1605, opened 2026-02-17
fix(molecule): update agent heartbeat on idle cycle timeout

olpie101:
## Summary

- Update `last_activity` timestamp when `await-signal` times out due to idle cycles
- Keeps watchers informed that the agent is still alive even when idle

## Problem

When an agent is idle (no signals arriving), `await-signal` would time out but not update the agent's `last_activity` timestamp. Watchers monitoring agent health could incorrectly assume the agent was dead/stuck.

## Solution

Call `updateAgentHeartbeat()` on idle cycle timeout to refresh the `last_activity` timestamp, signaling to watchers that the agent is still responsive.

### DIFF
--- a/internal/cmd/molecule_await_signal.go
+++ b/internal/cmd/molecule_await_signal.go
@@ -209,6 +209,13 @@ func runMoleculeAwaitSignal(cmd *cobra.Command, args []string) error {
 		} else {
 			result.IdleCycles = newIdleCycles
 		}
+		// Update last_activity so watchers know agent is still alive
+		if err := updateAgentHeartbeat(awaitSignalAgentBead, beadsDir); err != nil {
+			if !awaitSignalQuiet {
+				fmt.Printf("%s Failed to update agent heartbeat: %v\n",
+					style.Dim.Render("⚠"), err)
+			}
+		}
 		// Clear the backoff window — timeout completed normally
 		_ = clearAgentBackoffUntil(awaitSignalAgentBead, beadsDir)
 	} else if result.Reason == "signal" && awaitSignalAgentBead != "" {


julianknutsen:
## Adoption Review

Thanks for the contribution, @olpie101! This is a well-targeted fix for a real operational issue — idle agents tripping false staleness alerts is exactly the kind of subtle bug that degrades trust in the monitoring system. The implementation is clean and mirrors the existing signal-branch pattern perfectly.

### Review Summary
Triple-model review (Claude + Codex + Gemini) unanimously approved. The fix correctly adds `updateAgentHeartbeat()` to the idle-timeout branch of `await-signal`, preventing watchers from falsely flagging idle agents as dead when backoff approaches the 5-minute staleness threshold. All three reviewers confirmed the fix is correct with zero drift from existing patterns. Two minor findings noted below for your awareness — neither blocks merge.

### Findings (non-blocking, for awareness)

- **Error Handling & Resilience (minor, pre-existing):** `updateAgentHeartbeat` shells out to `bd` via `exec.Command(...).Run()` with no context deadline. A hung `bd` process could theoretically block the patrol loop. This is a pre-existing condition shared with the signal branch (identical call pattern) — your PR doesn't introduce or worsen it. A follow-up adding `exec.CommandContext` with a short timeout to `updateAgentHeartbeat` itself would harden both call sites.

- **Test Evidence Quality (minor):** No test currently exercises the heartbeat-on-timeout path. The code mirrors an existing pattern and shells out to `bd`, making unit testing non-trivial without a stubbed binary. A follow-up test would be welcome but is not required.

1 review pass performed.

---

PR #1606, opened 2026-02-17
fix(prime): use agent bead hook_bead field as primary hook source

olpie101:
## Summary

- Use agent bead's `hook_bead` DB column as primary source for detecting hooked work
- Falls back to querying by assignee if `hook_bead` is not set

## Problem

`detectSessionState()` was querying beads by assignee to find hooked work, which could be slow or return stale results. The authoritative hook information is already stored in the agent bead's `hook_bead` field (set by `bd slot` during sling).

## Solution

Read `hook_bead` from the agent bead first (same strategy as `gt hook` uses). Only fall back to the assignee query if the agent bead doesn't have a `hook_bead` set.

This aligns `gt prime` with `gt hook` for consistent hook detection.

### DIFF
--- a/internal/cmd/prime.go
+++ b/internal/cmd/prime.go
@@ -403,6 +403,8 @@ func checkSlungWork(ctx RoleContext) bool {
 }
 
 // findAgentWork looks up hooked or in-progress beads assigned to this agent.
+// Primary: reads hook_bead from the agent bead (same strategy as detectSessionState/gt hook).
+// Fallback: queries by assignee for agents without an agent bead.
 // Returns nil if no work is found.
 func findAgentWork(ctx RoleContext) *beads.Issue {
 	agentID := getAgentIdentity(ctx)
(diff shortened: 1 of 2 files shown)


julianknutsen:
## Adoption Review

Thanks for the contribution, @olpie101! Using the agent bead's `hook_bead` field as the primary source for hook detection in `detectSessionState` is the right call — it aligns `gt prime` with the authoritative strategy already used by `gt hook` and `gt mol status`, and the assignee-query fallback ensures backward compatibility.

### Initial Review
The initial triple-model review (Claude + Codex + Gemini) flagged three issues: (1) a stale `hook_bead` could force false "autonomous" state since the target bead's status wasn't verified, (2) the main agent startup path (`findAgentWork`) still used the old assignee-only strategy creating a split-brain with `--state`, and (3) the target bead wasn't checked for existence before use.

### Maintainer Changes
Pushed a fixup commit addressing all three findings:
- **Stale hook_bead safety:** Both `detectSessionState` and `findAgentWork` now resolve and load the target bead, requiring status `hooked` or `in_progress` before returning autonomous state. Stale references fall through to the assignee query.
- **findAgentWork alignment:** The main startup path now uses the same "agent bead first, assignee fallback" strategy as `detectSessionState`, `gt hook`, and `gt mol status`.
- **Existence check:** Mirrors `molecule_status.go` and `signal_stop.go` patterns by verifying the target bead exists before using it.

Changes: 2 files, +29 -3 lines.

### Final Review Status
All clear — iteration review returned **approve** with no blockers or majors. Minor follow-up items identified (missing `Type == "agent"` guard, code duplication opportunity, test coverage blocked by bd CLI bug) are tracked separately.

### Review Iterations
2 review passes performed.

---

PR #1607, opened 2026-02-17
fix(sling): proactively burn stale molecules before re-sling

olpie101:
## Summary

- Check for existing attached molecules BEFORE applying a new formula during sling
- With `--force`: burn old molecules and proceed with new formula
- Without `--force`: error out to prevent accidental molecule duplication

## Problem

When re-slinging a bead that already has an attached molecule (wisp), each sling would create a NEW molecule bonded to the bead, leaving orphaned molecules accumulating. The existing `DetectOrphanedMolecules` in witness catches these reactively, but by then the orphans already exist.

## Solution: Proactive vs Reactive

| Approach | Where | When | Action |
|----------|-------|------|--------|
| **Reactive** (existing) | Witness `DetectOrphanedMolecules` | After orphans exist | Detects and closes orphaned molecules |
| **Proactive** (this PR) | Sling time | Before creating new molecule | Burns existing molecules first |

This PR adds **proactive prevention** at sling time, complementing the existing reactive cleanup:

1. `collectExistingMolecules()` - finds molecules via both dependency bonds (ground truth) and description metadata
2. `burnExistingMolecules()` - detaches molecule from bead, then closes the orphaned wisp

## Changes

- `sling_helpers.go`: Extended `beadInfo` struct, added collection/burn helpers
- `sling.go`: Added pre-formula molecule guard
- `sling_batch.go`: Same guard for batch operations

### DIFF
--- a/internal/cmd/sling.go
+++ b/internal/cmd/sling.go
@@ -482,13 +482,19 @@ func runSling(cmd *cobra.Command, args []string) error {
 	// Checks both dependency bonds (ground truth) and description metadata.
 	// When re-slinging with --force, burn ALL existing molecules before creating a new one.
 	// Without this, each sling creates a new wisp bonded to the bead, leaving orphaned molecules.
+	// NOTE: Uses local `force` (not `slingForce`) to respect auto-force paths (dead agent detection).
 	if formulaName != "" {
 		existingMolecules := collectExistingMolecules(info)
 		if len(existingMolecules) > 0 {
-			if slingForce {
+			if slingDryRun {
+				fmt.Printf("  Would burn %d stale molecule(s): %s\n",
+					len(existingMolecules), strings.Join(existingMolecules, ", "))
+			} else if force {
 				fmt.Printf("  %s Burning %d stale molecule(s) from previous assignment: %s\n",
 					style.Warning.Render("⚠"), len(existingMolecules), strings.Join(existingMolecules, ", "))
-				burnExistingMolecules(existingMolecules, beadID, townRoot)
+				if err := burnExistingMolecules(existingMolecules, beadID, townRoot); err != nil {
+					return fmt.Errorf("burning stale molecules: %w", err)
+				}
 			} else {
 				return fmt.Errorf("bead %s already has %d attached molecule(s): %s\nUse --force to replace, or --hook-raw-bead to skip formula",
 					beadID, len(existingMolecules), strings.Join(existingMolecules, ", "))
--- a/internal/cmd/sling_batch.go
+++ b/internal/cmd/sling_batch.go
@@ -102,6 +102,29 @@ func runBatchSling(beadIDs []string, rigName string, townBeadsDir string) error 
 			continue
 		}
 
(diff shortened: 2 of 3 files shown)


julianknutsen:
## Adoption Review

Thanks for the contribution, @olpie101! This PR addresses a real problem — re-slinging with `--force` was silently creating orphaned molecules because new wisps got bonded without cleaning up old ones. The two-step burn approach (detach via Go API + force-close wisps) is the right design, and the dual detection path (dependency bonds + description metadata) is solid defensive programming.

### Initial Review

The initial triple-model review (Claude + Codex + Gemini) found **2 blockers and 3 majors**:

1. **Blocker:** Guard used `slingForce` instead of local `force`, breaking dead-agent auto-force
2. **Blocker:** Dry-run mode executed burn side-effects (detach + close)
3. **Major:** `bd mol detach` shell command had zero existing usage — all detach paths use Go API
4. **Major:** Burn errors swallowed, allowing sling to proceed after partial burn
5. **Major:** `DependsOn` field unreliable for molecule detection (codebase warns to use `Dependencies`)

Plus 3 minors: `bd close` without `--force`, batch guard not gated on `formulaName`, no tests.

### Maintainer Changes

All initial findings resolved in a single fixup commit (+67 -50 across 3 files):

- Switched from `slingForce` to local `force` to respect auto-force paths
- Added dry-run guard (prints "Would burn" instead of executing)
- Replaced `bd mol detach` with Go API `DetachMoleculeWithAudit` (proper locking + audit logging)
- Made `burnExistingMolecules` return error; sling aborts on detach failure
- Used `Dependencies []IssueDep` instead of unreliable `DependsOn` field
- Used `ForceCloseWithReason` for wisp closure (handles open children)
- Gated batch guard on `formulaName != ""` for consistency
- Moved batch molecule guard before polecat spawn to avoid wasted spawn/cleanup

### Final Review Status

Decision: **approve** — all initial blockers and majors resolved. Remaining findings are minor/nit follow-ups:

1. **Minor — Zombie dependencies after burn** (all three reviewers, high confidence): Closed wisps persist as dependency bonds after burn. Future slings may re-detect them. Mitigated by reactive `DetectOrphanedMolecules`. Suggested follow-up: filter `collectExistingMolecules` by `dep.Status != "closed"`.

2. **Minor — No unit tests for collect/burn helpers** (claude+codex, high confidence): `collectExistingMolecules` and `burnExistingMolecules` have no test coverage. Suggested follow-up: add table-driven tests covering empty deps, bonds-only, description-only, detach failure, and close failure cases.

3. **Nit — Missing `Agent` field in DetachOptions** (gemini, high confidence): Audit log entries will have empty `detached_by`. Can add `Agent: detectCurrentAgent()` inline.

4. **Nit — Batch uses `slingForce` instead of local `force`** (claude, high confidence): Functionally equivalent today since batch has no auto-force path. Document if batch auto-force is added later.

### Review Iterations

3 review passes performed (triple-model: Claude Opus 4.6 + GPT-5.3 Codex + Gemini 3 Pro).

---

PR #1611, opened 2026-02-17
fix(polecat): handle existing sessions gracefully in gt sling

athosmartins:
## Problem

`gt sling` failed with `ErrSessionRunning` when a tmux session already existed — for example, after a polecat crashed or the host machine restarted. The only recovery path was manual cleanup (`tmux kill-session`) before retrying.

Worse, the failure triggered a rollback that could destroy a worktree that had been successfully created, leaving the workspace in an inconsistent state.

## What Changed

Enhanced `SessionManager.Start()` to detect and handle four cases instead of hard-erroring:

| Session state | Worktree exists? | Action |
|---|---|---|
| Process dead (stale) | either | Kill session, proceed |
| Alive | No (zombie) | Kill session, proceed |
| Alive | Yes, but no valid pane (broken) | Kill session, proceed |
| Alive | Yes, with valid pane ✓ | **Reuse** — return nil |

Added `hasValidWorktree()` helper that checks for both the directory and `.git` file/directory to distinguish zombie sessions from reusable ones.

The rollback path no longer tears down a worktree that was successfully created before the session conflict was detected.

## Testing

- All existing polecat tests pass
- All existing sling tests pass
- Build succeeds with no compilation errors

Fixes: gt-m2hnr

### DIFF
--- a/internal/cmd/polecat_spawn.go
+++ b/internal/cmd/polecat_spawn.go
@@ -2,8 +2,8 @@
 package cmd
 
 import (
+	"errors"
 	"fmt"
-	"os"
 	"path/filepath"
 	"strings"
 	"time"
@@ -30,6 +30,7 @@ type SpawnedPolecatInfo struct {
 	Pane        string // Tmux pane ID (empty until StartSession is called)
 	DoltBranch  string // Dolt branch for write isolation (empty if not created)
 	BaseBranch  string // Effective base branch (e.g., "main", "integration/epic-id")
+	Reused      bool   // True if an existing session was reused instead of freshly spawned
 
 	// Internal fields for deferred session start
 	account string
@@ -275,15 +276,27 @@ func (s *SpawnedPolecatInfo) StartSession() (string, error) {
 			return "", err
 		}
 		startOpts.Command = cmd
+		startOpts.Agent = s.agent
 	}
 	if err := polecatSessMgr.Start(s.PolecatName, startOpts); err != nil {
-		if err == polecat.ErrSessionReused {
-			// Session reused — skip all startup state mutations.
-			// The agent is already running; get pane and return.
+		if errors.Is(err, polecat.ErrSessionReused) {
+			// Session reused — agent is already running. Update state for
+			// monitoring visibility, then return the pane. Callers must check
+			// s.Reused to know a nudge is needed (the session didn't get a
+			// StartupNudge from Start()).
 			pane, paneErr := getSessionPane(s.SessionName)
 			if paneErr != nil {
 				return "", fmt.Errorf("getting pane for reused session %s: %w", s.SessionName, paneErr)
 			}
+			polecatGit := git.NewGit(r.Path)
+			polecatMgr := polecat.NewManager(r, polecatGit, t)
+			if stateErr := polecatMgr.SetAgentStateWithRetry(s.PolecatName, "working"); stateErr != nil {
+				fmt.Printf("Warning: could not update agent state for reused session: %v\n", stateErr)
+			}
+			if stateErr := polecatMgr.SetState(s.PolecatName, polecat.StateWorking); stateErr != nil {
+				fmt.Printf("Warning: could not update issue status for reused session: %v\n", stateErr)
+			}
+			s.Reused = true
 			return pane, nil
 		}
 		return "", fmt.Errorf("starting session: %w", err)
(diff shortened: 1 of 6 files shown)


julianknutsen:

### Decision: request_changes

Thanks for tackling this, @athosmartins! The problem is real and painful — having to manually `tmux kill-session` after a crash is a terrible UX, and the rollback-destroying-worktree issue makes it worse. The four-case session state matrix is a solid approach to the problem.

That said, the review surfaced a few issues with the reuse path that need attention before this can merge safely. The core concern is that returning `nil` for reused sessions is indistinguishable from "freshly created" to callers, which triggers unintended state mutations. Please address the items below and re-submit — not all may require changes depending on your read of the severity.

---

### Fixes Validated
- **`ErrSessionRunning` hard-error after crash:** Confirmed correct. Stale, zombie, and broken sessions are now detected and cleaned up instead of hard-erroring.
- **Rollback destroying successfully-created worktree:** Confirmed correct. Reuse path returns nil, so sling's rollback is no longer triggered for live sessions.

---

### Major: Reuse path skips caller startup orchestration

When `Start()` returns nil via the reuse path, callers like `SpawnedPolecatInfo.StartSession()` continue to run `WaitForRuntimeReady`, `SetAgentStateWithRetry("working")`, and `getSessionPane` — overwriting the state of a session that may be mid-execution on different work. The sling caller then injects a start prompt into the old session's pane.

**Suggested fix:** Return a distinct signal from `Start()` (sentinel error like `ErrSessionReused`, or a boolean return) so callers can distinguish "freshly created" from "reused" and skip state mutation on reuse.

### Major: `--issue` flag silently ignored on reused sessions
**Source:** codex+claude, high confidence

`gt session start --issue X` reports "Session started" on reuse, but issue validation/hooking never executes for the reuse path. The user sees success while `--issue` is silently dropped.

**Suggested fix:** Same fix as above — callers must detect reuse and either reject `--issue` with a clear message or apply the issue to the reused session.

### Major: `hasValidWorktree` ignores `opts.WorkDir` contract

`hasValidWorktree` always checks `m.clonePath(polecat)`. If `Start` is called with a custom `WorkDir`, the helper misclassifies the session as zombie and kills a valid session with a worktree at the custom path.

**Suggested fix:** Pass the resolved work directory (respecting `opts.WorkDir` override) into `hasValidWorktree`.

### Major: Pane-alive does not imply agent-alive — incorrect reuse
**Source:** codex+claude, medium confidence

Reuse gating checks pane existence and pane-PID liveness, not agent-process liveness. A session with a live shell but dead agent is reused as "healthy," leading to a non-functional session that appears working.

**Suggested fix:** In the reuse path, verify agent liveness (e.g., via `GetAgentPaneID` or agent command check). Kill and recreate when no agent process is found.

### Major: No test coverage for four-case state machine
**Source:** all, high confidence

The four-branch dispatch (stale, zombie, broken, reusable) has no tests in `session_manager_test.go`. Only `TestStartPolecatNotFound` exists. Without tests, regressions could silently kill valid sessions or fail to clean up broken ones.

**Suggested fix:** Add table-driven tests covering all four session states, plus caller-facing behavior for reused sessions.

### Minor: Worktree validation logic triplicated

Three implementations of worktree/repo validation now exist: `git.IsRepo()`, `verifyWorktreeExists` in `polecat_spawn.go`, and the new `hasValidWorktree`. Consider consolidating into a shared helper.

### Minor: `ErrSessionRunning` now effectively dead code for polecats

No polecat `Start()` path returns `ErrSessionRunning` anymore. The sentinel and callers checking it are unreachable for polecats (still used by crew/dog). Document the contract change.

---

### Pre-Merge Checklist
- [ ] `Start()` returns a distinct signal for reused sessions so callers can branch
- [ ] `hasValidWorktree` respects `opts.WorkDir` override
- [ ] Reuse path verifies agent liveness, not just pane liveness
- [ ] Tests added for stale, zombie, broken, and reusable session states

---
_🤖 Generated by automated triple-model review pipeline (Claude Opus 4.6 + GPT-5.3 Codex + Gemini 3 Pro)_

julianknutsen (changes requested):
Triple-model review found 5 major issues with the session reuse path. See detailed comment above.

athosmartins:
Thanks for the detailed review @julianknutsen — all 5 issues addressed:

**1. ErrSessionReused sentinel** (`a706335c`)
Added `ErrSessionReused = errors.New("session reused")` so callers can distinguish a graceful reuse from a fresh start.

**2. WorkDir ignored in hasValidWorktree** (`53683898`)
`hasValidWorktree` now accepts `workDir string` directly instead of `polecat string`. `workDir` is resolved from `opts.WorkDir` (with `clonePath` fallback) *before* the `if running {` block, so the correct path is used throughout.

**3. Agent-liveness check** (`53683898`)
Added `!m.tmux.IsAgentAlive(sessionID)` check after `GetPaneID` succeeds. Pane alive but agent dead now kills and recreates the session (Case 4b) instead of falsely marking it as reusable.

**4. --issue flag applied on reuse** (`53683898`)
`validateIssue` + `hookIssue` are now called in the reuse path before returning `ErrSessionReused`, so the issue association is not silently dropped.

**5. Table-driven tests for 4 session states** (`ff81d71c`)
- `TestHasValidWorktree` — 4 subtests covering missing dir, dir without .git, .git file (worktree), .git dir (clone)
- `TestStartSessionStateMatrix` — 5 subtests covering stale, zombie, reusable, custom WorkDir override, and ErrSessionReused sentinel distinctness

Full suite passes (`go test ./...`, 0 failures).

julianknutsen:
## Adoption Review

Thanks for the contribution, @athosmartins! The session reuse logic is a real improvement — replacing the hard `ErrSessionRunning` failure with a 5-state classification (stale/zombie/broken/dead-agent/reusable) eliminates a whole class of `gt sling` failures that required manual cleanup. The approach is well-structured and the state machine is sound.

### Initial Review
Triple-model review (Claude + Codex + Gemini) identified 7 findings across the original commits: sling reuse path missing nudge (major), GT_AGENT not set for non-Claude agents (major), `gt session start` treating reuse as error, `errors.Is()` not used for sentinel comparisons, duplicate worktree validation, dead `ErrSessionRunning` check in `up.go`, and test naming/coverage gaps.

### Maintainer Changes
Fixup commit addressing all 7 initial findings plus 1 finding from iteration review:
- Sling reuse path now correctly sends nudge to reused sessions via `Reused` flag on `SpawnedPolecatInfo`
- GT_AGENT derived from `opts.Agent` (handles command overrides) with `runtimeConfig.Command` fallback
- `gt session start` and `gt session restart` both handle `ErrSessionReused` as success
- All sentinel comparisons switched to `errors.Is()`
- `HasValidWorktree` exported, `verifyWorktreeExists` delegates to it
- Dead `ErrSessionRunning` check removed from `up.go`, deprecation comment added
- Test improvements: renamed for clarity, added `errors.Is` compatibility test

### Final Review Status
All clear — no remaining blockers or majors. Minor accepted findings: tmux-dependent `Start()` paths lack test coverage (requires mock infrastructure), `verifyWorktreeExists` loses some error detail, and pre-existing sessions won't have `GT_AGENT` set (one-time migration, falls back to Claude process names).

### Review Iterations
4 review passes performed (1 PR-mode + 3 local branch iterations).

---

PR #1612, opened 2026-02-17
fix(beads): update agent bead validation to check gt:agent label

athosmartins:
## Problem

A previous change (gt-vja7b) migrated agent bead type from \`"agent"\` to \`"task"\` with a \`gt:agent\` label. Validation in \`molecule_status.go\` was still checking only \`type == "agent"\`, causing silent failures when looking up agent beads in the new format:

\`\`\`
// gt mol status silently skips hook display for any agent bead with type=task + gt:agent label
if err == nil && agentBead != nil && agentBead.Type == "agent" {  // ← misses new format
\`\`\`

This meant \`gt mol status\` would not show the hooked bead for agents using the new label-based format, producing errors like:
\`\`\`
Error: wa-whatsapp_automation-polecat-guzzle is not an agent bead (type=task)
\`\`\`

## What Changed

Added \`IsAgentBead()\` helper in \`beads.go\` that accepts both formats:
- Legacy: \`type == "agent"\` (backward compatible)
- Current: \`type == "task"\` with \`gt:agent\` label

Updated \`molecule_status.go\` to use \`IsAgentBead()\` — this is the actual bug fix.

Updated \`ready.go\`'s \`filterIdentityBeads\` to also use \`IsAgentBead()\` for consistency. Note: \`ready.go\` already handled \`gt:agent\` labeled beads correctly via a separate label-scanning loop, so this is a consolidation refactor rather than a bug fix — both paths produce identical behavior.

Added comprehensive test coverage for all combinations: nil input, legacy type, \`gt:agent\` label, both fields set, and non-agent beads.

Fixes: gt-dsw1e

### DIFF
--- a/internal/beads/beads_agent.go
+++ b/internal/beads/beads_agent.go
@@ -627,8 +627,8 @@ func (b *Beads) GetAgentBead(id string) (*Issue, *AgentFields, error) {
 		return nil, nil, err
 	}
 
-	if !HasLabel(issue, "gt:agent") {
-		return nil, nil, fmt.Errorf("issue %s is not an agent bead (missing gt:agent label)", id)
+	if !IsAgentBead(issue) {
+		return nil, nil, fmt.Errorf("issue %s is not an agent bead (type=%s)", id, issue.Type)
 	}
 
 	fields := ParseAgentFields(issue.Description)


julianknutsen:
## Adoption Review

Thanks for the contribution, @athosmartins! This is a well-scoped fix for a real bug — `gt mol status` silently skipping hook display for agents using the new label-based format was causing confusing errors. The `IsAgentBead()` helper is a clean approach that correctly handles both legacy and current formats, and the test coverage is solid.

### Initial Review
Triple-model review (Claude + Codex + Gemini) found the core fix in `molecule_status.go` correct and the `ready.go` consolidation behavior-preserving. One major finding: `GetAgentBead()` in `beads_agent.go:630` was still checking only `HasLabel(issue, "gt:agent")`, which would reject legacy `type=agent` beads that the new `IsAgentBead()` accepts — a semantic inconsistency in the same package. Two minor findings: `lifecycle.go` duplicates agent check logic inline (acceptable as follow-up), and `install.go`/`convoy.go` query by `--type=agent` only (not currently broken since `CreateAgentBead` sets both).

### Maintainer Changes
Updated `GetAgentBead()` to use `IsAgentBead()` for its validation check, replacing the strict `HasLabel(issue, "gt:agent")` guard. This aligns the retrieval function with the canonical agent bead definition introduced in this PR. 1 file changed (2 insertions, 2 deletions).

### Final Review Status
All clear — the `GetAgentBead` fix is a one-line swap from `HasLabel` to `IsAgentBead` in the same package, matching the exact pattern established by the PR.

### Review Iterations
1 review pass performed.

---

PR #1613, opened 2026-02-17
Fix: Use correct .opencode/plugins directory for OpenCode hooks

DerrickBarra:
## Summary
Fixes OpenCode plugin loading by correcting hooks directory from singular to plural.

## Changes
- `internal/config/agents.go`: Update `HooksDir` from `.opencode/plugin` to `.opencode/plugins`
- Update related comments to reflect correct path

## Problem
OpenCode CLI expects plugins in `.opencode/plugins/` (per [official docs](https://opencode.ai/docs/plugins/)), but Gastown was installing to `.opencode/plugin/`. This caused:
- Hooks to never load
- `session.created` event to never fire
- Persistent agents (mayor) to exit immediately (~10s)

## Verification
- Built and tested locally with OpenCode
- Mayor now stays running (5+ minutes vs. 10 seconds before)
- All tests pass: `go test ./...`

## Regression Note
Bug introduced in commit `0ae16ec3` during agent preset registry refactor.

## Testing
- [x] `go test ./...` passes
- [x] Manual testing with OpenCode runtime
- [x] Verified mayor stays running

 Closes Issue [#1614](https://github.com/steveyegge/gastown/issues/1614)

### DIFF
--- a/internal/config/agents.go
+++ b/internal/config/agents.go
@@ -98,7 +98,7 @@ type AgentPresetInfo struct {
 	// Empty or "none" means no hooks support.
 	HooksProvider string `json:"hooks_provider,omitempty"`
 
-	// HooksDir is the directory for hooks/settings (e.g., ".claude", ".opencode/plugin").
+	// HooksDir is the directory for hooks/settings (e.g., ".claude", ".opencode/plugins").
 	HooksDir string `json:"hooks_dir,omitempty"`
 
 	// HooksSettingsFile is the settings/plugin filename (e.g., "settings.json", "gastown.js").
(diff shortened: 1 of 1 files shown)


julianknutsen:
## Adoption Review

Thanks for the contribution, @DerrickBarra! This is a great catch — the singular/plural mismatch in the OpenCode hooks directory was silently breaking plugin loading and causing mayor to exit immediately. Solid debugging to trace it back to the agent preset registry refactor.

### Review Summary
**Approve.** Triple-model review (Claude + Codex + Gemini) confirmed the fix is correct: all three `agents.go` locations are updated, and the path propagates cleanly through `RuntimeConfigFromPreset` → `EnsurePluginAt()`. The embed path (`internal/opencode/plugin/gastown.js`) is a Go package path and correctly remains unchanged. No blockers or majors found — only two minor items:

- **Stale comment** in `hooks_test.go:229` still references `.opencode/plugin` (minor, all reviewers flagged)
- **Missing regression test** — `TestOpenCodeProviderDefaults` doesn't assert `defaultHooksDir("opencode")` like the Copilot equivalent does (minor, codex+gemini)

We'll address both as a fast-follow.

1 review pass performed.

---

PR #1623, opened 2026-02-17
fix(formula): use dynamic agent bead resolution in patrol formulas

olpie101:
## Summary

- Replace hardcoded `gt-<rig>-witness` and `gt-<rig>-refinery` patterns with dynamic agent bead lookup
- Patrol agents now resolve their agent bead ID once per session via `bd list --type=agent --desc-contains=...`
- Use the returned bead ID (`YOUR_AGENT_BEAD`) for `await-signal` and `gt agent state` commands

## Problem

The hardcoded pattern `gt-<rig>-<role>` doesn't always match actual agent bead IDs, especially in multi-rig or non-standard configurations where agent beads may have different naming conventions.

## Solution

Add a dynamic resolution step at the start of the `loop-or-exit` / `burn-or-loop` steps:

```bash
bd list --type=agent --desc-contains="witness" --json
# or for refinery:
bd list --type=agent --desc-contains="refinery" --json
```

The agent resolves this once per session and uses `YOUR_AGENT_BEAD` as a placeholder in subsequent commands.

## Files Changed

- `internal/formula/formulas/mol-witness-patrol.formula.toml`
- `internal/formula/formulas/mol-refinery-patrol.formula.toml`
- `.beads/formulas/mol-witness-patrol.formula.toml`
- `.beads/formulas/mol-refinery-patrol.formula.toml`


julianknutsen:
## Adoption Review

Thanks for the contribution, @olpie101! This PR correctly identifies a real problem — the hardcoded `gt-<rig>-witness` and `gt-<rig>-refinery` patterns in patrol formulas break agent bead resolution on non-gastown rigs. The dynamic `bd list` approach is the right fix.

### Initial Review
The initial triple-model review (Claude + Codex + Gemini) confirmed the core fix is correct but identified several issues: the original `--desc-contains` substring match was too broad, no fallback existed for empty results, and disambiguation for multi-rig environments relied on prose-only instructions.

### Maintainer Changes
The maintainer fixup commit addresses all review findings across 4 iterations:
- Use `--desc-contains="role_type: refinery/witness"` with exact role_type field match
- Add `--status=open` filter to avoid matching closed/duplicate beads
- Replace jq `contains()` with anchored regex `test("(?m)^\\s*rig: <YOUR_RIG>\\s*$")` for exact-line rig matching (prevents substring collisions like `town` matching `town2`)
- Hard error on zero or multiple matches — agents STOP and report rather than guessing
- Explicit `<YOUR_RIG>` substitution instructions with verification hint
- Unconditional resolution each patrol cycle

### Final Review Status
All clear — no remaining issues found. Both Claude and Codex independently confirmed all 11 review categories are clean.

### Review Iterations
5 review passes performed.

---

olpie101:
@julianknutsen I know this is closed. Just something I have noticed. If the agent bead is for some reason anything other than open this would mean patrol would effectively always stop. So if the agent bead is pinned for example this would cause an immediate stop.

PR #1635, opened 2026-02-17
feat: add session name validation to witness/refinery start commands

athosmartins:
## Summary

Prevent malformed session names from being created by validating format before session creation.

- Add `checkForLegacySession()` helper to detect old-format sessions
- Check for legacy format (prefix-rigname-role) before creating new sessions
- Return helpful error message suggesting `gt doctor --fix-session-names`
- Applies to both witness and refinery start commands

**Old format:** `gt-gastown-witness`
**New format:** `gt-witness`

## Motivation

Without this guard, starting witness/refinery with a stale config could silently create sessions with the old naming convention, leading to zombie-session accumulation that the doctor's `orphan-sessions` check can't detect.

## Test plan
- [ ] `go test ./internal/cmd/... -run TestWitness -v`
- [ ] `go test ./internal/cmd/... -run TestRefinery -v`
- [ ] Manual: attempt to start witness with legacy session name, confirm helpful error

### DIFF
--- a/internal/cmd/refinery.go
+++ b/internal/cmd/refinery.go
@@ -308,6 +308,11 @@ func runRefineryStart(cmd *cobra.Command, args []string) error {
 		return err
 	}
 
+	// Check for legacy session format before creating new session
+	if err := checkForLegacySession(rigName, "refinery"); err != nil {
+		return err
+	}
+
 	fmt.Printf("Starting refinery for %s...\n", rigName)
 
 	if err := mgr.Start(refineryForeground, refineryAgentOverride); err != nil {
--- a/internal/cmd/witness.go
+++ b/internal/cmd/witness.go
@@ -154,10 +154,36 @@ func getWitnessManager(rigName string) (*witness.Manager, error) {
 	return mgr, nil
 }
 
+// checkForLegacySession checks if a legacy-format session exists and returns an error if found.
+// Legacy format: {prefix}-{rigname}-{role} (e.g., gt-gastown-witness)
+// New format: {prefix}-{role} (e.g., gt-witness)
+func checkForLegacySession(rigName, role string) error {
+	prefix := session.PrefixFor(rigName)
+	legacySessionName := fmt.Sprintf("%s-%s-%s", prefix, rigName, role)
+
+	t := tmux.NewTmux()
+	exists, err := t.HasSession(legacySessionName)
+	if err != nil {
+		// Non-fatal: if we can't check, proceed anyway
+		return nil
+	}
+
+	if exists {
+		return fmt.Errorf("found legacy session %q\n\nSession naming format changed. Run this command to fix:\n  gt doctor --fix-session-names", legacySessionName)
+	}
+
+	return nil
+}
+
 func runWitnessStart(cmd *cobra.Command, args []string) error {
 	rigName := args[0]
 
 	if err := checkRigNotParkedOrDocked(rigName); err != nil {
+		return err
+	}
+
+	// Check for legacy session format before creating new session
+	if err := checkForLegacySession(rigName, "witness"); err != nil {
 		return err
 	}
 


julianknutsen:

### Decision: request_changes

Thanks for the contribution, @athosmartins! Preventing zombie sessions from accumulating under stale naming conventions is a real operational pain point, and adding a proactive guard before session creation is the right instinct. The code is clean and the error messaging approach (pointing users to a doctor fix) is a good UX pattern.

However, the review surfaced two blocking issues with the current implementation that need to be resolved before this can merge. Please address the items below — not all may require code changes, some may need design clarification.

---

### Blocker: "Legacy" format is actually the daemon's current canonical session name

The check constructs `gt-gastown-witness` (via `fmt.Sprintf("%s-%s-%s", prefix, rigName, role)`) and treats it as "legacy." However, this is the **current format** that the daemon's `identityToSession` produces when role config exists — it expands `SessionPattern: "gt-{rig}-witness"` to `gt-gastown-witness` (confirmed at `internal/daemon/lifecycle.go:318-319` and `internal/daemon/role_config_integration_test.go:35`). Meanwhile, `session.WitnessSessionName` returns `gt-witness`.

Both formats are actively produced by different subsystems. Blocking on `gt-gastown-witness` would cause `gt witness start` / `gt refinery start` to fail whenever the daemon has already created a session — a normal operational state.

**Suggested fix:** Clarify which naming convention (`gt-witness` vs `gt-gastown-witness`) is authoritative. The codebase has a real inconsistency here that should be resolved before adding validation that assumes one format is legacy.

---

### Blocker: Error message references non-existent `gt doctor --fix-session-names`

The error message directs users to run `gt doctor --fix-session-names`, but this flag does not exist on the doctor command (`internal/cmd/doctor.go:107-113` — flags are `--fix`, `--verbose`, `--rig`, `--restart-sessions`, `--slow`). A codebase-wide search for `fix-session-names` returned zero results. Users hitting this error would have no working remediation path.

**Suggested fix:** Either implement `gt doctor --fix-session-names` first, add session name migration to the existing `--fix` flow, or change the error message to provide instructions that work today (e.g., `tmux kill-session -t <name>`).

---

### Major: Guard only covers CLI `start` handlers; 10+ other call sites bypass it

The validation only runs in `runWitnessStart` and `runRefineryStart`, but many other paths create sessions via `Manager.Start()` — including `witness attach`, `gt start`, `gt up`, `gt rig` commands, and daemon restarts. Two sessions under different name formats can coexist, which is the exact scenario this PR aims to prevent.

**Suggested fix:** Move the validation into `witness.Manager.Start()` and `refinery.Manager.Start()` so all callers benefit.

---

### Major: No test coverage for new validation logic

The new `checkForLegacySession` function and its integration have zero test coverage. Once the design issues above are resolved, tests should cover legacy-present, legacy-absent, and tmux-error branches.

---

### Minor: Silent suppression of tmux errors disables validation

If `HasSession` returns an error, the check silently passes with no log output. Consider logging a warning so operators know the safety check is disabled.

---

### Nit: Helper defined in witness.go but shared across roles

`checkForLegacySession` is role-agnostic but lives in `witness.go`. Consider moving to a shared location like `internal/cmd/session_helpers.go`.

---

### Pre-Merge Checklist
- [ ] Resolve naming authority: determine whether `gt-witness` or `gt-gastown-witness` is the correct format
- [ ] Implement `gt doctor --fix-session-names` or replace error message with working remediation
- [ ] Move legacy check into `Manager.Start()` to cover all call sites
- [ ] Add test coverage for the validation logic
- [ ] Surface tmux errors as warnings instead of silent suppression

---

athosmartins:
Thank you for the thorough review. Both blockers are correct:

**Blocker 1 — Naming convention conflict**

The `identityToSession` function in `internal/daemon/lifecycle.go:318-319` confirms the concern: when a rig's role config has `SessionPattern: "gt-{rig}-witness"`, the daemon actively creates `gt-gastown-witness` sessions using `beads.ExpandRolePattern`. It only falls back to `session.WitnessSessionName()` (which returns `gt-witness`) when no `SessionPattern` is set.

So both formats ARE simultaneously produced by different subsystems — our guard would block legitimate daemon-managed sessions. The root fix needs to happen at the daemon level first: either remove legacy `SessionPattern` values from rigs.json, or change `identityToSession` to always use `session.WitnessSessionName` for witness/refinery roles.

**Blocker 2 — Non-existent flag**

Confirmed: `gt doctor --fix-session-names` doesn't exist. Should be `gt doctor --fix`.

**Proposed resolution**

Given that:
1. The naming convention conflict can't be safely resolved without changes to daemon session creation
2. PR #1637 (`gt doctor` check) already provides a mechanism to detect and rename outdated sessions via `gt doctor --fix`
3. Adding a pre-flight guard before the convention is settled risks breaking daemon-managed session workflows

I'm proposing to close this PR. The `gt doctor --fix` path in #1637 provides the same protection as a post-hoc cleanup rather than a pre-flight check. Once the daemon's session naming is unified (i.e., `SessionPattern` values updated or daemon updated to always call `session.WitnessSessionName`), a cleaner pre-flight check can be added inside `Manager.Start()` as you suggest.

Closing this PR and tracking the remainder as future work.

PR #1637, opened 2026-02-17
feat(doctor): add session-name-format check to detect outdated session names

athosmartins:
## Summary

Adds a new fixable `gt doctor` check that detects Gas Town tmux sessions whose names don't match the canonical format computed from their parsed identity.

This catches sessions that survived upgrades where the naming scheme changed (e.g., old `gt-whatsapp_automation-witness` vs new `wa-witness`).

## Changes

- **Run:** lists sessions with mismatched names vs `ParseSessionName().SessionName()`
- **Fix:** renames via `tmux rename-session` (crew sessions protected from auto-rename)
- **Registered** in `Cleanup` category alongside `orphan-sessions` and `zombie-sessions`
- **3 unit tests** covering empty, correct, and non-Gas Town session cases

## Motivation

During the `wa-nwgrr` zombie investigation, we found that session name format changed in Feb 2026. Old-format sessions survived upgrades and appeared as zombies. This check prevents the issue by detecting and fixing name mismatches before they cause problems.

## Test plan
- [ ] `go test ./internal/doctor/... -run TestSessionNameFormatCheck -v`
- [ ] `gt doctor` — verify new check appears in Cleanup category
- [ ] `gt doctor --fix` — verify old-format sessions get renamed


julianknutsen:

### Decision: request_changes

Thanks for the contribution, @athosmartins! This is a well-motivated addition — the zombie session problem from the Feb 2026 naming scheme change is real, and a `gt doctor` check is the right place to catch it. The code is clean, follows established doctor check patterns, and includes thoughtful safeguards (crew session protection, collision detection).

However, both reviewers independently identified a fundamental issue with the detection mechanism that needs to be resolved before this can merge.

---

### Blocker: Round-trip detection cannot catch legacy name migrations

The check relies on `ParseSessionName(s).SessionName() != s` to detect outdated names. For the stated scenario (`gt-whatsapp_automation-witness` → `wa-witness`):
- If `gt` is registered, the parser treats `whatsapp_automation-witness` as a polecat name under prefix `gt`, and `SessionName()` round-trips to the same string — no mismatch detected.
- If `gt` is not registered, `ParseSessionName` fails and the session is silently skipped as "not a Gas Town session."

In neither case is a mismatch reported. The `ParseSessionName` function matches by *prefix*, not by embedded rig name, so it fundamentally cannot map `gt-whatsapp_automation-witness` to `wa-witness`.

**Suggested fix:** Either implement explicit legacy-name detection (e.g., match rig names within session strings against the current prefix registry) or redefine the scope of what this check detects and update the description accordingly.

---

### Major: Tests pass vacuously due to missing registry initialization

`TestMalformedSessionNameCheck_Run_AllCorrect` uses sessions like `wa-crew-batista` which fail `ParseSessionName` (no `wa` prefix in the default registry) and are silently skipped. The test asserts OK and passes — but only because *all* sessions are skipped as unparseable, not because they were validated as correct.

**Suggested fix:** Call `setupTestRegistry(t)` (defined in `orphan_check_test.go` in the same package) at the start of each test. Use session names that actually parse under the test registry (e.g., `gt-witness`, `gt-refinery`, `gt-crew-batista`).

---

### Major: No test coverage for mismatch detection or Fix() path

The three tests cover empty, "correct" (vacuously), and non-Gas Town sessions. There is zero coverage for the core feature: detecting a mismatched name. The `Fix()` path (rename, collision skip, crew skip, error handling) is also untested.

**Suggested fix:** Add test(s) that produce a genuine `ParseSessionName(s).SessionName() != s` mismatch and validate detection. Add Fix() path tests for successful rename, collision skip, and crew skip.

---

### Minor: Crew sessions flagged then silently skipped in Fix()

`Run()` reports crew sessions as malformed with `FixHint: "Run 'gt doctor --fix'"`, but `Fix()` silently skips them. The user is told to run `--fix` for something that `--fix` won't fix.

**Suggested fix:** Either exclude crew sessions from findings in `Run()`, or emit a distinct message like "manual rename required for crew sessions."

---

### Minor: TOCTOU between Run() and Fix()

Between `Run()` and `Fix()`, earlier doctor checks (zombie-sessions) may kill sessions cached in `c.malformed`. The collision check handles target-exists but not source-gone.

**Suggested fix:** Add `HasSession(r.oldName)` guard before `RenameSession`.

---

### Pre-Merge Checklist
- [ ] Fix fundamental detection mechanism (explicit legacy detection or redefine scope)
- [ ] Add `setupTestRegistry(t)` calls in tests
- [ ] Add test for genuine name mismatch detection
- [ ] Add Fix() path tests (rename, collision skip, crew skip)
- [ ] Clarify crew-session UX in findings vs fix behavior

---

julianknutsen (changes requested):
See automated review comment above for details. Blocking on: round-trip detection mechanism cannot catch the stated legacy name migration use case.

athosmartins:
Thanks for the thorough review. All issues addressed:

**Blocker: Round-trip detection → explicit legacy matching**

Replaced the `ParseSessionName(s).SessionName() != s` round-trip with `detectLegacySessionNames()` that scans for `{any_prefix}-{registered_rig_name}-{role_suffix}` patterns by looking up all known rigs from `PrefixRegistry.AllRigs()`. Added `AllRigs() map[string]string` to the registry for this purpose.

The core test `TestMalformedSessionNameCheck_Run_LegacyWAWitness` directly validates the stated use case: `gt-whatsapp_automation-witness` → detected, canonical `wa-witness` reported.

**Major: Vacuous test pass**

Added `registryForTest *session.PrefixRegistry` injection to the check struct. All tests now pass an explicit registry via `testRegistryForNameCheck()` (registers `gt/gastown`, `nif/niflheim`, `wa/whatsapp_automation`). Sessions no longer fail to parse and get silently skipped.

**Major: Missing mismatch and Fix() path tests**

Added 7 new tests covering the core detection and Fix() paths:
- `TestMalformedSessionNameCheck_Run_DetectsMismatch` — genuine `gt-niflheim-{witness,refinery}` detection
- `TestMalformedSessionNameCheck_Run_LegacyWAWitness` — the exact stated use case  
- `TestMalformedSessionNameCheck_Run_CrewSession` — crew sessions detected + "manual rename" note
- `TestMalformedSessionNameCheck_Fix_Rename` — happy path rename
- `TestMalformedSessionNameCheck_Fix_SkipsCrew` — crew not renamed
- `TestMalformedSessionNameCheck_Fix_SkipsCollision` — target-exists guard
- `TestMalformedSessionNameCheck_Fix_TOCTOUGuard` — source-gone guard

Fix() tests use a `tmuxRenamer` interface + `mockTmux` (injected via `tmuxForTest`) so they don't need a real tmux session.

**Minor: Crew sessions silently skipped in Fix()**

`Run()` now labels crew session details with `"(crew session — manual rename required)"` and sets `FixHint` accordingly. `Fix()` still skips crew silently (the user was already told in `Run()`'s output). No more misleading "run --fix" hint when only crew sessions are outdated.

**Minor: TOCTOU guard**

Added `HasSession(r.oldName)` check before `RenameSession` in `Fix()`, guarding against the case where a prior doctor check killed the source session between `Run()` and `Fix()`.

---

Branch rebased onto current upstream `main` (clean cherry-pick of the single doctor commit).

julianknutsen:
## Adoption Review

Thanks for the contribution, @athosmartins! This is a well-designed doctor check that addresses a real problem discovered during the wa-nwgrr zombie investigation — legacy session names surviving upgrades and causing confusion. The explicit legacy-name matching approach, crew session protection, and TOCTOU guards show careful thought about the edge cases.

### Initial Review
Dual-model review (Claude + Codex) found 2 major issues and 3 minor ones:
1. **Check registration ordering** — session-name-format ran after orphan/zombie checks, so `--fix` could kill sessions before renaming them.
2. **No ownership filter** on `matchLegacyName` — any tmux session containing `-{rig_name}-witness|refinery` would be matched, including non-Gastown sessions.
3. Minor: `HasSession` errors silently ignored in `Fix()`.

### Maintainer Changes
Fixup commit addresses the 3 actionable findings (+114 -9 across 3 files):
1. Moved `NewMalformedSessionNameCheck()` registration before orphan and zombie checks so names are normalized first.
2. Added known-prefix ownership guard — `matchLegacyName` now requires the prefix before the rig name to be a registered Gastown prefix, filtering out non-Gastown sessions and polecat names with embedded rig substrings.
3. `HasSession` errors now propagate instead of silent `continue`.
4. Added 3 new test cases: non-Gastown rig substring, polecat rig substring, and HasSession error injection.

### Final Review Status
Iteration review confirmed all 3 fixes are correct. One remaining edge case identified: a polecat named exactly `{rig_name}-{role}` on a *different* rig (e.g., `wa-gastown-witness`) could still be misclassified. Accepted as extremely low probability — requires deliberately naming a polecat after another rig's role.

| 1 | Behavioral Correctness | major | high | codex+claude | Cross-rig prefix collision on polecat names (accepted as low probability) |
| 2 | Change Impact | minor | high | claude | Zombie check won't monitor renamed sessions (pre-existing) |

### Review Iterations
2 review passes performed.

---

PR #1648, opened 2026-02-17
fix(done): defer no-commits-ahead close to after BD_BRANCH unset

KeithWyatt:
## Summary

When a polecat completes with zero commits ahead (`aheadCount == 0`), the G15 fix calls `ForceCloseWithReason` to close the bead so it doesn't get stuck as HOOKED. However, at that point `BD_BRANCH` is still set in the environment, so the close writes `status=closed` to the **polecat's Dolt branch** instead of main.

The subsequent `MergePolecatBranch` at the `notifyWitness` label can then hit merge conflicts — main has `status=hooked` while the polecat branch has `status=closed` from different base states. When this happens, the bead is left permanently stuck as HOOKED with no recovery path.

**Fix:** Defer the close to after the `afterDoltMerge` label, where `BD_BRANCH` has already been unset. The polecat branch merges cleanly (no meaningful writes on it), then the close writes directly to Dolt main.

## Reproduction

1. Sling a bead to a polecat (creates Dolt branch with `BD_BRANCH` set)
2. The bug is already fixed or work was pushed directly to main
3. Polecat runs `gt done --cleanup-status clean`
4. `aheadCount == 0` path triggers, `ForceCloseWithReason` writes close to polecat branch
5. `MergePolecatBranch` hits conflict → bead stuck as HOOKED

## Test plan

- [x] `go test ./internal/cmd/ -run Done` — all tests pass
- [ ] Manual: sling a bead, push fix directly to main, run `gt done --cleanup-status clean` — bead should close successfully

### DIFF
--- a/internal/cmd/done.go
+++ b/internal/cmd/done.go
@@ -814,6 +814,11 @@ notifyWitness:
 	}
 
 afterDoltMerge:
+	// Ensure BD_BRANCH is unset before any post-merge bead operations.
+	// The normal path unsets it above after MergePolecatBranch, but the
+	// checkpoint-resume path (goto afterDoltMerge) bypasses that code.
+	os.Unsetenv("BD_BRANCH")
+
 	// Deferred close for no-commits-ahead path (G15 fix).
 	// The close is deferred to here because BD_BRANCH has now been unset,
 	// so ForceCloseWithReason writes directly to Dolt main. Closing earlier


julianknutsen:
## Adoption Review

Thanks for the contribution, @KeithWyatt! This is a great catch — the BD_BRANCH ordering issue in the no-commits-ahead path is a subtle but real bug that leaves beads permanently stuck as HOOKED with no recovery. The deferred-close approach is the right fix, and the PR description and reproduction steps are excellent.

This PR was adopted with a maintainer fix pushed directly to the contributor's branch.

### Initial Review
The initial dual-model review (Claude + Codex) confirmed the core fix is correct for the normal path: deferring `ForceCloseWithReason` to after `afterDoltMerge` ensures BD_BRANCH is unset before the close writes to Dolt. However, both reviewers identified that the **checkpoint-resume path** (`CheckpointDoltMerged` → `goto afterDoltMerge`) bypasses `os.Unsetenv("BD_BRANCH")`, recreating the same stuck-HOOKED scenario on resume.

### Maintainer Changes
Added unconditional `os.Unsetenv("BD_BRANCH")` at the `afterDoltMerge` label, before the deferred close block. This covers both the normal path (idempotent double-unset, harmless) and the resume path (first unset). +5 lines in `internal/cmd/done.go`.

### Final Review Status
All clear — both fixes validated as correct by the iteration review. No blockers or majors. Two minor pre-existing findings noted for follow-up: (1) same BD_BRANCH pattern in direct/owned+direct convoy close paths (mitigated by `--theirs` resolution), (2) no behavioral ordering test for the deferred close sequence.

### Review Iterations
2 review passes performed.

---

julianknutsen:
Hey @KeithWyatt — thanks for putting this together. The failure scenario you described (bead stuck as HOOKED after merge conflict) is a reasonable concern, and the fix was internally consistent, which is why it passed our multi-model review initially.

After closer analysis (cc @arttttt who investigated this in #1656), we believe `MergePolecatBranch` already handles this case. When `ForceCloseWithReason` writes `status=closed` to the polecat branch while main has `status=hooked`, Phase 2 of `MergePolecatBranch` kicks in:

```go
CALL DOLT_MERGE('%s');
CALL DOLT_CONFLICTS_RESOLVE('--theirs', '.');
```

`--theirs` means the polecat branch wins — so after merge, `status=closed` lands on main. No stuck HOOKED state.

The fix in this PR bypassed that resolution path by deferring the close to after `BD_BRANCH` is unset (writing directly to main), which works but breaks the branch isolation model that the rest of the polecat workflow relies on.

That said — we want to make sure we're not missing something. **Were you seeing actual stuck HOOKED beads in practice?** If so, we'd love to understand the reproduction conditions — there might be a different root cause (e.g., Phase 2 not triggering, or the merge path being skipped entirely) that we should investigate separately.

This was reverted in #1656. If there is a real failure mode we missed, we absolutely want to fix it — just through a path that preserves branch isolation.

PR #1654, opened 2026-02-17
Fix: Defensive handling for bd update --type with custom types

DerrickBarra:
### Summary

Makes `CreateOrReopenAgentBead` more resilient to a beads CLI bug where `bd update --type=<custom>` fails validation even though `bd create --type=<custom>` works.

### Problem

When reopening/updating an existing agent bead, Gastown tries to ensure the type is `agent`. But `bd update --type=agent` fails with:
```
Error: invalid issue type "agent". Valid types: bug, feature, task, epic, chore
```

This is a **beads bug** — custom types work for `create` but not `update`.

### Solution

1. **Check before updating**: Only run `bd update --type=agent` if the current type is NOT already `agent`
2. **Warn on failure**: If the update fails, print a warning and continue instead of failing entirely

The bead still has the `gt:agent` label, which is what Gastown actually uses for identification.

### Changes

- `internal/beads/beads_agent.go`: Modified `CreateOrReopenAgentBead()` to check existing type and warn instead of fail

### Testing

- Verified polecat spawning now proceeds past agent bead creation
- Convoy and formula instantiation work correctly
- End-to-end workflow progresses (blocked later at Dolt infrastructure, unrelated)

### Notes

This is a **defensive workaround** for a beads CLI bug. The proper fix is in beads to make `bd update --type` consistent with `bd create --type` for custom types.

### DIFF
--- a/internal/beads/beads.go
+++ b/internal/beads/beads.go
@@ -246,7 +246,7 @@ func (b *Beads) run(args ...string) ([]byte, error) {
 
 	err := cmd.Run()
 	if err != nil {
-		return nil, b.wrapError(err, stderr.String(), args)
+		return nil, b.wrapError(err, stderr.String(), args, stdout.String())
 	}
 
 	// Handle bd exit code 0 bug: when issue not found,
@@ -290,7 +290,7 @@ func (b *Beads) runWithRouting(args ...string) ([]byte, error) { //nolint:unpara
 
 	err := cmd.Run()
 	if err != nil {
-		return nil, b.wrapError(err, stderr.String(), args)
+		return nil, b.wrapError(err, stderr.String(), args, stdout.String())
 	}
 
 	if stdout.Len() == 0 && stderr.Len() > 0 {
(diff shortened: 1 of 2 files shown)


julianknutsen:
## Adoption Review

Thanks for the contribution, @DerrickBarra! This is a solid defensive fix for a real pain point — the `bd update --type` validation bug was causing agent bead reopens to fail entirely, and your approach of guarding with a type check + warn-on-failure keeps the system moving while the upstream bd CLI bug gets resolved.

### Review Summary
Decision: **approve**. Two reviewers (Claude + Codex) independently confirmed both fixes are correct: the `existing.Type != "agent"` guard avoids the problematic `bd update --type` call when unnecessary, and the `wrapError` stdout check correctly handles Dolt backend behavior. One major finding was identified (broad error suppression) but assessed as low practical risk given that the `gt:agent` label is the primary agent identification mechanism.

### Review Findings

| 1 | Behavioral Correctness | major | medium | claude+codex | Warn-and-continue catches all `bd update --type` failures, not just the known validation bug — unexpected errors (network, disk) would be silently swallowed |
| 2 | Test Evidence Quality | minor | high | claude+codex | No active test coverage for the new type-conditional branch or stdout-based not-found detection (pre-existing: tests already skipped) |
| 3 | Behavioral Correctness | minor | medium | claude | Slot operations silently degrade on the new warn path (absorbed by existing non-fatal handling) |
| 4 | Error Handling | nit | high | claude | Warning message text "already has gt:agent label" is slightly misleading |

1 review pass performed.

---

PR #648, opened 2026-01-17
feat(ratelimit): implement intelligent instance swapping for rate limit handling

karanb192:
## Summary
Implemented intelligent rate-limit aware instance swapping system.

## Related Issue
Fixes #232

## New Package: `internal/ratelimit/`

| File | Purpose |
|------|---------|
| `types.go` | Core types: InstanceProfile, RolePolicy, SwapResult, etc. |
| `detect.go` | Rate limit detection via regex patterns |
| `manager.go` | Profile selection, cooldown tracking, swap coordination |
| `detect_test.go` | 19 test cases for detection |
| `manager_test.go` | Tests for cooldown, fallback, stickiness |

## Features
- **Rate limit detection**: Pattern matching for 429, throttling, "overloaded", provider-specific errors
- **Instance profiles**: Named profiles with provider, auth, model settings
- **Fallback chains**: Per-role policies with ordered fallback profiles
- **Cooldown tracking**: Prevents rapid profile cycling with configurable cooldown periods
- **Daemon integration**: Monitors sessions during heartbeat for rate limit indicators

## Other Files Modified
- `internal/daemon/daemon.go` - Added rate limit check step in heartbeat
- `internal/daemon/ratelimit.go` - Daemon integration for monitoring sessions
- `internal/events/events.go` - Added event types for rate limit, swap, cooldown

## Testing
- [x] Unit tests added (detect_test.go, manager_test.go)
- [x] Unit tests pass (`go test ./...`) - verified locally
- [x] Build passes (`go build -o gt ./cmd/gt`)
- [x] go vet clean

## Checklist
- [x] Code follows project style
- [x] No breaking changes

Issue #232, opened 2026-01-06
ABSOLUTELY URGENT FEATURE: Intelligent instance swapping (rate-limit aware) + provider/auth profiles for Gastown


josephbasilework:
@steveyegge 
@claude 
@martinemde 
@daveio 
@kiwiupover 
## Problem Statement

Gastown is designed to run continuously with ephemeral sessions and persistent state (Beads/Hooks/Molecules). In practice, long-running harnesses (e.g., Cloud Code CLI / Claude Code-like sessions) frequently encounter provider throttling (HTTP 429 / “rate limit”) or get stuck retrying, which stalls progress.

Today, when a specific worker instance (and potentially other long-lived roles) gets rate-limited, there is no first-class mechanism to:

* Detect the condition in a structured way
* Swap that instance to a different provider/account/model configuration automatically
* Preserve continuity by resuming from persistent state (instead of requiring manual restart)
* Apply role-specific preferences (e.g., “Deacon should prefer staying on Anthropic / Opus 4.5 if possible”)
* Handle the “watcher problem” (Witness itself getting stuck or rate-limited)

The result is rigs that go idle during throttling, requiring manual intervention, which defeats “always-on” operation.

---

## Proposed Solution

Introduce **intelligent instance swapping** plus **provider/auth profiles**, with **role-specific policies** and **conditional transition hooks**, implemented as a lifecycle feature (primarily within Witness for workers) and supervised by a higher-level daemon/doctor layer for rig roles.

### 1) Provider/Auth “Profiles” (a first-class config object)

Add a new concept: **Instance Profile** (or “Harness Profile”) that defines how to launch a session.

Each profile should include:

* `provider`: anthropic | z-ai | minimax | …
* `auth_ref`: e.g., anthropicAcctA / anthropicAcctB / zaiKey1 / minimaxKey1
* `model_main` / `model_fast` (or similar mapping)
* `harness_command` / environment variables needed to run the harness (Cloud Code CLI wrapper, etc.)
* optional `skill_pack` (preloaded workflows/skills enabled only for that profile)

### 2) Role Policies (fallback chains + stickiness + cooldown)

Per-role configuration that determines what to do when a rate limit (or stuck/crash) occurs:

* Ordered `fallback_chain` of profiles
* “Stickiness” constraints (e.g., Deacon prefers Anthropic; only fail over to other providers if all Anthropic profiles are in cooldown/unavailable)
[...]

PR #1668, opened 2026-02-18
feat(quota): account quota rotation for rate-limit management

seanbearden:
## Problem

Gas Town operators managing multiple concurrent Claude Code sessions face a painful operational bottleneck when accounts hit rate limits. The current experience:

1. A session hits its rate limit and displays: `You've hit your limit · resets 7pm (America/Los_Angeles)`
2. The session becomes **completely stuck** — every interaction (including incoming mail notifications) triggers the same rate-limit prompt
3. The operator must **manually toggle to each blocked terminal**, one by one, to identify which sessions are affected
4. For each blocked session, the operator must run `/login` to re-authenticate, then `cont` to resume work
5. With 5+ concurrent sessions across multiple rigs, this becomes a multi-minute interruption that kills flow state

Real example from a mayor session hitting the limit:
```
❯ /rate-limit-options
  ⎿  You've hit your limit · resets 7pm (America/Los_Angeles)

❯ 📬 You have new mail from laser/witness. Subject: WITNESS_PING laser.
  ⎿  You've hit your limit · resets 7pm (America/Los_Angeles)

❯ 📬 You have new mail from laser/witness. Subject: WITNESS_PING laser.
  ⎿  You've hit your limit · resets 7pm (America/Los_Angeles)
```

The mayor is completely blocked — it can't even process incoming mail. Every ping just re-triggers the rate-limit prompt. This cascades: witnesses can't reach the mayor, crew can't get assignments, the whole town stalls.

Users who manage 3-5 Claude accounts can work around this by rotating to a fresh account, but doing so manually across many sessions is tedious and error-prone. There's currently no tooling to detect which sessions are blocked or to swap accounts programmatically.

## Relationship to #232 and #648

Issue #232 describes a comprehensive vision for intelligent instance swapping with provider profiles, fallback chains, role-specific policies, transition hooks, and "watch the watcher" supervision. PR #648 attempted to solve this via a daemon heartbeat that scanned every session's tmux pane output with regexes. It was [closed](https://github.com/steveyegge/gastown/pull/648#issuecomment-3894955909) by @julianknutsen for three reasons:

1. **Architecture mismatch**: A single daemon scanning all panes doesn't scale to the multi-rig topology
2. **Detection without application**: Rate limits were detected and logged but never actually applied — sessions were never swapped
3. **Merge conflicts**: Daemon heartbeat integration point had drifted

**This PR takes a deliberately different approach.** Rather than coupling rate-limit handling to the daemon, this is a **standalone CLI tool** (`gt quota`) that operators invoke on demand. It addresses the #648 feedback directly:

- **No daemon coupling**: Rotation is an explicit operator action, not a heartbeat side-effect. This avoids the single-bottleneck concern entirely.
- **End-to-end execution**: Unlike #648, this actually performs the swap — sets `CLAUDE_CONFIG_DIR`, kills the process, and respawns the pane with a fresh account. The full lifecycle works.
- **No daemon modification**: Zero changes to daemon code, no integration point conflicts.

This solves the immediate operational pain (manual account rotation across sessions) without prescribing how daemon-level automation should eventually work. The structured signal approach discussed in #648 review feedback (hook events, beads) remains the right path for automated detection — this PR simply provides the rotation machinery that any future detection system can call.

## Solution

New `gt quota` command tree that builds on the existing `gt account` system:

| Command | Description |
|---------|-------------|
| `gt quota status` | Show quota state for all registered accounts |
| `gt quota scan` | Detect rate-limited sessions by examining tmux pane content |
| `gt quota scan --update` | Scan and persist detected limits to `mayor/quota.json` |
| `gt quota rotate` | Swap blocked sessions to available accounts (LRU ordering) |
| `gt quota rotate --dry-run` | Preview rotation plan without executing |

### How it works

**Scan**: Captures the last 30 lines from each Gas Town tmux session and matches against configurable rate-limit patterns. Resolves which account each session is using by reading `CLAUDE_CONFIG_DIR` from the tmux session environment and mapping it back to registered account handles.

**Rotate**: Runs a scan, plans account assignments using least-recently-used ordering (so accounts get even usage), then executes all rotations atomically under a single file lock. Uses the battle-tested `KillPaneProcesses` → `RespawnPane` restart sequence from the handoff system. Updates `CLAUDE_CONFIG_DIR` in the tmux session environment before respawn so the new shell inherits the correct account. Multiple sessions are rotated concurrently (each targets an independent tmux pane).

**State**: Persisted to `mayor/quota.json` with crash-safe atomic writes (`util.AtomicWriteJSON`) and file-level locking (`gofrs/flock`). Tracks per-account status (available/limited/cooldown), limit detection timestamps, estimated reset times, and last-used timestamps for LRU ordering.

### Architecture

The rotation logic lives in `internal/quota/` behind injected interfaces (`TmuxClient` for read-only scanning, `TmuxExecutor` for pane lifecycle mutations, `Logger` for warning output). The `Rotator` struct takes all dependencies via constructor injection, making it testable without tmux or disk I/O. The `Execute` method holds the quota file lock for the entire rotation lifecycle — load, rotate all sessions concurrently, single save — eliminating TOCTOU races.

This separation means any future caller (daemon, witness, CLI) can use `quota.NewRotator(...)` and `rotator.Execute(plan, sessions)` without reimplementing the rotation logic.

### Design principles

- **Zero-config default**: Quota features are entirely opt-in. Users who don't use multiple accounts are unaffected — no new config files are created, no behavior changes.
- **Additive-only**: New subcommand tree (`gt quota`) under the existing `GroupServices` group. No changes to existing commands or behavior.
- **Builds on existing infrastructure**: Uses `AccountsConfig`, `CLAUDE_CONFIG_DIR` injection, `ResolveAccountConfigDir()`, and the handoff restart primitives that already exist.
- **Provider-agnostic patterns**: Rate-limit detection uses configurable regex patterns, not hardcoded to any specific provider's message format.
- **CLI-first, automation-ready**: Designed as an operator tool today, with domain-layer interfaces that future automation can consume.

## Changes

**New files:**
- `internal/quota/state.go` — `Manager` with flock-based locking (`WithLock`/`SaveUnlocked`), atomic writes, state machine operations
- `internal/quota/scan.go` — `Scanner` with `TmuxClient` interface, pattern matching, account resolution
- `internal/quota/rotate.go` — `PlanRotation()` with LRU assignment and same-account skip logic
- `internal/quota/executor.go` — `Rotator` with `TmuxExecutor` interface, concurrent `Execute`, validation-before-mutation ordering
- `internal/util/path.go` — `ExpandHome` utility with `sync.Once` cache
- `internal/cmd/quota.go` — Cobra command tree (status, scan, rotate) with text and JSON output
- 5 test files — 42 tests covering state management, scanning, rotation planning, execution, and path expansion

**Modified files:**
- `internal/config/types.go` — Added `QuotaState`, `AccountQuotaState`, `AccountQuotaStatus` enum
- `internal/constants/constants.go` — Added `FileQuotaJSON`, `MayorQuotaPath()`, `DefaultRateLimitPatterns`

## Test plan

- [x] All 42 tests pass (`go test -race ./internal/quota/... ./internal/util/...`)
- [x] Race detector clean
- [x] Full binary builds cleanly (`go build ./cmd/gt`)
- [x] Zero vet issues (`go vet`)
- [ ] Manual test: `gt quota status` with registered accounts
- [ ] Manual test: `gt quota scan` detects rate-limited session
- [ ] Manual test: `gt quota rotate --dry-run` shows correct plan
- [ ] Manual test: `gt quota rotate` successfully restarts session with new account

### DIFF
--- a/internal/cmd/quota.go
+++ b/internal/cmd/quota.go
@@ -495,6 +495,14 @@ func runQuotaRotate(cmd *cobra.Command, args []string) error {
 			fmt.Printf(" %s %d session(s) cannot be rotated (not enough available accounts)\n",
 				style.WarningPrefix, unassignable)
 		}
+		if len(plan.SkippedAccounts) > 0 {
+			fmt.Println()
+			for handle, reason := range plan.SkippedAccounts {
+				acct := acctCfg.Accounts[handle]
+				fmt.Printf(" %s Skipped %s — %s\n", style.WarningPrefix, handle, reason)
+				fmt.Printf("   Run: claude /login  (in CLAUDE_CONFIG_DIR=%s)\n", acct.ConfigDir)
+			}
+		}
 	}
 
 	if rotateDryRun {
--- a/internal/quota/keychain.go
+++ b/internal/quota/keychain.go
@@ -4,12 +4,15 @@ package quota
 
 import (
 	"crypto/sha256"
+	"encoding/base64"
 	"encoding/json"
 	"fmt"
+	"net/http"
 	"os"
 	"os/exec"
 	"path/filepath"
 	"strings"
+	"time"
 )
 
 const (
@@ -195,6 +198,83 @@ func RestoreOAuthAccount(targetConfigDir string, backup json.RawMessage) error {
 		return fmt.Errorf("marshaling target .claude.json: %w", err)
 	}
 	return os.WriteFile(targetPath, out, 0600)
+}
+
+// ValidateKeychainToken checks if the OAuth token for a config dir is still usable.
+// It attempts local validation first (JSON credential expiry, JWT expiry), then
+// falls back to a lightweight API call. Returns nil if the token appears valid
+// or if the token can't be read (the actual swap will fail clearly in that case).
+func ValidateKeychainToken(configDir string) error {
+	svc := KeychainServiceName(configDir)
+	raw, err := ReadKeychainToken(svc)
+	if err != nil {
+		// Can't read the token — don't block planning. The swap itself will
+		// fail with a clear error if the keychain entry doesn't exist.
+		return nil
+	}
+	if raw == "" {
+		return nil
+	}
+
+	// Strategy 1: Parse as JSON credential with expires_at field.
+	// Claude Code may store the full OAuth response including expiry.
+	var cred struct {
+		ExpiresAt int64 `json:"expires_at"`
+	}
+	if json.Unmarshal([]byte(raw), &cred) == nil && cred.ExpiresAt > 0 {
+		if time.Now().Unix() >= cred.ExpiresAt {
+			return fmt.Errorf("token expired at %s", time.Unix(cred.ExpiresAt, 0).Format(time.RFC3339))
+		}
+		return nil
+	}
+
+	// Strategy 2: Parse as JWT — decode payload, check exp claim.
+	parts := strings.Split(raw, ".")
+	if len(parts) == 3 {
+		payload, decErr := base64.RawURLEncoding.DecodeString(parts[1])
+		if decErr == nil {
+			var claims struct {
+				Exp int64 `json:"exp"`
+			}
+			if json.Unmarshal(payload, &claims) == nil && claims.Exp > 0 {
+				if time.Now().Unix() >= claims.Exp {
+					return fmt.Errorf("JWT expired at %s", time.Unix(claims.Exp, 0).Format(time.RFC3339))
+				}
+				return nil
+			}
+		}
+	}
+
+	// Strategy 3: HTTP validation — send a minimal malformed request to the
+	// Anthropic API. Auth is checked before request body: 401 = bad token,
+	// any other status (400, 422, etc.) = token accepted by auth layer.
+	// Network errors are treated as "can't determine" → assume valid.
+	return validateTokenHTTP(raw)
+}
+
+// validateTokenHTTP sends a minimal request to the Anthropic API to check if a
+// token is accepted by the auth layer. Returns error only for HTTP 401.
+func validateTokenHTTP(token string) error {
+	req, err := http.NewRequest("POST", "https://api.anthropic.com/v1/messages",
+		strings.NewReader("{}"))
+	if err != nil {
+		return nil // Can't create request → assume valid
+	}
+	req.Header.Set("Authorization", "Bearer "+token)
+	req.Header.Set("Content-Type", "application/json")
+	req.Header.Set("anthropic-version", "2023-06-01")
+
+	client := &http.Client{Timeout: 10 * time.Second}
+	resp, err := client.Do(req)
+	if err != nil {
+		return nil // Network error → assume valid
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode == http.StatusUnauthorized {
+		return fmt.Errorf("token rejected by API (HTTP 401)")
+	}
+	return nil
 }
 
 // expandTilde expands a leading ~/ to the user's home directory.
--- a/internal/quota/keychain_stub.go
+++ b/internal/quota/keychain_stub.go
@@ -22,3 +22,4 @@ func SwapKeychainCredential(_, _ string) (*KeychainCredential, error)           
 func RestoreKeychainToken(_ *KeychainCredential) error                             { return errNotDarwin }
 func SwapOAuthAccount(_, _ string) (json.RawMessage, error)                        { return nil, errNotDarwin }
 func RestoreOAuthAccount(_ string, _ json.RawMessage) error                        { return errNotDarwin }
+func ValidateKeychainToken(_ string) error                                         { return nil }
--- a/internal/quota/rotate.go
+++ b/internal/quota/rotate.go
@@ -33,6 +33,10 @@ type RotatePlan struct {
 	// One keychain swap per config dir, not per session.
 	// All sessions sharing a config dir get the same assignment.
 	ConfigDirSwaps map[string]string
+
+	// SkippedAccounts maps handle -> reason for accounts that were
+	// available by quota status but had invalid/expired tokens.
+	SkippedAccounts map[string]string `json:"skipped_accounts,omitempty"`
 }
 
 // PlanRotation scans for limited sessions and plans account assignments.
@@ -87,6 +91,25 @@ func PlanRotation(scanner *Scanner, mgr *Manager, acctCfg *config.AccountsConfig
 
 	// Get available accounts
 	available := mgr.AvailableAccounts(state)
+
+	// Validate tokens for available accounts — skip accounts with expired or
+	// revoked tokens. This prevents swapping a bad token into the target's
+	// keychain entry, which would leave the session non-functional.
+	skipped := make(map[string]string)
+	var validAvailable []string
+	for _, handle := range available {
+		acct, ok := acctCfg.Accounts[handle]
+		if !ok {
+			continue
+		}
+		configDir := util.ExpandHome(acct.ConfigDir)
+		if err := ValidateKeychainToken(configDir); err != nil {
+			skipped[handle] = err.Error()
+			continue
+		}
+		validAvailable = append(validAvailable, handle)
+	}
+	available = validAvailable
 
 	// Collect unique config dirs from limited sessions.
 	// Multiple sessions can share the same config dir (via the same account).
@@ -163,5 +186,6 @@ func PlanRotation(scanner *Scanner, mgr *Manager, acctCfg *config.AccountsConfig
 		AvailableAccounts: available,
 		Assignments:       assignments,
 		ConfigDirSwaps:    configDirSwaps,
+		SkippedAccounts:   skipped,
 	}, nil
 }


seanbearden:
@nohomedirectory @karanb192 I'd love your feedback on this! Going to test it locally today.

seanbearden:
## Manual Test: End-to-End Quota Rotation

Tested the full `gt quota` workflow against a live rate-limited crew session.

### 1. Check account status — all accounts available
```
$ gt quota status
Account Quota Status

   acct-a       available
   acct-b       available
   acct-c       available
 * acct-d       available

 Summary: 4 available, 0 limited
```

### 2. Scan detects rate-limited session (account unknown before first rotation)
```
$ gt quota scan

 ! gt-crew-bear              account: (unknown)

 Summary: 1 of 3 sessions rate-limited
```

### 3. Persist scan results
```
$ gt quota scan --update

 ! gt-crew-bear              account: (unknown)

 Summary: 1 of 3 sessions rate-limited
```

### 4. Preview rotation plan
```
$ gt quota rotate --dry-run

Rotation Plan

 → gt-crew-bear              (unknown) → acct-a

 (dry run — no changes made)
```

### 5. Execute rotation
```
$ gt quota rotate
Rotation Plan

 → gt-crew-bear              (unknown) → acct-a

 ✓ gt-crew-bear → acct-a
```

### 6. Session still rate-limited (hit limit on new account too)
```
$ gt quota scan

 ! gt-crew-bear              account: acct-a resets 11am (America/Los_Angeles)

 Summary: 1 of 3 sessions rate-limited
```

### 7. Rotate again — LRU picks next available account
```
$ gt quota rotate --dry-run

Rotation Plan

 → gt-crew-bear              acct-a → acct-b

 (dry run — no changes made)

$ gt quota rotate
Rotation Plan

 → gt-crew-bear              acct-a → acct-b

 ✓ gt-crew-bear → acct-b
```

### 8. First account now tracked as limited, second rotation succeeded
```
$ gt quota status
Account Quota Status

   acct-a       limited (resets 11am (America/Los_Angeles))
   acct-b       available
   acct-c       available
 * acct-d       available

 Summary: 3 available, 1 limited

$ gt quota scan

 ✓ No rate-limited sessions detected (3 scanned)
```

### Known issues filed
- **gt-hcxx**: Respawned session shows Bypass Permissions confirmation prompt — needs to carry over the acceptance flag
- **gt-zdvp**: Session context/history lost after respawn, `/resume` doesn't recover — manual `/login` preserves context without interruption

seanbearden:
@steveyegge @julianknutsen Heads up — the resume approach that landed in main (`executor.go`) has a fundamental issue that this branch fixes.

**The bug:** Main's `Rotator.executeOne` changes `CLAUDE_CONFIG_DIR` to the *new* account's config dir (line 204: `export CLAUDE_CONFIG_DIR=<newConfigDir>`), then tries to resume via `SessionLinker` symlinks. But `/resume` still can't reliably find the previous session transcripts because Claude Code resolves session files relative to the config dir at startup, and the symlink approach is fragile (noted in my earlier test as `gt-zdvp`: "Session context/history lost after respawn, `/resume` doesn't recover").

**The fix (this branch):** Instead of changing `CLAUDE_CONFIG_DIR`, we swap the macOS Keychain OAuth token from the available account *into* the rate-limited account's keychain entry, then respawn with the **same** config dir. Since the config dir doesn't change, `/resume` naturally finds all previous session transcripts. Tested live — both sessions recovered full conversation context and authenticated with the fresh account's credentials.

Key differences from main's approach:
- **Same config dir** → `/resume` just works (no symlinks needed)
- **Keychain swap** → one swap per config dir serves all sessions sharing it (config dir grouping)
- **Interactive `/resume`** → sends the slash command after respawn instead of `--resume` flag
- Backup/rollback support if the keychain write fails

The merge conflict resolution integrates main's good refactoring (deterministic ordering, `WithLock` lifecycle, `quotaLogger`) while replacing the rotation execution path with the keychain swap approach.

julianknutsen:
## Post-Merge Review — Triple-Model (Claude + Codex + Gemini)

### Decision: request_changes (post-merge)

Thanks for the contribution, @seanbearden! This is a well-architected addition — the scanner/planner/executor separation with injected interfaces is clean, the flock-scoped lifecycle eliminates TOCTOU races, and the 42 tests with race-detector coverage show real rigor. The operator pain point is real and this solves it.

Since this PR was already merged, the findings below will be addressed as **follow-up commits** prioritized by severity.

> **Follow-up plan:** Major findings (#1–#5) will be addressed in near-term follow-up commits. Minor findings (#6–#7) are tracked but lower priority. Nit (#8) is optional.

---

### Major: Rotate doesn't persist detected rate-limit statuses

`gt quota rotate` detects which accounts are limited during its scan phase but never writes those statuses to `quota.json`. A subsequent rotation has no memory of which accounts are rate-limited and can reassign a still-limited account to a new session.

**Suggested fix:** Persist detected limited statuses within the same lock scope as execution, reusing the `updateQuotaState` logic from the scan command.

---

### Major: `--json` flag broken for rotate no-op/dry-run branches

The `--json` flag is documented on `gt quota rotate`, but the no-op path (no limited sessions), no-assignment path (no available accounts), and `--dry-run` path all emit text or nothing rather than JSON. Machine consumers will break.

**Suggested fix:** Return structured JSON in all rotate exit branches when `--json` is set. Suppress text output in those branches.

---

### Major: Resume path drops role environment and working directory

When the resume path succeeds, `BuildResumeCommand` returns only `claude --dangerously-skip-permissions --resume <id>`, replacing the full restart command that includes `cd <workdir>`, `export GT_ROLE=...`, `export BD_ACTOR=...`, `export GIT_AUTHOR_NAME=...`, etc. The resumed session loses role identity, bead actor tracking, and git author attribution.

**Suggested fix:** Compose the resume command WITH the environment exports from `buildRestartCommand` — extract the `cd` + `export` prefix and prepend it to the resume command.

---

### Major: Default rate-limit scan pattern overly broad
**Source:** claude+gemini, medium confidence

The pattern `(?i)rate limit` matches any pane output containing those words — agent discussion, log messages, code comments. Since `gt quota rotate` scans AND executes, a false positive could kill a healthy session.

**Suggested fix:** Replace `"rate limit"` with a more specific pattern matching actual Claude rate-limit messages (e.g., `"You've hit your .*rate limit"`). The existing `"You've hit your limit"` and TUI-prompt patterns are already specific enough.

---

### Major: Session symlinks accumulated without cleanup
**Source:** claude+codex, medium confidence

When the resume path creates a session symlink, the cleanup function is deliberately discarded (`_ = cleanup`). `cleanupOrphanedSessionSymlinks()` only runs at seance start, never after rotation. Over many rotations, stale cross-account symlinks accumulate.

**Suggested fix:** Call `cleanupOrphanedSessionSymlinks()` as part of post-rotation cleanup, or document as a known limitation with a tracking issue.

---

### Minor: `ExpandHome` duplicates existing logic

Home directory expansion logic is now triplicated: `util.ExpandHome`, `config.expandPath`, and inline in `seance.go`. Since `seance.go` was modified in this PR, there's an opportunity to consolidate.

---

### Minor: Missing command-level tests for rotate branches
**Source:** codex+gemini, high confidence

No tests exercise the Cobra command layer's JSON output, dry-run formatting, or state persistence integration.

---

### Nit: Inconsistent indentation in rotation result display

Misleading indentation in the `if result.Rotated` block at `quota.go:433-443`.

---

### Pre-Merge Checklist (post-merge follow-ups)
- [ ] Persist detected rate-limit statuses during rotate
- [ ] Return JSON in all `gt quota rotate` exit branches when `--json` is set
- [ ] Compose resume command with full environment exports
- [ ] Tighten default rate-limit scan patterns
- [ ] Address session symlink accumulation
- [ ] Add command-level tests for rotate JSON/dry-run/persistence

---

seanbearden:
Thanks for the thorough triple-model review @julianknutsen! Here's what we addressed and our reasoning on the approach.

## Findings addressed

| # | Finding | Status |
|---|---------|--------|
| **1** | Rotate doesn't persist rate-limit statuses | **Fixed** — `updateQuotaState` called after `PlanRotation`, same lock-scoped persistence as `gt quota scan --update` |
| **2** | `--json` broken for no-op/dry-run | **Fixed** — all exit branches return structured JSON when `--json` is set (empty array for no-op, full plan for dry-run) |
| **3** | Resume drops role env/working dir | **Already fixed by design** — keychain swap uses `buildRestartCommand` (full `cd` + exports) for respawn. Role identity, bead actor, git author all preserved. |
| **4** | Overly broad scan patterns | **Fixed** — replaced bare `resets` pattern with `limit\s*·\s*resets` requiring rate-limit context. `You've hit your limit` already anchored. |
| **5** | Session symlinks accumulate | **N/A** — keychain swap keeps the same config dir, no symlinks needed |
| **6** | ExpandHome duplicated | **Fixed** — removed `expandAccountPath`, replaced inline expansion in `seance.go`, both now use `util.ExpandHome` |
| **7** | Missing command-level tests | **Fixed** — added state persistence, dry-run plan serialization, config dir grouping, and preemptive rotation tests |
| **8** | Inconsistent indentation | **Already fixed** — inline result display in the rotation loop eliminated the misleading nesting |

## Context-preserving rotation via `--continue`

Instead of sending interactive `/resume` after respawn (which was fragile and raced with the `gt prime` handoff hook), rotation now uses `claude --continue` to resume the most recent conversation at startup. This runs before any hooks fire, so the agent picks up exactly where it left off.

- **`buildRestartCommandWithOpts`** accepts `ContinueSession bool` — when true, injects `--continue` into the CLI command and uses a continuation prompt ("Your account was rotated to avoid a rate limit. Continue your previous task.") instead of the full `[GAS TOWN]` handoff beacon
- The old `sendResumeToSession` function (which sent `/resume` as interactive input) has been removed entirely

## Preemptive rotation (`--from` + `--idle`)

Swap sessions to a fresh account *before* hitting rate limits, targeting only idle agents to avoid disrupting active work.

```
gt quota rotate --from work --idle     # rotate only idle sessions on 'work' account
gt quota rotate --from work            # rotate all sessions on 'work' (careful!)
gt quota rotate --from work --dry-run  # preview what would be rotated
```

- **`--from <handle>`**: targets all sessions using the specified account, regardless of rate-limit status
- **`--idle`**: only rotates sessions at the idle prompt; uses Claude Code's `⏵⏵` status bar with "esc to interrupt" as the definitive busy/idle signal
- Preemptive rotations skip marking accounts as "limited" in state (since they aren't actually rate-limited)

## Full account identity swap (keychain + oauthAccount)

The keychain swap alone wasn't sufficient — Claude Code caches account identity (`accountUuid`, `organizationUuid`, `emailAddress`) in `.claude.json`'s `oauthAccount` field and sends it with API requests. The server rate-limits based on this identity, not just the auth token. Now rotation swaps both:

1. **Keychain token** — the OAuth credential used for authentication
2. **`oauthAccount` in `.claude.json`** — the cached account identity so the API sees the correct account

The `oauthAccount` swap is skipped gracefully when `.claude.json` doesn't exist (e.g., default `~/.claude` config dir).

## Unknown-account session support

Sessions that weren't started via our rotation (no `GT_QUOTA_ACCOUNT` env var, `CLAUDE_CONFIG_DIR` doesn't match any registered account) were previously skipped entirely. Now:

- **Scanner captures `CLAUDE_CONFIG_DIR`** for every session, independent of account resolution
- **Falls back to `~/.claude`** when `CLAUDE_CONFIG_DIR` isn't set (Claude Code's default)
- **`PlanRotation` includes unknown-account sessions** in the plan as long as they have a config dir to target for keychain swap
- The warning message correctly distinguishes "no CLAUDE_CONFIG_DIR" (truly unrotatable) from "not enough available accounts"

## Scanner false-positive fix

The scanner was checking all 30 captured pane lines for rate-limit patterns. After a rate limit is resolved (e.g., `/login`), the old "You've hit your limit" message stays in the scrollback. Now only the **bottom 10 lines** are checked — resolved rate limits scroll above the check window.

## Auto-expiry of rate-limited accounts

Added `ClearExpired` which parses human-readable reset times like "7pm (America/Los_Angeles)" and automatically marks accounts as available when their reset time passes. Runs on every `gt quota status` and during rotation planning.

## Account tracking after keychain swap

After rotation, the scanner needs to know which account is actually active (the config dir still maps to the old account). Now:

- **`GT_QUOTA_ACCOUNT`** env var is set in both the shell process (`export`) and tmux session environment (`SetEnvironment`)
- Scanner checks `GT_QUOTA_ACCOUNT` first, falls back to `CLAUDE_CONFIG_DIR` matching

## Why macOS Keychain swap instead of config dir switching

The original approach changed `CLAUDE_CONFIG_DIR` to the new account's directory, then tried to recover context via `SessionLinker` symlinks + `--resume <id>`. In live testing this failed — session context was lost. The root cause: Claude Code resolves session transcripts relative to the config dir at startup, and symlinks across config dirs are fragile.

The keychain swap approach solves this at the auth layer:
1. **Swap the OAuth token + account identity** in the rate-limited account's config dir
2. **Keep the same `CLAUDE_CONFIG_DIR`** — all session transcripts stay where Claude expects them
3. **Respawn with `--continue`** — naturally finds previous conversation, full context recovered

## macOS-only / Windows support

The keychain approach is macOS-only by design (`//go:build darwin` + `keychain_stub.go` for other platforms). `SwapOAuthAccount` and `RestoreOAuthAccount` are also stubbed for non-darwin builds. If Gas Town needs to run on other platforms, the stub provides a clear extension point — swap in the platform's credential store. For now, all Gas Town rigs run on macOS.

steveyegge:
Closing this draft PR for now. Feel free to reopen when it's ready for review — happy to take another look\! Thanks for the contribution.

PR #1541, opened 2026-02-16
fix(rig): create server-side database and set issue_prefix during gt rig add

julianknutsen:
## Summary

`gt rig add` was missing the `doltserver.InitRig()` call to create the server-side database in `.dolt-data/` before initializing beads. This caused `bd init --server` to write `metadata.json` pointing at the wrong database name (`beads_<prefix>` instead of `<rigName>`), and `issue_prefix` was set against the non-existent embedded database. After `EnsureMetadata` corrected `dolt_database`, the prefix remained on the wrong database, causing all bead creation to fail with "issue_prefix config is missing."

## Related Issue

Fixes #1540

## Changes

- Add `doltserver.InitRig()` call before `InitBeads` in `AddRig`, guarded by `exec.LookPath("dolt")` (matching the `gt install` pattern at `install.go:311-316`)
- Re-set `issue_prefix` and `types.custom` after `EnsureMetadata` corrects `dolt_database`, ensuring config targets the correct server-side database
- Add clarifying comments on the ordering dependency between `EnsureMetadata` and `issue_prefix` setup

## Testing

- [x] Unit tests pass (`go test ./...`)
- [x] Manual testing performed — validated in e2e test that `gt rig add` followed by bead creation (`bd create --id=<prefix>-xxx`) succeeds with the fix and fails without it
- [x] `go vet ./internal/rig/...` passes

## Checklist

- [x] Code follows project style
- [ ] Documentation updated (if applicable)
- [x] No breaking changes (or documented in summary)

PR #1679, opened 2026-02-18
fix(rig): don't create orphaned Dolt database for non-tracked rigs

Rome-1:
## Problem

When `gt rig add` runs on a repo with no tracked `.beads/` directory (the common case), two things happened in sequence that contradicted each other:

1. `initBeads` called `bd init --prefix <prefix> --backend dolt --server`, which created a standalone Dolt database named `beads_<prefix>` in `.dolt-data/`
2. `AddRig` then called `EnsureMetadataWithDB(..., "hq")` to implement the share-HQ design (gt-42zaq), overwriting `metadata.json` to point at the HQ database

The `beads_<prefix>` database had no owner and became a permanent orphan. This showed up as two doctor warnings after every `gt rig add` on a standard repository:

```
⚠  dolt-orphaned-databases  1 orphaned database(s) in .dolt-data/
    └─ Orphaned: beads_tr (125.1 KB)

⚠  database-prefix  1 database prefix mismatch(es) with routes.jsonl
    └─ Route 'testrig': routes.jsonl says 'tr', database has 'hq'
```

These warnings appeared in the e2e CI output on every run.

## Fix

For non-tracked rigs, skip `bd init` entirely. Write only `config.yaml` and `issues.jsonl`—the minimum needed for gastown's Go code to read the prefix. `EnsureMetadataWithDB` in `AddRig` writes `metadata.json` pointing at HQ as before, but now there's no abandoned database left behind.

The `bd init`-produced configuration (`types.custom`, `issue_prefix`, `migrate --update-repo-id`) was always targeting the orphaned database anyway—never the HQ database that `bd` actually connected to. Removing it changes nothing observable.

## Tests

- Add `assertRigBeadsNoOrphan` helper that checks three things after `gt rig add` on a non-tracked repo:
  - `testrig/.beads/config.yaml` exists and contains the prefix
  - `testrig/.beads/metadata.json` has `dolt_database: "hq"`
  - No `beads_*` directory exists in `.dolt-data/`
- Add `verify-rig-beads` subtest in both `TestInstallDoctorClean` and `TestInstallWithDaemon`
- Tighten both doctor subtests to fail on `dolt-orphaned-databases` warnings

## Notes

The `database-prefix` doctor warning (routes.jsonl says `tr`, HQ database says `hq`) is a separate design-level question about the share-HQ strategy and is not addressed here. The orphan is the concrete bug; the prefix mismatch is a known trade-off of the gt-42zaq design.

### DIFF
--- a/internal/cmd/install_integration_test.go
+++ b/internal/cmd/install_integration_test.go
@@ -433,6 +433,11 @@ func TestInstallDoctorClean(t *testing.T) {
 		assertDirExists(t, filepath.Join(rigPath, ".repo.git"), "testrig/.repo.git/")
 	})
 
+	// 5b. Verify rig beads: no orphaned Dolt database, metadata shares HQ
+	t.Run("verify-rig-beads", func(t *testing.T) {
+		assertRigBeadsNoOrphan(t, hqPath, "testrig", "tr")
+	})
+
 	// 6. Add a crew member
 	t.Run("crew-add", func(t *testing.T) {
 		runGTCmd(t, gtBinary, hqPath, env, "crew", "add", "jayne", "--rig", "testrig")
@@ -466,6 +471,10 @@ func TestInstallDoctorClean(t *testing.T) {
 			if strings.Contains(outStr, signal) {
 				t.Fatalf("doctor crashed with %s", signal)
 			}
+		}
+		// Regression: initBeads orphan bug should not create beads_<prefix> in .dolt-data/
+		if strings.Contains(outStr, "⚠  dolt-orphaned-databases") {
+			t.Errorf("doctor reports orphaned databases after rig-add (initBeads regression):\n%s", outStr)
 		}
 	})
 }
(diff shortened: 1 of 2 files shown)


l0g1x:
Good stuff, this was bothering me too

julianknutsen:
## Closing — already resolved by #1541

Thanks for investigating this, @Rome-1! The orphaned database issue you identified was real, but it's already been fixed by #1541 (merged Feb 16), which reordered `AddRig` to call `doltserver.InitRig()` before `InitBeads()`.

### Why orphans no longer occur

The current call sequence in `AddRig` is:

1. **`doltserver.InitRig(townRoot, rigName)`** — creates `.dolt-data/<rigName>/` and calls `EnsureMetadata`, which writes `.beads/metadata.json` with `dolt_mode: "server"` and `dolt_database: <rigName>`
2. **`InitBeads` → `bd init --prefix <prefix> --server`** — beads' safety guard (`checkExistingBeadsDataAt`) detects the existing `.beads/metadata.json` with `dolt_mode: "server"` and exits with "already initialized", which gastown catches and continues past
3. **`EnsureMetadata` (second pass)** — no-op since fields are already set

Because `bd init` never gets past the safety guard, it never reaches the code that would create a `beads_<prefix>` database or overwrite `dolt_database` in metadata.json. The metadata stays pointed at `<rigName>` — the correct server-side database.

I verified this with an integration test (`TestMetadataDatabaseAudit`) against current main that audits metadata after install + rig add:
- HQ: `dolt_database: "hq"`, databases: `[hq]`, zero orphans
- Rig: `dolt_database: "testrig"`, databases: `[hq, testrig]`, zero orphans

### Note on the test assertion

The test in this PR asserts `dolt_database: "hq"` for the rig, but `EnsureMetadata` sets `dolt_database` to the rig name (not "hq"), so that assertion would fail against current main.

Separately, I filed [beads#1862](https://github.com/steveyegge/beads/issues/1862) for adding a `bd init --server --database <name>` flag so the integration boundary is cleaner long-term.

Issue #1681, opened 2026-02-18
session: avoid stale cleanup false positives on PID reuse


davidahmann:
Active agents can be killed during stale-session cleanup when the OS reuses a PID after a restart, causing loss of in-flight work.

Why now:
- Restart-heavy flows are common during active orchestration.
- PID reuse creates intermittent false stale detection in operational paths.

Expected behavior:
- Stale cleanup should only reap truly stale sessions and should not kill active sessions that happen to share a reused PID.

Evidence packet:
- Commit under test: origin/main @ b1e71c86
- Runtime: Go 1.25.7, Darwin arm64 (macOS 26 kernel 25.3.0)
- Minimal repro:
  1. Start a session and persist PID metadata
  2. Simulate process restart where PID value is reused by a new active agent
  3. Run stale cleanup
- Expected: active session retained.
- Actual: active session can be misclassified stale and reaped.

Acceptance / validation:
- Add PID-reuse simulation test proving active session survives cleanup.
- Existing stale cleanup behavior for truly dead sessions remains unchanged.

julianknutsen:
Closed by #1682

PR #1682, opened 2026-02-18
session: avoid PID-reuse kills in tracked PID cleanup

davidahmann:
## Problem
Stale-session cleanup can terminate active processes when a previously tracked PID is reused by the OS, causing avoidable work loss.

## Why now
Restart-heavy flows make PID reuse realistic; unsafe cleanup during `gt down --all/--force` is high-impact.

## What changed
- Extended PID tracking records to include process start time when available.
- Added reuse guard in `KillTrackedPIDs` to skip termination when start time mismatches (PID reused).
- Added parser/behavior tests for tracked format and PID-reuse skip path.

## Validation
- `go test ./internal/session/...`

Refs #1681

### DIFF
--- a/internal/session/pidtrack.go
+++ b/internal/session/pidtrack.go
@@ -12,6 +12,8 @@ import (
 	"github.com/steveyegge/gastown/internal/tmux"
 )
 
+// pidStartTimeFunc is overridden in tests. This package's tests must NOT use
+// t.Parallel() because they mutate this package-level variable without synchronization.
 var pidStartTimeFunc = processStartTime
 
 type trackedPID struct {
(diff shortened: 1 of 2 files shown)


davidahmann:
Implemented PID-reuse hardening for tracked PID cleanup.

- PID records now include process start-time metadata when available.
- Cleanup verifies current PID start time before SIGTERM to avoid killing a reused PID.
- Added coverage for parsed record format and reuse mismatch skip behavior.
- Validation: `go test ./internal/session/...` passed.

This contribution was informed by patterns from Gait: https://github.com/Clyra-AI/gait

julianknutsen:
## Adoption Review

Thanks for the contribution, @davidahmann! This PR adds a critical safety guard to the PID tracking cleanup path — preventing `KillTrackedPIDs` from terminating unrelated processes when PIDs get reused by the OS. The approach of recording process birth time at track and comparing at kill is exactly right for defense-in-depth orphan cleanup.

### Initial Review
The initial triple-model review (Claude + Codex + Gemini) identified 4 findings:
- **Major:** `startErr` conflated with mismatch — lookup errors deleted tracking files, creating silent orphan leaks
- **Major:** `ps -o lstart=` output is locale-sensitive — format differences between track/kill cause false-positive PID-reuse detection
- **Minor:** PID 1 test was non-deterministic on some platforms
- **Minor:** `pidStartTimeFunc` global needs `t.Parallel()` warning

### Maintainer Changes
The fixup commit (2 files, +64 -5) addresses all initial review findings:
1. Split `startErr` from mismatch — lookup errors now preserve PID file and report via `errSessions`
2. Pinned `LC_ALL=C` on `processStartTime` `exec.Cmd`
3. Routed warnings through `errSessions` return channel (not `log.Printf`)
4. Fixed reuse test to use `os.Getpid()` with `startTimeCalled` assertion
5. Added `TestKillTrackedPIDs_PreservesFileOnLookupError`
6. Added `t.Parallel()` warning comment and Windows/container docs

### Final Review Status
All clear — third review pass returned **approve** with no blockers or majors. Remaining observations are minor/nit and optional follow-ups:
- PID-only fallback degrades silently (pre-existing behavior, not a regression)
- Hardcoded `.runtime` string vs `constants.DirRuntime` (optional cleanup)

### Review Iterations
3 review passes performed (1 PR-mode, 2 local branch iterations).

---

PR #1685, opened 2026-02-18
fix(polecat): close TOCTOU window in AllocateName() with .pending reservation

athosmartins:
## Problem

Two concurrent `gt sling` processes could silently allocate the same polecat name. The pool flock protected name selection, but the polecat directory wasn't created until **after** the flock was released. A second process running `reconcilePoolInternal` in that window would see no directory and reallocate the same name.

This was observed in the wild: hq-ypvza lost polecat `furiosa` — it was silently overwritten by a concurrent allocation.

Tracked in: gt-601kx

## Fix: `.pending` reservation marker

Write a `.pending` file **inside** the pool lock scope in `AllocateName()`, before releasing the flock. The file contains the allocating process's PID for diagnostics.

Three-part fix:

**1. `AllocateName()` — write marker under lock**
After saving the pool state (still holding the flock), write `polecats/<name>.pending`. This closes the window: any concurrent `reconcilePoolInternal` that runs after the lock is released will see the marker and treat the name as in-use.

**2. `reconcilePoolInternal()` — include pending names in in-use set**
Scan `polecats/` for `*.pending` files and add those names to `namesWithDirs` before feeding into `ReconcilePoolWith`. Without this, reconcile would see no directory and treat the name as available.

**3. Cleanup in two places**
- `AddWithOptions()`: removes `<name>.pending` after `os.MkdirAll` succeeds — the directory is now the authoritative source of truth.
- `cleanupOrphanPolecatState()`: removes `.pending` files older than `pendingMaxAge` (5 minutes). Handles the crash-after-AllocateName-before-AddWithOptions case, so names are never permanently reserved by dead processes.

## Tests

Two new tests in `manager_test.go`:
- `TestPendingMarkerBlocksReallocation`: verifies a `.pending` marker prevents concurrent allocation of the same name.
- `TestStalePendingMarkerIsCleanedUp`: verifies `cleanupOrphanPolecatState` removes markers older than `pendingMaxAge`.

## Checklist

- [x] All existing polecat tests pass (`go test ./internal/polecat/...`)
- [x] Two new regression tests added
- [x] No changes to external API or CLI behavior
- [x] Fix is backwards-compatible (stale markers from before this change are cleaned up after 5 min)

### DIFF
--- a/internal/polecat/manager.go
+++ b/internal/polecat/manager.go
@@ -1034,8 +1034,11 @@ func (m *Manager) AllocateName() (string, error) {
 	// directory until AddWithOptions removes it after os.MkdirAll succeeds.
 	// Stale markers (process crashed before AddWithOptions) are cleaned up by
 	// cleanupOrphanPolecatState after pendingMaxAge.
-	if err := os.MkdirAll(filepath.Join(m.rig.Path, "polecats"), 0755); err == nil {
-		_ = os.WriteFile(m.pendingPath(name), []byte(fmt.Sprintf("%d", os.Getpid())), 0644)
+	if err := os.MkdirAll(filepath.Join(m.rig.Path, "polecats"), 0755); err != nil {
+		return "", fmt.Errorf("creating polecats dir for reservation marker: %w", err)
+	}
+	if err := os.WriteFile(m.pendingPath(name), []byte(fmt.Sprintf("%d", os.Getpid())), 0644); err != nil {
+		return "", fmt.Errorf("writing reservation marker: %w", err)
 	}
 
 	// Kill any lingering tmux session for this name (gt-pqf9x).


julianknutsen:
## Adoption Review

Thanks for the contribution, @athosmartins! This is a well-analyzed fix for a real concurrency bug — the TOCTOU window between `AllocateName()` releasing the pool flock and `AddWithOptions()` creating the directory was a genuine race that caused silent polecat name collisions in the wild. The `.pending` marker approach is clean, minimal, and follows the existing filesystem-as-source-of-truth pattern.

### Initial Review
The triple-model review (Claude + Codex + Gemini) validated the TOCTOU fix as correct and identified one major issue: the `.pending` marker write was best-effort — if it failed, `AllocateName()` silently degraded to the pre-fix vulnerable behavior. Two minor issues were also flagged: `gt polecat identity add` leaving orphaned markers until stale cleanup, and lack of end-to-end lifecycle tests.

### Maintainer Changes
Made the `.pending` marker write fail-closed: if `os.MkdirAll` or `os.WriteFile` fails, `AllocateName()` now returns an error instead of silently continuing without TOCTOU protection. (+5 -2 in `manager.go`)

### Final Review Status
All clear — re-review approved with no blockers or majors. Remaining findings are all minor/nit: a theoretical split-snapshot race in `reconcilePoolInternal`'s two-pass scan (vanishingly narrow, self-healing), a transient name leak on marker write failure (self-healing via reconcile), and optional test coverage for the fail-closed path.

### Review Iterations
2 review passes performed.

---

PR #1686, opened 2026-02-18
fix(boot): execute pending death warrants in degraded triage

athosmartins:
## Problem

Boot's degraded mode (`gt boot triage --degraded`) checks Deacon health but never scans `~/gt/warrants/` for pending death warrants. As a result, 7 warrants have accumulated unexecuted since the warrant system was introduced.

The warrant system docs say:
> Boot picks up the warrant during triage cycles and executes it.

This was true for the AI code path (where the formula prompts Claude to check warrants) but never implemented in the mechanical degraded path.

## Solution

Add `executeWarrants()` helper that scans the warrants directory as a side effect at the start of `runDegradedTriage`, before the normal Deacon health decision.

**Execution order in `runDegradedTriage`:**
1. Shutdown check (existing — abort if `gt down` in progress)
2. **Warrant scan and execution** (new — side effect, no return value)
3. Deacon health check (existing — returns triage action to caller)

**Design choices:**
- Non-fatal: individual warrant failures are logged and skipped, not fatal to triage
- Already-executed warrants are skipped (`Executed == true`)
- Missing warrants directory is handled gracefully (not an error)
- Reuses existing `targetToSessionName()` and `Warrant` type from `warrant.go`
- `tmux.Tmux` instance is reused from the Boot struct (already initialized)

## Tests

5 new unit tests in `boot_test.go`:
- `TestExecuteWarrants_MarksPendingAsExecuted` — pending warrant marked executed when session dead
- `TestExecuteWarrants_SkipsAlreadyExecuted` — already-executed warrants not modified
- `TestExecuteWarrants_MissingDir` — missing warrants dir handled gracefully
- `TestExecuteWarrants_EmptyDir` — empty warrants dir handled gracefully
- `TestExecuteWarrants_IgnoresNonWarrantFiles` — non-`.warrant.json` files ignored

Tests use real `tmux.NewTmux()` with non-existent session names, which takes the "already dead" code path (tmux returns `ErrSessionNotFound` → `HasSession` returns `false, nil`).

## Related

Companion to PR fixing `AllocateName()` TOCTOU race (separate PR by @athosmartins/digo).

### DIFF
--- a/internal/cmd/boot.go
+++ b/internal/cmd/boot.go
@@ -510,27 +510,10 @@ func executeWarrants(warrantDir string, tm *tmux.Tmux) {
 			continue
 		}
 
-		sessionName, err := targetToSessionName(w.Target)
-		if err != nil {
-			fmt.Printf("Warning: warrant for %s has invalid target: %v\n", w.Target, err)
+		if err := executeOneWarrant(&w, path, tm); err != nil {
+			fmt.Printf("Warning: executing warrant for %s: %v\n", w.Target, err)
 			continue
 		}
-
-		if has, _ := tm.HasSession(sessionName); has {
-			if err := tm.KillSession(sessionName); err != nil {
-				fmt.Printf("Warning: killing session %s for warrant %s: %v\n", sessionName, w.Target, err)
-				continue
-			}
-			fmt.Printf("Warrant executed: terminated session %s (%s)\n", sessionName, w.Target)
-		} else {
-			fmt.Printf("Warrant executed: session %s already dead (%s)\n", sessionName, w.Target)
-		}
-
-		now := time.Now()
-		w.Executed = true
-		w.ExecutedAt = &now
-		data, _ = json.MarshalIndent(w, "", "  ")
-		_ = os.WriteFile(path, data, 0644)
 	}
 }
 
--- a/internal/cmd/warrant.go
+++ b/internal/cmd/warrant.go
@@ -303,34 +303,69 @@ func runWarrantExecute(cmd *cobra.Command, args []string) error {
 		return nil
 	}
 
-	// Determine session name from target
-	sessionName, err := targetToSessionName(target)
-	if err != nil {
-		return fmt.Errorf("determining session name: %w", err)
-	}
-
-	// Kill the session if it exists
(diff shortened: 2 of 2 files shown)


athosmartins:
Fixed the CI failure — the `internal/config` test failures (`TestPiProviderDefaults`, `TestFillRuntimeDefaults/pi_command_gets_hooks_and_tmux_defaults`) were pre-existing on main but I should have fixed them before submitting. Root cause: `AgentPi` preset had `SupportsHooks: true` but was missing `HooksProvider`/`HooksDir`/`HooksSettingsFile`, so `fillRuntimeDefaults()` never auto-filled hooks for the pi command. Added the three fields in the follow-up commit.

julianknutsen:
## Adoption Review

Thanks for the contribution, @athosmartins! This PR fills an important gap — degraded triage was supposed to execute warrants per the docs, but the mechanical path never actually did. The approach is clean: `executeWarrants` as a non-fatal side effect before the Deacon health decision, reusing existing warrant infrastructure. The tests are thorough and the Pi preset fix is a nice bonus.

### Initial Review
The triple-model review (Claude + Codex + Gemini) found one blocker and two majors:
1. **Blocker:** `HasSession` errors were silently discarded — transient tmux failures would permanently consume warrants without killing the target session.
2. **Major:** `KillSession` used instead of `KillSessionWithProcesses`, leaving orphan processes that ignore SIGHUP.
3. **Major:** Warrant execution logic was duplicated between `boot.go` (batch) and `warrant.go` (CLI), creating drift risk.

### Maintainer Changes
Extracted shared `executeOneWarrant()` into `warrant.go` that both code paths now call. This function:
- Handles `HasSession` errors explicitly (warrants NOT marked executed when status is unknown)
- Uses `KillSessionWithProcesses` for full process tree cleanup
- Propagates write-back errors instead of silently discarding them

The fix also improves the existing `gt warrant execute` CLI command, which had the same issues.

### Final Review Status
All three initial blockers/majors resolved. Re-review decision: **approve** with only minor/nit findings remaining (shutdown race cosmetics, ad-hoc Printf logging, test coverage for error branches). None blocking.

### Review Iterations
2 review passes performed.

---

PR #1691, opened 2026-02-18
fix(mol): add --jitter flag to gt mol squash to reduce Dolt lock contention

athosmartins:
## Problem

Multiple patrol agents (deacon, witness, refinery) all call `bd mol squash` at
the end of their patrol cycles. When these cycles align, all agents try to acquire
Dolt's exclusive write lock simultaneously, causing:

- `bd mol current` spinning at 90-101% CPU for 10-30 minutes
- All other `bd` commands failing with `dolt access lock timeout (exclusive, 15s)`
- System-wide lock contention blocking every rig

## Fix

Add a `--jitter <duration>` flag to `gt mol squash` that sleeps a random duration
from 0 to the given value **before** acquiring any Dolt lock. This desynchronizes
concurrent squash calls without changing semantics.

- `molecule.go`: add `--jitter` flag definition
- `molecule_lifecycle.go`: implement jitter in `runMoleculeSquash`
- `mol-{witness,deacon,refinery}-patrol.formula.toml`: updated to use `gt mol squash --jitter 10s`
- `patrol_backoff_test.go`: enforce that all patrol formulas use `--jitter`

## Testing

```bash
go test ./internal/formula/... # passes
```

Patrol agents now stagger their squash calls by up to 10 seconds, distributing
Dolt lock acquisitions over time rather than bunching at cycle boundaries.


julianknutsen:
This looks to have a few unrelated commits on it. Please rebase and resubmit.

athosmartins:
Rebased. PR now contains only the jitter commit — the 4 unrelated commits (session name validation, hooks registry docs, two bd timeout fixes) have been dropped.

PR #1702, opened 2026-02-18
fix: cache last good status in watch mode to prevent empty bubbles

### LINKED ISSUE #1701: gt status --watch intermittently shows empty bubbles
## Problem

`gt status --watch` intermittently shows all agents as not running (empty ○ indicators) even when agents are actively running. The status output shows blank bubbles with no mail info or hook data, despite `gt status` (single-run) showing correct data at the same moment.

## Root Cause

Watch mode polls `gatherStatus()` every 2 seconds. Each iteration spawns multiple tmux subprocesses via `ListSessions()` and parallel `IsAgentAlive()` calls (2+ subprocess spawns per session). Under load, the tmux server intermittently fails to respond, causing:

1. `ListSessions()` failure → empty session map → all agents show ○ (not running)
2. Individual `IsAgentAlive()` failures → specific agents show ○
3. Beads query errors (silently swallowed) → no hook/work info displayed
4. Mail query errors (silently swallowed) → no mail counts

The single-run `gt status` uses the identical `gatherStatus()` code path but only calls it once, so transient failures are less likely to be observed.

## Expected Behavior

Watch mode should show consistent agent status, falling back to cached data during transient failures rather than displaying a blank/empty screen.

## Reproduction

Run `gt status --watch` in a workspace with several running agents. Over time (seconds to minutes), the display will intermittently flash to show all agents as not running before recovering on the next poll cycle.

ekoziol:
## Summary

- Add retry-on-failure and cache-based fallback to `runStatusWatch()` to prevent transient tmux failures from showing empty agent bubbles
- When `gatherStatus()` returns zero running agents after previously having some, retry once then fall back to cached data (up to 5 missed cycles)
- Display a staleness note when using cached data so the user knows the display may be slightly behind
- Add `countRunningAgents()` helper with table-driven tests

Closes #1701

## Test plan

- [x] `TestCountRunningAgents` covers: empty status, global-only, rig-only, mixed, all-not-running
- [x] All existing status tests pass (`TestRunStatusWatch_*`, `TestDiscoverRigAgents_*`, `TestBuildStatusIndicator_*`)
- [x] Full build passes (`go build ./...`)
- [ ] Manual: run `gt status --watch` with active agents and verify stable display

### DIFF
--- a/internal/cmd/status.go
+++ b/internal/cmd/status.go
@@ -431,6 +431,14 @@ func runStatusWatch(_ *cobra.Command, _ []string) error {
 
 	isTTY := term.IsTerminal(int(os.Stdout.Fd()))
 
+	// Cache the last successful status to handle transient tmux/beads
+	// failures. Watch mode spawns many tmux subprocesses per iteration;
+	// under load the tmux server can intermittently fail, causing all
+	// agents to appear as not running (empty bubbles).
+	var cachedStatus *TownStatus
+	var cachedAt time.Time
+	maxStale := time.Duration(statusInterval) * time.Second * 5
+
 	for {
 		var buf bytes.Buffer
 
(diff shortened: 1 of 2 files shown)


steveyegge (approved):
Clean fix — retry + cache fallback for transient tmux failures in watch mode. Good tests.

PR #1711, opened 2026-02-18
fix(doctor): use --ephemeral and verify DB state in misclassified-wisps check

jason-curtis:
## Summary
- **Fix**: Use `bd update --ephemeral` instead of `bd close` when fixing misclassified wisps. This marks issues as ephemeral (preserving audit trail) rather than permanently closing them. Requires bd 0.52.0+
- **Check improvement**: Verify current DB state before flagging JSONL candidates. JSONL may be stale when the bd daemon isn't running, causing false positives for issues already closed or marked ephemeral in the live database

## Test plan
- [ ] Run `gt doctor` — misclassified-wisps check should not flag already-ephemeral issues
- [ ] Run `gt doctor --fix` — should use `bd update --ephemeral` instead of `bd close`
- [ ] Verify no regression with bd versions that support `--ephemeral`

### DIFF
--- a/internal/doctor/misclassified_wisp_check.go
+++ b/internal/doctor/misclassified_wisp_check.go
@@ -66,10 +66,12 @@ func (c *CheckMisclassifiedWisps) Run(ctx *CheckContext) *CheckResult {
 	}
 
 	var details []string
+	var totalProbeErrors int
 
 	for _, rigName := range rigs {
 		rigPath := filepath.Join(ctx.TownRoot, rigName)
-		found := c.findMisclassifiedWisps(rigPath, rigName)
+		found, probeErrors := c.findMisclassifiedWisps(rigPath, rigName)
+		totalProbeErrors += probeErrors
 		if len(found) > 0 {
 			c.misclassified = append(c.misclassified, found...)
 			c.misclassifiedRigs[rigName] = len(found)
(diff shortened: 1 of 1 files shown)


julianknutsen:
## Adoption Review

Thanks for the contribution, @jason-curtis! Switching `gt doctor --fix` from `bd close` to `bd update --ephemeral` is a meaningful improvement — preserving the audit trail for misclassified wisps instead of permanently closing them is the right call. The stale-JSONL guard is a smart addition that prevents false positives when the daemon isn't running.

### Initial Review
The initial dual-model review (Claude + Codex) found a blocker and a major: `isIssueStillOpen` was missing `cmd.Dir` (querying the wrong beads DB), and the fail-open error semantics preserved exactly the false positives the PR aimed to eliminate.

### Maintainer Changes
Fixup commit addresses all review findings across 3 iterations:
- Set `cmd.Dir` on the DB probe for rig-scoped queries
- Added `--allow-stale` flag consistent with all other `bd` invocations
- Changed `CombinedOutput()` to `Output()` to prevent stderr warnings from poisoning JSON parse
- Changed `isIssueStillOpen` to return errors so probe failures are tracked and surfaced in `CheckResult.Details`
- Updated stale doc comment on `Fix` method

### Final Review Status
All clear — no remaining blockers or majors. Two minor non-blocking items noted: `BEADS_DIR` env pinning (pre-existing pattern gap across doctor package) and test coverage (pre-existing gap).

### Review Iterations
3 review passes performed.

---

PR #1714, opened 2026-02-18
fix(config): set GT_AGENT for role_agents to prevent witness auto-nuke

arttttt:
## Problem

Polecats running non-Claude agents (e.g., opencode) get auto-nuked by witness patrol while actively working.

### How we discovered this

Four polecats were dispatched via `gt sling shiny-enterprise` to work on Phase 2 tasks in a terminal rig. The rig was configured with `role_agents.polecat = "codex"` (an opencode-based custom agent using GPT 5.3). All four polecats started successfully, began writing code, and then silently died — no crash logs, no kill events, no error output. Worktrees were cleaned up and agent beads showed `agent_state: nuked`.

### What was happening

The witness patrol runs `DetectZombiePolecats` periodically, which calls `IsAgentAlive` for each polecat. `IsAgentAlive` reads `GT_AGENT` from the tmux session environment to determine which process names to look for. When `GT_AGENT` is empty, it falls back to `["node", "claude"]`.

The bug: `BuildStartupCommand` never set `GT_AGENT` when the agent was resolved through `role_agents` config (as opposed to an explicit `--agent` CLI override). So polecats running opencode had no `GT_AGENT` in their session. Witness looked for `node`/`claude` processes, found none, concluded the polecat was a zombie, and nuked it.

```
role_agents.polecat = "codex"
    → ResolveRoleAgentConfig("polecat") → RuntimeConfig{Command: "opencode", ...}
    → BuildStartupCommand() → GT_AGENT not set (agent name lost during resolution)
    → IsAgentAlive() → GetProcessNames("") → ["node", "claude"]
    → process "opencode" not found → zombie → auto-nuke
```

### 1. Propagate resolved agent name through RuntimeConfig

Added `ResolvedAgent string` field (runtime-only, `json:"-"`) to `RuntimeConfig`. Set it in `resolveRoleAgentConfigCore` and `resolveAgentConfigInternal` when the agent name is determined — before it gets lost in the lookup chain.

### 2. Export GT_AGENT in all startup paths

`BuildStartupCommand` now sets `GT_AGENT` from `rc.ResolvedAgent`. `BuildStartupCommandWithAgentOverride` uses explicit override if present, falls back to `ResolvedAgent`.

### 3. Fail-fast validation

Added a check in polecat `session_manager.Start()`: after session startup, verify `GT_AGENT` is set in the tmux environment. If missing, kill the session immediately and return a clear error explaining the problem. This prevents silent failures if the resolution logic is broken in the future.

## Changes

| File | Change |
|------|--------|
| `config/types.go` | Add `ResolvedAgent` field to `RuntimeConfig` |
| `config/loader.go` | Set `ResolvedAgent` during agent resolution; export `GT_AGENT` from it in startup commands |
| `polecat/session_manager.go` | Validate `GT_AGENT` presence after session start |
| `config/loader_test.go` | Update existing test, add 2 regression tests for role_agents with built-in and custom agents |

## Test plan

- [x] `go test ./internal/config/` — all pass
- [x] `go test ./internal/polecat/` — all pass
- [x] `go test ./internal/witness/` — all pass
- [x] Manual: configure `role_agents.polecat = "opencode"`, run `gt sling`, verify `GT_AGENT=opencode` in tmux env
- [x] Manual: verify witness patrol does not auto-nuke running opencode polecats

### DIFF
--- a/internal/config/loader.go
+++ b/internal/config/loader.go
@@ -936,8 +936,11 @@ func resolveAgentConfigInternal(townRoot, rigPath string) *RuntimeConfig {
 
 	// Backwards compatibility: if Runtime is set directly, use it
 	if rigSettings != nil && rigSettings.Runtime != nil {
-		rc := rigSettings.Runtime
-		return fillRuntimeDefaults(rc)
+		rc := fillRuntimeDefaults(rigSettings.Runtime)
+		if rc.ResolvedAgent == "" {
+			rc.ResolvedAgent = inferAgentName(rc)
+		}
+		return rc
 	}
 
 	// Load town settings for agent lookup
@@ -988,8 +991,11 @@ func resolveAgentConfigWithOverrideInternal(townRoot, rigPath, agentOverride str
 
 	// Backwards compatibility: if Runtime is set directly, use it (but still report agentOverride if present)
 	if rigSettings != nil && rigSettings.Runtime != nil && agentOverride == "" {
-		rc := rigSettings.Runtime
-		return fillRuntimeDefaults(rc), "", nil
+		rc := fillRuntimeDefaults(rigSettings.Runtime)
+		if rc.ResolvedAgent == "" {
+			rc.ResolvedAgent = inferAgentName(rc)
+		}
+		return rc, "", nil
 	}
 
 	// Load town settings for agent lookup
@@ -1439,6 +1445,7 @@ func fillRuntimeDefaults(rc *RuntimeConfig) *RuntimeConfig {
 		Command:       rc.Command,
 		InitialPrompt: rc.InitialPrompt,
 		PromptMode:    rc.PromptMode,
+		ResolvedAgent: rc.ResolvedAgent,
 	}
 
 	// Deep copy Args slice to avoid sharing backing array
@@ -1543,6 +1550,19 @@ func fillRuntimeDefaults(rc *RuntimeConfig) *RuntimeConfig {
 	}
 
 	return result
+}
+
+// inferAgentName determines the agent name from a legacy RuntimeConfig.
+// It mirrors the preset resolution logic in fillRuntimeDefaults:
+// use Provider if set, otherwise Command, falling back to "claude".
+func inferAgentName(rc *RuntimeConfig) string {
+	if rc.Provider != "" {
+		return rc.Provider
+	}
+	if rc.Command != "" {
+		return rc.Command
+	}
+	return "claude"
 }
 
 // GetRuntimeCommand is a convenience function that returns the full command string
--- a/internal/config/loader_test.go
+++ b/internal/config/loader_test.go
@@ -2326,6 +2326,7 @@ func TestFillRuntimeDefaults(t *testing.T) {
 			Env:           map[string]string{"OPENCODE_PERMISSION": `{"*":"allow"}`},
 			InitialPrompt: "test prompt",
 			PromptMode:    "none",
+			ResolvedAgent: "opencode",
 			Session: &RuntimeSessionConfig{
 				SessionIDEnv: "OPENCODE_SESSION_ID",
 			},
@@ -2371,6 +2372,9 @@ func TestFillRuntimeDefaults(t *testing.T) {
 		}
 		if result.Instructions == nil || result.Instructions.File != input.Instructions.File {
 			t.Errorf("Instructions: got %+v, want %+v", result.Instructions, input.Instructions)
+		}
+		if result.ResolvedAgent != input.ResolvedAgent {
+			t.Errorf("ResolvedAgent: got %q, want %q", result.ResolvedAgent, input.ResolvedAgent)
 		}
 	})
 
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -1502,6 +1502,15 @@ func (d *Daemon) restartPolecatSession(rigName, polecatName, sessionName string)
 		_ = d.tmux.SetEnvironment(sessionName, k, v)
 	}
 
+	// Set GT_AGENT in tmux session env so tools querying tmux environment
+	// (e.g., witness patrol) can detect non-Claude agents.
+	// BuildStartupCommand sets GT_AGENT in process env via exec env, but that
+	// isn't visible to tmux show-environment.
+	rc := config.ResolveAgentConfig(d.config.TownRoot, rigPath)
+	if rc.ResolvedAgent != "" {
+		_ = d.tmux.SetEnvironment(sessionName, "GT_AGENT", rc.ResolvedAgent)
+	}
+
 	// Apply theme
 	theme := tmux.AssignTheme(rigName)
 	_ = d.tmux.ConfigureGasTownSession(sessionName, theme, rigName, polecatName, "polecat")


julianknutsen:

### Decision: request_changes

Thanks for tracking this down, @arttttt — the bug analysis is thorough and the failure chain (`role_agents` → missing `GT_AGENT` → witness patrol falls back to Claude process names → auto-nuke) is exactly right. The `ResolvedAgent` field approach is the correct fix, and the propagation through config resolution is well-implemented.

However, both reviewers independently found a critical issue: `GT_AGENT` is set in **process env** (`exec env`) but both consumers — the fail-fast validation and `IsAgentAlive` — read from **tmux session env** (`tmux show-environment`). These are separate namespaces. Please address the items below and re-submit.

---

### Fixes Validated
- **GT_AGENT missing when agent resolved via `role_agents`:** Confirmed correct. Both reviewers independently traced the full failure path end-to-end — `resolveRoleAgentConfigCore` lost the agent name, `GetProcessNames("")` fell back to `["node", "claude"]`, witness patrol nuked the session.
- **`BuildStartupCommandWithAgentOverride` no-override path:** Confirmed correct. Now emits `GT_AGENT` from resolved config.

---

### Blocker: tmux namespace mismatch — `GT_AGENT` never reaches session env

`GT_AGENT` is placed into the `exec env` prefix of the startup command (`loader.go:1748→1772`), which sets **process-level** env vars inherited by the agent binary. But both consumers read **tmux session env** via `tmux show-environment`:

- `IsAgentAlive` (`tmux.go:1544`) calls `GetEnvironment(session, "GT_AGENT")` — tmux session env
- The fail-fast check (`session_manager.go:398`) calls the same `GetEnvironment` — tmux session env

`AgentEnv()` (`env.go:38`) does not include `GT_AGENT`, and `session_manager.Start()` never calls `SetEnvironment` for it. The existing `SetEnvironment` calls (lines 312-322) cover `GT_BRANCH`, `GT_POLECAT_PATH`, `GT_TOWN_ROOT` etc., but not `GT_AGENT`.

Impact: (a) the fail-fast check will find `GT_AGENT=""` for **every** polecat and kill the session; (b) witness patrol still won't see `GT_AGENT` and will continue auto-nuking non-Claude agents.

**Suggested fix:** Add `m.tmux.SetEnvironment(sessionID, "GT_AGENT", runtimeConfig.ResolvedAgent)` in `session_manager.Start()` alongside the existing `SetEnvironment` calls (after line 322). Keep the `exec env` changes — they serve process-internal consumers like `os.Getenv("GT_AGENT")` in `handoff.go:545`.

### Major: `ResolvedAgent` not populated on all resolver branches

`tryResolveFromEphemeralTier` returns `fillRuntimeDefaults(rc)` (`loader.go:1206`) without setting `ResolvedAgent`. If the `SetEnvironment` fix is applied, this path would set `GT_AGENT=""` — benign for Claude agents (falls back to correct detection) but inconsistent and fragile.

**Suggested fix:** Set `ResolvedAgent` before returning from `tryResolveFromEphemeralTier`, or at the call site in `resolveRoleAgentConfigCore`.

### Major: Tests validate command string but not tmux session env

All three new tests check `strings.Contains(cmd, "GT_AGENT=opencode")` — validating process env injection. None verify that `GT_AGENT` reaches tmux session env (the namespace that actually matters for `IsAgentAlive` and the fail-fast check). The fail-fast validation itself has zero test coverage.

**Suggested fix:** After fixing the blocker, add a test verifying `SetEnvironment` is called with `GT_AGENT`. Add at least one test for the fail-fast validation path. Document that `GT_AGENT` must be set in both namespaces for different consumers.

---

### Pre-Merge Checklist
- [ ] Add `SetEnvironment(sessionID, "GT_AGENT", runtimeConfig.ResolvedAgent)` in `session_manager.Start()`
- [ ] Set `ResolvedAgent` in `tryResolveFromEphemeralTier` return path
- [ ] Add test coverage for fail-fast validation and tmux session env propagation

---

arttttt:
Thanks for the thorough review. Addressing each point:

### 1. Blocker: tmux namespace mismatch — fixed

Already addressed in commit 993a8f1b (pushed before this review). Added `SetEnvironment(sessionID, "GT_AGENT", runtimeConfig.ResolvedAgent)` in `session_manager.Start()` after the existing env var block (line ~335). Both namespaces are now covered:
- `exec env` prefix → process env for `os.Getenv("GT_AGENT")` consumers (`handoff.go`, `runtime.go`)
- `tmux set-environment` → session env for `IsAgentAlive` and the fail-fast validation

### 2. Major: `ResolvedAgent` not set in `tryResolveFromEphemeralTier` — fixed

Fixed in 3ac77646. `fillRuntimeDefaults(rc)` now gets `ResolvedAgent = agentName` before returning. Agreed this was fragile — even though cost tiers currently only manage Claude models, the inconsistency would be a latent bug if non-Claude agents were added to tiers.

### 3. Major: Tests validate command string but not tmux session env

Valid observation. The current `session_manager_test.go` tests use real tmux (no mock interface) and the existing `Start` tests only cover edge cases (polecat not found). A full integration test for `Start` that verifies `SetEnvironment` calls would require a real rig, worktree, and git repo setup — that's a significant test infrastructure effort beyond this PR's scope.

The fail-fast validation is tested indirectly: it caught the exact bug during real sling operations (GT_AGENT was empty → session killed → rollback triggered). But I agree a unit test would be better — this is a candidate for follow-up when the session_manager test infrastructure supports it.

### Pre-merge checklist update
- [x] `SetEnvironment(sessionID, "GT_AGENT", ...)` in `session_manager.Start()` — 993a8f1b
- [x] `ResolvedAgent` in `tryResolveFromEphemeralTier` — 3ac77646
- [ ] Test coverage for tmux session env — deferred (requires test infra work)

julianknutsen:
Completely understand on the testing. Auto-reviews love pointing this out, but our architecture has some gaps that make this hard right now. Thanks for the quick iteration.

julianknutsen:

### Decision: request_changes

Thanks for this excellent contribution, @arttttt! The bug you found is real and impactful — non-Claude polecats getting silently auto-nuked by witness patrol is a nasty failure mode, and your root-cause analysis tracing it through the `ResolveRoleAgentConfig` → `BuildStartupCommand` → `IsAgentAlive` chain is thorough. The `ResolvedAgent` field approach and the fail-fast validation are solid design choices. However, the review surfaced a few issues that need attention before merge.

---

### Fixes Validated
- **GT_AGENT not set for role_agents-resolved agents:** Confirmed correct. All resolution paths (`resolveRoleAgentConfigCore`, `resolveAgentConfigInternal`, `tryResolveFromEphemeralTier`) now propagate the agent name via `ResolvedAgent`, and both process env and tmux session env are correctly populated.

---

### Blocker: Legacy `rigSettings.Runtime` path causes hard startup failure

The fail-fast validation in `session_manager.Start()` will kill sessions for rigs that still use the deprecated `rigSettings.Runtime` field. The `resolveAgentConfigInternal` function has an early return at line 938-940 for this legacy path that does NOT set `ResolvedAgent`. The fail-fast check then reads empty `GT_AGENT` from tmux session env and kills the session — turning a previously working polecat startup into a hard failure.

**Suggested fix:** Set `ResolvedAgent` in the legacy `rigSettings.Runtime` path (infer from `rc.Command` or `rc.Provider`, defaulting to `"claude"`). Also apply the same fix to `resolveAgentConfigWithOverrideInternal` at line 988-990. Add a test covering legacy `Runtime` config + fail-fast interaction.

### Major: Custom alias polecats still auto-nuked

When `codex` is configured as a custom alias running `opencode`, `GT_AGENT` is set to `"codex"` and `GetProcessNames("codex")` resolves codex's default process names — not `opencode`. The polecat remains vulnerable to witness auto-nuke in this scenario, which is the motivating use case from the PR description.

**Suggested fix:** Either export the actual runtime process names alongside `GT_AGENT`, derive liveness identity from the resolved command rather than the alias name, or document this as a known limitation if it won't be fixed in this PR.

### Major: Daemon respawn doesn't set GT_AGENT in tmux session env

After a daemon-triggered respawn (`daemon.restartPolecatSession`), the fresh tmux session (via `EnsureSessionFresh`, which wipes env) will have `GT_AGENT` in process env but NOT in tmux session env. Non-Claude polecats respawned by the daemon remain vulnerable.

**Suggested fix:** Add `tmux.SetEnvironment(sessionName, "GT_AGENT", rc.ResolvedAgent)` to `daemon.restartPolecatSession`, mirroring the logic added to `session_manager.Start()`.

### Major: `fillRuntimeDefaults` drops `ResolvedAgent` field

`fillRuntimeDefaults` copies `RuntimeConfig` field-by-field but doesn't include `ResolvedAgent`. Currently `ResolvedAgent` is set after `fillRuntimeDefaults`, but this is a fragile ordering invariant that creates a maintenance trap.

**Suggested fix:** Add `result.ResolvedAgent = rc.ResolvedAgent` to `fillRuntimeDefaults`.

---

### Pre-Merge Checklist
- [ ] Set `ResolvedAgent` in legacy `rigSettings.Runtime` paths (`loader.go:938-940`, `loader.go:988-990`)
- [ ] Add test for legacy `Runtime` config + fail-fast interaction
- [ ] Address custom alias liveness gap (fix or document as known limitation)
- [ ] Add `tmux.SetEnvironment` for GT_AGENT in `daemon.restartPolecatSession`
- [ ] Add `result.ResolvedAgent = rc.ResolvedAgent` to `fillRuntimeDefaults`

---

arttttt:
## Addressed review feedback (commit f288254e)

Fixed 3 of 4 points from the automated review:

### 1. Legacy `rigSettings.Runtime` doesn't set `ResolvedAgent` — FIXED
Both `resolveAgentConfigInternal` and `resolveAgentConfigWithOverrideInternal` now call `inferAgentName(rc)` on the legacy `Runtime` path, deriving the agent name from `Provider` → `Command` → `"claude"` fallback.

### 2. Custom alias `GetProcessNames("codex")` doesn't resolve to opencode — DEFERRED
This is a design limitation: custom aliases are not supported in `GetProcessNames`. Would require a separate mapping layer. Filed as separate work.

### 3. Daemon respawn doesn't set GT_AGENT in tmux session env — FIXED
`restartPolecatSession` now calls `config.ResolveAgentConfig()` and sets `GT_AGENT` via `tmux.SetEnvironment()`, making it visible to `tmux show-environment` queries (used by witness patrol).

### 4. `fillRuntimeDefaults` drops `ResolvedAgent` — FIXED
Added `ResolvedAgent` to the deep copy struct initialization. Also added test coverage in `TestFillRuntimeDefaults/preserves_all_fields`.

julianknutsen:
## Adoption Review

Thanks for the contribution, @arttttt! This is a really well-diagnosed bug — the trace from `role_agents` resolution through `BuildStartupCommand` to `IsAgentAlive`'s fallback behavior clearly explains why non-Claude polecats were getting silently killed. The fail-fast validation in `session_manager.Start()` is a particularly nice addition that will catch future regressions early.

### Review Summary
Decision: approve. Dual-model review (Claude + Codex) confirmed the core fix is correct across all 7 resolution paths. Two findings were identified — a daemon restart path using the wrong resolution function, and an alias/command identity gap — both of which will be addressed in a follow-up PR rather than blocking this contribution.

### Review Findings

| 1 | Behavioral Correctness | blocker | high | claude+codex | `daemon.go:1509` calls `ResolveAgentConfig` instead of `ResolveRoleAgentConfig("polecat", ...)` — restarted polecats get wrong GT_AGENT |
| 2 | Contract & Interface Fidelity | major | medium | codex | GT_AGENT set to alias ("codex") while command is "opencode" — liveness depends on preset ProcessNames completeness |
| 3 | Contract & Interface Fidelity | minor | medium | claude | `resolveAgentConfigWithOverrideInternal` doesn't set `ResolvedAgent` |
| 4 | Test Evidence Quality | minor | high | codex | No test coverage for restart path or liveness integration |

Findings #1 and #3 will be addressed in a follow-up fixup PR. #2 and #4 are tracked for future work.

1 review pass performed.

---

PR #1718, opened 2026-02-18
fix: filter StashCount by current branch for worktree correctness

arttttt:
## How we found this

While working on PR #1714 (GT_AGENT fix for role_agents), we added a fail-fast validation in `session_manager.go` that kills the session if `GT_AGENT` is not set. When this validation fired during `gt sling`, the rollback tried to clean up the spawned polecat — and failed:

```
⚠ Session failed, rolling back spawned polecat furiosa...
  ○ Unhooked bead te-lo57
  Warning: Could not clean up orphaned polecat furiosa: polecat furiosa has uncommitted work: 1 stash(es)
```

Furiosa was freshly allocated. No code was written. The session never even started. Where did the stash come from?

## Root cause

**Git stashes are repo-wide, not per-worktree.** They're stored in `.git/refs/stash` (or in Gas Town's case, `.repo.git/refs/stash`) and shared across all worktrees. `StashCount()` calls `git stash list` which returns every stash in the entire repo.

Investigation revealed the stash belonged to a branch called `temp` that no longer exists:

```
$ cd terminal/polecats/furiosa/terminal
$ git stash list
stash@{0}: WIP on temp: c746c8d Switch candlestick chart to DexScreener candles

$ git branch --show-current
polecat/furiosa/te-lo57@mlso3hda
```

Every polecat worktree saw this orphan stash:

```
$ cd terminal/polecats/slit/terminal
$ git stash list
stash@{0}: WIP on temp: c746c8d Switch candlestick chart to DexScreener candles
```

Confirmed with a standalone test:
```
$ cd /tmp && git init main-repo && cd main-repo
$ echo "x" > f && git add . && git commit -m init
$ echo "y" > f && git stash
$ git worktree add ../wt -b wt-branch
$ cd ../wt && git stash list
stash@{0}: WIP on master: abc1234 init   # ← worktree sees main repo's stash
```

## The bug flow

1. `gt sling` → `SpawnPolecatForSling` → allocates polecat name, creates worktree
2. Formula instantiation → hooks bead to polecat
3. `StartSession` → session start fails (any reason)
4. `rollbackSlingArtifacts` → `cleanupSpawnedPolecat` → `Remove(name, force=true)`
5. `RemoveWithOptions(force=true)` → `CheckUncommittedWork()` → `StashCount()`
6. `git stash list` returns stashes from other worktrees/deleted branches → `StashCount > 0`
7. `force=true` bypasses uncommitted changes but **blocks on stashes** (line 825):
   ```go
   if status.StashCount > 0 || status.UnpushedCommits > 0 {
       return &UncommittedWorkError{...}
   }
   ```
8. Rollback fails → orphaned polecat → blocks subsequent sling attempts

## Impact

This affects **any** `Remove(force=true)` call when any stash exists anywhere in the shared repo:
- Sling rollback on session failure
- `gt polecat nuke --force`
- `gt done` cleanup
- Refinery merge queue processing

One orphan stash from a deleted branch can block force-removal of every polecat in the rig.

## Fix

Filter `StashCount()` results by the current branch name. Git stash reflog messages always include the branch name in the format `"WIP on <branch>:"` or `"On <branch>:"`. By matching against the current worktree's branch, we only count stashes that actually belong to this worktree.

Falls back to counting all stashes when the branch cannot be determined (detached HEAD) — better to over-count than silently lose work.

## Test plan

- [x] `TestStashCount_FiltersByBranch` — verifies:
  - Worktree sees 0 stashes when all stashes belong to other branches
  - Worktree sees 1 after creating its own stash
  - Main repo only counts its own stashes, not worktree stashes
- [x] All existing `internal/git/` tests pass
- [x] `go build ./...` passes

### DIFF
--- a/internal/git/git.go
+++ b/internal/git/git.go
@@ -1082,7 +1082,12 @@ func (g *Git) CountCommitsBehind(ref string) (int, error) {
 	return count, nil
 }
 
-// StashCount returns the number of stashes in the repository.
+// StashCount returns the number of stashes belonging to the current branch.
+// Git stashes are stored in the main repo (.git/refs/stash) and shared across
+// all worktrees. Counting all stashes is incorrect for worktree-based polecats:
+// a fresh polecat worktree would inherit stash count from siblings, blocking
+// Remove(force=true) on work it never created. Filter by current branch name
+// to only count stashes that actually belong to this worktree.
 func (g *Git) StashCount() (int, error) {
 	out, err := g.run("stash", "list")
 	if err != nil {
@@ -1093,13 +1098,26 @@ func (g *Git) StashCount() (int, error) {
 		return 0, nil
 	}
 
-	// Count lines in the stash list
+	// Get current branch to filter stashes.
+	// If we can't determine the branch (detached HEAD, error), count all
+	// stashes as a safe fallback — better to over-count than silently lose work.
+	branch, branchErr := g.CurrentBranch()
+	filterByBranch := branchErr == nil && branch != "" && branch != "HEAD"
+
 	lines := strings.Split(out, "\n")
 	count := 0
 	for _, line := range lines {
-		if line != "" {
-			count++
+		if line == "" {
+			continue
 		}
+		if filterByBranch {
+			// Stash reflog messages use "WIP on <branch>:" or "On <branch>:" format.
+			// Only count stashes that were created on the current branch.
+			if !strings.Contains(line, "on "+branch+":") && !strings.Contains(line, "On "+branch+":") {
+				continue
+			}
+		}
+		count++
 	}
 	return count, nil
 }
--- a/internal/git/git_test.go
+++ b/internal/git/git_test.go
@@ -1310,3 +1310,84 @@ func TestGetPushURL_NoPushURL(t *testing.T) {
 		t.Errorf("GetPushURL = %q, want fetch URL when no push URL configured", got)
 	}
 }
+
+// TestStashCount_FiltersByBranch verifies that StashCount only counts stashes
+// belonging to the current branch, not stashes from other worktrees/branches.
+// Git stashes are repo-wide (stored in .git/refs/stash), so without filtering
+// a worktree would see sibling stashes and block Remove(force=true).
+func TestStashCount_FiltersByBranch(t *testing.T) {
+	t.Parallel()
+	dir := initTestRepo(t)
+	g := NewGit(dir)
+
+	// Create a stash on the default branch
+	if err := os.WriteFile(filepath.Join(dir, "dirty.txt"), []byte("dirty"), 0644); err != nil {
+		t.Fatal(err)
+	}
+	cmd := exec.Command("git", "add", ".")
+	cmd.Dir = dir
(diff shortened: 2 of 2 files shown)


julianknutsen:

### Decision: request_changes

Thanks for the contribution, @arttttt! This is a well-investigated bug — the repo-wide stash sharing across worktrees is a subtle git behavior, and your PR description does an excellent job documenting the root cause and the bug flow through `Remove(force=true)`. The fix approach (filtering by branch name in the reflog message) is sound, and the test creates a real multi-worktree scenario to verify isolation.

Both reviewers independently confirmed the bug is real and the core fix is correct. Two issues need attention before merge.

---

### Fixes Validated
- **Cross-worktree stash pollution in `Remove(force=true)`:** Confirmed correct. Both reviewers traced the full path: `StashCount()` → `CheckUncommittedWork()` → `manager.Remove()` at `manager.go:825` where `status.StashCount > 0` blocks removal even with `force=true`. The branch-filtered counting with detached HEAD fallback is the right approach.

---

### Major: Duplicate stash counting in `getGitState()` not updated

`getGitState()` in `internal/cmd/polecat.go` (around line 887) has its own inline `git stash list` counting that counts ALL stashes without branch filtering — the exact same bug this PR fixes in `StashCount()`. This feeds into recovery status determination (`runPolecatRecoveryStatus()`) and `gt polecat git-state` display, which can produce false "NEEDS_RECOVERY" verdicts and spurious "has N stash(es)" warnings for worktrees with zero stashes of their own.

**Suggested fix:** Replace the inline stash counting in `getGitState()` with `git.NewGit(worktreePath).StashCount()` to ensure branch filtering applies uniformly. Alternatively, this can be filed as a tracked follow-up if you consider it out of scope.

### Major: Substring match can false-positive from commit messages

The `strings.Contains(line, "on "+branch+":")` check in the new filtering logic searches the entire stash line, including the trailing commit message. Git stash list format is `stash@{N}: WIP on <branch>: <short-hash> <commit-message>`. A stash from branch "develop" with commit message "work on fix: edge case" would match when the current branch is "fix". While gastown's polecat-style branch names make this unlikely in practice, it's a correctness gap in the new code.

**Suggested fix:** Anchor the match to the reflog prefix position. Check for `": WIP on "+branch+":"` or `": On "+branch+":"` (adding the `": "` that always precedes the branch name in stash reflog output), or use a regex anchored to the start of the format.

### Minor: Edge cases untested

The new test covers the happy path well (auto-stash isolation across worktrees) but misses: (1) detached HEAD fallback where `CurrentBranch()` returns "HEAD" and all stashes should be counted, (2) custom stash messages (`git stash push -m "msg"`) which produce `On <branch>:` vs `WIP on <branch>:` format.

**Suggested fix:** Add at minimum a detached-HEAD test case. The format variant test is nice-to-have.

---

### Pre-Merge Checklist
- [ ] Anchor `strings.Contains` match to reflog prefix position to prevent commit-message false positives
- [ ] Address `getGitState()` inline stash counting (fix in this PR or create tracked follow-up)
- [ ] Add detached-HEAD fallback test case

---

PR #1725, opened 2026-02-19
fix(tmux): preserve user keybindings instead of overriding them

### LINKED ISSUE #1548: gt install silently overrides user tmux keybindings without consent
## Problem

`gt install` (and/or `gt mayor attach`) silently overrides existing user tmux prefix keybindings without warning or opt-in. This is destructive to users who have carefully configured their tmux environment.

## Affected bindings

Gas Town overwrites at least four prefix-table bindings:

| Binding | Gas Town maps to | Previous user binding |
|---|---|---|
| `prefix g` | `gt agents` | Whatever the user had (common: lazygit, goto, etc.) |
| `prefix a` | `gt feed --window` | Whatever the user had |
| `prefix n` | `gt cycle next` | `next-window` (tmux default) |
| `prefix p` | `gt cycle prev` | `previous-window` (tmux default) |

Additionally, a `MouseDown1StatusRight` root binding is added.

## The specific issue

The bindings include an `if-shell` guard that checks for Gas Town session names (`^(gt|hq)-`), which is good. However, the **else branch** for non-gt sessions displays an error message ("C-b g is for Gas Town sessions only") instead of falling through to the user's original binding. This means the user's binding is destroyed in ALL tmux sessions, not just Gas Town sessions.

## Expected behavior

Gas Town should do one of the following (in order of preference):

1. **Never override user bindings without explicit opt-in** — e.g., `gt install --tmux-bindings` or a config option
2. **If overriding, preserve and chain the original binding** — the else branch of the `if-shell` should invoke the user's original binding, not show an error
3. **At minimum, warn the user** during `gt install` that tmux bindings will be modified, and list which ones

## Additional context

- The bindings appear to be injected from the compiled `gt` binary, not from a user-editable config file, making them impossible to prevent without post-hoc workarounds
- Users who have extensively customized their tmux environment (plugin bindings, custom workflows) lose functionality silently
- `prefix n` and `prefix p` override tmux's own defaults (`next-window` / `previous-window`), which may confuse users who rely on standard tmux navigation

## Environment

- Gas Town: v0.6.0
- tmux: 3.5a
- macOS (Homebrew install)

---
*Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>*

claycantrell:
## Summary
- Capture existing tmux keybindings before overwriting them, and use the original binding as the fallback in non-Gas Town sessions
- Adds `if-shell` guard to `SetMailClickBinding` (previously unconditional)
- Removes "C-b X is for Gas Town sessions only" error messages that blocked user bindings in all sessions

Closes #1548

## Root Cause

Gas Town's keybinding functions (`SetFeedBinding`, `SetAgentsBinding`, `SetCycleBindings`, `SetMailClickBinding`) use global `bind-key` commands that affect ALL tmux sessions, not just Gas Town sessions. While an `if-shell` guard correctly detected non-GT sessions, the else branch for `prefix g` and `prefix a` displayed an error message instead of invoking the user's original binding, effectively destroying it.

`SetMailClickBinding` had no `if-shell` guard at all, unconditionally overriding `MouseDown1StatusRight` in every session.

## Changes

1. **`getKeyBinding()` helper** — Queries `tmux list-keys` to capture the current binding for a key before overwriting it. Detects existing Gas Town bindings (containing "gt ") to prevent recursive `if-shell` wrapping on repeated `ConfigureGasTownSession` calls.

2. **`SetFeedBinding` (C-b a)** — Else branch changed from `display-message 'C-b a is for Gas Town sessions only'` to the user's original binding (or nop if none existed).

3. **`SetAgentsBinding` (C-b g)** — Same fix as above.

4. **`SetCycleBindings` (C-b n/p)** — Now captures user's original bindings instead of hardcoding `next-window`/`previous-window`. Falls back to tmux defaults if no prior binding exists.

5. **`SetMailClickBinding` (MouseDown1StatusRight)** — Added `if-shell` guard matching the pattern used by other bindings. Falls back to user's original binding or nop in non-GT sessions.

## Test plan
- [x] `go build ./...` succeeds
- [x] `go test ./internal/tmux/` — all tests pass
- [x] `TestGetKeyBinding_NoExistingBinding` — returns empty for unbound keys
- [x] `TestGetKeyBinding_CapturesDefaultBinding` — correctly captures `next-window` for prefix-n
- [x] `TestGetKeyBinding_CapturesDefaultBindingWithArgs` — parses multi-word commands (e.g., `choose-tree -Zs`)
- [x] `TestGetKeyBinding_SkipsGasTownBindings` — returns empty for existing GT bindings (requires tmux server)
- [x] `TestGetKeyBinding_CapturesUserBinding` — captures custom user bindings (requires tmux server)
- [ ] Manual: verify C-b g/a/n/p work in GT sessions and fall back correctly in non-GT sessions

### DIFF
--- a/internal/tmux/tmux.go
+++ b/internal/tmux/tmux.go
@@ -1894,6 +1894,10 @@ func IsInsideTmux() bool {
 // MouseDown1StatusRight binding (if any) is preserved.
 // See: https://github.com/steveyegge/gastown/issues/1548
 func (t *Tmux) SetMailClickBinding(session string) error {
+	// Skip if already configured — preserves user's original fallback from first call
+	if t.isGTBinding("root", "MouseDown1StatusRight") {
+		return nil
+	}
 	ifShell := fmt.Sprintf("echo '#{session_name}' | grep -Eq '%s'", sessionPrefixPattern())
 	fallback := t.getKeyBinding("root", "MouseDown1StatusRight")
 	if fallback == "" {
@@ -1974,6 +1978,20 @@ func (t *Tmux) SetTownCycleBindings(session string) error {
 	return t.SetCycleBindings(session)
 }
 
+// isGTBinding checks if the given key already has a Gas Town if-shell binding.
+// Used to skip redundant re-binding on repeated ConfigureGasTownSession calls,
+// preserving the user's original fallback captured on the first call.
+func (t *Tmux) isGTBinding(table, key string) bool {
+	output, err := t.run("list-keys", "-T", table, key)
+	if err != nil || output == "" {
+		return false
+	}
+	// GT bindings use if-shell with a run-shell/display-popup invoking "gt ".
+	// Require both "if-shell" and "gt " to avoid false positives on user
+	// bindings that happen to contain "gt " without the if-shell guard.
+	return strings.Contains(output, "if-shell") && strings.Contains(output, "gt ")
+}
+
 // getKeyBinding returns the current tmux command bound to the given key in the
 // specified key table. Returns empty string if no binding exists or if querying
 // fails. This is used to capture user bindings before overwriting them, so the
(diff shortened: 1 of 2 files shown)


julianknutsen:
## Adoption Review

Thanks for the contribution, @claycantrell! This is a well-thought-out fix for a real usability pain point — Gas Town's keybindings were silently overriding user tmux configurations in all sessions, which is the kind of thing that makes users distrust the tool. Your approach of capturing bindings before overwriting and using if-shell guards is the right pattern.

### Initial Review
The initial dual-model review (Claude + Codex) confirmed all three fix validations as correct: the if-shell guard for SetMailClickBinding, the removal of error-message else-branches, and the dynamic capture of cycle bindings. One major finding was identified: on repeated `ConfigureGasTownSession` calls (6+ call sites in normal operation), `getKeyBinding` would detect the existing GT binding, return empty, and the fallback would silently revert to defaults — losing the user's original binding after the first session creation.

### Maintainer Changes
Added `isGTBinding()` guard to skip re-binding when a GT if-shell guard already exists, preventing loss of user's original fallback on subsequent calls. Also tightened GT binding detection to require both "if-shell" and "gt " (reduces false positives), documented assumed `list-keys` format, added tests for idempotency, and fixed existing tests for GT-active environments. (+115 -6 across 2 files)

### Final Review Status
All clear — both reviewers independently confirmed the fix validations and the idempotency fix as correct. Remaining findings are all minor trade-offs strictly better than pre-PR behavior (stale rig prefix on skip, single-key guard for dual-key bind, list-keys parsing limitations). No blockers or unmitigated majors.

### Review Iterations
2 review passes performed.

---

Issue #1141, opened 2026-01-31
Observability-Aware Agent Orchestration for Gastown


pae23:
## Problem Statement

Gastown orchestrates agents around tasks (Beads) and code (git), but it has no first-class access to **runtime observability signals** such as logs, metrics, traces, or test execution output.

In modern systems—especially **swarms of microservices**—the ground truth does not live in a single repository or a single service. Failures are often:

* distributed across multiple services,
* cascading (one service causes others to fail),
* visible only at runtime (latency, retries, timeouts),
* and surfaced indirectly through end-to-end or integration tests.

Without access to observability data, agents:

* reason from incomplete information,
* confuse symptoms with root causes,
* and create Beads that are misrouted (wrong rig, wrong owner, wrong priority).

In development environments, there is an additional issue:
the same agent often **writes code, runs tests, and declares success**, which weakens validation and increases false positives.

---

## Proposed Solution

Introduce an **Observability Layer** in Gastown, along with a **separation of concerns for test execution in dev**.

### 1) Observability Sources as First-Class Inputs

Allow Gastown to read from **anonymized observability sources**, independently of concrete container names or infrastructure details.

Each observability source declares:

* `scope`: dev or prod
* `service_id`: abstract identifier (e.g. svc-A, svc-B)
* `env_id`: abstract environment (env-dev, env-staging, env-prod)
* `source_kind`: runtime_logs, scheduler_events, traces, metrics, test_output
* `redaction_policy`: masking and minimization rules
* `routing_rules`: how detected issues map to rigs/crews

The observability data is streamed, normalized, fingerprinted, and correlated before being used to create or annotate Beads.

### 2) Dev-Only: Test Runner vs Test Controller

When `scope=dev`, split responsibilities between two agents:

* **Test Runner Agent**: executes tests or launches programs and produces artifacts.
* **Test Controller Agent**: decides what to run, enforces policies (stop-on-fail, retries, flaky handling), validates results, and creates Beads when failures are actionable.

This prevents a single agent from both executing and validating its own work.

---

## Alternatives Considered

* **Relying only on code and static analysis**
  Insufficient for distributed systems where many failures are runtime-only.

* **Pushing observability interpretation entirely to humans**
  Does not scale in agent-driven workflows and breaks the automated loop.

* **Embedding observability logic directly inside each agent**
  Leads to duplicated logic, inconsistent interpretations, and poor cross-service correlation.

* **Single-agent test execution and validation**
[...]

PR #1730, opened 2026-02-19
feat: add observability-aware agent orchestration foundation

claycantrell:
## Summary

Foundation PR for issue #1141 (observability-aware agent orchestration). This adds the **opt-in** infrastructure for reading runtime signals (logs, metrics, traces, test output) and flowing them through the existing feed pipeline. Zero impact on existing users who don't configure it.

### What's in this PR

- **Configuration schema** (`internal/config/types.go`): `ObservabilityConfig`, `ObservabilitySourceConfig`, `ObservabilityRoutingRules`, `TestOrchestrationConfig` types added to `TownSettings` with `omitempty` — invisible when not configured
- **Event types** (`internal/events/events.go`): 7 new `observe_*` constants and 3 payload helpers (`ObserveLogPayload`, `ObserveTestPayload`, `ObserveSourcePayload`)
- **`internal/observe/` package**: File-tailing `Source` implementing `feed.EventSource`, plus a `Redactor` with three policy levels (`none`/`standard`/`strict`)
- **`gt observe` command group**: `list`, `status`, `add`, `remove`, `tail` subcommands for managing observability sources
- **Feed integration**: observe events display in TUI (`buildEventMessage`), curator summaries (`generateSummary`), and optional source creation in `runFeedTUI`
- **Beads exempt**: `observe` added to `beadsExemptCommands` in `root.go`

### Bugs caught and fixed during review

- **Race condition in `Source.Close()`**: `file.Close()` could race with the tail goroutine still reading via `bufio.Scanner`. Fixed with `sync.WaitGroup` — `Close()` now waits for the goroutine to exit before closing the file
- **Double-close panic**: Added `sync.Once` guard so `Close()` is idempotent
- **Scanner buffer overflow**: Default `bufio.Scanner` (64KB) would permanently stop reading on long lines. Set 1MB buffer and recover by re-creating the scanner on error
- **`inferSeverity` silently dropped unrecognized lines**: Lines with no severity keyword (e.g., `"just a plain line"`) were classified as `"info"` and then dropped by `warn`/`error` thresholds. Now returns empty string for unknown severity, which always passes filters
- **PII leak via `Event.Raw` field**: `Raw` stored the original un-redacted line. Now stores the redacted version
- **Silent failures in feed integration**: Failed observe sources were silently skipped. Now logs warnings to stderr so users can diagnose misconfigured paths

### Test coverage (47 test cases)

**`internal/observe/`** (29 tests):
- Source: tails new lines, applies redaction, Raw field is redacted, severity filtering, unknown severity passes through, close stops goroutine deterministically, double-close safe, missing file, nil config, empty path, event type/actor, no-service-id fallback, default redaction policy
- Redactor: none/standard/strict policies, unknown policy treated as none, multiple PII on same line, version string behavior, unicode content, empty string, long lines
- `inferSeverity`: error/warn/debug/info detection, no-severity returns empty, empty input
- `passesSeverityFilter`: nil rules, empty threshold, unknown/unrecognized severity, invalid threshold, all rank comparisons, case insensitivity

**`internal/cmd/`** (18 tests):
- List: empty, populated, Observability-not-nil-but-Sources-nil edge case
- Add: 10 validation cases (valid + invalid scope/kind/redaction/empty), duplicate detection, nil container initialization
- Remove: basic, cleanup of empty containers, preserves TestOrchestration when removing last source, nonexistent source
- Config: JSON round-trip, omitempty behavior, file-based save/load round-trip, `DefaultObservabilitySourceConfig` defaults

**`internal/feed/`** (10 new cases in existing `TestCurator_GeneratesSummary`):
- observe_log with source_id, observe_log without source_id, observe_log nil payload
- observe_test with name+status, observe_test nil payload
- observe_source_up/down with and without source_id
- observe_metric with source_id

### What's deferred to follow-up PRs

Per #1141, the following are **not in scope** for this foundation PR:
- **Agent orchestration wiring** for test runner/controller roles (schema is defined but not wired)
- **Structured parsers** for metric, trace, and test_output source kinds (currently all treated as log lines)
- **Cross-service correlation engine**
- **Automatic bead creation** from observability signals
- Broader severity inference (currently substring-based; structured log format detection deferred)

### Known limitations documented in tests

- IP regex matches standalone quad-dotted version numbers like `3.2.1.0` at word boundaries (conservative: better to over-redact). Version-prefixed strings like `v3.2.1.0` do NOT match
- `inferSeverity` uses substring matching which can false-positive on content words (e.g., "the ERROR_CODE table")
- Strict redaction mode catches Unix timestamps (10+ digit numbers) — documented trade-off for strict policy

## Test plan

- [x] `go build ./...` compiles cleanly
- [x] `go test ./internal/observe/...` — 29 tests pass
- [x] `go test ./internal/cmd/...` — all tests pass including 18 new observe tests
- [x] `go test ./internal/feed/...` — all tests pass including 10 new curator summary cases
- [x] `go test ./internal/config/...` — existing tests pass, no regressions
- [ ] Manual: `gt observe list` with no config prints "No observability sources configured"
- [ ] Manual: `gt observe add test-logs --path /tmp/test.log --kind log --service-id svc-a` adds source
- [ ] Manual: `gt observe tail test-logs` shows events as lines are appended

### DIFF
--- a/internal/cmd/observe_list.go
+++ b/internal/cmd/observe_list.go
@@ -66,7 +66,7 @@ func runObserveList(cmd *cobra.Command, args []string) error {
 		}
 		fmt.Fprintf(w, "%s\t%s\t%s\t%s\t%s\n", name, scope, kind, svc, src.Path)
 	}
-	w.Flush()
+	_ = w.Flush()
 
 	return nil
 }
--- a/internal/cmd/observe_status.go
+++ b/internal/cmd/observe_status.go
@@ -60,7 +60,7 @@ func runObserveStatus(cmd *cobra.Command, args []string) error {
 		}
 		fmt.Fprintf(w, "%s\t%s\t%s\t%s\n", name, status, modTime, src.Path)
 	}
-	w.Flush()
+	_ = w.Flush()
 
 	return nil
 }
--- a/internal/cmd/observe_tail.go
+++ b/internal/cmd/observe_tail.go
@@ -59,7 +59,7 @@ func runObserveTail(cmd *cobra.Command, args []string) error {
 	if err != nil {
 		return fmt.Errorf("creating source %q: %w", name, err)
 	}
-	defer src.Close()
+	defer func() { _ = src.Close() }()
 
 	fmt.Fprintf(os.Stderr, "Tailing %s (%s)... press Ctrl-C to stop\n", name, srcCfg.Path)
 


steveyegge:
Thanks for the thorough implementation and great test coverage, Clay! The self-caught bugs (race condition, double-close, PII leak) show real attention to quality.

After evaluating this against the existing codebase, we think the current feed/patrol/costs infrastructure already covers gastown's observability needs today. The feed system already tails events in real-time, the curator handles dedup/aggregation, and patrol monitors agent health.

The vision in #1141 (multi-service observability, distributed trace correlation) is interesting longer-term, but gastown isn't orchestrating across production microservices yet. If we do need to consume external signals in the future, we'd likely extend the existing feed pipeline rather than build a parallel system.

We'll keep #1141 open as a future direction marker. If a concrete, narrower need comes up, we'd love a targeted PR extending the feed system. Thanks for the contribution!

PR #1615, opened 2026-02-17
fix(daemon): rewrite convoy manager for multi-rig event polling and lazy store init

l0g1x:
> **Depends on:** [beads#1815](https://github.com/steveyegge/beads/pull/1815) — adds `beads.OpenFromConfig(ctx, beadsDir)`. This PR's `go.mod` has a temporary `replace` directive pointing to the fork. Once beads#1815 merges, the `replace` should be removed and the beads version bumped.

## Summary

This PR improves convoy reliability by replacing the subprocess-based `ConvoyWatcher` with a SDK-based `ConvoyManager` that polls all beads stores, centralizing convoy observation in the daemon and removing the scattered observer call sites from the witness and refinery.

## Problem: Convoy reliability

The upstream convoy system has several architectural limitations that compound to make convoy progression unreliable:

**1. Single-store event visibility.** The `ConvoyWatcher` tails events via a `bd activity --follow --town` subprocess, which only watches the town-level (hq) beads store. Close events for rig-level beads (`gt-*`, `bd-*`, `sh-*`) live in each rig's own store and are invisible to the watcher. A convoy tracking `gt-issue1` can only auto-progress if the close event happens to appear in hq — which it doesn't. The only fallback is the stranded scan, which runs every 2 minutes, making convoy advancement slow and unpredictable.

**2. Fragile subprocess event tailing.** The watcher spawns a long-running `bd activity --follow` process and reads its stdout via pipe. Any interruption — Dolt server restart, OOM, broken pipe — kills the stream. The watcher retries after 5 seconds, but events during the gap are silently dropped (there's no high-water mark or replay). Every Dolt restart produces a repeating error loop in daemon logs until `bd activity` can reconnect.

**3. Scattered observation with a latent bug.** To compensate for the watcher's limited visibility, 6 `CheckConvoysForIssue` call sites are spread across the witness (5 in `handlers.go`) and refinery (1 in `engineer.go`). These are pure side-effect hooks that don't influence their host's control flow — convoy tracking is an implicit concern of components whose core job is polecat lifecycle and merge queue management. The refinery call site has a latent bug: it passes `e.rig.Path` (a rig worktree path) instead of the town root, so the convoy lookup is called with the wrong directory and silently fails every time.

[...]

PR #1759, opened 2026-02-19
fix(convoy): add type filtering, blocks dep checking, and dispatch failure iteration

l0g1x:
## Summary

The convoy-manager-rewrite (merged in PR #1615) introduced event-driven feeding that dispatches convoy issues within 5 seconds of a close event. The feeder lacks type and dependency guards, so it eagerly dispatches epics and blocked tasks that the old slow Deacon patrol path never reached. Additionally, batch sling (`gt sling bead1 bead2 ... rig`) creates a separate convoy per bead instead of one convoy tracking the whole batch, never persists ConvoyID on the beads, and requires the rig to be specified explicitly even though it can be inferred from bead prefixes. This PR adds safety guards to the feeder, fixes batch sling convoy handling, and adds rig auto-resolution.

## Problem: convoy feeder dispatches issues it shouldn't

The event-driven feeder in `feedNextReadyIssue` (operations.go:141) and the stranded scan feeder in `feedFirstReady` (convoy_manager.go:308) both select issues based solely on `status == "open" && assignee == ""`. This creates three problems:

**1. Epics get slung to polecats.** Any open, unassigned issue tracked by a convoy is eligible for dispatch -- including epics, sub-epics, and other container types that have no meaningful polecat work. Before the rewrite, these rarely surfaced because the old witness-based feed path was slow enough that epics typically got assigned or closed before the Deacon patrol reached them.

**2. Blocked tasks dispatch immediately.** Issues with `blocks` or `conditional-blocks` dependencies are dispatched without checking whether their blockers have closed. A task that depends on another task completing first gets slung within 5 seconds of the convoy receiving a close event, regardless of which issue actually closed. The feeder has no concept of execution ordering.

**3. First dispatch failure stalls the convoy for 30 seconds.** `feedFirstReady` (stranded scan path, convoy_manager.go:312) only tries `ReadyIssues[0]`. If that issue targets a parked rig, has no route, or the sling command fails, the entire convoy waits until the next stranded scan cycle (30s). The event-driven path (`feedNextReadyIssue`) has the same problem -- on dispatch failure it logs the error and returns, abandoning the remaining ready issues.

## Problem: batch sling creates N convoys instead of 1

**4. One convoy per bead in batch mode.** `runBatchSling` (sling_batch.go) calls `createAutoConvoy` inside the per-bead loop, creating a separate convoy for each bead. When a user runs `gt sling bead1 bead2 bead3 rig`, they get 3 independent convoys instead of 1 convoy tracking all 3 beads. This defeats the purpose of batch slinging -- the convoy manager treats each bead as an independent unit with no shared lifecycle. @steveyegge [confirmed](https://discord.com/channels/1462817445562814505/1473745882120720557/1473895486728700129) the expected behavior is a single convoy for multiple beads slung at once.

**5. ConvoyID never stored on beads in batch mode.** The `beadFieldUpdates` struct in the batch loop never sets `ConvoyID` or `MergeStrategy`. When `gt done` runs, `getConvoyInfoFromIssue` (the fast path) reads these fields from the bead -- but they're empty, so the fast path fails and falls back to the slow convoy scan. In single-sling mode these fields are set correctly; the omission is batch-specific.

## Problem: batch sling requires explicit rig despite having prefix information

**6. Rig must be specified explicitly.** `gt sling gt-task1 gt-task2 gt-task3 gastown` requires the user to type `gastown` even though every bead's `gt-` prefix already maps to `gastown` via `routes.jsonl`. The infrastructure for prefix-to-rig resolution exists (`beads.GetRigNameForPrefix`) and is used by the convoy manager and deacon redispatch, but sling never uses it for target resolution.

### Type filtering via `IsSlingableType` (addresses #1)

Adds a `slingableTypes` allowlist (`task`, `bug`, `feature`, `chore`, and empty string for beads that default to task). The exported `IsSlingableType` function is called in two places:

- `feedNextReadyIssue` (event-driven path) -- skips issues with non-slingable types before attempting dispatch
- `findStrandedConvoys` in cmd/convoy.go (stranded scan path) -- excludes non-slingable types from the ready issues list before the convoy manager ever sees them

The `trackedIssue` struct and `getConvoyTrackedIssues` now thread `IssueType` through from the beads store so the filter has data to work with.

### Blocks dependency checking via `isIssueBlocked` (addresses #2)

Adds `isIssueBlocked` which calls `store.GetDependenciesWithMetadata` and checks for unclosed `blocks`, `conditional-blocks`, or `waits-for` dependencies. `parent-child` is intentionally excluded -- a child task is dispatchable even if its parent epic is open, consistent with `bd ready` and molecule step behavior.

Called in `feedNextReadyIssue` after the type filter, before rig resolution. Fail-open on store errors (assumes not blocked) to avoid stalling convoys on transient Dolt issues.

Note: the stranded scan path in cmd/convoy.go already has its own blocked check via `isReadyIssue` which reads `t.Blocked` from issue details. The `isIssueBlocked` function here covers the event-driven path which operates on the lighter `trackedIssue` data from convoy dependencies.

### Dispatch failure iteration (addresses #3)

Both feed paths now iterate past failures:

- `feedNextReadyIssue`: on dispatch failure, `continue` to try the next ready issue instead of returning
- `feedFirstReady`: restructured from single-attempt (`ReadyIssues[0]`) to `for range ReadyIssues` with `continue` on each skip/failure condition and `return` only on first successful dispatch. Logs "no dispatchable issues (all N skipped)" when every issue fails.

### Single convoy for batch sling (addresses #4, #5)

Replaces the per-bead `createAutoConvoy` call with a new `createBatchConvoy` function called once before the per-bead loop:

- Creates exactly one convoy with title "Batch: N beads to \<rig\>" and adds `tracks` deps for all beads
- Stores `batchConvoyID` and `MergeStrategy` in each bead's `beadFieldUpdates`, fixing the `gt done` fast-path lookup
- Pre-loop conflict check: if any bead is already tracked by a convoy, errors with rich diagnostics (`printConvoyConflict`) showing all beads in the conflicting convoy with statuses and 4 recommended courses of action
- Empty convoy cleanup: if all beads fail to sling, closes the convoy with reason "all beads failed to sling"

### Rig auto-resolution for batch sling (addresses #6)

Adds `resolveRigFromBeadIDs()` which resolves the target rig from bead prefixes via `routes.jsonl`. Batch mode detection in `sling.go` now handles two patterns:

- **`gt sling gt-abc gt-def gt-ghi`** (no rig) -- auto-resolves rig from prefixes. All beads must resolve to the same rig.
- **`gt sling gt-abc gt-def gt-ghi gastown`** (explicit rig) -- still works, prints deprecation warning: "gt sling now auto-resolves the rig from bead prefixes. You no longer need to explicitly specify \<gastown\>."

Error cases with suggested actions:
- Mixed prefixes resolving to different rigs -- lists each bead's resolved rig, suggests slinging per-rig batches separately
- Unmapped prefix -- suggests checking `routes.jsonl` or specifying rig explicitly
- Town-level prefix (path=".") -- suggests creating bead from the target rig directory
- Cross-rig mismatch (explicit rig) -- batch-specific error with options to remove the bead, sling it separately, or `--force`

### Additional: convoy agent skill and documentation

- Added `docs/skills/convoy/SKILL.md` -- convoy agent skill covering architecture, safety guards, CLI commands, batch sling behavior, rig resolution, test commands, test patterns, and key invariants
- Added `docs/design/convoy/testing.md` -- test plan for daemon convoy management (failure modes, coverage gaps, harness scorecard)
- Added `docs/design/convoy/daemon.md` -- convoy manager flow diagram
- Updated `docs/design/convoy/roadmap.md` -- reflects merged rewrite, PR status
- Updated `docs/design/convoy/convoy-lifecycle.md` -- batch sling section updated for single-convoy, auto-rig-resolution, and deprecation
- Updated `docs/design/convoy/spec.md` -- fixed 9 stale `observer.go` references → `operations.go`, corrected intervals, removed deleted functions
- Updated `docs/concepts/convoy.md` -- corrected `gt convoy add` as implemented (was incorrectly marked "not yet implemented")

### Review fixes (commit `8c8d1a8`) — addresses [review feedback](https://github.com/steveyegge/gastown/pull/1759#issuecomment-3931080119)

Six findings from automated triple-model review (Claude + Codex + Gemini), all resolved:

| Finding | Severity | Fix |
|---------|----------|-----|
| ConvoyID stamped on beads not tracked by convoy on partial dep failure | Major | `createBatchConvoy` now returns `(string, []string, error)` — the tracked bead set. `runBatchSling` only stamps ConvoyID/MergeStrategy on beads in that set. |
| `--force` suggestion in mixed-rig error is unreachable | Major | Error message now suggests specifying an explicit rig (`gt sling gt-a bd-c <rig>`) instead of `--force`. |
| Convoy creation failure silently regresses to no-convoy behavior | Minor | Convoy creation failure is now a hard error that aborts the batch. Message suggests `--no-convoy` if the user wants to proceed without convoy tracking. |
| `TestIsIssueBlocked_BlockedByOpenBlocker` has no assertion | Minor | Added real assertion with `t.Skip` fallback for embedded Dolt limitation. |
| `append(beadIDs, rigName)` mutates shared backing array | Minor | Replaced with `make`+`copy` to avoid slice aliasing. |
| `TestIsIssueBlocked_NoStore` has empty body | Nit | Added nil store guard in `isIssueBlocked` + test verifying fail-open on nil store. |

## Key design decisions

| Decision | Rationale |
|----------|-----------|
| `parent-child` is not blocking | Consistent with `bd ready`, beads SDK, and molecule step dispatch. Epics are organizational containers, not execution prerequisites. |
| Empty IssueType treated as slingable | Beads default to type "task" when IssueType is unset. Treating empty as non-slingable would break all legacy beads. |
| Fail-open on `isIssueBlocked` store errors | A transient Dolt error should not permanently stall a convoy. The next feed cycle will retry with fresh store state. |
| No capacity plumbing in this PR | `isRigAtCapacity` callback adds a parameter to every function in the call chain with no runtime effect until Phase 2 commands exist. Deferred to reduce review surface. |
| Batch conflict is a hard error | A bead already tracked by another convoy is an error, not a skip. This prevents silent duplicate tracking and gives the user clear options to resolve the conflict. |
| Partial dep tracking with precise metadata | If `bd dep add` fails for one bead during batch convoy creation, the convoy still tracks the other beads. ConvoyID is only stamped on beads that were successfully tracked, preventing `gt done` from referencing a convoy with no knowledge of the bead. |
| Deprecation warning, not removal | Explicit rig arg still works to avoid breaking existing scripts. Warning nudges users toward the simpler auto-resolve form. |
| Auto-resolve requires all beads same rig | Mixed-rig batch sling is an error, not a split. Users should be explicit about cross-rig operations. |

## Testing

```bash
go test ./internal/convoy/... ./internal/daemon/... ./internal/cmd/... -count=1
```

New tests:

| Test | What it proves |
|------|---------------|
| `TestIsSlingableType` | Allowlist correctly identifies task/bug/feature/chore/empty as slingable, rejects epic/convoy/sub-epic/decision/unknown |
| `TestReadyIssueFilterLogic_SkipsNonSlingableTypes` | Type filter integrated into the ready-issue selection predicate skips epics and convoys |
| `TestFeedFirstReady_IteratesPastDispatchFailure` | When first sling fails, feedFirstReady continues to next issue and dispatches it |
| `TestFeedFirstReady_AllIssuesFail_LogsNoneDispatchable` | When all issues fail to dispatch, logs the "no dispatchable issues" summary with count |
| `TestCreateBatchConvoy_CreatesOneConvoyTrackingAllBeads` | Core contract: exactly 1 `bd create` + N `bd dep add` commands for N beads |
| `TestCreateBatchConvoy_OwnedLabel` | `--owned` flag propagates `gt:owned` label to the convoy |
| `TestCreateBatchConvoy_MergeStrategyInDescription` | Merge strategy appears in convoy description |
| `TestCreateBatchConvoy_EmptyBeadIDs` | Returns error when called with no beads |
| `TestCreateBatchConvoy_TitleIncludesBeadCount` | Convoy title matches "Batch: N beads to \<rig\>" format |
| `TestCreateBatchConvoy_PartialDepFailureContinues` | One dep add failure doesn't abort -- other beads still tracked |
| `TestBatchSling_ConvoyIDStoredInBeadFieldUpdates` | ConvoyID and MergeStrategy are set in each bead's field updates |
| `TestBatchSling_ErrorsOnAlreadyTrackedBead` | Pre-loop conflict check detects already-tracked bead and errors |
| `TestBatchSling_EmptyConvoyCleanupOnAllFailures` | When all beads fail, convoy is closed with cleanup reason |
| `TestAllBeadIDs_TrueWhenAllBeadIDs` | Syntactic check correctly distinguishes bead IDs from rig names and paths |
| `TestResolveRigFromBeadIDs_AllSamePrefix` | All beads with same prefix resolve to correct rig |
| `TestResolveRigFromBeadIDs_MixedPrefixes_Errors` | Beads from different rigs error with suggested actions |
| `TestResolveRigFromBeadIDs_UnmappedPrefix_Errors` | Unmapped prefix errors with diagnostic info |
| `TestResolveRigFromBeadIDs_TownLevelPrefix_Errors` | Town-level prefix (path=".") errors correctly |

Coverage tests (targeting the ~152 uncovered patch lines flagged by Codecov):

**operations_test.go (15 tests):**

| Test | What it proves |
|------|---------------|
| `TestBlockingDepTypes_ContainsExpectedTypes` | `blockingDepTypes` includes blocks, conditional-blocks, waits-for |
| `TestBlockingDepTypes_ExcludesParentChild` | `parent-child` is not in the blocking set |
| `TestBlockingDepTypes_ExactSize` | No unexpected dep types in the map |
| `TestIsIssueBlocked_NoDeps` | Issue with no dependencies is not blocked |
| `TestIsIssueBlocked_BlockedByOpenBlocker` | Open `blocks` dependency correctly blocks dispatch |
| `TestIsIssueBlocked_NotBlockedByClosedBlocker` | Closed `blocks` dependency does not block |
| `TestIsIssueBlocked_ParentChildDoesNotBlock` | `parent-child` dependency never blocks (organizational only) |
| `TestIsIssueBlocked_FailOpenOnNonexistentIssue` | Store error returns not-blocked (fail-open) |
| `TestRigForIssue_ValidPrefix` | Known prefix resolves to correct rig |
| `TestRigForIssue_EmptyPrefix` | Empty prefix returns error |
| `TestRigForIssue_EmptyIssueID` | Empty issue ID returns error |
| `TestRigForIssue_UnknownPrefix` | Unknown prefix returns error |
| `TestRigForIssue_NoRoutesFile` | Missing routes file returns error |
| `TestRigForIssue_TownLevelPrefix` | Town-level prefix (path=".") returns error |

**sling_batch_test.go (14 tests):**

| Test | What it proves |
|------|---------------|
| `TestSlingGenerateShortID_Format` | Short ID is 8 hex characters |
| `TestSlingGenerateShortID_Unique` | 1000 generated IDs are all unique |
| `TestConvoyInfo_IsOwnedDirect` | 5 table-driven cases: nil receiver, owned+direct, owned+mr, not-owned+direct, not-owned+empty |
| `TestCreateAutoConvoy_BasicSuccess` | Happy path creates convoy and adds dep |
| `TestCreateAutoConvoy_FlagLikeTitleReturnsError` | Flag-like titles (starting with `-`) are rejected |
| `TestCreateAutoConvoy_OwnedLabel` | `--owned` flag adds `gt:owned` label |
| `TestCreateAutoConvoy_DepFailCleansUpOrphan` | Failed dep add closes the orphaned convoy |
| `TestConvoyTracksBead_ExactMatch` | Bead found as exact `tracks` dependency |
| `TestConvoyTracksBead_ExternalWrappedMatch` | Bead found in `external(...)` wrapper |
| `TestConvoyTracksBead_NoMatch` | Non-tracked bead returns false |
| `TestConvoyTracksBead_BdError` | Store error returns false (fail-open) |
| `TestBatchSling_CrossRigGuardRejectsPrefix` | Cross-rig mismatch is detected and rejected |

**Review fix tests (commit `8c8d1a8`):**

| Test | What it proves |
|------|---------------|
| `TestCreateBatchConvoy_ReturnsTrackedBeadSet` | Only successfully-tracked bead IDs returned; failed deps excluded |
| `TestBatchSling_DoesNotSuggestForce` | Mixed-rig error suggests explicit rig, not `--force` |
| `TestBatchSling_ConvoyCreationFailureIsHardError` | Convoy creation failure aborts batch with `--no-convoy` hint |
| `TestBatchSling_SliceAliasingInCrossRigGuard` | `append` in cross-rig path doesn't corrupt original slice |
| `TestIsIssueBlocked_BlockedByOpenBlocker` | _(updated)_ Real assertion that open blocker returns blocked=true |
| `TestIsIssueBlocked_NoStore` | _(updated)_ Nil store guard returns not-blocked (fail-open) |

All existing tests pass unchanged -- the behavioral changes are backward-compatible (previously-dispatched issue types still dispatch, iteration is strictly more permissive than single-attempt, explicit rig arg still accepted with deprecation warning).

### DIFF
--- a/internal/formula/formulas/mol-polecat-work.formula.toml
+++ b/internal/formula/formulas/mol-polecat-work.formula.toml
@@ -34,7 +34,7 @@ for step details — do NOT read this file directly.
 | base_branch | sling vars | The base branch to rebase on (default: main) |
 | setup_command | rig config | Setup/install command (e.g., `pnpm install`). Empty = skip. |
 | typecheck_command | rig config | Type check command (e.g., `tsc --noEmit`). Empty = skip. |
-| test_command | rig config | Test command (default: `go test ./...`) |
+| test_command | rig config | Test command. Empty = skip. Rig must configure for its language. |
 | lint_command | rig config | Lint command (e.g., `eslint .`). Empty = skip. |
 | build_command | rig config | Build command (e.g., `go build ./...`). Empty = skip. |
 
@@ -531,7 +531,7 @@ default = ""
 
 [vars.test_command]
 description = "Command to run tests (auto-detected from rig settings)"
-default = "go test ./..."
+default = ""
 
 [vars.lint_command]
 description = "Command to run linting. Empty = skip."
--- a/internal/formula/formulas/mol-sync-workspace.formula.toml
+++ b/internal/formula/formulas/mol-sync-workspace.formula.toml
@@ -32,7 +32,7 @@ Polecats can file a bead and delegate. Refinery must resolve inline.
 | setup_command | config | Setup/install command (empty = skip) |
 | typecheck_command | config | Type check command (empty = skip) |
 | lint_command | config | Lint command (empty = skip) |
-| test_command | config or default | Test command (default: `go test ./...`) |
+| test_command | rig config | Test command. Empty = skip. Rig must configure for its language. |
 | build_command | config | Build command (empty = skip) |
 
 ## Failure Modes
@@ -515,7 +515,7 @@ default = ""
 
 [vars.test_command]
 description = "Command to run tests (auto-detected from rig settings)"
-default = "go test ./..."
+default = ""
 
 [vars.build_command]
 description = "Command to run build. Empty = skip."


julianknutsen:

### Decision: request_changes

Thanks for this thorough contribution, @l0g1x! The convoy feeder safety guards are a real need — the event-driven feeding introduced in the rewrite (#1615) clearly needs type and dependency gating to prevent epics and blocked tasks from dispatching. The batch sling single-convoy fix and rig auto-resolution are solid quality-of-life improvements too. The PR description and test coverage are exceptionally detailed.

Two contract-level issues surfaced during review that should be addressed before merge. Please take a look at the items below — not all may require changes depending on your read of the severity.

---

### Fixes Validated
- **Epics dispatched to polecats:** Confirmed correct. `IsSlingableType` allowlist applied in both event-driven and stranded-scan paths.
- **Blocked tasks dispatch immediately:** Confirmed correct. `isIssueBlocked` checks `blocks`, `conditional-blocks`, `waits-for` with fail-open semantics.
- **First dispatch failure stalls convoy:** Confirmed correct. Both paths now loop with `continue` on failure.
- **One convoy per bead in batch mode:** Confirmed correct. `createBatchConvoy` called once before loop.
- **ConvoyID never stored in batch mode:** Incomplete — stored in happy path, but partial dep failure leaves metadata inconsistent (see finding below).
- **Batch requires explicit rig:** Confirmed correct. `resolveRigFromBeadIDs` infers rig from prefixes.

---

### Major: ConvoyID stamped on beads not tracked by convoy

`createBatchConvoy` tolerates per-bead `dep add` failures (partial tracking is acceptable), but `runBatchSling` stamps `ConvoyID` on all successful beads regardless. This means a bead can reference a convoy that doesn't actually track it, causing `gt done` fast-path to find a convoy with no knowledge of the bead.

**Suggested fix:** Return the set of successfully-tracked bead IDs from `createBatchConvoy` and only write `ConvoyID`/`MergeStrategy` for beads in that set.

### Major: `--force` suggestion in mixed-rig auto-resolve error is unreachable

The mixed-prefix error in `resolveRigFromBeadIDs` suggests using `--force`, but `resolveRigFromBeadIDs` runs before `--force` is checked in `runBatchSling`. Users following the CLI's own guidance get the same error again.

**Suggested fix:** Either check `slingForce` before calling `resolveRigFromBeadIDs`, or change the suggestion to require an explicit rig: `gt sling gt-a bd-c <rig> --force`.

### Minor: Convoy creation failure silently regresses to pre-PR behavior

When `createBatchConvoy` fails, batch proceeds with empty `batchConvoyID`. Every bead gets empty `ConvoyID` — the exact problem this PR fixes for the happy path. The warning is easily missed.

**Suggested fix:** Either make convoy creation failure a hard error, or escalate the warning to explain the consequence.

### Minor: `TestIsIssueBlocked_BlockedByOpenBlocker` has no assertion

The most important behavioral test for the blocking check logs the result but never asserts. A regression that always returns `false` would pass the test suite silently.

**Suggested fix:** Add `t.Skipf` when result doesn't match expected (embedded Dolt limitation), or a proper assertion when `GetDependenciesWithMetadata` succeeds.

### Minor: `append(beadIDs, rigName)` mutates shared backing array

In the explicit-rig path, `beadIDs = args[:len(args)-1]` shares the backing array with `args`. The `append(beadIDs, rigName)` writes into shared memory. Works today by coincidence (immediate return), fragile to future refactoring.

**Suggested fix:** Copy before appending: `allArgs := append([]string{}, beadIDs...)`.

### Nit: `TestIsIssueBlocked_NoStore` has empty body

Empty test body provides zero coverage. Either delete it or add a minimal call with mock/error verification.

---

### Pre-Merge Checklist
- [ ] Return successfully-tracked bead IDs from `createBatchConvoy`; only stamp `ConvoyID` on those beads
- [ ] Fix `--force` suggestion in mixed-rig error to be actionable
- [ ] Escalate convoy-creation-failure warning or make it a hard error
- [ ] Add assertion or explicit skip to `TestIsIssueBlocked_BlockedByOpenBlocker`
- [ ] Fix slice aliasing in `runBatchSling` error path

---

julianknutsen (changes requested):
Two contract-level issues found: (1) ConvoyID stamped on beads not tracked by convoy on partial dep failure, (2) --force suggestion in mixed-rig error is unreachable. See detailed review comment above.

l0g1x:
Thanks for the thorough review, @julianknutsen — all six findings addressed in `8c8d1a8`.

---

### Major: ConvoyID stamped on beads not tracked by convoy

Fixed. `createBatchConvoy` now returns `(string, []string, error)` where the second value is the set of bead IDs that were successfully added as `tracks` deps. `runBatchSling` only stamps `ConvoyID` and `MergeStrategy` on beads in that tracked set. Beads where `dep add` failed get no convoy metadata — they'll fall through to the slow convoy scan path in `gt done` (which is correct, since the convoy may or may not actually track them).

New test: `TestCreateBatchConvoy_ReturnsTrackedBeadSet` — verifies that failed deps are excluded from the returned set. Five existing `createBatchConvoy` tests updated for the 3-value return signature.

### Major: `--force` suggestion in mixed-rig error is unreachable

Fixed. The mixed-rig error message now suggests specifying an explicit rig (`gt sling gt-a bd-c <rig>`) instead of the unreachable `--force` flag.

New test: `TestBatchSling_DoesNotSuggestForce` — verifies the error output contains no `--force` suggestion.

### Minor: Convoy creation failure silently regresses

Fixed. Convoy creation failure is now a hard error that aborts the batch. The error message suggests `--no-convoy` if the user wants to proceed without convoy tracking. This is strictly better than the silent regression — the user sees exactly what happened and has an explicit opt-out.

New test: `TestBatchSling_ConvoyCreationFailureIsHardError` — verifies abort + hint.

### Minor: `TestIsIssueBlocked_BlockedByOpenBlocker` has no assertion

Fixed. The test now has a real assertion: `require.True(t, blocked)`. Kept the `t.Skip` fallback path for environments where `GetDependenciesWithMetadata` returns an error (embedded Dolt limitation), but when the call succeeds, the assertion fires.

### Minor: `append(beadIDs, rigName)` mutates shared backing array

Fixed. Replaced with explicit `make` + `copy` to create an independent slice before appending the rig name. No shared backing array.

New test: `TestBatchSling_SliceAliasingInCrossRigGuard` — verifies the original slice is not mutated.

### Nit: `TestIsIssueBlocked_NoStore` has empty body

Fixed. Added a nil guard in `isIssueBlocked` (returns `false, nil` when store is nil) and the test now calls through with a nil store and asserts not-blocked. Consistent with the fail-open principle.

---

All six items on the pre-merge checklist are done. PR description updated with a review fixes section and the new test entries. Let me know if anything else needs attention.

PR #1764, opened 2026-02-19
fix(nudge): use immediate delivery with consistent MQ-oriented messages

### LINKED ISSUE #1763: Refinery not waking up after `gt done` — nudge queue replaced direct tmux delivery
## Summary

After `gt done`, the Refinery no longer wakes up to process the merge queue. The `POLECAT_DONE` mail is sent to Witness, and a nudge is enqueued for the Refinery, but the Refinery never actually receives it because the nudge sits in a file-based queue that requires an active turn boundary to drain.

## Root Cause

Commit `93f71cd9` (`feat(nudge): add non-destructive delivery modes`) replaced **all** direct `tmux.NudgeSession()` calls with `nudge.Enqueue()` — a file-based cooperative queue.

Introduced in PR #1405 (`seanbearden/fix/non-destructive-nudge-delivery`), merged as `1eaf4ac4`.

The PR contains two commits:
- `93f71cd9` — `feat(nudge): add non-destructive delivery modes (queue, wait-idle)` — blanket-replaced all `tmux.NudgeSession()` with `nudge.Enqueue()`
- `d405c852` — `fix(nudge): address review findings — session validation, crash recovery, docs`

**Before** (`sling_helpers.go`):
```go
t := tmux.NewTmux()
_ = t.NudgeSession(refinerySession, message)  // immediate tmux send-keys
```

**After**:
```go
nudge.Enqueue(townRoot, refinerySession, nudge.QueuedNudge{...})  // file written to disk
// Direct tmux nudge only as fallback when townRoot is empty (never in practice)
```

The file-based queue relies on the target agent calling `nudge.Drain()` at its next turn boundary (via UserPromptSubmit hook). But if the Refinery is idle — waiting for input, sleeping between patrol cycles — there is no turn boundary, so `Drain()` is never called and the nudge file rots on disk.

The same issue affects `wakeRigAgents()` which nudges the Witness after polecat dispatch.

## Affected Flow

1. Polecat runs `gt done`
2. `gt done` creates MR bead, pushes branch, merges Dolt branch
3. `gt done` calls `nudgeRefinery()` → `nudge.Enqueue()` writes a file (`done.go:846`)
4. `gt done` sends `POLECAT_DONE` mail to Witness (`done.go:882`)
5. **Refinery is idle** → nudge file sits unread → MR queue not processed
6. Witness processes `POLECAT_DONE` on next patrol → sends `MERGE_READY` mail → also enqueues nudge for Refinery → **same problem**

Neither path delivers the nudge if the Refinery session has no active turn.

## Expected Behavior

After `gt done` with a pending MR, the Refinery should immediately wake up and begin processing the merge queue — as it did before when `tmux.NudgeSession()` (direct send-keys) was used.

## Suggested Fix

For event-driven signals like `POLECAT_DONE` → Refinery nudge, use `immediate` delivery mode (direct tmux send-keys) instead of `queue` mode. The cooperative queue makes sense for non-urgent messages where interrupting a tool call is harmful, but waking an idle Refinery is the exact case where immediate delivery is needed.

The commit itself introduced three modes (`immediate`, `queue`, `wait-idle`) but then blanket-switched all callers to `queue`, removing the direct path for time-sensitive signals.

## References

- Introducing PR: #1405 by @seanbearden (Feb 13, 2026)
- Affected files: `internal/cmd/sling_helpers.go` (`nudgeRefinery`, `wakeRigAgents`), `internal/witness/handlers.go` (`nudgeRefinery`)

arttttt:
## Summary

Fixes #1763 — refinery not waking up after `gt done`.

Two problems fixed:

### 1. Wait-idle delivery causes stuck nudges

The original wait-idle approach (`WaitForIdle` → direct if idle, `Enqueue` if busy) has a fatal flaw: nudges queued while the agent is busy are **never drained**. `Drain()` only runs at `UserPromptSubmit` hook boundaries, and an idle agent never submits prompts. This causes queued nudges to be stuck forever — visible as `N queued` in `gt status` while the refinery sits idle.

**Fix**: Replace with immediate delivery via `NudgeSession()`. Sends text directly to the tmux pane. If idle — agent wakes up immediately. If busy — text buffers in tmux input and is processed at next prompt. No `Drain()` dependency.

### 2. Nudge message leads refinery to bypass merge queue

`done.go` and `mq_submit.go` sent messages like `"MR submitted: te-1615 branch=polecat/furiosa/te-r1qt@mlu360ha"`, which led the refinery to operate on the git branch directly instead of following the `mol-refinery-patrol` workflow (`gt mail inbox` → `gt mq list` → process).

**Fix**: All refinery nudge messages now use `"MERGE_READY received - check inbox for pending work"` — consistent with the witness nudge — directing the refinery through its merge queue patrol loop.

## Files changed

- **`internal/cmd/sling_helpers.go`** — `wakeRigAgents()` and `nudgeRefinery()`: removed wait-idle + queue logic, replaced with direct `NudgeSession()`
- **`internal/witness/handlers.go`** — `nudgeRefinery()`: same simplification
- **`internal/cmd/done.go`** — nudge message changed from `"MR submitted: ..."` to `"MERGE_READY received - check inbox for pending work"`
- **`internal/cmd/mq_submit.go`** — same message change

## Test plan

- [x] `go vet ./...` passes
- [x] `go test ./internal/cmd/... ./internal/witness/...` passes
- [x] Manual: `gt done` multiple times → refinery wakes up each time, no stuck `queued`
- [x] Manual: refinery follows inbox → mq list workflow, not direct git branch operations

### DIFF
--- a/internal/cmd/done.go
+++ b/internal/cmd/done.go
@@ -843,7 +843,7 @@ afterDoltMerge:
 	// on the polecat branch and refinery won't find it on main.
 	// If no branch existed (crew worker), MR bead is already on main.
 	if mrID != "" && !mergeFailed {
-		nudgeRefinery(rigName, fmt.Sprintf("MR submitted: %s branch=%s", mrID, branch))
+		nudgeRefinery(rigName, "MERGE_READY received - check inbox for pending work")
 	}
 
 	// Notify Witness about completion
--- a/internal/cmd/mq_submit.go
+++ b/internal/cmd/mq_submit.go
@@ -233,7 +233,7 @@ func runMqSubmit(cmd *cobra.Command, args []string) error {
 		}
 
 		// Nudge refinery to pick up the new MR
-		nudgeRefinery(rigName, fmt.Sprintf("MR submitted: %s branch=%s", mrIssue.ID, branch))
+		nudgeRefinery(rigName, "MERGE_READY received - check inbox for pending work")
 	}
 
 	// Success output
--- a/internal/cmd/sling_helpers.go
+++ b/internal/cmd/sling_helpers.go
@@ -16,7 +16,6 @@ import (
 	"github.com/steveyegge/gastown/internal/cli"
 	"github.com/steveyegge/gastown/internal/config"
 	"github.com/steveyegge/gastown/internal/constants"
-	"github.com/steveyegge/gastown/internal/nudge"
 	"github.com/steveyegge/gastown/internal/session"
 	"github.com/steveyegge/gastown/internal/style"
 	"github.com/steveyegge/gastown/internal/tmux"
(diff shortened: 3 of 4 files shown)


Xexr:
+1 for this, my refineries are effectively permanently idle on the current build

l0g1x:
nice!

arttttt:
tested it locally, works as intended

julianknutsen:
Adopted and merged via #1774 (new branch, since maintainer edits weren't enabled on this PR). Your commits are preserved — thank you @arttttt!

PR #1768, opened 2026-02-19
fix: correct liveness detection for custom agents shadowing built-in presets

arttttt:
## Problem

`gt polecat list` shows `○ done` and `gt status` shows `running: false` for actively working polecats, while `gt polecat status` correctly reports `State: working`.

## Root Cause

Custom agents defined in `settings/config.json` can shadow built-in preset names. In this case, a custom `"codex"` agent uses `opencode` as the command:

```json
"codex": { "command": "opencode", "args": ["-m", "openai/gpt-5.3-codex"] }
```

But the built-in `AgentCodex` preset has `Command: "codex"` and `ProcessNames: ["codex"]`.

The chain of failure:

1. `session_manager.Start()` sets `GT_AGENT=codex` in tmux session env
2. `IsAgentAlive()` reads `GT_AGENT=codex`
3. `GetProcessNames("codex")` finds the **built-in** `AgentCodex` preset → returns `["codex"]`
4. `IsRuntimeRunning()` looks for a process named `codex` in the session's process tree
5. The actual process is `opencode` → **not found** → returns `false`
6. `CheckSessionHealth()` → `AgentDead` → `IsRunning()` → `false`

The reconciliation logic in `gt polecat list` then overrides the beads state: "session dead + beads says working → must be done". Correct logic, but wrong liveness input.

`gt polecat status` works because it uses `HasSession()` (simple tmux existence check), not `IsAgentAlive()`/`CheckSessionHealth()`.

### Blast radius

`IsAgentAlive()` is called in 20+ places: witness zombie patrol, daemon lifecycle, doctor checks, refinery/witness managers, crew management, status display. All affected when a custom agent name collides with a built-in preset.

### New: `ResolveProcessNames(agentName, command)` in `config/agents.go`

Cross-references the agent name with the actual command binary:

1. If agent name matches a built-in preset AND the preset's Command matches → use preset's ProcessNames (no mismatch)
2. If command differs → find a built-in preset whose Command matches the actual binary → use its ProcessNames
3. Fallback: `[command]` for fully custom binaries

### New: `GT_PROCESS_NAMES` tmux env var

Set at session startup alongside `GT_AGENT`. Contains the comma-separated resolved process names (e.g., `opencode,node,bun`).

### New: `resolveSessionProcessNames()` helper in `tmux/tmux.go`

Shared by both `IsAgentAlive()` and `FindAgentPane()`. Prefers `GT_PROCESS_NAMES` if set, falls back to `GT_AGENT`-based lookup for legacy sessions.

## Why not alternative approaches

- **Fix `GetProcessNames` to be config-aware**: Would need `townRoot`/`rigPath` context that it doesn't have, and the function is called from tmux package which shouldn't depend on full config resolution.
- **Change `GT_AGENT` to the command name**: Breaks `handoff.go` and `sling_helpers.go` which use `GT_AGENT` as agent config name (not process name).
- **Register custom agents in global registry**: Timing issues (registry must be populated before any liveness check) and shared state conflicts between rigs with different custom configs.
- **Resolve from pane command at runtime**: Unreliable — pane command can be `bash` (shell wrapper), not the agent binary.

## Files changed

| File | Change |
|------|--------|
| `config/agents.go` | Add `ResolveProcessNames()` |
| `config/agents_test.go` | 8 test cases for `ResolveProcessNames` |
| `tmux/tmux.go` | Add `resolveSessionProcessNames()` helper, update `IsAgentAlive()` and `FindAgentPane()` |
| `polecat/session_manager.go` | Set `GT_PROCESS_NAMES` at polecat startup |
| `daemon/daemon.go` | Set `GT_PROCESS_NAMES` at daemon polecat restart |

## Test plan

- [x] `go vet ./...` passes
- [x] `TestResolveProcessNames` — 8 cases including the key scenario (`codex` + `opencode` → `["opencode", "node", "bun"]`)
- [x] All existing `config` tests pass
- [ ] Manual: sling a polecat with custom `codex` agent, verify `gt polecat list` and `gt status` show correct running state</body>
</invoke>

### DIFF
--- a/internal/config/agents.go
+++ b/internal/config/agents.go
@@ -574,6 +574,44 @@ func GetProcessNames(agentName string) []string {
 	return info.ProcessNames
 }
 
+// ResolveProcessNames determines the correct process names for liveness detection
+// given an agent name and the actual command binary. This handles custom agents
+// that shadow built-in preset names (e.g., a custom "codex" agent that runs
+// "opencode" instead of the built-in "codex" binary).
+//
+// Resolution order:
+//  1. If agentName matches a built-in preset AND the preset's Command matches
+//     the actual command → use the preset's ProcessNames (no mismatch).
+//  2. Otherwise, find a built-in preset whose Command matches the actual command
+//     and use its ProcessNames (custom agent using a known launcher).
+//  3. Fallback: [command] (fully custom binary).
+func ResolveProcessNames(agentName, command string) []string {
+	ensureRegistry()
+	registryMu.RLock()
+	defer registryMu.RUnlock()
+
+	// Check if agentName matches a built-in/registered preset with matching command
+	if info, ok := globalRegistry.Agents[agentName]; ok && len(info.ProcessNames) > 0 {
+		if info.Command == command || command == "" {
+			return info.ProcessNames
+		}
+	}
+
+	// Agent name doesn't match or command differs — look up by command
+	if command != "" {
+		for _, info := range globalRegistry.Agents {
+			if info.Command == command && len(info.ProcessNames) > 0 {
+				return info.ProcessNames
+			}
+		}
+		// Unknown command — use the binary name itself
+		return []string{command}
+	}
+
+	// No command provided, agent not in registry — Claude defaults
+	return []string{"node", "claude"}
+}
+
 // MergeWithPreset applies preset defaults to a RuntimeConfig.
 // User-specified values take precedence over preset defaults.
 // Returns a new RuntimeConfig without modifying the original.
--- a/internal/config/agents_test.go
+++ b/internal/config/agents_test.go
@@ -258,6 +258,82 @@ func TestGetProcessNamesRespectsRegistryOverride(t *testing.T) {
 	}
 }
 
+func TestResolveProcessNames(t *testing.T) {
+	t.Parallel()
+	ResetRegistryForTesting()
+	t.Cleanup(ResetRegistryForTesting)
+
+	tests := []struct {
+		name      string
+		agentName string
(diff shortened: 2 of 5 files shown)


julianknutsen:
## Adoption Review

Thanks for the contribution, @arttttt! This is a well-researched fix for a real and impactful bug — the root cause analysis tracing the full `IsAgentAlive` → `GetProcessNames` → built-in preset mismatch chain is excellent, and the architectural approach of resolving at startup time (when both agent name and command are available) is the right call.

### Review Summary
Decision: request_changes. The core fix logic is correct and the approach is sound, but the review identified a blocker-severity regression and two major gaps that need addressing before this can safely merge.

### Review Findings

| 1 | Behavioral Correctness | blocker | high | claude+codex | `ResolveProcessNames` fails when command is a resolved path (e.g., `/home/user/.claude/local/claude`) — breaks liveness for all new Claude polecats. Needs `filepath.Base()` normalization. |
| 2 | Change Impact / Blast Radius | major | high | codex+gemini | `GT_PROCESS_NAMES` only stamped in polecat/daemon; Crew, Deacon, Refinery, Dog sessions remain vulnerable to shadowing. |
| 3 | Behavioral Correctness | major | high | gemini | `handoff.go:buildRestartCommand` preserves `GT_AGENT` but not `GT_PROCESS_NAMES` — sessions revert to broken path after handoff. |
| 4 | Test Evidence Quality | minor | high | claude+codex | No integration tests for `GT_PROCESS_NAMES` precedence/fallback chain in `resolveSessionProcessNames`. |

1 review pass performed.

---

PR #1788, opened 2026-02-20
fix(daemon): skip restart for polecats in agent_state=spawning (#1752)

### LINKED ISSUE #1752: Boot triage double-spawns polecat sessions (race condition kills polecats)
## Summary

Every polecat spawn produces two Claude Code processes for the same worktree, consistently killing the polecat within 1-2 minutes.

## Evidence

Events log (`.events.jsonl`) shows two `session_start` events for every single spawn, ~7 seconds apart:

```
12:42:12  spawn   (gt)
12:42:12  sling   (mayor → cli/polecats/furiosa)
12:42:13  session_start  PID furiosa-28211  cwd: .../furiosa/cli   ← from sling
12:42:20  session_start  PID furiosa-32476  cwd: .../furiosa/cli   ← from boot triage
```

This is 100% reproducible - checked 10+ spawns in the events log, every single one has the double-start pattern with the same ~7s gap.

## Root Cause

1. `gt sling` creates worktree, sets `agent_state=spawning`, calls `gt session start` → Claude process #1 starts
2. ~7 seconds later, the **witness boot triage** runs, sees polecat in `agent_state=spawning` with no confirmed-alive session, fires `gt session start` again → Claude process #2 starts
3. Two Claude processes share the same `.jsonl` session file and fight each other → one (or both) die

The binary contains `'Witness for %s already running, skipping spawn'` - a guard exists for witness but the equivalent for polecats either doesn't exist or the alive-check races with startup.

## Fix Needed

`gt session start` / boot triage should verify whether a Claude process is **already running in the tmux pane** before firing a second one. Needs to handle the startup race: tmux session may exist but Claude not yet confirmed alive.

## Environment

- `gt` binary: built with `go build` (not `make build`, per warning on every invocation)
- Claude Code v2.1.47
- Polecat: `cli/polecats/furiosa`

Bochenski:
## Problem

Every polecat spawn produces two `session_start` events ~4-8s apart. 13/13 spawns were affected.

**Root cause**: `checkPolecatHealth` in the daemon fires during the race window between:
1. `AddWithOptions` (sets `hook_bead` + `agent_state=spawning` atomically)
2. The actual tmux session launch by `gt sling`

During this window the tmux session doesn't exist yet, so the daemon sees:
- session dead ✓
- hook_bead set ✓
→ "CRASH DETECTED" → `restartPolecatSession` → second Claude process starts alongside the one gt sling is about to launch

Both processes share the same session file and fight, causing premature `done` events and process death.

## Fix

Add a spawning guard in `checkPolecatHealth` that returns early when `agent_state == "spawning"`:

```go
if info.State == "spawning" {
    d.logger.Printf("Skipping restart for %s/%s: agent_state=spawning (gt sling in progress)",
        rigName, polecatName)
    return
}
```

This is placed after the `hook_bead == ""` check, so polecats with no work are still handled by the existing orphan detection path.

**Safety**: Polecats stuck in `spawning` due to a `gt sling` crash are still caught by the existing GUPP violation mechanism after 30 minutes.

## Test Plan

- [x] Added `TestCheckPolecatHealth_SkipsSpawning`: verifies spawning polecats are not restarted (no CRASH DETECTED)
- [x] Added `TestCheckPolecatHealth_DetectsCrashedPolecat`: verifies non-spawning polecats with dead sessions still trigger crash detection (guard doesn't regress normal behaviour)
- [x] All daemon package tests pass
- [x] Build passes: `go build ./...`

Fixes #1752

### DIFF
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -1461,6 +1461,17 @@ func (d *Daemon) checkPolecatHealth(rigName, polecatName string) {
 		return
 	}
 
+	// Spawning guard: skip polecats being actively started by gt sling.
+	// agent_state='spawning' means the polecat bead was created (with hook_bead
+	// set atomically) but the tmux session hasn't been launched yet. Restarting
+	// here would create a second Claude process alongside the one gt sling is
+	// about to start, causing the double-spawn bug (issue #1752).
+	if info.State == "spawning" {
+		d.logger.Printf("Skipping restart for %s/%s: agent_state=spawning (gt sling in progress)",
+			rigName, polecatName)
+		return
+	}
+
 	// TOCTOU guard: re-verify session is still dead before restarting.
 	// Between the initial check and now, the session may have been restarted
 	// by another heartbeat cycle, witness, or the polecat itself.
--- a/internal/daemon/polecat_health_test.go
+++ b/internal/daemon/polecat_health_test.go
@@ -1,0 +1,103 @@
+package daemon
+
+import (
+	"log"
+	"os"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/steveyegge/gastown/internal/tmux"
+)
+
+// writeFakeTestTmux creates a shell script in dir named "tmux" that simulates
+// "session not found" for has-session calls and fails on anything else.
+func writeFakeTestTmux(t *testing.T, dir string) {
+	t.Helper()
+	script := "#!/bin/sh\n" +
+		"case \"$*\" in\n" +
+		"  *has-session*) echo \"can't find session\" >&2; exit 1;;\n" +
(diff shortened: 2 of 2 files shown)


julianknutsen:
Thanks for this contribution, @Bochenski! Your fix correctly identified and resolved the double-spawn race condition in the daemon's health check — a subtle timing bug between bead creation and tmux session launch that was causing duplicate Claude processes.

We've merged this via #1789, which includes your original commit (with your authorship preserved) plus some maintainer fixups that came out of the review (reading agent_state from the DB column instead of stale description text, adding a 5-minute time-bound on the spawning guard, and Windows CI compatibility). We created a new PR because we needed to add commits but couldn't push directly to your branch.

Your commit is in the merged history. We appreciate the contribution and look forward to more!

PR #1789, opened 2026-02-20
fix(daemon): skip restart for polecats in agent_state=spawning (#1752)

julianknutsen:
* fix(daemon): skip restart for polecats in agent_state=spawning (#1752)

The daemon's checkPolecatHealth fired during the race window between
AddWithOptions (sets hook_bead + agent_state=spawning atomically) and
the actual tmux session launch by gt sling. Because the session didn't
exist yet, the daemon classified the polecat as a crash and called
restartPolecatSession, starting a second Claude process that shared the
same session file with the one gt sling was about to launch.

Fix: add a spawning guard that returns early when agent_state=spawning.
Polecats stuck in spawning due to a gt sling crash are still caught by
the existing GUPP violation mechanism after 30 minutes.

Regression tests added in polecat_health_test.go.

Fixes: steveyegge/gastown#1752

Co-Authored-By: Claude Sonnet 4.6 <noreply@anthropic.com>

* fix: read agent_state from DB column, add spawning guard time-bound

Address three review findings:

1. BLOCKER: getAgentBeadInfo read info.State from the description text,
   but UpdateAgentState only updates the DB column — so after a polecat
   transitions from "spawning" to "working", the description is stale
   and crash detection is permanently disabled. Fix: read State from the
   DB agent_state column (matching the HookBead pattern), with fallback
   to description for legacy beads.

2. MINOR: The spawning guard had no time-bound, so a polecat stuck in
   "spawning" due to a gt sling crash would be invisible to the daemon
   indefinitely. Fix: skip crash detection only if the bead was updated
   within the last 5 minutes; after that, fall through to normal crash
   detection.

3. MINOR: Tests synced description and DB agent_state, masking the
   divergence that causes the blocker. Fix: add test for expired
   spawning guard and test for description/DB state divergence.

4. Windows CI: Add runtime.GOOS skip guards for all 4 tests since they
   use Unix shell script mocks for tmux and bd.

---

Supersedes #1788 by @Bochenski (original fix preserved with original authorship).
Credit: Original fix by @Bochenski.

### DIFF
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -1466,10 +1466,26 @@ func (d *Daemon) checkPolecatHealth(rigName, polecatName string) {
 	// set atomically) but the tmux session hasn't been launched yet. Restarting
 	// here would create a second Claude process alongside the one gt sling is
 	// about to start, causing the double-spawn bug (issue #1752).
+	//
+	// Time-bound: only skip if the bead was updated recently (within 5 minutes).
+	// If gt sling crashed during spawn, the polecat would be stuck in 'spawning'
+	// indefinitely. The Witness patrol also catches spawning-as-zombie, but a
+	// time-bound here makes the daemon self-sufficient for this edge case.
 	if info.State == "spawning" {
-		d.logger.Printf("Skipping restart for %s/%s: agent_state=spawning (gt sling in progress)",
-			rigName, polecatName)
-		return
+		if updatedAt, err := time.Parse(time.RFC3339, info.LastUpdate); err == nil {
+			if time.Since(updatedAt) < 5*time.Minute {
+				d.logger.Printf("Skipping restart for %s/%s: agent_state=spawning (gt sling in progress, updated %s ago)",
+					rigName, polecatName, time.Since(updatedAt).Round(time.Second))
+				return
+			}
+			d.logger.Printf("Spawning guard expired for %s/%s: agent_state=spawning but last updated %s ago (>5m), proceeding with crash detection",
+				rigName, polecatName, time.Since(updatedAt).Round(time.Second))
+		} else {
+			// Can't parse timestamp — be safe, skip restart during spawning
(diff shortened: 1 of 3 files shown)


julianknutsen:
## Adoption Review

This PR supersedes #1788 by @Bochenski, which correctly identified and fixed the double-spawn race condition in the daemon's health check (issue #1752). This PR includes that fix and adds maintainer fixups.

Credit: Original fix by @Bochenski (commit preserved with original authorship).

### Initial Review
Decision: **block** — triple-model review (Claude + Codex + Gemini) identified a blocker: `getAgentBeadInfo` reads `info.State` from description text, but `UpdateAgentState` only writes to the DB column. After a polecat transitions spawning→working, the stale description permanently disables crash detection. Two additional minor findings: no time-bound on spawning guard, and tests masking the description/DB divergence.

### Maintainer Changes
- Read `agent_state` from DB column (matching existing `HookBead` pattern), with fallback to description for legacy beads
- Add 5-minute time-bound on spawning guard so stuck-in-spawning polecats are detected
- Rewrite test helpers with separate `descState`/`dbState` parameters to expose divergence
- Add `TestCheckPolecatHealth_SpawningGuardExpires` and `TestCheckPolecatHealth_DBStateOverridesDescription` regression tests
- Add `runtime.GOOS` skip guards for Windows CI (tests use Unix shell script mocks)

### Final Review Status
Decision: **approve** — all 3 fix validations confirmed correct by all reviewers. One remaining major/medium finding: unparseable `updated_at` while `agent_state=spawning` defaults to skip-restart (permanently bypassing crash detection). Mitigated by: (a) three independent failures must compound, (b) Witness patrol provides backup, (c) 5-min bound covers normal path. Recommended as follow-up task.

### Review Iterations
2 review passes performed.

---
_Adopted via `/adopt-pr` workflow (Path C — new PR). Original: #1788. Contributor commits preserved._

PR #1797, opened 2026-02-20
Fix: Unset BD_BRANCH when reading hooked work

zfogg:
Critical fix for polecat work discovery: BD_BRANCH environment variable was restricting beads CLI to read from the polecat's isolated Dolt branch, preventing it from seeing hooked work.

When polecats run `gt hook` or `gt mol status`, they now temporarily unset BD_BRANCH to read from the main rig branch where work assignments live, then operations can use BD_BRANCH for write isolation.

## Polecat Session Startup Fixes

This completes the sequence of fixes for polecat work discovery:

1. ✓ Working directory: polecat starts in root (not worktree subdirectory)
2. ✓ Startup instructions: sent via beacon + nudges to agent to run `gt prime` and discover hooked work
3. ✓ Beads redirect: .beads/redirect in root finds shared database
4. ✓ BD_BRANCH: unset for reads to access main branch work [NEW]

**Fixes the 'Propulsion Principle'** where hooked work should auto-execute on polecat startup.

### DIFF
--- a/internal/cmd/prime.go
+++ b/internal/cmd/prime.go
@@ -412,6 +412,13 @@ func findAgentWork(ctx RoleContext) *beads.Issue {
 		return nil
 	}
 
+	// Unset BD_BRANCH so we read from the main rig branch where work is assigned,
+	// not the polecat's isolated branch. Polecats have BD_BRANCH set for write isolation,
+	// but they need to read hooked work from the main branch.
+	if bdBranch := os.Getenv("BD_BRANCH"); bdBranch != "" {
+		os.Unsetenv("BD_BRANCH")
+	}
+
 	b := beads.New(ctx.WorkDir)
 
 	// Primary: agent bead's hook_bead field (authoritative, set by bd slot set during sling)


julianknutsen:
I looked into this change and I'm not convinced it's the right fix. I've opened #1815 to revert it — here's my analysis:

### The sling sequence guarantees hooked data is on the polecat branch

The `CreateDoltBranch()` method in `polecat_spawn.go:329-347` explicitly:
1. **Flushes** the working set to HEAD via `CommitServerWorkingSet` (line 338)
2. **Forks** the branch from that HEAD via `CreatePolecatBranch` (line 342)

This is called *after* all sling writes (hook, formula, fields) complete. The comments in `sling_formula.go:215-217`, `sling.go:654-656`, and `sling_batch.go:229-231` all emphasize this ordering. So the hooked bead data should exist on both main and the polecat branch by construction.

### `os.Unsetenv("BD_BRANCH")` breaks write isolation

The unset is permanent for the process — it's never restored. Since `beads.run()` passes `os.Environ()` to the `bd` CLI (see `beads.go:254`), any subsequent `bd` writes in the same `gt` process invocation will go to main instead of the polecat's isolated branch. This silently defeats the purpose of branch-per-polecat isolation.

### The fix is in the wrong layer

If there were a legitimate need to read from main, it should be handled once in `beads.run()` (e.g., stripping `BD_BRANCH` from the env for read operations) rather than copy-pasted across 3 call sites (`hook.go`, `molecule_status.go`, `prime.go`).

### Question for @mayor / the author

What specific error or behavior were you seeing that motivated this change? Was the polecat unable to find its hooked work on startup? If so, the root cause is more likely:

1. The flush not completing before the branch fork
2. The `bd` CLI not properly reading from the `BD_BRANCH` branch
3. New work being hooked to an already-running polecat from outside (which writes to main after the branch already exists — a legitimate scenario, but one that needs a different solution than globally unsetting the env var)

Understanding the reproduction case would help identify the actual fix.

PR #1804, opened 2026-02-20
fix(refinery): preserve role_agents override across handoff/respawn

olpie101:
## Summary
- Fix refinery handoff/respawn reverting to "claude" when using role_agents override

## Problem
When refinery starts without `--agent` flag, `ResolveRoleAgentConfig` correctly resolves the agent from `role_agents` in settings. However, the agent name was never persisted as `GT_AGENT` in the tmux environment. On handoff, the code falls back to `ResolveAgentConfig` which ignores `role_agents`, reverting to "claude".

## Solution
- Resolve `effectiveAgent` before building startup command (checks --agent flag, then role_agents)
- Set `GT_AGENT` in tmux environment to preserve the agent across handoff/respawn

## Changes
- `internal/refinery/manager.go`: +19/-2 lines

## Testing
- Start refinery with `role_agents.refinery = "opencode"` in settings
- Verify `GT_AGENT=opencode` is set in tmux: `tmux show-environment -t <refinery-session>`
- Trigger handoff and verify agent is preserved

### DIFF
--- a/internal/deacon/manager_test.go
+++ b/internal/deacon/manager_test.go
@@ -2,10 +2,13 @@ package deacon
 
 import (
 	"errors"
+	"os"
 	"path/filepath"
+	"strings"
 	"testing"
 	"time"
 
+	"github.com/steveyegge/gastown/internal/config"
 	"github.com/steveyegge/gastown/internal/tmux"
 )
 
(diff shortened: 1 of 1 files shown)


julianknutsen:
Superseded by #1813, which takes a more centralized approach — adding `ResolvedAgent` to `AgentEnvConfig` so `AgentEnv()` handles the GT_AGENT fallback in one place for all roles, rather than resolving the effective agent in each manager's `Start()` method individually.

#1813 has been merged. Closing this PR.

PR #1806, opened 2026-02-20
fix(done): preserve validation errors instead of swallowing them

olpie101:
## Summary
- Fix `gt done` silently swallowing validation errors due to deferred cleanup overwriting retErr

## Problem
The deferred cleanup in `runDone()` unconditionally set `retErr = NewSilentExit(0)`, which overwrote any validation errors that occurred during execution. This caused validation failures to be silently ignored, making debugging difficult.

## Solution
- Only set `SilentExit(0)` when `retErr` is nil (no prior error)
- Added `SilenceUsage: true` to prevent usage output on errors

## Changes
- `internal/cmd/done.go`: +6/-3 lines

## Testing
- Existing tests pass
- Validation errors now propagate correctly instead of being overwritten

### DIFF
--- a/internal/cmd/done_test.go
+++ b/internal/cmd/done_test.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"os/exec"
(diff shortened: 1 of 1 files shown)


olpie101:
Not sure if this the appropriate approach, currently "work on my machine" but I couldn't easily see what the ramification of this change could be. I can't tell idf this could result in silent errors

PR #1809, opened 2026-02-20
fix(handoff): kill old processes before respawning to prevent duplicates

olpie101:
## Summary
- Fix duplicate agent processes caused by self-handoff not killing old processes properly

## Problem
Self-handoff relied on `respawn-pane -k` to kill the old process, but respawn-pane only sends SIGHUP which opencode/Node.js ignores. This caused the old process to survive while a new one spawned, creating duplicate agent processes.

## Solution
1. **handoff.go**: Call `KillPaneProcessesExcluding()` before `RespawnPane()` to explicitly kill all pane processes (except gt handoff itself) using SIGTERM/SIGKILL.

2. **daemon.go**: Change hung session handling from `KillSession()` to `KillSessionWithProcesses()` for witness and refinery.

3. **witness/manager.go**: Change `Stop()` and zombie detection to use `KillSessionWithProcesses()`.

4. **witness/handlers.go**: Change polecat nuke to use `KillSessionWithProcesses()`.

5. **refinery/manager.go**: Change `Stop()` and zombie detection to use `KillSessionWithProcesses()`.

`KillSession()` only sends SIGHUP which agents ignore, leaving orphan processes. `KillSessionWithProcesses()` sends SIGTERM/SIGKILL to ensure all descendant processes are terminated.

## Changes
- `internal/cmd/handoff.go`: +15/-10 lines
- `internal/daemon/daemon.go`: +2/-2 lines
- `internal/refinery/manager.go`: +3/-3 lines
- `internal/witness/manager.go`: +3/-3 lines
- `internal/witness/handlers.go`: +2/-2 lines

## Testing
- Build passes
- Existing tests pass

### DIFF
--- a/internal/cmd/handoff.go
+++ b/internal/cmd/handoff.go
@@ -295,15 +295,14 @@ func runHandoff(cmd *cobra.Command, args []string) error {
 		style.PrintWarning("could not set remain-on-exit: %v", err)
 	}
 
-	// NOTE: For self-handoff, we do NOT call KillPaneProcesses here.
-	// That would kill the gt handoff process itself before it can call RespawnPane,
-	// leaving the pane dead with no respawn. RespawnPane's -k flag handles killing
-	// atomically - tmux kills the old process and spawns the new one together.
-	// See: https://github.com/steveyegge/gastown/issues/859 (pane is dead bug)
-	//
-	// For orphan prevention, we rely on respawn-pane -k which sends SIGHUP/SIGTERM.
-	// If orphans still occur, the solution is to adjust the restart command to
-	// kill orphans at startup, not to kill ourselves before respawning.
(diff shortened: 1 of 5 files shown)


julianknutsen:
Follow-up review fixups merged via #1819. These address issues identified during the adoption review of this PR:

- Moved `GetPaneWorkDir` before `KillPaneProcessesExcluding` to prevent read-after-kill race
- Fixed 100ms grace period to `processKillGracePeriod` (2s) in `KillPaneProcessesExcluding`
- Added `collectReparentedGroupMembers` to `KillPaneProcessesExcluding` (matching sibling functions)
- Fixed `molecule_step.go` to use `KillPaneProcessesExcluding` with self-PID exclusion
- Added pane PID exclusion to prevent SIGHUP regression when session leader is killed

PR #1817, opened 2026-02-20
fix: set GT_PROCESS_NAMES for Deacon and unified session lifecycle

### LINKED ISSUE #1808: Deacon auto-respawn hook fails with OpenCode: GT_PROCESS_NAMES not set in tmux session env
**Bug Description**

When using OpenCode CLI as the agent (instead of Claude), the Deacon's auto-respawn hook intermittently fails with:

This error completely replaces the attached Mayor view in tmux, requiring users to detach and reattach.

**Root Cause Analysis**

The issue is in how `GT_PROCESS_NAMES` environment variable is handled:

1. `BuildStartupCommand()` correctly sets `GT_PROCESS_NAMES=opencode,node,bun` in the startup command's `exec env` prefix
2. However, `AgentEnv()` does NOT include `GT_PROCESS_NAMES` in its returned map
3. The Deacon manager calls `t.SetEnvironment()` with the `AgentEnv()` result, which overwrites the tmux session table WITHOUT `GT_PROCESS_NAMES`
4. Later, `IsAgentAlive()` calls `GetEnvironment(session, "GT_PROCESS_NAMES")` which reads from tmux's session table, not the process environment
5. Since `GT_PROCESS_NAMES` is missing from the session table, `IsAgentAlive` falls back to default `["node", "claude"]`
6. The Deacon is running `opencode`, not `claude`, so process detection fails and `IsAgentAlive` returns false
7. This triggers the auto-respawn hook, which fails because the session is actually alive

**Relevant Code**

`internal/config/env.go` - `AgentEnv()` does not include GT_PROCESS_NAMES:
https://github.com/steveyegge/gastown/blob/main/internal/config/env.go#L39-L139

`internal/deacon/manager.go` - Sets environment from AgentEnv() after session creation:
https://github.com/steveyegge/gastown/blob/main/internal/deacon/manager.go#L102-L107

`internal/tmux/tmux.go` - `resolveSessionProcessNames()` reads from session environment:
https://github.com/steveyegge/gastown/blob/main/internal/tmux/tmux.go#L1014-L1027

**Environment**

- Gastown configured to use OpenCode CLI (`command: opencode` in settings)
- Agent preset: `kimi-k2.5-free` (or any OpenCode-based agent)
- Town-level agents (Mayor, Deacon) affected

**Expected Behavior**

`GT_PROCESS_NAMES` should be set in the tmux session environment so that `IsAgentAlive` correctly detects OpenCode processes.

**Possible Fix**

Option 1: Add `GT_PROCESS_NAMES` to `AgentEnv()` by resolving the agent config and including the process names.

Option 2: In role managers (deacon, mayor), explicitly set `GT_PROCESS_NAMES` via `SetEnvironment` after resolving the runtime config.

Option 3: In `resolveSessionProcessNames`, also check the pane's actual running command (via `GetPaneCommand`) as a fallback when `GT_PROCESS_NAMES` is empty.

**Workaround**

Users can manually set the environment after starting Deacon:
`gt deacon env GT_PROCESS_NAMES=opencode,node,bun`

Or ignore the error — the daemon will still restart the Deacon via heartbeat monitoring.

DerrickBarra:
## Problem

The fix from #1785 added GT_PROCESS_NAMES to daemon and polecat session managers, but missed:
1. **Deacon manager** (town-level agent)
2. **Unified session lifecycle** (StartSession used by Mayor, Boot, etc.)

This caused IsAgentAlive to fail for non-Claude agents (opencode, etc.) running as Mayor or Deacon, because:
- GT_PROCESS_NAMES was set in the process env via "exec env ..." in the startup command
- But NOT in the tmux session table via SetEnvironment()
- IsAgentAlive reads from tmux session table, which defaulted to ["node", "claude"]
- The actual process was "opencode", not "claude", so liveness check failed
- This triggered false "respawn-pane" errors and broke agent detection

## Solution

Add GT_PROCESS_NAMES environment variable setting to:
- internal/deacon/manager.go - explicit Deacon startup path
- internal/session/lifecycle.go - unified StartSession() used by Mayor, Boot, Witness, Refinery

## Testing

- All existing tests pass: go test ./internal/deacon/... ./internal/session/...
- Verified GT_PROCESS_NAMES is correctly set for opencode agents
- Fix confirmed working in live Gastown deployment

## Related

- Fixes #1808
- Related to #1785 (previous fix for daemon/polecat)

Co-authored-by: Cookie (OpenClaw Agent)

### DIFF
--- a/internal/deacon/manager.go
+++ b/internal/deacon/manager.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
+	"strings"
 	"time"
 
 	"github.com/steveyegge/gastown/internal/config"
@@ -126,6 +127,11 @@ func (m *Manager) Start(agentOverride string) error {
 		Role:     "deacon",
 		TownRoot: m.townRoot,
 	})
+	// FIX: GT_PROCESS_NAMES must be set for correct IsAgentAlive detection
+	// when using non-Claude agents (opencode, codex, etc.)
+	// See: https://github.com/steveyegge/gastown/issues/1808
+	processNames := config.GetProcessNames(runtimeConfig.ResolvedAgent)
+	envVars["GT_PROCESS_NAMES"] = strings.Join(processNames, ",")
 	for k, v := range envVars {
 		_ = t.SetEnvironment(sessionID, k, v)
 	}
--- a/internal/session/lifecycle.go
+++ b/internal/session/lifecycle.go
@@ -4,6 +4,7 @@ package session
 import (
 	"fmt"
 	"sort"
+	"strings"
 	"time"
 
 	"github.com/steveyegge/gastown/internal/config"
@@ -193,6 +194,11 @@ func StartSession(t *tmux.Tmux, cfg SessionConfig) (*StartResult, error) {
 		TownRoot:         cfg.TownRoot,
 		RuntimeConfigDir: cfg.RuntimeConfigDir,
 	})
+	// FIX: GT_PROCESS_NAMES must be set for correct IsAgentAlive detection
+	// when using non-Claude agents (opencode, codex, etc.)
+	// See: https://github.com/steveyegge/gastown/issues/1808
+	processNames := config.GetProcessNames(runtimeConfig.ResolvedAgent)
+	envVars["GT_PROCESS_NAMES"] = strings.Join(processNames, ",")
 	for _, k := range mapKeysSorted(envVars) {
 		_ = t.SetEnvironment(cfg.SessionID, k, envVars[k])
 	}


julianknutsen:
Thanks for this contribution, @DerrickBarra! Your fix correctly identified that `GT_PROCESS_NAMES` was missing from both the Deacon manager and the unified `StartSession` lifecycle, which caused `IsAgentAlive` to fall back to legacy detection and fail for non-Claude agents like opencode and codex.

We've merged this via #1821, which includes your original commit (with your authorship preserved) plus a maintainer fixup that changed `GetProcessNames` → `ResolveProcessNames` to handle custom agents that shadow built-in preset names. We created a new PR because we needed to add commits but couldn't push directly to your branch.

Your commit is in the merged history — `git log --author="DerrickBarra"` will show it. We appreciate the contribution and look forward to more!

PR #1816, opened 2026-02-20
fix(done): prevent stale polecat from force-closing reassigned beads

julianknutsen:
## Summary
- Add `isStalePolecatDone` ownership guard before the G15 force-close path in `gt done`
- Prevents respawned polecats with stale branch state from closing beads assigned to other polecats
- Two independent checks: agent hook mismatch (primary) and assignee mismatch (backup)

## Related Issue
Fixes gt-frf61: polecat furiosa was respawned for gt-pjox2 but retained stale branch `polecat/furiosa/gt-frf61@mlurugd2`. Its `gt done` found 0 commits ahead, hit the G15 force-close path, and closed gt-frf61 with "Completed with no code changes" — while cheedo was actively working on it. The branch with 2 real review-fix commits (`polecat/furiosa/gt-frf61@mlupw65o`) was never merged.

### Root cause timeline
1. Furiosa completed gt-frf61 review fixes at 10:18 (branch `@mlupw65o` pushed to origin)
2. Furiosa respawned for gt-pjox2, completed it at 10:37/10:51
3. Witness recovered orphaned gt-frf61 at 10:52, mayor dispatched to **cheedo** at 10:55
4. Furiosa respawned at 10:53 — worktree created with stale frf61 branch `@mlurugd2`
5. Furiosa's session started at 10:57 (for MERGED mail about pjox2)
6. **Bug**: Within 41s, furiosa ran `gt done` on the stale branch → 0 commits ahead → G15 `ForceCloseWithReason` closed gt-frf61 at 10:58
7. Auto-flush captured closed state; cheedo's later merge lost to conflict resolution

### Why the guard works
`isStalePolecatDone` runs two checks before `ForceCloseWithReason`:
1. **Hook mismatch**: Agent's `hook_bead` (authoritative current work) ≠ `issueID` from branch name → stale
2. **Assignee mismatch**: Bead's assignee doesn't contain this polecat's name → stale

In the gt-frf61 case: furiosa's hook was empty/pointed to pjox2, and the bead was assigned to cheedo. Either check catches it.

## Changes
- `internal/cmd/done.go`: Add `isStalePolecatDone` guard before G15 force-close, add helper function
- `internal/cmd/done_test.go`: Add `TestIsStalePolecatDoneLogic` with 8 test cases covering hook match/mismatch, assignee formats, and edge cases

## Testing
- [x] Unit tests pass (`go test ./internal/cmd/`)
- [x] `go vet ./internal/cmd/` clean
[...]

PR #1818, opened 2026-02-20
fix(done): detect stale branch before push/MR/close (gt-frf61)

julianknutsen:
## Summary

When a polecat is respawned for new work, its worktree branch retains the old issue ID while `hook_bead` points to the new assignment. If the stale polecat calls `gt done`, it incorrectly pushes, creates MRs, or force-closes the old bead — which may now belong to another polecat. This is the root cause of gt-frf61 being prematurely closed as "not implemented" despite having real commits on origin.

Companion to #1816 which adds a defense-in-depth guard on the G15 force-close path specifically. This PR catches the problem earlier, before *any* done operations execute.

## Related Issue

Root cause of gt-frf61 premature closure.

## Changes

- Add early stale-branch detection at the top of the `ExitCompleted` block in `done.go`
- Compare branch-parsed `issueID` against the agent's `hook_bead` via existing `getIssueFromAgentHook`
- On mismatch, log a warning and `goto notifyWitness` to skip push/MR/close and cleanly tear down the session
- Add `TestStaleBranchDetectionLogic` with 5 test cases covering match, mismatch, empty hook, empty issue, and both-empty scenarios

## Testing

- [x] Unit tests pass (`go test ./internal/cmd/...`)
- [x] `go vet ./internal/cmd/...` clean
- [x] New test `TestStaleBranchDetectionLogic` covers all guard branches

## Checklist
- [x] No breaking changes
- [x] No merge conflicts with #1816 (touches different code paths)

### DIFF
--- a/internal/cmd/done.go
+++ b/internal/cmd/done.go
@@ -375,6 +375,7 @@ func runDone(cmd *cobra.Command, args []string) (retErr error) {
 	var pushFailed bool
 	var doneErrors []string
 	var convoyInfo *ConvoyInfo // Populated if issue is tracked by a convoy
+	var staleBranchDetected bool
 
 	// Pre-declare variables used between notifyWitness and afterAgentStateUpdate
 	// so that goto notifyWitness / goto afterAgentStateUpdate don't jump over declarations.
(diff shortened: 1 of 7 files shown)


julianknutsen:
## Adoption Review

Thanks for the contribution, @julianknutsen! This PR correctly identifies the stale-branch scenario where a respawned polecat's old worktree retains a stale issue ID while `hook_bead` points to the new assignment — causing `gt done` to push/create MRs/force-close the wrong bead. The guard-based approach with early detection and `goto notifyWitness` is clean and follows the existing `done.go` control flow patterns.

### Initial Review
The initial triple-model review (Claude + Codex + Gemini) identified 5 findings across the original commit:
1. **Major:** `updateAgentStateOnDone` could close the new assignment's bead on the stale path
2. **Major:** `--issue` flag could bypass the stale-branch guard
3. **Major:** Test reimplements guard logic inline instead of exercising the real code path
4. **Minor:** No structured logging or witness distinction for stale detection
5. **Minor:** Guard fails open when bead lookup errors (swallowed by `getIssueFromAgentHook`)

### Maintainer Changes
The fixup commit (`fix(done): harden stale-branch guard`) addresses findings 1, 2, 4, and 5:
- Wires `StaleBranch` through the witness protocol (`PolecatDonePayload` field + parser + handler check) so stale POLECAT_DONE is not routed through pending-MR path
- Suppresses WORK_DONE mail, LogDone, and TypeDone event on stale path
- Skips done checkpoint writes on stale path (prevents next `gt done` from skipping required stages)
- Clears done-intent label and checkpoints on stale path (prevents witness patrol misclassification)
- Uses `info.Issue` (branch-parsed) instead of `issueID` to prevent `--issue` bypass
- Guards `updateAgentStateOnDone` with `!staleBranchDetected`
- Adds `TypeStaleBranch` structured event with diagnostic payload
- Documents generic branch (`polecat/<worker>-<timestamp>`) limitation as follow-up

Finding 3 (test quality) was downgraded to minor — the existing test validates the guard logic correctly, though an integration-level `runDone` test would add confidence. Tracked as a follow-up.

+123 -25 across 6 files.

### Final Review Status
The final review (iteration 3) returned **approve** with no blockers or majors. Three minor/nit findings remain, none requiring changes before merge:
1. **Nit:** Dolt merge runs unnecessarily on stale path (idempotent, harmless)
2. **Minor:** Test doesn't exercise `agentBeadID` non-empty + `getIssueFromAgentHook` returning `""` edge case (logic is correct, test is imprecise)
3. **Minor:** `Issue:` field in stale payload contains the stale issue ID (not functional — `StaleBranch` flag causes witness to short-circuit)

### Review Iterations
3 review passes performed.

---

PR #1821, opened 2026-02-20
fix: set GT_PROCESS_NAMES for Deacon and unified session lifecycle

### LINKED ISSUE #1808: Deacon auto-respawn hook fails with OpenCode: GT_PROCESS_NAMES not set in tmux session env

julianknutsen:
## Summary

Supersedes #1817 by @DerrickBarra. This PR includes the original fix plus a maintainer fixup that came out of the adoption review.

Credit: Original fix by @DerrickBarra (commit preserved with original authorship).

The fix from #1785 added GT_PROCESS_NAMES to daemon and polecat session managers, but missed:
1. **Deacon manager** — explicit Deacon startup path
2. **Unified session lifecycle** — `StartSession()` used by Mayor, Boot, Dog, etc.

This caused `IsAgentAlive` to fail for non-Claude agents (opencode, codex, etc.) running as Mayor or Deacon, because `resolveSessionProcessNames` reads GT_PROCESS_NAMES from the tmux session table, which was never set for these paths.

**Maintainer fixup:** Changed `GetProcessNames` → `ResolveProcessNames` to cross-reference both agent name and command binary, matching the pattern used by all other GT_PROCESS_NAMES callsites (daemon.go, polecat/session_manager.go, loader.go, handoff.go).

## Related Issue

Fixes #1808
Supersedes #1817
Related to #1785

## Changes

- Add GT_PROCESS_NAMES to `internal/deacon/manager.go` using `config.ResolveProcessNames()`
- Add GT_PROCESS_NAMES to `internal/session/lifecycle.go` using `config.ResolveProcessNames()`

## Testing

- Existing tests pass (build failure in `config/agents.go:338` is pre-existing on main — duplicate struct fields unrelated to this PR)
- Pattern matches established usage in daemon.go and polecat/session_manager.go
- Fix confirmed working in live Gastown deployment (per original PR)

## Checklist
- [x] No breaking changes
- [x] Matches established GT_PROCESS_NAMES pattern across codebase

### DIFF
--- a/internal/deacon/manager.go
+++ b/internal/deacon/manager.go
@@ -132,7 +132,7 @@ func (m *Manager) Start(agentOverride string) error {
 	// FIX: GT_PROCESS_NAMES must be set for correct IsAgentAlive detection
 	// when using non-Claude agents (opencode, codex, etc.)
 	// See: https://github.com/steveyegge/gastown/issues/1808
-	processNames := config.GetProcessNames(runtimeConfig.ResolvedAgent)
+	processNames := config.ResolveProcessNames(runtimeConfig.ResolvedAgent, runtimeConfig.Command)
 	envVars["GT_PROCESS_NAMES"] = strings.Join(processNames, ",")
 	for k, v := range envVars {
 		_ = t.SetEnvironment(sessionID, k, v)
--- a/internal/session/lifecycle.go
+++ b/internal/session/lifecycle.go
@@ -199,7 +199,7 @@ func StartSession(t *tmux.Tmux, cfg SessionConfig) (*StartResult, error) {
 	// FIX: GT_PROCESS_NAMES must be set for correct IsAgentAlive detection
 	// when using non-Claude agents (opencode, codex, etc.)
 	// See: https://github.com/steveyegge/gastown/issues/1808
-	processNames := config.GetProcessNames(runtimeConfig.ResolvedAgent)
+	processNames := config.ResolveProcessNames(runtimeConfig.ResolvedAgent, runtimeConfig.Command)
 	envVars["GT_PROCESS_NAMES"] = strings.Join(processNames, ",")
 	for _, k := range mapKeysSorted(envVars) {
 		_ = t.SetEnvironment(cfg.SessionID, k, envVars[k])


julianknutsen:
## Adoption Review

This PR supersedes #1817 by @DerrickBarra, which correctly identified that `GT_PROCESS_NAMES` was missing from the Deacon manager and unified `StartSession()` lifecycle — causing `IsAgentAlive` to fail for non-Claude agents (opencode, codex, etc.) in Mayor, Boot, Dog, and Deacon roles. This PR includes that fix and adds a maintainer fixup.

Credit: Original fix by @DerrickBarra (commit preserved with original authorship).

### Initial Review
The initial triple-model review (Claude + Codex + Gemini) identified 1 blocker and 2 majors:
- **Blocker:** `beads_types.go` contained +104 lines of out-of-scope database init code conflicting with main (already superseded by PR #1777)
- **Major:** Used `GetProcessNames` instead of `ResolveProcessNames` — doesn't handle custom agents shadowing built-in names
- **Major:** Branch was stale with `AgentEnvConfig` struct divergence guaranteeing merge conflicts

### Maintainer Changes
1. Rebased onto current main — dropped stale `beads_types.go` changes and 2 redundant commits already on main
2. Changed `GetProcessNames` → `ResolveProcessNames(runtimeConfig.ResolvedAgent, runtimeConfig.Command)` in both callsites to match established pattern in `daemon.go:1619`, `polecat/session_manager.go:344`, `loader.go:1777,1953`, `handoff.go:787,858,868`

### Final Review Status
All clear — triple-model re-review (Claude + Codex + Gemini) returned **approve** with no blockers or majors. All three reviewers independently validated the fix as correct. Two minor notes tracked as follow-ups:
- Refinery/Witness/Crew managers bypass `StartSession()` and remain unpatched (filed as gt-k1wib)
- No regression test for GT_PROCESS_NAMES (low risk given mechanical 3-line pattern addition)

### Review Iterations
2 review passes performed.

---
_Adopted via `/adopt-pr` workflow (Path C — new PR). Original: #1817. Contributor commits preserved._

PR #1826, opened 2026-02-21
fix: add polecats/ prefix to nudge address resolution

timyarm:
## Summary

- Short addresses like `rig/name` resolve crew-first in `gt nudge`, which causes nudges intended for polecats to land in crew sessions when both share the same name
- Adds explicit `polecats/` prefix handling (alongside existing `crew/` prefix) so callers can disambiguate: `rig/polecats/name` always targets the polecat session
- Backward compatible — `rig/name` still resolves crew-first

## Context

Witnesses nudge polecats to report startup hook failures, but when a crew worker shares the same name (e.g., `furiosa` exists as both `VAS-crew-furiosa` and `VAS-furiosa`), the nudge lands in the crew session instead. This causes crew workers to receive misdirected messages like "Startup hook failed. Run gt prime."

## Test plan

- [x] Existing `TestNudge*` tests pass (no regressions)
- [ ] Manual: `gt nudge vastal/polecats/furiosa "test"` targets polecat, not crew
- [ ] Manual: `gt nudge vastal/crew/furiosa "test"` still targets crew
- [ ] Manual: `gt nudge vastal/furiosa "test"` still targets crew (backward compat)


julianknutsen:
Thanks for this contribution, @timyarm! Your fix correctly identified that `rig/polecats/name` addresses weren't handled in `gt nudge`, causing nudges to target nonexistent sessions instead of the intended polecat.

We've merged this via #1830, which includes your original commit (with your authorship preserved) plus maintainer fixups that came out of the review — specifically, the same `polecats/` prefix handling needed to be added to `addressToAgentBeadID` in both `nudge.go` and `router.go` to prevent DND bypass, plus test coverage. We created a new PR because your branch had additional unrelated commits that expanded the scope.

Your commit is in the merged history — `git log --author="Tim Yarm"` will show it. We appreciate the contribution and look forward to more!

PR #1832, opened 2026-02-21
feat: file-based event channel system (await-event + emit-event)

timyarm:
## Summary

- Add `gt mol step await-event` — blocks on channel-specific file events using inotifywait, with exponential backoff and agent bead idle tracking (mirrors `await-signal` pattern)
- Add `gt mol step emit-event` — programmatic Go counterpart to `emit-event.sh` for creating event files on named channels
- Wire `nudgeRefinery()` to also emit `MQ_SUBMIT` file events, so `gt mq submit` automatically triggers the event system
- Update refinery patrol formula to use `await-event --channel refinery` instead of generic `await-signal`
- Update witness patrol formula to emit `PATROL_WAKE` events when waking the refinery

## Motivation

The refinery currently uses `await-signal` which subscribes to the generic beads activity feed — every bead change wakes it, even when irrelevant. This causes unnecessary patrol cycles and credit burn.

The event channel system gives each subscriber a dedicated channel (`~/gt/events/<channel>/`) so they only wake for events they care about:
- `MERGE_READY` — from witness when polecat branch is pushed
- `PATROL_WAKE` — from witness when MRs waiting but refinery idle  
- `MQ_SUBMIT` — from `gt mq submit` via `nudgeRefinery()`

## Test plan

- [x] 22 unit tests covering timeout calculation, event file I/O, backoff math, JSON marshaling, payload parsing
- [x] E2E: `emit-event` wakes blocking `await-event` within ~1s
- [x] E2E: `--cleanup` flag auto-deletes processed events
- [x] E2E: Multiple events batched and returned together
- [x] E2E: Timeout returns correctly with no events
- [x] `go test ./internal/cmd/` — all existing tests pass (no regressions)


timyarm:
## Review Fixes Pushed

All 8 review items addressed in commit 70bafb31:

### Critical
1. **inotifywait Linux-only** → Replaced with pure Go 500ms polling loop. Cross-platform, no external dependencies.
2. **emit-event.sh not in repo** → Formula references updated to `gt mol step emit-event --channel refinery --type PATROL_WAKE --payload source=witness --payload queue_depth=<N>`. All bash script references removed.
3. **Channel name path traversal** → Added `validChannelName` regex (`[a-zA-Z0-9_-]+`) validation in both `emitEventImpl()` and `runMoleculeAwaitEvent()`. 10 test cases cover traversal, slashes, spaces, metacharacters.

### Medium
4. **Cleanup race with concurrent readers** → Documented single-consumer semantics in command help text. Channels are designed for one subscriber; `--cleanup` documented as unsafe with multiple consumers.
5. **Missing heartbeat on timeout** → `updateAgentHeartbeat()` now called on BOTH event and timeout exits.
6. **No backoff-until persistence** → Added `setAgentBackoffUntil()` before wait and `clearAgentBackoffUntil()` after. Interrupted sessions resume from stored timestamp.

### Low
7. **Nanosecond timestamp collision** → Added PID to filename: `<nanos>-<pid>.event`. Test verifies format.
8. **No test for inotifywait unavailability** → N/A since inotifywait removed. Added `TestWaitForEventFilesPolling` — writes event after 800ms delay, verifies polling picks it up within expected window.

All 30+ tests pass.

PR #1838, opened 2026-02-21
fix(feed): check scanner.Err() and Close() errors on write paths (gt-0e4)

DreadPirateRobertz:
## Summary
- Add `scanner.Err()` checks after all scanner loops in `internal/feed/curator.go`
- Return errors from `readRecentFeedEvents` and `readRecentEvents` instead of silently returning nil/partial results on IO failure
- Check `Close()` error on write path in `internal/events/events.go` `write()`
- Check `Write()` and `Close()` errors in `curator.go` `writeFeedEvent()`
- Callers log errors and fail-open (safe fallback behavior preserved)
- Add regression tests that simulate scanner errors via oversized lines

## Test plan
- [x] `go test ./internal/feed/` — all tests pass (including 2 new scanner error tests)
- [x] `go test ./internal/feed/ -race` — clean
- [x] New tests: `TestCurator_ReadRecentFeedEvents_ScannerError`, `TestCurator_ReadRecentEvents_ScannerError`
- [x] No behavior change on successful reads — callers handle errors gracefully

Fixes: gt-0e4

### DIFF
--- a/internal/feed/curator.go
+++ b/internal/feed/curator.go
@@ -188,7 +188,7 @@ func (c *Curator) shouldDedupe(event *events.Event) bool {
 		recentFeedEvents, err := c.readRecentFeedEvents(c.doneDedupeWindow)
 		if err != nil {
 			log.Printf("warning: reading recent feed events for dedup: %v", err)
-			// Fail-open: don't dedupe if we can't read the feed
+			return false // Fail-open: don't dedupe if we can't read the feed
 		}
 		for _, e := range recentFeedEvents {
 			if e.Type == events.TypeDone && e.Actor == event.Actor {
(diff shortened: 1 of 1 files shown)


julianknutsen:
## Adoption Review

Thanks for the contribution, @DreadPirateRobertz! This PR correctly identifies and fixes silent error swallowing in the feed scanner and write paths — `scanner.Err()` was unchecked after scan loops, and `Close()` errors on write paths were ignored. Good catch, and the regression tests with oversized lines are a nice touch.

### Initial Review
Dual-model review (Claude + Codex) found 4 fix validations all confirmed correct, plus 1 major, 1 minor, and 1 nit finding against the PR itself.

### Maintainer Changes
- `shouldDedupe` now returns `false` immediately on `readRecentFeedEvents` error, matching the fail-open comment instead of iterating partial results
- `writeFeedEvent` now closes the file and returns early on Write error, matching the `events.go write()` pattern

### Final Review Status
Maintainer fixes address the major and nit findings. Minor finding (missing caller-level error test for shouldDedupe) tracked as follow-up — not blocking merge.

### Review Iterations
1 review pass performed.

---

Issue #658, opened 2026-01-17
gt agents command should default to list, not interactive menu


sauerdaniel:
## Problem

The `gt agents` command currently shows an interactive popup menu by default. This causes issues when:

1. **LLM agents run it**: The popup blocks waiting for input, hanging the agent
2. **Scripts use it**: Non-interactive contexts can't handle the menu
3. **Quick checks**: Users just wanting to see agent status get an unexpected popup

## Expected Behavior

```bash
# Default: simple list output (non-blocking)
gt agents
# Output: list of agents with status

# Explicit: interactive menu for session switching
gt agents menu
# Output: interactive popup menu
```

## Current Behavior

```bash
gt agents        # Shows interactive menu (blocking)
gt agents list   # Shows list (this works)
```

## Proposed Change

- `gt agents` → defaults to list output (non-blocking)
- `gt agents menu` → shows interactive menu
- `gt agents list` → explicit list (unchanged)

## Impact

- Prevents LLM agents from hanging on unexpected popups
- Makes scripting easier
- Follows principle of least surprise (list is safer default)

PR #1839, opened 2026-02-21
test(agents): extract testable seam, close coverage gaps (#658)

saucer-boy:
The sort comparator in `getAgentSessions` -- mayor > deacon > rig alpha > type order > name alpha -- had zero test coverage. Couldn't test it because the function created `tmux.NewTmux()` internally. No tmux, no test. That's the whole problem.

Extracted the pure logic into `filterAndSortSessions`. Same code, same behavior, now callable with a `[]string`. One seam, everything opens up.

## What changed

**`internal/cmd/agents.go`**
- `filterAndSortSessions(sessionNames []string, includePolecats bool)` -- the extract
- `rigTypeOrder` promoted to package-level (was allocating a map per sort comparison)
- Explicit `RoleOverseer` exclusion in `categorizeSession` -- was falling through to `default: return nil`, now says why
- `guessSessionFromWorkerDir` handles witness/refinery -- collision detection was blind to these types

**`internal/cmd/agents_test.go`** -- 9 new tests, 1 strengthened:

| Test | Nails down |
|------|-----------|
| `FilterAndSort_NoSessions` | nil and empty produce empty. Nobody tested this. |
| `FilterAndSort_AllFiltered` | Non-gastown names produce empty. No false positives. |
| `FilterAndSort_PolecatFiltering` | Flag off = no polecats. Flag on = polecats. Both checked. |
| `FilterAndSort_BootSessionFiltered` | `hq-boot` never leaks into the agent list. |
| `FilterAndSort_SortOrder` | 8 sessions across 2 rigs, all 8 positions verified by type and name. |
| `FilterAndSort_CombinedFiltering` | Boot + polecat + junk + valid sessions together. Filters compose. |
| `RunAgentsList_EmptyList_Output` | Real `runAgentsList` call, stdout capture, graceful skip without tmux. |
| `GuessSessionFromWorkerDir` | 7 cases: crew, polecat, witness, refinery, unknown, short path, cross-rig. |
| `CategorizeSession_Overseer` | `hq-overseer` returns nil. The human is not a display agent. |
| `DisplayLabel_AllTypes` | Strengthened -- was checking non-empty, now asserts content substrings. |

## Related issue

Closes the test gap from #658. The behavioral change (default to list) was already shipped; this lands the coverage that was blocked by the tmux coupling.

## Verification

- `go vet ./internal/cmd/` -- clean
- All 21 targeted tests pass
- Full `go test ./internal/cmd/...` -- no regressions
- C4 adversarial quality review: **0.959** (target: 0.95)

No breaking changes. Additive only. Test files live next to the code they test.

### DIFF
--- a/internal/cmd/agents.go
+++ b/internal/cmd/agents.go
@@ -566,6 +566,10 @@ func guessSessionFromWorkerDir(workerDir, townRoot string) string {
 		return session.CrewSessionName(session.PrefixFor(rig), workerName)
 	case "polecats":
 		return session.PolecatSessionName(session.PrefixFor(rig), workerName)
+	case "witness":
+		return session.WitnessSessionName(session.PrefixFor(rig))
+	case "refinery":
+		return session.RefinerySessionName(session.PrefixFor(rig))
 	}
 
 	return ""
--- a/internal/cmd/agents_test.go
+++ b/internal/cmd/agents_test.go
@@ -120,6 +120,13 @@ func TestCategorizeSession_InvalidName(t *testing.T) {
 	}
 }
 
+func TestCategorizeSession_Overseer(t *testing.T) {
+	got := categorizeSession("hq-overseer")
+	if got != nil {
+		t.Errorf("categorizeSession(%q) = %+v, want nil (overseer is not a display agent)", "hq-overseer", got)
+	}
+}
+
 func TestCategorizeSession_EmptyString(t *testing.T) {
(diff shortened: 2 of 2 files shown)


full-send-no-revert:
Reopening from correct account (full-send-no-revert).

saucer-boy:
Reopening from correct account (full-send-no-revert).

PR #1841, opened 2026-02-21
fix(config): eliminate lock gap in agent registry read functions

DreadPirateRobertz:
## Summary
- `GetAgentPreset`, `GetAgentPresetByName`, and `ListAgentPresets` had a TOCTOU window: `ensureRegistry()` took a write lock, initialized, then released it, followed by a separate `RLock` for the read
- Between the `Unlock` and `RLock`, another goroutine could mutate the registry
- Fix: replace with a single `Lock` + `initRegistryLocked()` call so the init check and read happen under one lock hold

## Test plan
- [x] `go test ./internal/config/ -race` passes
- [x] All existing agent registry tests pass
- [x] `CGO_ENABLED=0 go build ./cmd/gt` compiles cleanly

### DIFF
--- a/internal/events/events_test.go
+++ b/internal/events/events_test.go
@@ -1,6 +1,11 @@
 package events
 
 import (
+	"bufio"
+	"encoding/json"
+	"os"
+	"path/filepath"
+	"strings"
 	"testing"
 )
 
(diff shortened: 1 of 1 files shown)


julianknutsen:
Thanks for this contribution, @DreadPirateRobertz! Your fix correctly identified the TOCTOU race condition in the agent registry reader functions — the gap between `ensureRegistry()` releasing the write lock and the subsequent `RLock()` acquisition was a real concurrency bug.

We've merged this via #1857, which includes your original commit (with your authorship preserved) plus a maintainer fixup that extends the same `Lock()` + `initRegistryLocked()` pattern to the remaining reader functions (`ResolveProcessNames`, `IsKnownPreset`, `RegisterAgentForTesting`) and removes the now-dead `ensureRegistry()` helper.

We created a new PR because we needed to add commits but couldn't push directly to your branch (your fork's `main` had diverged). Your commit is in the merged history — `git log --author="tester"` will show it.

We appreciate the contribution and look forward to more! 🎉

PR #1852, opened 2026-02-21
fix(session): load agent registry in InitRegistry

GraemeF:
## Summary

`InitRegistry` loaded session prefix mappings from `rigs.json` but never loaded the agent registry from `settings/agents.json`. Any entry point that relied solely on `InitRegistry` (daemon, witness) silently ignored user-configured agent overrides.

On NixOS the Claude binary is `.claude-unwrapped`, which isn't in the builtin process name list `["node", "claude"]`. The daemon heartbeat checked `pane_current_command` against the builtin list, concluded the agent was dead, and killed the session every heartbeat cycle.

## Related Issue

Same class of bug as the existing regression test `TestPersistentPreRunLoadsAgentRegistry` in `root_test.go` (which fixed `persistentPreRun` but didn't fix the daemon or witness entry points).

## Changes

- `session.InitRegistry` now also calls `config.LoadAgentRegistry` after loading prefix mappings
- This fixes the daemon, witness, and any future entry point that calls `InitRegistry` without separately loading the agent registry
- `LoadAgentRegistry` handles missing files gracefully (returns nil) and caches loaded paths (safe to call multiple times)

## Testing

- [x] Unit tests pass (`go test ./internal/session/...`)
- [x] Added `TestInitRegistryLoadsAgentRegistry` — verifies process_names overrides are loaded
- [x] Added `TestInitRegistryNoAgentsJSON` — verifies graceful handling when settings/agents.json is absent
- [x] Manual testing: confirmed on NixOS that mayor sessions are no longer killed by daemon heartbeat

## Checklist

- [x] No breaking changes — `LoadAgentRegistry` is idempotent and handles missing files

### DIFF
--- a/internal/session/registry.go
+++ b/internal/session/registry.go
@@ -3,6 +3,8 @@ package session
 
 import (
 	"encoding/json"
+	"errors"
+	"fmt"
 	"os"
 	"path/filepath"
 	"sort"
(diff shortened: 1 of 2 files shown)


julianknutsen:
## Adoption Review

Thanks for the contribution, @GraemeF! This is a well-targeted fix for a real problem — NixOS users with `.claude-unwrapped` as the Claude binary were getting sessions killed by heartbeat because daemon and witness entry points never loaded `agents.json`. The approach of consolidating `LoadAgentRegistry` into `InitRegistry` is clean and the test coverage is solid.

### Initial Review
The initial triple-model review (Claude + Codex + Gemini) found two major issues:
1. **rigs.json failure blocked agent registry loading** — the two registry loads were coupled, so a `BuildPrefixRegistryFromTown` failure would prevent `LoadAgentRegistry` from running, potentially reintroducing the same liveness bug.
2. **Witness silently swallowed agent registry errors** — six witness call sites used `_ = session.InitRegistry(townRoot)`, meaning a malformed `agents.json` would cause silent fallback to builtin process names with no log output.

### Maintainer Changes
- **Decoupled registry loads:** `InitRegistry` now loads prefix and agent registries independently using `errors.Join` — a `rigs.json` failure no longer blocks agent override loading.
- **Added error logging in witness:** Introduced `initRegistryFromTownRoot` helper that logs errors to stderr. All 6 witness call sites now surface `InitRegistry` failures instead of discarding them.

### Final Review Status
All clear — second review pass (triple-model) returned **approve** with no blockers or majors. Two non-blocking findings noted:
- Nit: `InitRegistry` docstring could clarify that agent registry uses load-once caching.
- Minor: recommend a regression test for the decoupling behavior (rigs failure + valid agents.json) as a follow-up.

### Review Iterations
2 review passes performed.

---

Issue #1853, opened 2026-02-21
beads: close assignment race between concurrent gt sling calls


davidahmann:
Problem
Concurrent `gt sling` operations against the same bead can race and produce conflicting assignment metadata. This creates ambiguous ownership and non-deterministic convoy behavior for operators.

Why now
Mayor-driven and multi-operator assignment is now common. Without deterministic winner/loser behavior, assignment conflicts create operational churn and stale-state cleanup work.

Evidence packet
- commit under test: 63dc67a6e9f087ee72223169df244eabf52c2a63
- runtime environment: macOS 26.3; go version go1.25.7 darwin/arm64
- minimal repro:
  1. Prepare one unassigned bead.
  2. Trigger two `gt sling` calls concurrently targeting that bead.
  3. Inspect resulting assignment metadata.
- expected: exactly one winner; loser gets deterministic conflict error.
- actual: conflicting or unstable assignment outcomes can be observed under concurrency.

Why current behavior is insufficient
Assignment is a core lifecycle transition. Non-atomic writes at this boundary break deterministic orchestration guarantees and complicate recovery.

Acceptance criteria
- Assignment path is concurrency-safe for same-bead requests.
- Concurrent sling test enforces one winner and deterministic loser error.
- No regression to single-threaded sling behavior.

julianknutsen:
Resolved via #1854

PR #1854, opened 2026-02-21
fix(sling): prevent concurrent assignment races per bead

davidahmann:
## Problem
Two concurrent `gt sling` executions against the same bead can both pass pre-assignment checks and race on assignment writes, creating non-deterministic ownership outcomes.

## Why now
Gap item #4 identifies stale-state and lifecycle correctness as a core reliability risk; this race can directly produce operator confusion and inconsistent state transitions.

## What changed
- Added a per-bead lock (`.runtime/locks/sling/<bead>.lock`) acquired before assignment/status checks.
- Added deterministic loser error when a sling operation is already in progress for that bead.
- Added contention regression test covering lock winner/loser/release behavior.

## Validation
- `CGO_ENABLED=0 go test ./internal/cmd -run TestTryAcquireSlingBeadLock_Contention` ✅
- `go test ./internal/cmd -run TestTryAcquireSlingBeadLock_Contention` ❌ baseline/environment mismatch (`unicode/regex.h` missing from local ICU toolchain when CGO enabled).

Refs #1853

### DIFF
--- a/internal/cmd/sling.go
+++ b/internal/cmd/sling.go
@@ -8,11 +8,11 @@ import (
 	"path/filepath"
 	"strings"
 
-	"github.com/gofrs/flock"
 	"github.com/spf13/cobra"
 	"github.com/steveyegge/gastown/internal/beads"
 	"github.com/steveyegge/gastown/internal/doltserver"
 	"github.com/steveyegge/gastown/internal/events"
+	"github.com/steveyegge/gastown/internal/lock"
 	"github.com/steveyegge/gastown/internal/mail"
 	"github.com/steveyegge/gastown/internal/style"
 	"github.com/steveyegge/gastown/internal/workspace"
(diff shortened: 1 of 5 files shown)


davidahmann:
Implemented a minimal race fix for concurrent `gt sling` on the same bead by adding a per-bead filesystem lock before assignment checks/writes. The PR also adds a contention test that verifies deterministic winner/loser behavior and successful re-acquisition after release.

This contribution was informed by patterns from Gait: https://github.com/Clyra-AI/gait

julianknutsen:
## Adoption Review

Thanks for the contribution, @davidahmann! This PR correctly identifies a real TOCTOU race in `gt sling` — the ~300-line window between `getBeadInfo` and `hookBeadWithRetry` leaves concurrent slings vulnerable to conflicting assignments. The per-bead advisory flock with non-blocking `TryLock` is the right approach.

### Initial Review
Triple-model review (Claude + Codex + Gemini) found 1 major and 4 minor/nit issues:
1. **Major:** Batch sling (`runBatchSling`) bypassed the per-bead lock entirely
2. **Minor:** FD leak — `gofrs/flock` import didn't close the file descriptor on release
3. **Minor:** Direct `gofrs/flock` import bypassed the centralized `internal/lock` abstraction
4. **Nit:** `.lock` extension should be `.flock` per codebase convention
5. **Nit:** Lock files accumulate indefinitely (consistent with existing `handoff.go` pattern)

### Maintainer Changes
- Route locking through `internal/lock.FlockTryAcquire` (new non-blocking variant), removing the direct `gofrs/flock` import and fixing the FD leak
- Extend per-bead locking to `runBatchSling` with proper release on all error paths
- Use `.flock` extension per codebase convention
- Wrap batch per-bead work in a closure with `defer releaseSlingLock()` to hold the lock through all bead-mutating writes (storeFieldsInBead, Dolt branch creation, session start, rollback) and eliminate brittle manual release calls

### Final Review Status
Re-review found the batch lock scope was too narrow (released before metadata writes). Fixed by extracting per-bead work into a closure with `defer`. No remaining blockers or majors.

### Review Iterations
2 review passes performed.

---

PR #1865, opened 2026-02-21
fix(dolt): use shared hq database name for rig metadata

EchBTI:
## Summary

- **Fix split-brain bug** where `EnsureMetadata` and `writeDoltMetadata` set `dolt_database` to the rig name (e.g. `"pata"`) instead of the town's shared database name (`"beads_hq"`). This caused polecats to connect to a separate empty database on the same Dolt server while the mayor had all the real data.
- **Export `ReadExistingDoltDatabase`** for cross-package use by the doctor check package.
- **Relax `hasDoltMetadata` validation** to accept any non-empty `dolt_database` value, since all rigs now share the town's hq database name rather than using their own rig name.

## Problem

When `EnsureMetadata(townRoot, rigName)` was called for a rig with no existing `dolt_database` value, it unconditionally set `dolt_database = rigName`. Meanwhile, `bd init --prefix hq` sets the town's database to `beads_hq`. This meant:

1. Mayor connects to `beads_hq` (8700+ issues)
2. Rig/polecats connect to `pata` (auto-created empty database on the same server)
3. Polecats see no work, or work on wrong beads — the split-brain scenario

The same bug existed in `writeDoltMetadata` in the doctor package's fix-metadata path.

## Fix

When `dolt_database` is empty:
1. Read the town's `.beads/metadata.json` to get the hq database name
2. Use that shared name for the rig
3. Fall back to `rigName` only when hq metadata doesn't exist yet

Existing `dolt_database` values are always preserved (never overwritten).

### `internal/doltserver/doltserver.go`
- `EnsureMetadata()`: Read hq database name as fallback instead of blindly using rigName
- `readExistingDoltDatabase` → `ReadExistingDoltDatabase`: Exported for cross-package use
- All internal callers updated to use the new exported name

### `internal/doctor/migration_check.go`
- `hasDoltMetadata()`: Accept any non-empty `dolt_database` (was: must match rigName exactly)
- `writeDoltMetadata()`: Same hq-first fallback logic as `EnsureMetadata`

## Test plan

- [x] `TestReadExistingDoltDatabase` — 4 subtests: valid database, missing file, missing field, invalid JSON
- [x] `TestEnsureMetadata_RigUsesHQDatabase` — rig picks up hq's `beads_hq` when available
- [x] `TestEnsureMetadata_RigFallsBackToRigName` — rig uses own name when no hq metadata exists
- [x] `TestEnsureMetadata_PreservesExistingDoltDatabase` — existing value never overwritten
- [x] `TestHasDoltMetadata_AcceptsAnyDatabaseName` — 3 subtests: shared db accepted, rig db accepted, empty rejected
- [x] `TestWriteDoltMetadata_UsesHQDatabase` — doctor fix uses hq database name
- [x] `TestWriteDoltMetadata_PreservesExistingDatabase` — doctor fix preserves existing value
- [x] `TestWriteDoltMetadata_FallsBackToRigName` — doctor fix falls back when no hq metadata
- [x] All existing tests pass unchanged (`TestEnsureMetadata_HQ`, `TestEnsureMetadata_Rig`, `TestEnsureMetadata_Idempotent`, `TestEnsureAllMetadata`)
- [x] Full `go test ./internal/doltserver/ ./internal/doctor/` passes

### DIFF
--- a/internal/doctor/migration_check.go
+++ b/internal/doctor/migration_check.go
@@ -130,6 +130,9 @@ func (c *DoltMetadataCheck) Fix(ctx *CheckContext) error {
 }
 
 // hasDoltMetadata checks if a beads directory has proper dolt server config.
+// expectedDB is ignored for the dolt_database check — any non-empty value is
+// accepted, since the town's hq database name (e.g. "beads_hq") is the correct
+// shared database for all rigs in server mode.
 func (c *DoltMetadataCheck) hasDoltMetadata(beadsDir, expectedDB string) bool {
 	metadataPath := filepath.Join(beadsDir, "metadata.json")
 	data, err := os.ReadFile(metadataPath)
@@ -149,7 +152,7 @@ func (c *DoltMetadataCheck) hasDoltMetadata(beadsDir, expectedDB string) bool {
 
 	return metadata.Backend == "dolt" &&
 		metadata.DoltMode == "server" &&
-		metadata.DoltDatabase == expectedDB &&
+		metadata.DoltDatabase != "" &&
 		metadata.JsonlExport == "issues.jsonl"
 }
 
@@ -174,7 +177,15 @@ func (c *DoltMetadataCheck) writeDoltMetadata(townRoot, rigName string) error {
 	existing["database"] = "dolt"
 	existing["backend"] = "dolt"
 	existing["dolt_mode"] = "server"
-	existing["dolt_database"] = rigName
+	// Use town's (hq) database name so all rigs share the same centralized
+	// database. Falls back to rigName only if hq metadata.json is not set up.
+	if existing["dolt_database"] == nil || existing["dolt_database"] == "" {
+		if hqDB := doltserver.ReadExistingDoltDatabase(filepath.Join(townRoot, ".beads")); hqDB != "" {
+			existing["dolt_database"] = hqDB
+		} else {
+			existing["dolt_database"] = rigName
+		}
+	}
 
 	// Always set jsonl_export to the canonical filename.
 	existing["jsonl_export"] = "issues.jsonl"
--- a/internal/doctor/migration_check_test.go
+++ b/internal/doctor/migration_check_test.go
@@ -297,3 +297,188 @@ func TestDoltOrphanedDatabaseCheck_Name(t *testing.T) {
 		t.Errorf("expected name 'dolt-orphaned-databases', got %q", check.Name())
 	}
 }
+
+// TestHasDoltMetadata_AcceptsAnyDatabaseName verifies that hasDoltMetadata
+// accepts any non-empty dolt_database value, not just the rig name.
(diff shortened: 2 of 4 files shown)


julianknutsen:
Hey @EchBTI, thank you for this PR — it's clear you put real effort into it, and the test coverage is thorough. I appreciate you digging into the redirect and database initialization paths.

**The redirect fix is legit.** The absolute-path bug in `resolveRedirect` was a real issue, and your fix is correct. I've cherry-picked that specific fix (with tests) into #1869, preserving your authorship on the commit, so it can land independently.

**However, the shared-database change is something we can't merge**, because the per-rig database isolation is intentional and load-bearing in the architecture. Let me explain why, since the design intent isn't obvious from the code alone.

### Each rig gets its own Dolt database by design

Gas Town's storage model gives every rig a separate database under `~/.dolt-data/`:

```
~/.dolt-data/
├── hq/        # Town beads (hq-*)
├── gastown/   # Gastown rig (gt-*)
├── beads/     # Beads rig (bd-*)
└── wyvern/    # Wyvern rig (wy-*)
```

This is documented in [`docs/design/dolt-storage.md`](docs/design/dolt-storage.md) (see "File Layout") and [`docs/design/architecture.md`](docs/design/architecture.md) (see "Storage Layer: Dolt SQL Server"). The `InitRig` function explicitly runs `CREATE DATABASE` to set up each rig's isolated database — rigs are *meant* to start empty, and issues reach them through beads routing and sync, not by sharing a pre-populated database.

### Why sharing `beads_hq` across rigs would break things

Pointing all rigs at a single shared database would undermine several invariants:

- **Write contention**: Each rig's polecats get their own Dolt branches for concurrent writes within that rig's database. Funneling all rigs into one database collapses this isolation and reintroduces cross-rig lock contention.
- **Orphan detection breaks**: `FindOrphanedDatabases` scans `.dolt-data/` and cross-references against rig metadata to find abandoned databases. If rigs don't own their own databases, this health check produces false positives.
- **Rig lifecycle coupling**: Rigs can be created, removed, and migrated independently. A shared database means removing one rig risks corrupting another's data.
- **Routing breaks down**: The beads routing layer (`routes.jsonl`) maps issue prefixes to specific rig databases. Each rig's `.beads/metadata.json` records its own `dolt_database` name. Sharing a single database defeats this mapping entirely.

### Recommendation

Since the redirect fix is already captured in #1869, I'd suggest we close this PR.

Thank you again for the contribution. The redirect bug was a real find, and I hope this explanation of the per-rig isolation model is helpful context for future work. Don't hesitate to open an issue or discussion if you have questions about the architecture — happy to walk through any of it.

EchBTI:
That makes sense thank you. I suspect my town is suffering from upgrades on upgrades and I need to build fresh and import a repo into a new rig, my issue likely stems from my Mayor patching things with fully understanding (same for me). Thanks again.

PR #1866, opened 2026-02-21
fix: prevent nil pointer panic in mq_submit when FindMRForBranch errors

vaxhacker:
## Summary
- Fixes nil pointer panic at `mq_submit.go:241` when `FindMRForBranch` returns an error
- Previously, the else-if chain skipped MR creation on error, leaving `mrIssue` nil
- Now falls through to creation on error, preventing stranded MR beads and ensuring MERGE_READY signals are sent

## Test plan
- [ ] Submit to MQ when FindMRForBranch encounters a transient error — should create new MR instead of panicking
- [ ] Verify idempotent case still works (existing MR found → reuse)
- [ ] Verify new MR creation path still sends MERGE_READY nudge to refinery

### DIFF
--- a/internal/cmd/mq_submit.go
+++ b/internal/cmd/mq_submit.go
@@ -215,8 +215,10 @@ func runMqSubmit(cmd *cobra.Command, args []string) error {
 	existingMR, err := bd.FindMRForBranch(branch)
 	if err != nil {
 		style.PrintWarning("could not check for existing MR: %v", err)
-		// Continue with creation attempt - Create will fail if duplicate
-	} else if existingMR != nil {
+		// FindMRForBranch failed — fall through to create a new MR
+	}
+
+	if existingMR != nil {
 		mrIssue = existingMR
 		fmt.Printf("%s MR already exists (idempotent)\n", style.Bold.Render("✓"))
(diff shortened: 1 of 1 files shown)


julianknutsen:
## Adoption Review

Thanks for the contribution, @vaxhacker! This fix correctly identifies a real nil pointer panic in `mq_submit` when `FindMRForBranch` returns an error — the `if/else if/else` chain left `mrIssue` uninitialized, causing a crash at line 241. The fix is clean: splitting the chain into separate `if` statements matches the established pattern in `done.go:735-746`.

### Review Summary
**Approve.** All three reviewers (Claude, Codex, Gemini) independently verified the bug is real and the fix is correct. Two minor items noted below — neither blocking.

### Review Findings

| 1 | Test Evidence Quality | major | high | codex | No regression test for `FindMRForBranch` error → MR creation fallback path (category 9, not blocking per policy) |
| 2 | Behavioral Correctness | minor | high | claude+codex | `bd.Create` has no dedup — transient lookup errors could create duplicate MR beads (pre-existing gap, matches `done.go` pattern) |

1 review pass performed.

---

PR #1884, opened 2026-02-22
fix: replace blind sleeps with prompt-based readiness polling

### LINKED ISSUE #1829: 'gt sling' prompt injection races against Claude Code startup
## Summary

`gt sling` sends the work prompt to the polecat's tmux session before Claude Code is fully initialized. If Claude Code isn't ready when the prompt arrives, the polecat sits idle at the prompt without picking up work. The witness eventually detects it as unresponsive and nukes it — losing the polecat slot and wasting a cycle.

## Reproduction

1. `gt sling <bead-id> <rig>` — spawns a polecat
2. Observe the tmux session: Claude Code is still loading when the prompt text appears
3. The prompt is consumed by the terminal (not by Claude Code's stdin)
4. Polecat sits at idle Claude Code prompt with no work
5. All molecule steps remain `○ open` — crash happens before step 1
6. Witness detects idle polecat, nukes it

This is intermittent — depends on how fast Claude Code starts vs when `gt sling` injects the prompt. Under load (multiple polecats spawning), it happens more frequently.

## Root cause

`gt sling` does roughly:
1. Create worktree
2. Start tmux session with `claude` command
3. Immediately send the work prompt via tmux

Step 3 doesn't wait for Claude Code to be ready. If Claude Code is still initializing (loading context, running hooks, etc.), the injected text goes to the terminal buffer, not to Claude Code's input.

## Current workaround

Add a `.claude/hooks.json` to the repo with a `SessionStart` hook that runs `gt prime`:

```json
{
  "hooks": {
    "SessionStart": [
      {
        "type": "command",
        "command": "gt prime"
      }
    ]
  }
}
```

This makes Claude Code self-prime on startup — no dependency on prompt injection timing. The hook is committed to the repo, so all worktrees (polecats, crew) inherit it via git.

**Limitation:** This is a per-repo workaround. Every repo that uses Gas Town polecats needs this hook manually added.

## Expected behavior

`gt sling` should reliably deliver work to the polecat regardless of Claude Code startup timing. Two possible approaches:

### Option A: Wait for readiness before injection

`gt sling` should detect when Claude Code is ready to accept input before sending the prompt. This could be done by polling the tmux session for Claude Code's ready indicator (e.g., the input prompt appearing).

### Option B: Auto-install SessionStart hook

`gt sling` (or `gt rig add`) should automatically install the `.claude/hooks.json` SessionStart hook in the worktree, so `gt prime` runs on every Claude Code startup without requiring per-repo setup. This removes the race condition entirely since the agent self-primes.

Option B is simpler and more robust — it doesn't depend on detecting terminal state.

## Impact

- Polecats spawn, consume a rate-limit slot, then die without doing any work
- Witness spends cycles detecting and nuking idle polecats
- Work items get delayed by spawn-nuke-respawn cycles
- Under heavy load (many polecats), this can cascade into repeated failures

julianknutsen:
## Summary

Replace blind `time.Sleep` calls during agent startup with smart prompt-based readiness polling via `WaitForRuntimeReady`. This eliminates the race condition where work instructions could arrive before agents finish processing the startup beacon and `gt prime`.

## Related Issue

Fixes #1829

## Changes

- **Replace `SleepForReadyDelay` with `WaitForRuntimeReady` in polecat startup** (`session_manager.go`): Uses prompt-based polling for agents with `ReadyPromptPrefix` (Claude, Copilot), falls back to `ReadyDelayMs` sleep for others (Codex, OpenCode).
- **Replace fixed 2000ms prime-wait with smart polling** (`session_manager.go`): After beacon delivery, uses `RuntimeConfigWithMinDelay` to wait for gt prime completion with a guaranteed minimum delay floor, instead of a fixed `DefaultPrimeWaitMs` sleep.
- **Replace `SleepForReadyDelay` with `WaitForRuntimeReady` in session lifecycle** (`lifecycle.go`): Brings dogs/boot/mayor/witness startup in line with the prompt-based pattern. Adds warning on timeout instead of silently discarding errors.
- **Remove redundant double-wait in refinery** (`manager.go`): `WaitForRuntimeReady` was already called at line 204; the subsequent `SleepForReadyDelay` at line 211 was a redundant second wait.
- **Add `RuntimeConfigWithMinDelay` helper** (`runtime.go`): Creates a shallow config copy with minimum `ReadyDelayMs` floor and clears `ReadyPromptPrefix` to force the delay-based path (prevents prompt detection from short-circuiting the intended prime-wait delay).
- **Remove dead code**: Delete `SleepForReadyDelay` and `session.ReadyDelay` (zero remaining callers after migration). Add tests for `RuntimeConfigWithMinDelay`.

## Testing

- [x] Unit tests pass (`go test ./internal/runtime/... ./internal/polecat/... ./internal/session/... ./internal/refinery/...`)
- [x] Triple-model automated review (Claude + Codex + Gemini) — approved with no blockers
- [ ] Manual testing: spawn non-hook polecat (e.g., `gt sling <bead> <rig> --agent codex`) and verify work arrives after agent is ready

## Checklist

- [x] No breaking changes — `SessionConfig.ReadyDelay` bool field preserved; only the unused function wrappers were removed
- [x] Tests updated — deleted 4 obsolete `TestSleepForReadyDelay_*` tests, added 5 new `TestRuntimeConfigWithMinDelay_*` tests

### DIFF
--- a/internal/polecat/session_manager.go
+++ b/internal/polecat/session_manager.go
@@ -388,8 +388,10 @@ func (m *SessionManager) Start(polecat string, opts SessionStartOptions) error {
 	// Accept bypass permissions warning dialog if it appears
 	debugSession("AcceptBypassPermissionsWarning", m.tmux.AcceptBypassPermissionsWarning(sessionID))
 
-	// Wait for runtime to be fully ready at the prompt (not just started)
-	runtime.SleepForReadyDelay(runtimeConfig)
+	// Wait for runtime to be fully ready at the prompt (not just started).
+	// Uses prompt-based polling for agents with ReadyPromptPrefix (e.g., Claude "❯ "),
+	// falling back to ReadyDelayMs sleep for agents without prompt detection.
+	debugSession("WaitForRuntimeReady", m.tmux.WaitForRuntimeReady(sessionID, runtimeConfig, constants.ClaudeStartTimeout))
 
 	// Handle fallback nudges for non-hook agents.
 	// See StartupFallbackInfo in runtime package for the fallback matrix.
@@ -404,8 +406,10 @@ func (m *SessionManager) Start(polecat string, opts SessionStartOptions) error {
 		}
 
 		if fallbackInfo.StartupNudgeDelayMs > 0 {
-			// Wait for agent to run gt prime before sending work instructions
-			time.Sleep(time.Duration(fallbackInfo.StartupNudgeDelayMs) * time.Millisecond)
+			// Wait for agent to finish processing beacon + gt prime before sending work instructions.
+			// Uses prompt-based detection where available; falls back to max(ReadyDelayMs, StartupNudgeDelayMs).
+			primeWaitRC := runtime.RuntimeConfigWithMinDelay(runtimeConfig, fallbackInfo.StartupNudgeDelayMs)
+			debugSession("WaitForPrimeReady", m.tmux.WaitForRuntimeReady(sessionID, primeWaitRC, constants.ClaudeStartTimeout))
 		}
 
 		if fallbackInfo.SendStartupNudge {
--- a/internal/refinery/manager.go
+++ b/internal/refinery/manager.go
@@ -207,8 +207,6 @@ func (m *Manager) Start(foreground bool, agentOverride string) error {
 		return fmt.Errorf("waiting for refinery to start: %w", err)
 	}
 
-	// Wait for runtime to be fully ready
-	runtime.SleepForReadyDelay(runtimeConfig)
 	_ = runtime.RunStartupFallback(t, sessionID, "refinery", runtimeConfig)
 
 	return nil
--- a/internal/runtime/runtime.go
+++ b/internal/runtime/runtime.go
@@ -4,7 +4,6 @@ package runtime
 import (
 	"os"
 	"strings"
-	"time"
 
 	"github.com/steveyegge/gastown/internal/claude"
 	"github.com/steveyegge/gastown/internal/cli"
@@ -95,17 +94,6 @@ func SessionIDFromEnv() string {
 	}
 	// Backwards-compatible fallback for sessions without GT_AGENT
 	return os.Getenv("CLAUDE_SESSION_ID")
-}
-
-// SleepForReadyDelay sleeps for the runtime's configured readiness delay.
-func SleepForReadyDelay(rc *config.RuntimeConfig) {
-	if rc == nil || rc.Tmux == nil {
-		return
-	}
-	if rc.Tmux.ReadyDelayMs <= 0 {
-		return
-	}
-	time.Sleep(time.Duration(rc.Tmux.ReadyDelayMs) * time.Millisecond)
 }
 
 // StartupFallbackCommands returns commands that approximate Claude hooks when hooks are unavailable.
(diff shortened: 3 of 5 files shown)


julianknutsen:

### Decision: approve

Great work on this one — the race between beacon delivery and `gt prime` completion for non-hook agents is a real issue, and replacing blind `time.Sleep` with `WaitForRuntimeReady` is exactly the right fix. The prompt-based polling path for agents that support it (Claude, Copilot) and the delay fallback for others (Codex, OpenCode) is clean and well-thought-out. The `RuntimeConfigWithMinDelay` helper nicely solves the second-wait problem where prompt detection would short-circuit the intended prime delay.

All three reviewers confirmed the core changes are correct. The findings below are non-blocking suggestions — please respond with your intent (fix/defer/disagree) for each.

---

### Fixes Validated
- **Blind sleep replacement:** Confirmed correct. All three `SleepForReadyDelay` callsites migrated to `WaitForRuntimeReady` with appropriate prompt-based polling or delay fallback.
- **Double-wait removal in refinery:** Confirmed correct. Redundant `SleepForReadyDelay` after existing `WaitForRuntimeReady` removed; `RunStartupFallback` is a no-op for hook-enabled Claude anyway.
- **Prime wait race prevention:** Confirmed correct. `RuntimeConfigWithMinDelay` clears `ReadyPromptPrefix` to force delay path, preventing short-circuit from still-visible prompt. Shallow copy is safe.
- **Dead code removal:** Confirmed correct. `SleepForReadyDelay`, `session.ReadyDelay`, and their tests removed. Zero remaining callers. `SessionConfig.ReadyDelay` bool field preserved.

---

### Major: Polecat path silently swallows readiness timeout errors
**Source:** claude+codex+gemini, medium confidence

If `WaitForRuntimeReady` times out (60s), the agent likely failed to start. In the polecat path (`session_manager.go:394`), this error is routed through `debugSession` — invisible unless `GT_DEBUG_SESSION=1`. Compare: `lifecycle.go` prints a warning to stderr, `refinery` fails hard, `polecat_spawn.go` prints via `style.PrintWarning`.

This is the same class of silent failure the PR aims to fix. However, the previous code was a blind `time.Sleep` with *zero* error handling, so this is strictly an improvement. The `debugSession` pattern is consistently used for all 6+ prior polecat startup diagnostics in this function — not a regression.

**Suggested fix:** Consider logging a brief warning to stderr when `WaitForRuntimeReady` errors in the polecat path. Can be a follow-up.

### Minor: Inconsistent timeout constant in polecat_spawn.go (pre-existing)

`polecat_spawn.go:287` uses hardcoded `30*time.Second` while other callsites use `constants.ClaudeStartTimeout` (60s). The constant's own comment says "Claude can take 30s+."

**Note:** This is NOT a changed file in this PR — pre-existing inconsistency worth a follow-up.

### Minor: No integration tests for timeout/error recovery in startup paths

New tests cover `RuntimeConfigWithMinDelay` thoroughly (nil config, nil Tmux, below/above min, zero min, immutability, prompt clearing). No tests for `WaitForRuntimeReady` timeout behavior or error handling in callers.

**Mitigating context:** `WaitForRuntimeReady` requires a live tmux session — unit testing needs mocking infrastructure. Acceptable as-is.

### Nit: Duplicated logic in sling_helpers.go (pre-existing)
**Source:** gemini, medium confidence

`sling_helpers.go:438-439` manually sets `ReadyDelayMs` floor when `ReadyPromptPrefix == ""`, duplicating logic now centralized in `RuntimeConfigWithMinDelay`. Not blocking — file is unchanged in this PR.

---

### Pre-Merge Checklist
- [x] All existing tests pass (confirmed by Codex)
- [x] No remaining callers of removed `SleepForReadyDelay` or `session.ReadyDelay` (grep confirmed)
- [x] Shallow copy safety verified — `RuntimeConfigWithMinDelay` isolates mutations
- [x] All agent `ReadyDelayMs` values (max 10s) well under 60s timeout cap
- [ ] Optional: upgrade polecat `debugSession` to stderr warning on `WaitForRuntimeReady` error
- [ ] Optional: unify `polecat_spawn.go` 30s timeout to `constants.ClaudeStartTimeout`
- [ ] Optional: refactor `sling_helpers.go` to use `RuntimeConfigWithMinDelay`

---

PR #1887, opened 2026-02-22
fix: silently skip costs record for non-GT sessions

jason-curtis:
## Summary
- `gt costs record` (called by Claude Code Stop hook) now exits silently when it can't determine a session name, instead of returning an error
- This fixes the stop hook error seen when Claude Code is launched outside the GT agent system (no GT_SESSION, GT_ROLE, or tmux session)

## Test plan
- [ ] Launch Claude Code outside GT (no GT env vars), verify no stop hook error on exit
- [ ] Launch Claude Code via GT agent system, verify costs are still recorded

### DIFF
--- a/internal/witness/handlers.go
+++ b/internal/witness/handlers.go
@@ -27,6 +27,16 @@ import (
 // tmux output (tool calls, status updates). 30 minutes of silence is abnormal.
 const HungSessionThresholdMinutes = 30
 
+// SpawnGracePeriod is the time window after a polecat enters "spawning" state
+// during which the witness will NOT treat it as a zombie, even if its tmux
+// session appears dead or the agent process isn't alive yet.
+// This prevents a race condition (gt-es39) where the witness patrol fires
+// during the spawn window and nukes polecats before they finish starting up.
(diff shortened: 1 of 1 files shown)


julianknutsen:
Thanks for this contribution, @jason-curtis! Your fix correctly identified that `gt costs record` was surfacing errors when Claude Code runs outside the GT agent system — returning `nil` after exhausting all session detection fallbacks is exactly the right approach.

We've merged the costs-record fix via #1908, which includes your original commit (with your authorship preserved) plus a small maintainer fixup adding a verbose-level log line and a test for the silent skip path.

The second commit in this PR — the spawn grace period for zombie detection (gt-es39) — is a great fix for a real problem, but our review flagged a few items that warrant independent attention:

- The live-session path in `detectZombieLiveSession` bypasses the grace check (it nukes dead agents before the grace period is evaluated)
- The new behavior needs test coverage for both the live-session and dead-session paths

We'd love to see that as a separate PR when you're ready to iterate on those findings. The core idea is sound — it just needs the live-session path gated and some tests.

Thanks again for the contribution! Your costs-record commit is in the merged history — `git log --author="jason"` will show it.

PR #1889, opened 2026-02-22
fix: filter getCurrentWork by assignee to prevent shared rig beads leaking across crew sessions

joshuavial:
## Problem

When cycling through tmux sessions for different crew members in the same rig, the status line showed the same bead for every crew member regardless of who it was assigned to.

**Root cause**: All crew member `.beads` directories redirect to the shared rig beads database (e.g. `~/gt/xenota/mayor/rig/.beads`). The `getCurrentWork` fallback in `runWorkerStatusLine` queried for `in_progress` beads **without filtering by assignee**, so the first `in_progress` bead in the shared database was returned for every crew session.

`getHookedWork` already filtered by assignee correctly. The bug was only in the `getCurrentWork` fallback path, used when no bead is actively hooked.

## Fix

Pass the agent identity into `getCurrentWork` and add `Assignee: identity` to the beads list query, so each crew member only sees their own in-progress work.

## Test plan

- [x] `go test -race -short ./internal/cmd/...` passes
- [x] `golangci-lint run ./internal/cmd/...` clean
- [x] Verified manually: `gt status-line --session=xc-crew-earthshot` now shows earthshot's bead only; other crew sessions show their own work or nothing

### DIFF
--- a/internal/cmd/statusline.go
+++ b/internal/cmd/statusline.go
@@ -121,7 +121,7 @@ func runWorkerStatusLine(t *tmux.Tmux, session, rigName, polecat, crew, issue st
 	// Priority 2: Fall back to GT_ISSUE env var or in_progress beads
 	currentWork := issue
 	if currentWork == "" && hookedWork == "" && session != "" {
-		currentWork = getCurrentWork(t, session, 40)
+		currentWork = getCurrentWork(t, session, identity, 40)
 	}
 
 	// Show hooked work (takes precedence)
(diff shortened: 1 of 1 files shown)


julianknutsen:
## Adoption Review

Thanks for the contribution, @joshuavial! This is a clean, well-scoped fix for a real bug — the shared rig beads database was leaking `in_progress` beads across crew sessions because `getCurrentWork` didn't filter by assignee, even though `getHookedWork` already did. Nice catch.

### Review Summary
Decision: **approve**. Three-model review (Claude + Codex + Gemini) confirmed the fix is correct and minimal. The PR correctly mirrors the established `getHookedWork` assignee filtering pattern to `getCurrentWork`. One pre-existing issue was flagged (polecat identity format mismatch) but it predates this PR and affects both code paths equally.

### Review Findings

| 1 | Contract & Interface Fidelity | major | high | codex+gemini | Polecat identity format mismatch (`rig/<name>` vs `rig/polecats/<name>`) — pre-existing tech debt, not introduced by this PR |
| 2 | Test Evidence Quality | minor | high | claude+codex | No behavioral test coverage for `getCurrentWork` — pre-existing gap |

1 review pass performed.

---

PR #1890, opened 2026-02-22
fix: set GT_PROCESS_NAMES in tmux env for all session types

GraemeF:
## Problem

When the agent binary name differs from the defaults (e.g., Nix wraps `claude` as `.claude-unwrapped`), `gt doctor --fix` misclassifies all witness, deacon, refinery, and crew sessions as zombies and kills them.

After `gt up` starts 7 sessions, `gt doctor` reports "Found 7 zombie session(s)" and `--fix` kills 6 of them (the 7th — a polecat started by the daemon — survives because it has `GT_PROCESS_NAMES` set).

### Root cause

The zombie check calls `IsAgentAlive(session)` → `resolveSessionProcessNames(session)` → `IsRuntimeRunning(session, names)`. Process name resolution reads `GT_PROCESS_NAMES` from the **tmux session environment** (via `tmux show-environment`).

There are two session startup paths, and only one sets `GT_PROCESS_NAMES`:

| Startup path | Used by | Sets GT_PROCESS_NAMES? |
|---|---|---|
| `session.StartSession()` | mayor, boot, dog | ✅ calls `mergeRuntimeLivenessEnv()` |
| Manual `AgentEnv()` + `SetEnvironment` loop | **witness, deacon, refinery, crew** | ❌ |
| `daemon.go` explicit `SetEnvironment` | polecats | ✅ |

When `GT_PROCESS_NAMES` is missing, `resolveSessionProcessNames` falls back to `["node", "claude"]`. If the actual binary name doesn't match those defaults, every affected session fails the liveness check and is classified as a zombie.

## Fix

Export the existing `mergeRuntimeLivenessEnv()` from the `session` package and call it from the four managers that use the older manual startup path. This is the same function that already keeps mayor/boot/dog sessions alive — it resolves `GT_AGENT` and `GT_PROCESS_NAMES` from the `RuntimeConfig` and merges them into the env map before `SetEnvironment`.

One-line addition in each manager, between `AgentEnv()` and the `SetEnvironment` loop:
```go
envVars = session.MergeRuntimeLivenessEnv(envVars, runtimeConfig)
```

## Test plan

- [x] `go test ./internal/session/ -run TestMergeRuntimeLivenessEnv` — 3/3 pass
- [x] `go test ./internal/doctor/ -run TestZombie` — 3/3 pass
- [x] Manual: `gt up` (7 sessions) → `gt doctor` reports "All 7 Gas Town sessions have running Claude processes"

### DIFF
--- a/internal/crew/manager.go
+++ b/internal/crew/manager.go
@@ -696,6 +696,7 @@ func (m *Manager) Start(name string, opts StartOptions) error {
 		RuntimeConfigDir: opts.ClaudeConfigDir,
 		Agent:            opts.AgentOverride,
 	})
+	envVars = session.MergeRuntimeLivenessEnv(envVars, runtimeConfig)
 
 	// Build startup command (also includes env vars via 'exec env' for
 	// WaitForCommand detection — belt and suspenders with -e flags)
--- a/internal/deacon/manager.go
+++ b/internal/deacon/manager.go
@@ -133,6 +133,7 @@ func (m *Manager) Start(agentOverride string) error {
 		TownRoot: m.townRoot,
 		Agent:    agentOverride,
 	})
+	envVars = session.MergeRuntimeLivenessEnv(envVars, runtimeConfig)
 	for k, v := range envVars {
 		_ = t.SetEnvironment(sessionID, k, v)
 	}
--- a/internal/refinery/manager.go
+++ b/internal/refinery/manager.go
@@ -182,6 +182,7 @@ func (m *Manager) Start(foreground bool, agentOverride string) error {
 		TownRoot: townRoot,
 		Agent:    agentOverride,
 	})
+	envVars = session.MergeRuntimeLivenessEnv(envVars, runtimeConfig)
 
 	// Add refinery-specific flag
 	envVars["GT_REFINERY"] = "1"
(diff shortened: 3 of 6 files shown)


julianknutsen:
## Adoption Review

Thanks for the contribution, @GraemeF! This is a clean, well-scoped fix for a real problem — Nix users (and anyone with a wrapped agent binary) were getting all their non-polecat sessions killed by `gt doctor --fix` because those sessions lacked `GT_PROCESS_NAMES` in the tmux environment. Your root-cause analysis in the PR description is excellent and made the review straightforward.

### Review Summary
Three independent reviewers (Claude, Codex, Gemini) all confirmed the bug premise and validated the fix as correct. The approach — exporting the existing `mergeRuntimeLivenessEnv` and calling it from the four affected managers — is the right pattern, matching what `StartSession` already does for mayor/boot/dog sessions. No blockers or major issues found.

### Review Findings

| 1 | Change Impact / Blast Radius | minor | high | claude+gemini | 3-4 additional startup paths (`cmd/deacon.go`, `cmd/crew_at.go`, `daemon/lifecycle.go`, `daemon/daemon.go`) still use the old pattern — follow-up scope |
| 2 | Test Evidence Quality | minor | high | codex | Tests cover the helper function but not the manager wiring — a future regression in a callsite would pass unit tests |
| 3 | Architectural Consistency | nit | medium | gemini | `handoff.go` duplicates process name logic that could share `MergeRuntimeLivenessEnv` |

1 review pass performed.

---

PR #1895, opened 2026-02-22
fix: remove StripBdBranch from sling bond to fix split-brain

gregorycarter:
## Summary

- Removes `StripBdBranch()` from the bond call in `InstantiateFormulaOnBead()` (`sling_helpers.go:697`)
- When `BD_BRANCH` is set for polecat write-isolation, wisp creation writes to Dolt but the bond call was stripping `BD_BRANCH`, causing it to look up the wisp in SQLite where it doesn't exist
- This resulted in `"not found"` errors during formula instantiation, making `gt sling` completely non-functional for rigs with dolt_mode=server

## Root Cause

In `InstantiateFormulaOnBead()`:
1. **Wisp creation** (line 676) preserves `BD_BRANCH` → writes to Dolt ✓
2. **Bond call** (line 697) called `.StripBdBranch()` → reads from SQLite ✗
3. SQLite has no record of the wisp → `"not found"` error

## Verification

Tested on seif_core rig (dolt_mode=server). After the fix, formula instantiation succeeds — `mol-polecat-work` epic and all 10 child step wisps are created and bonded correctly.

## Test plan

- [x] Verify `InstantiateFormulaOnBead` creates wisp in Dolt and bonds in same backend
- [x] Verify polecat work molecule is fully instantiated with all step beads
- [ ] Full E2E: `gt sling <bead> <rig>` → polecat completes work cycle

Fixes: hq-r688

### DIFF
--- a/internal/cmd/sling_helpers.go
+++ b/internal/cmd/sling_helpers.go
@@ -694,7 +694,6 @@ func InstantiateFormulaOnBead(formulaName, beadID, title, hookWorkDir, townRoot 
 		Dir(formulaWorkDir).
 		WithAutoCommit().
 		WithGTRoot(townRoot).
-		StripBdBranch().
 		Output()
 	if err != nil {
 		return nil, fmt.Errorf("bonding formula to bead: %w", err)


julianknutsen:
Closing in favor of #1897, which addresses the same root cause (split-brain in `InstantiateFormulaOnBead`) but with a more comprehensive fix.

**Analysis:** Both PRs diagnose the same symptom — "not found" errors during formula instantiation when `BD_BRANCH` is set. This PR removes `StripBdBranch()` from the bond step so everything targets the polecat's branch. #1897 instead adds `StripBdBranch()` to the cook and wisp steps so everything targets main.

In the **normal sling flow**, neither strip actually matters — `InstantiateFormulaOnBead` runs from Mayor/Deacon context (no `BD_BRANCH` set), and the Dolt branch is created *after* all sling writes complete (`sling.go:698`). The commit-then-branch ordering means the wisp/bond are always written to main's working set first, then included in the branch fork. So `StripBdBranch()` is defensive here, not load-bearing.

The bug only manifests in **edge cases** where `BD_BRANCH` leaks into the caller's environment. Both PRs fix that edge case — this one by removing the strip (consistent on the leaked branch), #1897 by adding strips everywhere (consistent on main). Since the correct intent for orchestration writes is to target main, #1897 is the safer direction. It also adds a fallback bond path for `bd` version compatibility and better output parsing, which makes it the more complete fix.

Thank you for the contribution and the clear split-brain diagnosis — it was spot on and helped confirm the fix direction.

PR #1902, opened 2026-02-22
fix: clean up stale molecules during polecat nuke to unblock re-sling

vaxhacker:
## Summary
- Adds molecule cleanup to `nukePolecatFull` — detaches stale molecules from beads before nuke completes
- Adds assignee check in `sling_dispatch.go` to prevent dispatching already-claimed beads
- Fixes the catch-22 where nuked polecats left orphan molecule refs that blocked re-dispatch

Fixes gt-npzy — stale molecule attachments were blocking `gt sling` with "bead already has attached molecule(s)" error, requiring `--hook-raw-bead` workaround.

## Test plan
- [ ] Verify `gt sling` works on a bead after its polecat was nuked (no stale molecule error)
- [ ] Verify nuke cleans molecule attachment from bead
- [ ] Verify dispatch skips already-claimed beads

### DIFF
--- a/internal/cmd/molecule_lifecycle.go
+++ b/internal/cmd/molecule_lifecycle.go
@@ -346,6 +346,17 @@ squashed_at: %s
 // closeDescendants recursively closes all descendant issues of a parent.
 // Returns the count of issues closed. Logs warnings on errors but doesn't fail.
 func closeDescendants(b *beads.Beads, parentID string) int {
+	return closeDescendantsImpl(b, parentID, false)
+}
+
+// forceCloseDescendants is like closeDescendants but uses force-close,
+// which succeeds even for beads in invalid states. Use in destructive
+// paths (nuke, burn) where we must clean up regardless.
+func forceCloseDescendants(b *beads.Beads, parentID string) {
+	closeDescendantsImpl(b, parentID, true)
+}
+
+func closeDescendantsImpl(b *beads.Beads, parentID string, force bool) int {
 	children, err := b.List(beads.ListOptions{
 		Parent: parentID,
 		Status: "all",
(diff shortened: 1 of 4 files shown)


julianknutsen:
## Adoption Review

Thanks for the contribution, @vaxhacker! This PR correctly identifies and fixes a real catch-22 where nuked polecats left orphan molecule refs that blocked re-dispatch — a frustrating failure requiring the `--hook-raw-bead` workaround. The molecule cleanup approach is sound and the `executeSling` dead-agent auto-force closes a real parity gap with `runSling`.

### Initial Review
Triple-model review (Claude + Codex + Gemini) confirmed both fixes are correct. Key findings: (1) dead-agent auto-force could accidentally bypass the deferred-bead gate, (2) `burnExistingMolecules` in sling_helpers.go diverged from the nuke cleanup pattern (missing `closeDescendants`, non-force close), (3) `closeDescendants` used non-force close which could leave orphans in destructive paths.

### Maintainer Changes
Addresses all three findings from the initial review:
1. **Deferred gate bypass** — capture `explicitForce` before dead-agent auto-force so deferred beads still require explicit `--force`
2. **Pattern alignment** — `burnExistingMolecules` now uses `forceCloseDescendants` and matches the nuke pattern order (close descendants → detach → close roots)
3. **Force-close descendants** — added `forceCloseDescendants` variant in `molecule_lifecycle.go` for destructive paths; wired into both nuke and burn

### Final Review Status
Iteration review confirmed all three initial findings resolved. Two architecture-level findings remain as tracked follow-ups:
- **Molecule discovery contract divergence** — nuke checks `attached_molecule` in description only while sling also checks dependency bonds. Low risk since the description path covers the common case.
- **Regression test coverage** — no targeted behavioral tests for the new code paths. Acceptable given best-effort semantics and the arch census test update.

### Review Iterations
2 review passes performed.

---

PR #1905, opened 2026-02-22
Fix handoff restart to honor role_agents (gt-uoq)

Henry-E:
## Summary
- add a regression test for self-handoff restart command generation when `role_agents` is configured
- update `buildRestartCommand` to build the runtime command from role-aware config when `GT_AGENT` override is not set
- keep explicit `GT_AGENT` override behavior unchanged

## Why
`buildRestartCommand` previously used `GetRuntimeCommandWithPrompt` in the no-override path, which resolves from default agent settings and can drop role-specific model flags. This caused self-handoff to lose `role_agents` selection.

## Tests
- `go test ./internal/cmd -run TestBuildRestartCommand_UsesRoleAgentsWhenNoAgentOverride -count=1`
- `go test ./...`

## Red/Green
- before fix: `TestBuildRestartCommand_UsesRoleAgentsWhenNoAgentOverride` fails (missing `--model sonnet`)
- after fix: same test passes

### DIFF
--- a/internal/cmd/handoff_test.go
+++ b/internal/cmd/handoff_test.go
@@ -107,14 +107,19 @@ func TestBuildRestartCommand_UsesRoleAgentsWhenNoAgentOverride(t *testing.T) {
 	origGTAgent := os.Getenv("GT_AGENT")
 	origTownRoot := os.Getenv("GT_TOWN_ROOT")
 	origRoot := os.Getenv("GT_ROOT")
+
+	// TempDir must be called BEFORE registering the Chdir cleanup so that
+	// LIFO ordering restores the working directory before TempDir removal.
+	// On Windows the directory cannot be deleted while the process CWD is
+	// inside it.
+	townRoot := t.TempDir()
+
 	t.Cleanup(func() {
 		_ = os.Chdir(origCwd)
 		_ = os.Setenv("GT_AGENT", origGTAgent)
 		_ = os.Setenv("GT_TOWN_ROOT", origTownRoot)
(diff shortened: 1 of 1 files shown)


julianknutsen:
## Adoption Review

Thanks for the contribution, @Henry-E! This PR correctly identifies and fixes a real inconsistency in `buildRestartCommand` where `runtimeCmd` was resolved via `ResolveAgentConfig` (ignoring `RoleAgents` mappings) while `runtimeConfig` resolved via `ResolveRoleAgentConfig` (honoring them). Aligning both through `ResolveRoleAgentConfig` makes restart match the authoritative `BuildStartupCommand` startup logic — nice catch.

### Initial Review
Decision: **approve**. All three reviewers (Claude, Codex, Gemini) independently confirmed the bug is real and the fix is correct. Gemini verified the pattern against 66 existing call sites — no drift detected. Three minor findings: implicit `--settings` flag / cost-tier gain (beneficial but undocumented), duplicate `ResolveRoleAgentConfig` call, and missing `GT_AGENT` override test coverage.

### Maintainer Changes
Fixed Windows CI failure in the new test: reordered `t.TempDir()` before `t.Cleanup()` so LIFO ordering restores the working directory before temp directory removal. On Windows, a directory can't be deleted while the process CWD is inside it.

`internal/cmd/handoff_test.go` — 1 file changed, +7 −2

### Final Review Status
All clear — no remaining blocking issues. The three minor review findings (undocumented side-effects, duplicate resolver call, missing override tests) are acceptable as-is or can be addressed in follow-ups.

### Review Iterations
1 review pass performed.

---

PR #1925, opened 2026-02-23
feat: stream tmux pane output to VictoriaLogs (gt pane-log)

pae23:
## Summary

Opt-in feature (`GT_LOG_PANE_OUTPUT=true`) that streams Claude's tmux pane output to VictoriaLogs, giving operators visibility into what agents are actually doing beyond what OTEL tool events capture.

Three commits:
1. **`session/pane_logging_unix.go`** — `activatePaneLogging()` launches `gt pane-log` as a detached daemon with PID file tracking; kills the previous process on restart to prevent accumulation
2. **`telemetry/recorder.go`** — adds `paneOutputTotal` counter + `RecordPaneOutput()` 
3. **`cmd/pane_log.go`** — hidden `gt pane-log --session <id>` command: polls `capture-pane` every 200ms, filters UI chrome (borders, spinners, status bars, token counts), emits new content via `RecordPaneOutput`

## Activation

```bash
# In environment or .env
GT_LOG_PANE_OUTPUT=true
GT_OTEL_LOGS_URL=http://localhost:9428/insert/opentelemetry/v1/logs
```

New sessions started after this is set will auto-activate. For an existing session:
```bash
gt pane-log --session hq-mayor &
```

## Query in VictoriaLogs

```
_msg:pane.output
```

## UI chrome filtered

- Box-drawing borders, block elements
- Spinner-only lines (braille, symbols)
- Status bar fragments (token counts, timing, `bypass permissions`, etc.)
- Claude Code welcome splash

### DIFF
--- a/internal/cmd/pane_log.go
+++ b/internal/cmd/pane_log.go
@@ -1,0 +1,106 @@
+package cmd
+
+import (
+	"context"
+	"regexp"
+	"strings"
+	"time"
+
+	"github.com/spf13/cobra"
+	"github.com/steveyegge/gastown/internal/telemetry"
+	"github.com/steveyegge/gastown/internal/tmux"
+)
+
+var paneLogSession string
+
+const paneLogPollInterval = 200 * time.Millisecond
+
+// paneUIRe matches lines that are pure Claude Code UI chrome with no content value.
+var paneUIRe = regexp.MustCompile(
+	// Box-drawing lines and blank lines (light + heavy + double + extensions)
+	`^[\s─━═⎯│║╭╮╰╯├└┘┐┤┬┴┼╔╗╚╝╠╣╦╩╬╟╞╡╢╤╧╪╫]+$` +
+		// Block elements only (borders, progress fill)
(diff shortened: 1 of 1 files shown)


steveyegge:
Closing -- the feature (agent pane visibility) is worth having, but the implementation should be rethought.

**Core issue:** Line-count-based diffing with capture-pane is fundamentally fragile. capture-pane returns the visible pane content, so when new output pushes old lines off the top, the line count can stay the same or shrink -- silently missing everything that scrolled away. For long-running agent sessions producing real output, that is most of the interesting stuff.

**Better primitive:** tmux pipe-pane gives you a stream of everything written to the pane -- no polling, no diffing, no missed output. One line to set up, zero CPU overhead vs 5 subprocess spawns/second.

**Other notes:**
- The UI chrome regex is clever but fragile -- every Claude Code UI update could break it
- activatePaneLogging() is defined but never called (no wiring into session lifecycle)
- 200ms polling across many sessions adds up

Would welcome a v2 built on pipe-pane. The telemetry plumbing (RecordPaneOutput, counter, 8KB truncation) is solid and can be reused.

PR #1926, opened 2026-02-23
fix: make KillSession idempotent when session already gone

pae23:
## Summary

- `KillSession` was returning `ErrSessionNotFound`/`ErrNoServer` when the tmux session was already dead
- `RecordSessionStop` received a non-nil error → logged as `SeverityError` in VictoriaLogs (8 false-positive errors/hour)
- Root cause: race between `gt done` (polecat self-kill) and witness/daemon trying to kill the same session
- If the session is already gone, the goal is achieved — return `nil` so telemetry logs `Info` not `Error`

Several callers already guarded against this manually (`crew_at.go`: `killErr != tmux.ErrSessionNotFound`); this centralises the idempotency at the source and removes the need for per-caller guards.

## Test plan

- [ ] Kill a session that doesn't exist: `gt session stop <nonexistent>` → no error
- [ ] VictoriaLogs: `session.stop` events with `session not found` disappear from error queries

### DIFF
--- a/internal/tmux/tmux.go
+++ b/internal/tmux/tmux.go
@@ -221,9 +221,13 @@ func (t *Tmux) EnsureSessionFresh(name, workDir string) error {
 }
 
 // KillSession terminates a tmux session.
+// Returns nil if the session does not exist (idempotent: goal is already achieved).
 func (t *Tmux) KillSession(name string) (retErr error) {
 	defer func() { telemetry.RecordSessionStop(context.Background(), name, retErr) }()
 	_, retErr = t.run("kill-session", "-t", name)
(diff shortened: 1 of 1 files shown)


steveyegge:
Closing — this is a snowball PR carrying changes from 8+ already-merged PRs (#1916, #1917, #1919, #1920, #1921, #1922, #1923) plus pane logging code from #1925 which was closed (we want a v2 built on tmux pipe-pane instead of poll-and-diff).

The core change (4-line KillSession idempotency fix in tmux.go) is correct and welcome. Please rebase onto main and resubmit with just the tmux.go change.

PR #1941, opened 2026-02-23
fix: collapse agent bead ID when prefix equals rig name

### LINKED ISSUE #1877: Agent bead IDs double-prefixed when rig name equals derived prefix
## Bug

## Reproduction

2. Run `gt doctor`

```sql
-- ff-ff-refinery
-- ff-ff-witness
```

## Root Cause

```go
```

## Suggested Fix

```go
        if name == "" {
            return prefix + "-" + role
        }
    }
    if name == "" {
    }
}
```

## Environment

- Gas Town v0.7.0
- Installed via Homebrew
- Linux x86_64

DreadPirateRobertz:
## Summary
- When `deriveBeadsPrefix()` returns the same string as the rig name (e.g., rig `ff` → prefix `ff`), `AgentBeadIDWithPrefix` produced stuttered IDs like `ff-ff-refinery`
- This caused `gt doctor` to report missing beads because the routing layer strips the prefix during lookup
- Fix: when `prefix == rig`, omit the redundant rig component (`ff-refinery` instead of `ff-ff-refinery`)

## Test plan
- [x] Added `TestAgentBeadIDWithPrefix_CollapsesPrefixEqualRig` — 8 cases covering collapsed, normal, and town-level forms
- [x] All existing `internal/beads` tests pass
- [ ] Manual: `gt rig add` with short name (≤3 chars), verify `gt doctor` finds agent beads

Closes #1877

### DIFF
--- a/internal/beads/agent_ids.go
+++ b/internal/beads/agent_ids.go
@@ -324,10 +324,17 @@ func ValidateAgentID(id string) error {
 // For town-level agents (mayor, deacon), pass empty rig and name.
 // For rig-level singletons (witness, refinery), pass empty name.
 // For named agents (crew, polecat), pass all three.
+//
+// When prefix == rig (happens for short rig names like "ff" where
+// deriveBeadsPrefix returns the name itself), the rig component is omitted
+// to avoid stuttered IDs like "ff-ff-refinery". Instead: "ff-refinery".
 func AgentBeadIDWithPrefix(prefix, rig, role, name string) string {
-	if rig == "" {
-		// Town-level agent: prefix-mayor, prefix-deacon
-		return prefix + "-" + role
+	if rig == "" || rig == prefix {
+		// Town-level agent or collapsed prefix==rig: prefix-role[-name]
+		if name == "" {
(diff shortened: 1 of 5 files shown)


DreadPirateRobertz (commented):
## Review: LGTM

**Agent bead ID fix**: When `prefix == rig` (short rig names like "ff"), collapses the rig component to avoid stuttered IDs like `ff-ff-refinery`. Tests cover collapsed, normal, and town-level cases.

**Stale PID cleanup** (`stale_pid.go`): Defensive removal of stale `dolt-server.pid` files via signal-0 probe. Prevents bd from connecting to wrong Dolt server on port 3307. Well-tested.

**Mailbox query change**: Switches from single broad `bd list` to per-identity `--assignee` queries, pushing filtering to Dolt. Reduces memory under concurrent load. Timeout bumped 30s→60s for contention. More bd spawns but memory savings worth it.

All three changes are solid.

DreadPirateRobertz:
Superseded by #1950 (cleaner approach)

PR #1948, opened 2026-02-23
fix(beads): deduplicate agent bead IDs when prefix equals rig name

### LINKED ISSUE #1877: Agent bead IDs double-prefixed when rig name equals derived prefix

DreadPirateRobertz:
## Summary
- When `deriveBeadsPrefix()` returns the same string as the rig name (short rig names like `"ff"`), `AgentBeadIDWithPrefix` produced stuttered IDs like `ff-ff-refinery` instead of `ff-refinery`
- This caused `gt doctor` to report agent beads as missing and `--fix` to create beads with the same double prefix
- Added a `prefix == rig` collapse check in `AgentBeadIDWithPrefix` — the parser and validator already handle the collapsed form correctly

## Test plan
- [x] Added 7 regression test cases covering singleton, named, and town-level agents with same/different prefix-rig combinations
- [x] All test cases verify round-trip through `ParseAgentBeadID`
- [x] Existing `TestValidateAgentID` (30 cases), `TestParseAgentBeadID` (23 cases), and `TestExtractAgentPrefix` (15 cases) all pass
- [x] Doctor package tests pass

Closes #1877

### DIFF
--- a/internal/beads/agent_ids.go
+++ b/internal/beads/agent_ids.go
@@ -329,6 +329,15 @@ func AgentBeadIDWithPrefix(prefix, rig, role, name string) string {
 		// Town-level agent: prefix-mayor, prefix-deacon
 		return prefix + "-" + role
 	}
+	// When prefix == rig (short rig names where deriveBeadsPrefix returns
+	// the rig name itself), omit the rig component to avoid stuttered IDs
+	// like "ff-ff-refinery". See #1877.
+	if prefix == rig {
+		if name == "" {
+			return prefix + "-" + role
+		}
+		return prefix + "-" + role + "-" + name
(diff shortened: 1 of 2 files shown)


DreadPirateRobertz:
Superseded by #1950 (cleaner approach)

PR #1950, opened 2026-02-23
fix: deduplicate agent bead IDs when prefix equals rig name

### LINKED ISSUE #1877: Agent bead IDs double-prefixed when rig name equals derived prefix

DreadPirateRobertz:
## Summary

Fixes #1877.

- When `deriveBeadsPrefix()` returns the same string as the rig name (e.g., rig `"ff"` → prefix `"ff"`), `AgentBeadIDWithPrefix` produced stuttered IDs like `ff-ff-refinery`
- The routing layer then fails to find these beads because prefix stripping yields a key that doesn't match the stored double-prefixed ID
- Collapse the rig component when `prefix == rig` so `ff-refinery` is produced instead of `ff-ff-refinery`

## Test plan

- [x] New test `TestAgentBeadIDWithPrefix` covers normal, town-level, and collapsed cases
- [x] New test `TestAgentBeadIDWithPrefix_NoStutter` specifically reproduces the #1877 bug scenario
- [x] All existing agent ID tests pass (ParseAgentBeadID, ValidateAgentID, IsAgentSessionBead)
- [x] Doctor package tests pass
- [x] Full `go build ./...` clean
- [ ] Manual verification with a short-named rig (e.g., `ff`) that `gt doctor` no longer reports missing agent beads

### DIFF
--- a/internal/beads/agent_ids.go
+++ b/internal/beads/agent_ids.go
@@ -324,10 +324,22 @@ func ValidateAgentID(id string) error {
 // For town-level agents (mayor, deacon), pass empty rig and name.
 // For rig-level singletons (witness, refinery), pass empty name.
 // For named agents (crew, polecat), pass all three.
+//
+// When prefix == rig (which happens for short rig names where deriveBeadsPrefix
+// returns the rig name itself, e.g. rig "ff" with prefix "ff"), the rig component
+// is collapsed to avoid double-prefixed IDs like "ff-ff-refinery". Instead,
+// "ff-refinery" is produced. See github.com/steveyegge/gastown/issues/1877.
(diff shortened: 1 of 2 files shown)


steveyegge:
Closing in favor of #1951 which is the properly scoped fix on a clean feature branch. See review comment for details.

PR #1951, opened 2026-02-23
fix: deduplicate agent bead IDs when prefix equals rig name

### LINKED ISSUE #1877: Agent bead IDs double-prefixed when rig name equals derived prefix

DreadPirateRobertz:
## Summary

Fixes #1877 — Agent bead IDs get double-prefixed when the rig name equals the derived prefix (e.g., short rig names like "ff" where `deriveBeadsPrefix("ff")` returns `"ff"`).

- `AgentBeadIDWithPrefix` now collapses the rig component when `prefix == rig`, producing `ff-witness` instead of `ff-ff-witness`
- `ValidateAgentID` updated to accept collapsed-prefix forms (e.g., `ff-witness`, `ff-crew-bob`)
- `ParseAgentBeadID` scan loop extended to `i >= 0` to correctly parse collapsed named agents (e.g., `ff-crew-bob`)

## Test plan

- [x] New `TestAgentBeadIDWithPrefix_CollapsedPrefix` — verifies deduplication for all agent types
- [x] New `TestParseAgentBeadID_CollapsedPrefix` — verifies parsing of collapsed IDs
- [x] Updated `TestValidateAgentID` — added collapsed-prefix valid/invalid cases
- [x] All existing agent ID tests pass (no regressions)
- [x] Full test suite passes (only pre-existing daemon/convoy failures unrelated to this change)
- [ ] Manual: create a rig with short name (e.g., `ff`), run `gt doctor` — should no longer report missing agent beads

### DIFF
--- a/internal/beads/agent_ids.go
+++ b/internal/beads/agent_ids.go
@@ -199,13 +199,23 @@ func ValidateAgentID(id string) error {
 		return fmt.Errorf("agent ID must include content after prefix (got %q)", id)
 	}
 
-	// Case 1: Single part after prefix - must be town-level role
+	// Case 1: Single part after prefix - town-level role, or collapsed prefix
+	// rig-level singleton (when prefix==rig, e.g., "ff-witness"). See GH #1877.
 	if len(parts) == 1 {
 		role := parts[0]
 		if isTownLevelRole(role) {
 			return nil // Valid town-level agent
 		}
+		// Rig-level singleton roles are also valid here because
+		// AgentBeadIDWithPrefix collapses the rig when prefix==rig,
+		// producing IDs like "ff-witness" instead of "ff-ff-witness".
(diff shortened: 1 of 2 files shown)


steveyegge:
Closing — the fix for #1877 is already on main (commit e4257210) with the full scope of changes (AgentBeadIDWithPrefix, ValidateAgentID, ParseAgentBeadID). This PR's changes are now redundant. Thank you for the clean implementation\!

PR #1956, opened 2026-02-24
Harden dashboard API against cross-origin and unauthenticated access

alexhop:
## Summary

- Bind dashboard server to `127.0.0.1` instead of `0.0.0.0` to prevent remote network access
- Remove `Access-Control-Allow-Origin: *` CORS headers from all API and SSE endpoints
- Add per-session CSRF token (generated via `crypto/rand` at startup) validated on all POST requests via `X-Dashboard-Token` header
- Enforce server-side `Confirmed` check for commands marked `Confirm: true` in `AllowedCommands`, preventing automated execution of dangerous commands even if other protections are bypassed

The CSRF token is embedded in the HTML via a `<meta name="dashboard-token">` tag and sent automatically by a `fetch()` interceptor in both the main dashboard JS and setup page JS.

## Testing Done

- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] All existing `internal/web` tests pass (12 files changed, test signatures updated)
- [x] New test: `TestAPIHandler_Run_MissingCSRFToken` — POST without token returns 403
- [x] New test: `TestAPIHandler_Run_WrongCSRFToken` — POST with wrong token returns 403
- [x] New test: `TestAPIHandler_Run_ConfirmRequired` — confirm command without `confirmed: true` returns 403

### DIFF
--- a/internal/cmd/dashboard.go
+++ b/internal/cmd/dashboard.go
@@ -119,7 +119,7 @@ func runDashboard(cmd *cobra.Command, args []string) error {
 	fmt.Printf("  launching dashboard at %s  •  api: %s/api/  •  ctrl+c to stop\n", url, url)
 
 	server := &http.Server{
-		Addr:              fmt.Sprintf(":%d", dashboardPort),
+		Addr:              fmt.Sprintf("127.0.0.1:%d", dashboardPort),
 		Handler:           handler,
 		ReadHeaderTimeout: 10 * time.Second,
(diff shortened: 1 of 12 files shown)


DreadPirateRobertz:
Reviewed — well-executed security hardening. Localhost bind, CORS removal, CSRF token (crypto/rand 256-bit), server-side confirm enforcement. All tests updated + 3 new test cases. Clean and correct. Ready to merge.

PR #1984, opened 2026-02-24
fix(doctor): detect stale types sentinel in EnsureCustomTypes

Xexr:
## Summary

- Fix `gt doctor --fix` silently failing to create agent beads for beads rig crew members
- Add version-aware sentinel validation in `EnsureCustomTypes`
- Add defense-in-depth type error detection and retry in `CreateAgentBead`

## Problem

`EnsureCustomTypes` trusted the `.gt-types-configured` sentinel file without verifying its content matched the current types list. When new custom types (like `agent`) were added after the sentinel was written, the sentinel became stale. `bd create --type=agent` would fail with "invalid issue type: agent", but since `EnsureCustomTypes` returned nil, the JSONL fallback wasn't triggered. This caused `gt doctor --fix` to silently fail when creating agent beads for beads rig crew members (korben, ruby, zorg).

## Solution

1. **Version-aware sentinel**: `EnsureCustomTypes` now computes a fingerprint from the sorted types list and compares it against the sentinel file content. Stale sentinels (including legacy `v1`) trigger reconfiguration.

2. **Defense-in-depth in `CreateAgentBead`**: When `bd create` fails with "invalid issue type", the code now invalidates the sentinel, re-runs `EnsureCustomTypes`, and retries once. This catches cases where the database state changed after startup.

3. **`InvalidateSentinel()` API**: New function for programmatic cache clearing (both in-memory and on-disk).

## Testing

- `TestEnsureCustomTypes` — updated for fingerprint-based sentinel, added stale sentinel test
- `TestTypesFingerprint` — deterministic, versioned, differs from legacy
- `TestInvalidateSentinel` — clears both in-memory cache and sentinel file
- `TestIsTypeValidationError` — error detection helper
- Full test suite passes (`go test ./...` — 62 packages)
- Manual verification: confirmed `bd create --type=agent` succeeds after types reconfiguration

(gt-uaq)

### DIFF
--- a/internal/beads/beads.go
+++ b/internal/beads/beads.go
@@ -388,6 +388,17 @@ func isSubprocessCrash(err error) bool {
 		strings.Contains(errStr, "panic:")
 }
 
+// isTypeValidationError returns true if the error indicates a custom type
+// validation failure (e.g., "invalid issue type: agent"). This typically means
+// the types sentinel file is stale and custom types need to be reconfigured
+// in the target database (gt-uaq).
+func isTypeValidationError(err error) bool {
+	if err == nil {
+		return false
+	}
+	return strings.Contains(err.Error(), "invalid issue type")
+}
+
 // buildRunEnv builds the environment for run() calls.
 // In isolated mode: strips all beads-related env vars for test isolation.
 // Otherwise: strips inherited BEADS_DIR so the caller can append the correct value.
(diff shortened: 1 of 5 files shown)


steveyegge:
Closing as superseded — main already has a simpler sentinel staleness approach (raw types list string) that was merged before this PR. The fingerprint/sha256 format in this PR conflicts with what's on main.

The `InvalidateSentinel()` function and the `CreateAgentBead` retry logic from this PR are good ideas — if you'd like to contribute those pieces, a fresh PR rebased on main with just those additions would be welcome. Thanks for the thorough work!

PR #2008, opened 2026-02-25
feat: add context-budget tap guard for automatic handoff at threshold

dannomayernotabot:
## Summary

- Adds `gt tap guard context-budget` command that monitors Claude Code context window consumption via transcript file size and enforces automatic handoff at configurable thresholds
- Fires warnings at 75% (info), 85% (soft gate with handoff instructions), and 92% (hard gate that blocks tool calls for mayor/deacon roles)
- Thresholds configurable via `GT_CONTEXT_BUDGET_*` environment variables; registered as PreToolUse hook in default base config

## Files Changed

- `internal/cmd/tap_guard.go` — registered new subcommand
- `internal/cmd/tap_guard_context_budget.go` — core implementation (403 lines)
- `internal/cmd/tap_guard_context_budget_test.go` — 6 tests covering config, parsing, and disable behavior
- `internal/hooks/config.go` — added hook registration in base config

## Test plan

- [x] All context-budget tests pass (`go test ./internal/cmd/ -run ContextBudget`)
- [x] Hooks tests pass (`go test ./internal/hooks/`)
- [x] Full build succeeds (`go build ./...`)
- [x] CI passes on PR

Closes hq-48p

### DIFF
--- a/internal/cmd/tap_guard.go
+++ b/internal/cmd/tap_guard.go
@@ -24,6 +24,9 @@ Available guards:
   mol-patrol         - Block mol patrol from agent contexts
   dangerous-command  - Block rm -rf, force push, hard reset, git clean
 
+External guards (standalone scripts, not compiled into gt):
+  context-budget   - scripts/guards/context-budget-guard.sh
+
 Example hook configuration:
   {
     "PreToolUse": [{
--- a/scripts/guards/context-budget-guard.sh
+++ b/scripts/guards/context-budget-guard.sh
@@ -1,0 +1,155 @@
+#!/usr/bin/env bash
+#
+# context-budget-guard.sh — Monitor context window usage, enforce handoff thresholds.
+#
+# A standalone guard script for Claude Code hooks. Reads the active session's
+# transcript to track token usage and enforces configurable thresholds.
+#
+# Exit codes:
+#   0 — Allow (under threshold, disabled, or fail-open on any error)
+#   2 — Block (hard-gate threshold exceeded for a hard-gated role)
+#
+# Configuration (environment variables):
+#   GT_CONTEXT_BUDGET_DISABLE=1                — Disable the guard entirely
+#   GT_CONTEXT_BUDGET_WARN=0.75                — Warning threshold (default: 0.75)
+#   GT_CONTEXT_BUDGET_SOFT_GATE=0.85           — Soft gate threshold (default: 0.85)
+#   GT_CONTEXT_BUDGET_HARD_GATE=0.92           — Hard gate threshold (default: 0.92)
+#   GT_CONTEXT_BUDGET_MAX_TOKENS=200000        — Max context tokens (default: 200000)
+#   GT_CONTEXT_BUDGET_HARD_GATE_ROLES=mayor,deacon,witness,refinery
+#                                              — Comma-separated roles that get blocked
+#                                                at hard gate (default shown above)
+#
+# Hook configuration example (UserPromptSubmit — fires once per user turn):
+#   {
+#     "UserPromptSubmit": [{
+#       "matcher": "",
+#       "hooks": [{"type": "command", "command": "/path/to/context-budget-guard.sh"}]
+#     }]
+#   }
+#
+# Dependencies: jq (fails open if not found)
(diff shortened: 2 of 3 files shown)


steveyegge:
Sheriff review: This is a substantial new feature (703 lines) that adds context-budget monitoring with automatic handoff thresholds, registered as a PreToolUse hook in the base config.

A few concerns:
1. **Plugin candidate**: Tap guards are a natural fit for the plugin/marketplace system. Baking this into core means every user gets it whether they want it or not. Consider shipping as an opt-in plugin instead.
2. **Size**: 703 lines of new code for a single tap guard feels heavy. The threshold logic, config parsing, and file-size-based estimation could likely be simplified.
3. **Base config modification**: Registering in the default base hooks config is a significant decision that affects all installations.

Leaving for maintainer review on direction. The concept is valuable — the question is whether it belongs in core vs. as a plugin.

steveyegge:
Thanks for this — context-budget monitoring is genuinely useful and the implementation is clean.

However, we'd like to see this as an **external guard script** rather than a compiled Go subcommand. The tap-guard hook system already supports arbitrary commands — any executable that reads stdin hook JSON and exits 0 (allow) or 2 (block) is a valid guard. This is a great candidate for proving out that plugin surface.

Specific feedback:

1. **Ship as a script, not Go**: A standalone `context-budget-guard.sh` (or Python, etc.) that users drop into a guards directory. The hook config just points at it.

2. **Role-gating belongs in registry.toml**: The hardcoded `HardGated` logic (mayor/deacon/witness = blocked, crew/polecat = warn-only) should be configuration, not compiled code. The registry already has `roles = [...]` scoping for this.

3. **Transcript parsing is fragile**: The JSONL format is a Claude Code implementation detail that could change. Worth noting in the script as a known limitation, and ideally abstracting the token-count source behind an env var or helper.

The guard itself is a good idea — we just want it to set the right precedent for how new guards are added. The next guard after this shouldn't need a Go PR.

Happy to help iterate on the script-based approach if you want to rework it.

DreadPirateRobertz:
### Code review

Found 2 issues:

1. `parseLastInputTokens` reads only `InputTokens` to estimate context window usage, ignoring `CacheCreationInputTokens` and `CacheReadInputTokens`. The sibling `TranscriptUsage` struct in `costs.go` defines all three fields, and `parseTranscriptUsage()` in the same file sums all three. In cache-heavy Gas Town sessions (where `gt prime --hook` injects large system prompts served from Anthropic's prompt cache), `input_tokens` alone may show ~8k when the actual context window is at ~170k. The guard would effectively never fire.

https://github.com/steveyegge/gastown/blob/e46827b26da564bf9818d51aa75e91673a092dd7/internal/cmd/tap_guard_context_budget.go#L192-L197

Compare with the existing usage tracking that accounts for all token types:

https://github.com/steveyegge/gastown/blob/e46827b26da564bf9818d51aa75e91673a092dd7/internal/cmd/costs.go#L205-L211

2. The context-budget hook is registered in `DefaultBase().PreToolUse` with `Matcher: ""` (empty). Every existing `PreToolUse` entry uses a specific `Bash(...)` glob matcher (e.g. `Bash(gh pr create*)`, `Bash(rm -rf /*)`) because `PreToolUse` fires per tool invocation. An empty matcher causes the guard to fire on every tool call (Read, Write, Edit, Glob, etc.), not just Bash, adding a full transcript file scan on each invocation. Additionally, `validateUniqueMatchers` enforces unique matchers per event type, so this empty matcher blocks any future catch-all `PreToolUse` entry. Consider using `UserPromptSubmit` (fires once per user turn) instead, or a specific matcher.

https://github.com/steveyegge/gastown/blob/e46827b26da564bf9818d51aa75e91673a092dd7/internal/hooks/config.go#L757-L764

Existing PreToolUse pattern for comparison (all use specific matchers):

https://github.com/steveyegge/gastown/blob/e46827b26da564bf9818d51aa75e91673a092dd7/internal/hooks/config.go#L717-L759

Generated with [Claude Code](https://claude.ai/code)

- If this code review was useful, please react with :+1:. Otherwise, react with :-1:.

PR #2011, opened 2026-02-25
fix(tmux): harden session creation, nudge delivery, and test isolation

l0g1x:
## Summary

- **Session creation hardening**: two-step `new-session` + `respawn-pane` with post-creation health check detects binary-not-found, syntax errors, and fast exits before callers dispatch work. Validates workDir and session names.
- **Nudge reliability**: sanitize control characters, exit copy/scroll mode before sending, fix ESC→Enter timing to exceed readline's keyseq-timeout, chunk large messages (>512B) to avoid argument limits.
- **WakePane fix**: use `resize-window` instead of `resize-pane` — single-pane sessions can't resize the pane smaller than the window, so SIGWINCH was never delivered.
- **Test isolation**: per-test tmux server sockets prevent interference with user sessions and cross-test contamination.

## Bugs Fixed

| Bug | Root Cause | Fix |
|-----|-----------|-----|
| Blank tmux windows (binary not found, bad path, syntax error) | `new-session -d` with a command that fails immediately exits silently | Two-step creation: spawn shell first, set `remain-on-exit`, `respawn-pane` with command, health-check `pane_dead`+`pane_dead_status` |
| Nudges swallowed in copy/scroll mode | `send-keys -l` goes to copy-mode input, not the underlying process | Check `pane_in_mode`, send `send-keys -X cancel` before delivery |
| ESC+Enter interpreted as M-Enter | 100ms delay between ESC and Enter is within readline's 500ms keyseq-timeout | Increase to 600ms |
| Control chars corrupt `send-keys -l` delivery | ESC triggers terminal sequences, CR acts as premature Enter, BS deletes chars | `sanitizeNudgeMessage()` strips ESC/CR/BS/DEL/control chars, replaces TAB with space |
| Large nudges (>512B) dropped by tmux | `send-keys -l` argument length limits | Chunk into 512-byte pieces with 10ms inter-chunk delays |
| `WakePane` SIGWINCH never fires | `resize-pane -y -1` fails silently on single-pane sessions (pane = window) | Use `resize-window` width ±1 dance instead |
| Tests kill user's tmux sessions | All tests shared the default tmux server | `newTestTmux(t)` creates isolated per-test server via `-L` flag |
| `GT_TEST_ALLOW_CLEANUP_TEST` guard | Added when tests used default server; now redundant with isolation | Removed — isolated server can never see production sessions |

## Test Results

**151 PASS, 0 FAIL, 1 SKIP** (up from ~110 PASS on upstream/main)

The single skip (`TestLargeNudge_MultiLine`) documents a known tmux limitation: `send-keys -l` treats `\n` as Enter, splitting multiline messages.

65+ new tests in `session_creation_test.go` covering blank window prevention, copy-mode delivery, control char sanitization, escape sequence handling, 10KB message delivery, multi-pane targeting, nudge lock contention, session death characterization, SIGWINCH verification, and environment variable edge cases.

## Files Changed

| File | Changes |
|------|---------|
| `internal/tmux/tmux.go` | Production fixes: session creation, nudge sanitization, chunking, WakePane |
| `internal/tmux/session_creation_test.go` | New file: 65+ edge case and regression tests |
| `internal/tmux/tmux_test.go` | Test isolation via `newTestTmux(t)`, remove env-var guard |

### DIFF
--- a/internal/tmux/session_creation_test.go
+++ b/internal/tmux/session_creation_test.go
@@ -1,0 +1,2657 @@
+package tmux
+
+import (
+	"fmt"
+	"os"
+	"strings"
+	"testing"
+	"time"
+)
+
+// These tests investigate tmux session creation edge cases that can cause:
+// 1. Blank window - session created but window shows nothing
+// 2. Command visible at top - the exec command is printed before the actual program renders
+//
+// Root causes identified:
+// - Blank window: command fails silently (binary not found, syntax error, workdir missing)
+// - Command visible: shell echo mode (set -x), shell startup scripts printing, or
+//   tmux default-shell behavior differences
+
+// =============================================================================
+// BLANK WINDOW TESTS
+// =============================================================================
+
+// TestBlankWindow_CommandNotFound reproduces blank window when binary doesn't exist.
+// The session is created by tmux, but dies immediately when command fails.
+// This is the "blank window" symptom - user sees nothing useful.
+func TestBlankWindow_CommandNotFound(t *testing.T) {
+	tm := newTestTmux(t)
+	sessionName := "gt-test-blank-notfound-" + t.Name()
+
+	_ = tm.KillSession(sessionName)
+	defer func() { _ = tm.KillSession(sessionName) }()
+
+	// Use a non-existent binary - this simulates claude-code not being in PATH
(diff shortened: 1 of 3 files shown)


l0g1x:

PR #2024, opened 2026-02-25
fix(daemon): use NewSessionWithCommand to prevent empty polecat windows

l0g1x:
## Summary

- Replace unsafe `EnsureSessionFresh` + `SendKeys` pattern with `EnsureSessionFreshWithCommand` in both daemon restart paths
- Add `EnsureSessionFreshWithCommand` to tmux package — kills zombie sessions then creates new session with command as initial pane process
- Eliminates race condition where shell isn't ready to receive keystrokes, producing empty/blank tmux windows

## Problem

The daemon's `restartPolecatSession` (daemon.go) and `restartSession` (lifecycle.go) used a two-step pattern:

1. `EnsureSessionFresh(name, workDir)` — creates session with empty shell
2. `SendKeys(name, startCmd)` — sends the `exec env ... claude ...` command as keystrokes

If the shell isn't fully initialized when `SendKeys` fires (race window between session creation and shell readiness), the command is lost — producing an empty tmux window with no agent running.

## Fix

New `EnsureSessionFreshWithCommand(name, workDir, command)` method:
1. Checks for existing zombie sessions and kills them
2. Creates session via `NewSessionWithCommand` — the command becomes the pane's initial process (no shell intermediary)
3. Returns `ErrSessionRunning` if session already has a healthy agent

This matches the safe pattern already used by `session_manager.Start()` for fresh polecat spawns via `gt sling`.

## Testing

2 new integration tests:
- `TestEnsureSessionFreshWithCommand_NoExisting` — fresh session with command as initial process
- `TestEnsureSessionFreshWithCommand_KillsZombie` — zombie replacement with command

All existing `EnsureSessionFresh` tests pass unchanged.

### DIFF
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -3,6 +3,7 @@ package daemon
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"log"
 	"os"
@@ -1104,7 +1105,6 @@ func (d *Daemon) isRigOperational(rigName string) (bool, string) {
 	return true, ""
 }
 
-
 // processLifecycleRequests checks for and processes lifecycle requests.
 func (d *Daemon) processLifecycleRequests() {
 	d.ProcessLifecycleRequests()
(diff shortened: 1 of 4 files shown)


l0g1x:

PR #2031, opened 2026-02-25
fix(tmux): prevent auto-respawn hook from racing with daemon and leaking output

l0g1x:
## Summary

Fixes three bugs in `SetAutoRespawnHook` that caused the hook's error messages to take over unrelated tmux sessions:

- **Race with daemon**: Hook fires on pane death, sleeps 3s, then blindly runs `respawn-pane -k`. During the 3s window, the daemon's heartbeat can restart the session. The hook then kills the daemon's fresh agent. Fix: check `#{pane_dead}` before respawning (using `##` escape so tmux evaluates the format at query time, not hook fire time).
- **Output leak**: `run-shell` without `-b` displays errors on the attached client's active pane. Fix: use `run-shell -b` for background execution.
- **Missing socket flag**: Hook's embedded tmux commands used bare `tmux`, targeting the default server instead of the town socket. Fix: include `-L` flag when configured.
- **No error suppression**: Append `|| true` to always exit 0.

## Key Discovery

tmux's `run-shell` expands format variables (`#{pane_dead}`) at hook fire time, before the shell even runs. The 3-second `sleep` in the shell command runs AFTER expansion. So `#{pane_dead}` is always `1` (dead at hook time). Using `##{pane_dead}` escapes the first expansion — tmux converts `##` to `#`, and the nested `tmux list-panes -F '#{pane_dead}'` query evaluates the format at runtime, giving the current pane state.

## Testing

7 tests added in `internal/tmux/respawn_hook_test.go`:

| Test | What it proves |
|------|---------------|
| `TestAutoRespawnHook_RespawnWorks` | E2E: hook respawns dead pane on isolated socket |
| `TestAutoRespawnHook_SocketFlagInHookCmd` | E2E: hook works with socket isolation |
| `TestAutoRespawnHook_SkipsAlreadyAlive` | E2E: hook does NOT kill daemon-respawned pane (race test) |
| `TestAutoRespawnHook_SilentOnSessionKilled` | E2E: no error output when session was killed |
| `TestAutoRespawnHookCmd_Format` (x3) | Unit: `-b` flag, `pane_dead` guard, `|| true` present |

The race test (`SkipsAlreadyAlive`) directly reproduces the reported issue: PID changes when the hook fires after daemon restart (before fix), PID stays stable (after fix).

### DIFF
--- a/internal/tmux/respawn_hook_test.go
+++ b/internal/tmux/respawn_hook_test.go
@@ -2,7 +2,9 @@ package tmux
 
 import (
 	"fmt"
+	"os"
 	"os/exec"
+	"path/filepath"
 	"strings"
 	"testing"
 	"time"
@@ -16,6 +18,9 @@ func testSocket(t *testing.T) string {
 	t.Cleanup(func() {
 		// Kill the entire tmux server on this socket
 		_ = exec.Command("tmux", "-L", socket, "kill-server").Run()
+		// Remove the socket file — tmux kill-server leaves it behind
+		socketPath := filepath.Join(fmt.Sprintf("/tmp/tmux-%d", os.Getuid()), socket)
+		_ = os.Remove(socketPath)
 	})
 	return socket
 }


l0g1x:

PR #2034, opened 2026-02-25
fix(session): derive tmux socket from town name, not $TMUX

l0g1x:
## Summary

- **Bug**: `InitRegistry()` parsed `$TMUX` to determine the tmux socket name, causing all sessions to land on the `default` server when `gt up` was run from an interactive terminal. The daemon (no `$TMUX`) left the socket empty, also falling through to the default server.
- **Fix**: Replace `$TMUX` parsing with `sanitizeTownName(filepath.Base(townRoot))`, which deterministically derives the socket from the town directory name (e.g., `/home/user/gt` → `-L gt`). All callers — CLI, daemon, agents — now target the same dedicated tmux server.
- The `sanitizeTownName` function already existed with tests but was never wired into `InitRegistry`.

## Root Cause

Commit `33362a75` (socket migration) introduced per-town sockets but `InitRegistry` at `session/registry.go:121-126` trusted `$TMUX` blindly:

```go
if tmuxEnv := os.Getenv("TMUX"); tmuxEnv != "" {
    if socketPath, _, ok := strings.Cut(tmuxEnv, ","); ok && socketPath != "" {
        tmux.SetDefaultSocket(filepath.Base(socketPath))
    }
}
```

When running from a terminal on the default tmux server, `$TMUX` is `/tmp/tmux-1000/default,...` → socket = `"default"` → sessions created on wrong server.

## Test Plan

New `TestInitRegistry_SocketFromTownName` with 5 cases:
1. Inside default tmux, town=gt → socket should be `"gt"` (was `"default"`)
2. Inside gt tmux, town=gt → socket should be `"gt"` (already worked)
3. Outside tmux (daemon), town=gt → socket should be `"gt"` (was `""`)
4. Town name with spaces → sanitized correctly
5. Town name with caps → lowercased correctly

Closes gt-qkekp

### DIFF
--- a/internal/session/registry.go
+++ b/internal/session/registry.go
@@ -111,19 +111,18 @@ func SetDefaultRegistry(r *PrefixRegistry) {
 func InitRegistry(townRoot string) error {
 	var errs []error
 
-	// Set tmux socket for session discovery.
-	// If we're inside tmux, parse the socket from $TMUX (e.g. /tmp/tmux-501/default,pid,idx)
-	// so we query the same server our session lives on. When not inside tmux
-	// (e.g., daemon process), leave the socket unset so tmux uses its default
-	// server — the same one the user's interactive sessions live on.
-	// Session names already include rig prefixes (hq-deacon, gt-witness, etc.)
-	// so there's no collision risk on the default server.
-	if tmuxEnv := os.Getenv("TMUX"); tmuxEnv != "" {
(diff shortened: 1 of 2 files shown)


l0g1x:
Superseded by #2040 which consolidates this PR with #2037 into a single tmux socket migration PR.

PR #2037, opened 2026-02-25
fix(tmux): cross-socket agent menu and keybindings

l0g1x:
## Summary

- **Menu actions**: `buildMenuAction()` generates cross-socket-aware tmux commands — tries `switch-client` first (instant, same socket), falls back to `detach-client -E "tmux -L <socket> attach"` (cross-socket)
- **Binding propagation**: `gt up` calls `EnsureBindingsOnSocket("default")` so `prefix+g` and `prefix+a` work when the user is attached to the default tmux server
- **Guard removal**: When no pre-existing user binding exists on the target socket, skips the `if-shell` session-name guard (on the default socket, no sessions match GT prefixes, so the guard prevented the menu from ever appearing)

## Context

After the socket migration (PR #2034), agent sessions live on `-L gt` while user terminals are on the `default` socket. This broke `prefix+g` (agent switcher) because:

1. The menu action used `switch-client -t 'session'` which only works within the same tmux server
2. The keybinding was only set on the gt socket (via `gt prime`), not on the default socket
3. Even if set on the default socket, the `if-shell` guard checked `#{session_name}` against GT prefixes — no sessions on `default` matched

## Test Plan

- `TestBuildMenuAction_CrossSocket` — verifies action format with/without socket
- `TestEnsureBindingsOnSocket` — binding set without guard when no user binding
- `TestEnsureBindingsOnSocket_PreservesUserBinding` — if-shell guard wraps user's custom binding
- All 41 existing agents tests pass

### DIFF
--- a/internal/cmd/agents.go
+++ b/internal/cmd/agents.go
@@ -28,6 +28,7 @@ const (
 	AgentRefinery
 	AgentCrew
 	AgentPolecat
+	AgentPersonal // Non-GT session (user's terminal session)
 )
 
 // AgentSession represents a categorized tmux session.
@@ -36,6 +37,7 @@ type AgentSession struct {
 	Type      AgentType
 	Rig       string // For rig-specific agents
 	AgentName string // e.g., crew name, polecat name
+	Socket    string // tmux socket name this session lives on
 }
 
 // AgentTypeColors maps agent types to tmux color codes.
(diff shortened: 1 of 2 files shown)


l0g1x:
Superseded by #2040 which consolidates this PR with #2034 into a single tmux socket migration PR.

PR #2040, opened 2026-02-25
fix(tmux): socket migration — derive socket from town name + cross-socket agent menu

l0g1x:
## Summary

Consolidates #2034 and #2037 into a single PR — they are two halves of the same tmux socket migration change.

**Problem**: `InitRegistry()` parsed `$TMUX` to set the default socket. When `gt up` ran from a terminal on the default tmux server, all sessions landed on socket `"default"` instead of a dedicated town socket. After fixing that, the agent switcher menu (`prefix+g`) couldn't switch across tmux servers, and keybindings were missing from the user's default socket.

**Fix** (2 commits):

### Commit 1: Derive socket from town name
- Replace `$TMUX` parsing with `sanitizeTownName(filepath.Base(townRoot))` so `/home/user/gt` → socket `"gt"`
- `sanitizeTownName` already existed with tests but was never wired into `InitRegistry()`
- All callers (CLI, daemon, agents) now target the same dedicated tmux server regardless of launch context

### Commit 2: Cross-socket agent menu and bindings
- `buildMenuAction`: try `switch-client` first (instant, same socket), fall back to `detach-client -E "tmux -L <socket> attach -t <session>"` for cross-socket switches
- `getAllSocketSessions`: list sessions from all known sockets (town + default), grouped by socket with non-selectable headers
- Socket groups shown with descriptive headers (e.g. "hq" for town socket), personal sessions in magenta
- `--` flag terminator before menu items so dash-prefixed session names aren't parsed as tmux flags
- Divider lines only between socket groups, not between rig sub-headers
- `EnsureBindingsOnSocket`: set `prefix+g` and `prefix+a` on a specified socket; called from `gt up` to ensure bindings exist on `"default"` socket
- Skip `if-shell` guard when no user binding to preserve

## Testing

- `TestInitRegistry_SocketFromTownName` (5 cases)
- `TestSanitizeTownName` (10 cases)
- `TestDisplayLabel_PersonalSession`
- `TestBuildMenuAction_PerSessionSocket`
- `TestBuildMenuAction_CrossSocket` (2 cases)
- `TestEnsureBindingsOnSocket`
- `TestEnsureBindingsOnSocket_PreservesUserBinding`

All tests pass. `go vet` and `golangci-lint` clean.

Supersedes #2034 and #2037.

### DIFF
--- a/internal/cmd/agents.go
+++ b/internal/cmd/agents.go
@@ -28,6 +28,7 @@ const (
 	AgentRefinery
 	AgentCrew
 	AgentPolecat
+	AgentPersonal // Non-GT session (user's terminal session)
 )
 
 // AgentSession represents a categorized tmux session.
@@ -36,6 +37,7 @@ type AgentSession struct {
 	Type      AgentType
 	Rig       string // For rig-specific agents
 	AgentName string // e.g., crew name, polecat name
+	Socket    string // tmux socket name this session lives on
 }
 
 // AgentTypeColors maps agent types to tmux color codes.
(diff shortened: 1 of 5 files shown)


l0g1x:

PR #2016, opened 2026-02-25
fix(deacon): respawn dead pane instead of kill+recreate to prevent crash loop

l0g1x:
## Summary

- The deacon exits cleanly after each ~3-minute patrol cycle, but the daemon's 3-minute heartbeat races with the auto-respawn hook's 3-second sleep. The daemon finds the dead pane, treats it as a zombie, kills + recreates the session, and increments the crash counter. After 5 restarts in 15 minutes → crash loop.
- When the deacon manager finds a dead pane (process exited, `remain-on-exit` on), use `respawn-pane` to restart in place instead of killing the session. Returns `ErrAlreadyRunning` → daemon calls `RecordSuccess()` instead of `RecordRestart()`.
- Also adds `IsPaneDead()` and `RespawnPaneDefault()` to the tmux package, and fixes `SetAutoRespawnHook` to include `-L <socket>` in embedded tmux commands (missed by the multi-town socket migration in 33362a75).

## Changes

- `internal/deacon/manager.go`: Dead pane → try `RespawnPaneDefault` before falling through to kill+recreate
- `internal/deacon/manager_test.go`: 3 new tests for dead pane vs zombie shell recovery paths
- `internal/tmux/tmux.go`: Added `IsPaneDead()`, `RespawnPaneDefault()`, fixed socket flag in `SetAutoRespawnHook`
- `internal/tmux/respawn_hook_test.go`: Integration tests for auto-respawn hook on isolated socket servers

## Testing

- All existing deacon tests pass
- 3 new unit tests verify respawn-instead-of-kill, fallthrough on failure, and zombie-still-kills
- 5 new tmux integration tests verify hook respawn, socket isolation, `IsPaneDead`, `RespawnPaneDefault`
- `go vet` and `golangci-lint` clean

PR #2042, opened 2026-02-25
fix(tmux): session reliability, socket migration, and test isolation

l0g1x:
## Summary

Fixes five interconnected tmux session management bugs — session creation losing commands, deacon crash-looping on restart, auto-respawn hooks racing with the daemon, sessions landing on the wrong tmux server, and agent menu unable to switch across servers — and adds per-package test socket isolation so tmux tests never interfere with the user's terminal or with each other.

## Problem: tmux session management is unreliable

**1. Empty polecat windows.** `EnsureSessionFresh` in `internal/tmux/tmux.go` creates a session with a shell, then sends the command via `SendKeys`. If the shell hasn't started by the time `SendKeys` fires, the command is lost and the pane shows an empty shell. This is a race condition visible in production as blank polecat windows after `gt up`.

**2. Deacon crash loop.** `Start()` in `internal/deacon/manager.go` kills and recreates the session on every restart attempt. The daemon's 3-minute heartbeat also triggers restarts. When both fire simultaneously, the session is destroyed and recreated in a loop — the deacon never stabilizes.

**3. Auto-respawn hook races with daemon.** `SetAutoRespawnHook` in `internal/tmux/tmux.go` sets a pane-died hook that sleeps 3 seconds then blindly runs `respawn-pane -k`. If the daemon restarts the session during the 3-second window, the hook kills the fresh agent. Additionally, `run-shell` without `-b` leaks error messages to the user's active pane when the target session has been killed, and the hook uses bare `tmux` without a `-L` socket flag.

**4. Sessions on wrong tmux server.** `InitRegistry()` in `internal/session/registry.go:121-126` parses `$TMUX` to determine the tmux socket. When `gt up` runs from a terminal on the default tmux server, all sessions land on socket `"default"` instead of a dedicated town socket. This defeats multi-town isolation and makes sessions visible to (and killable by) the user's interactive tmux.

**5. Tests share the user's tmux server.** All tmux integration tests across `internal/tmux`, `internal/cmd`, `internal/config`, `internal/polecat`, and `internal/daemon` use `NewTmux()` which targets the default tmux server. Tests create and destroy sessions on the user's live server, and packages running in parallel during `go test ./...` can interfere with each other.

### Session creation hardening (addresses #1)

New `EnsureSessionFreshWithCommand(name, dir, cmd)` creates sessions with the command as the initial pane process, bypassing the shell entirely. No `SendKeys` race. `ErrSessionRunning` added for callers that need to detect healthy existing sessions.

- `daemon.restartPolecatSession` and `lifecycle.restartSession` updated to use the new API
- 65 new tests in `session_creation_test.go` covering blank windows, command visibility, concurrent creation, rapid create/destroy, special characters, Unicode, nudge timing, and race conditions

### Deacon respawn (addresses #2)

`Start()` now checks `IsPaneDead()` first. If the pane is dead, it uses `RespawnPaneDefault()` to restart in place instead of kill+recreate. This avoids racing with the daemon's heartbeat.

### Auto-respawn hook hardening (addresses #3)

New `buildAutoRespawnHookCmd()` generates a hook command with three safety measures:
- `run-shell -b`: background execution prevents output leaking to user's active pane
- `#{pane_dead}` guard: checks the pane is still dead before respawning (avoids killing daemon-restarted agents)
- `|| true`: suppresses errors when the session has been killed entirely
- `-L <socket>` flag: targets the correct tmux server after socket migration

### Socket derivation from town name (addresses #4)

Replace `$TMUX` parsing in `InitRegistry()` with `sanitizeTownName(filepath.Base(townRoot))`. Socket is always derived from the town directory name — identical behavior inside tmux, outside tmux, or on a different tmux server.

Since sessions now live on a dedicated socket, the agent switcher menu needs cross-socket awareness:
- `buildMenuAction`: try `switch-client` first, fall back to `detach+reattach` for cross-socket switches
- `getAllSocketSessions`: list sessions from town + default sockets with grouped headers
- `EnsureBindingsOnSocket`: set `prefix+g` and `prefix+a` on the default socket from `gt up`
- `--` flag terminator before menu items for dash-prefixed session names

### Per-package test socket isolation (addresses #5)

Every package that interacts with tmux now gets an isolated socket via `TestMain`:

| Package | Socket name | Cleanup |
|---------|-------------|---------|
| `internal/tmux` | `gt-test-tmux-<pid>` | `kill-server` + socket file removal |
| `internal/cmd` | `gt-test-cmd-<pid>` | `kill-server` + socket file removal |
| `internal/config` | `gt-test-config-<pid>` | `kill-server` + socket file removal |
| `internal/polecat` | `gt-test-polecat-<pid>` | `kill-server` + socket file removal |
| `internal/daemon` | `gt-test-daemon-<pid>` | `kill-server` + socket file removal |

- `internal/config` uses a package-level `testTmuxSocket` variable instead of importing `tmux` (import cycle: config <- tmux)
- `config/integration_test.go`'s raw `exec.Command("tmux", ...)` calls converted to `tmuxCmd()` helper that injects `-L` flag
- `respawn_hook_test.go` converted from per-test sockets to shared package socket with per-test session cleanup

## Key Design Decisions

| Decision | Rationale |
|----------|-----------|
| Per-package socket, not per-test | Tests within a package are sequential (no `t.Parallel()` in tmux tests). Per-package avoids 120+ server create/destroy cycles (~25ms each) while still isolating across packages that run in parallel. |
| `TestMain` manages socket lifecycle | `TestMain` runs before any tests and cleans up after all complete — correct lifecycle for a shared resource. `t.Cleanup` in individual tests would race with each other. |
| `kill-server` + explicit socket file removal | `tmux kill-server` does NOT remove the socket file from `/tmp/tmux-UID/`. Without explicit `os.Remove`, socket files accumulate across test runs. |
| `sanitizeTownName` for socket, not `$TMUX` | `$TMUX` parsing created three code paths (inside tmux, outside, wrong server). Town name derivation is deterministic regardless of environment. |
| `switch-client` then `detach+reattach` fallback | `switch-client` is instant with no flicker when already on the target socket. Detach fallback only fires for actual cross-socket switches. |

## Testing

```bash
# Session creation + reliability (65 tests)
go test ./internal/tmux/... -count=1 -v -run 'TestEnsureSession|TestNewSession|TestSession'

# Respawn hook (10 tests, ~30s due to 3s hook sleeps)
go test ./internal/tmux/... -count=1 -v -run 'TestAutoRespawn|TestIsPaneDead|TestRespawnPane'

# Socket derivation
go test ./internal/session/... -count=1 -v -run 'Socket|SanitizeTownName'

# Cross-socket menu
go test ./internal/cmd/... -count=1 -v -run 'DisplayLabel_PersonalSession|BuildMenuAction'

# Bindings
go test ./internal/tmux/... -count=1 -v -run 'EnsureBindingsOnSocket'

# Verify socket cleanup (should be 0 after test run)
ls /tmp/tmux-$(id -u)/gt-test-* 2>/dev/null | wc -l
```

## Production evidence

Running on a live town. All sessions on dedicated `-L gt` socket, `prefix+g` working from both sockets, deacon stable with no crash loops:

```
$ tmux -L gt list-sessions
bcc-refinery: 1 windows ...
bcc-witness: 1 windows ...
gt-refinery: 1 windows ...
gt-witness: 1 windows ...
hq-boot: 1 windows ...
hq-deacon: 1 windows ...
hq-mayor: 1 windows ...
```

Supersedes #2011, #2016, #2024, #2031, #2040.

### DIFF
--- a/internal/tmux/respawn_hook_test.go
+++ b/internal/tmux/respawn_hook_test.go
@@ -474,3 +474,92 @@ func TestAutoRespawnHookCmd_Format(t *testing.T) {
 		})
 	}
 }
+
+// TestIsPaneDead_And_RespawnPaneDefault is an integration test that exercises
+// the exact code path the deacon uses to recover from a dead pane:
+//
+//  1. Create a session with remain-on-exit on, running a short-lived process
+//  2. Wait for the process to exit → pane becomes dead
+//  3. Verify IsPaneDead returns true
+//  4. Call RespawnPaneDefault to restart the pane
+//  5. Verify the pane comes back alive with a new PID
+//
+// This tests the deacon's "respawn instead of kill+recreate" path against a
+// real tmux server. The deacon uses this to avoid incrementing the daemon's
+// crash counter when a patrol cycle completes normally.
+func TestIsPaneDead_And_RespawnPaneDefault(t *testing.T) {
+	socket := requireTestSocket(t)
+	session := "test-respawn-default"
+
+	// Create session running a fast-exiting command.
+	testSession(t, socket, session, "sleep 1")
+	defer func() { _ = exec.Command("tmux", "-L", socket, "kill-session", "-t", session).Run() }()
+
+	// Enable remain-on-exit so the pane stays visible after the process exits.
+	if out, err := exec.Command("tmux", "-L", socket, "set-option", "-t", session, "remain-on-exit", "on").CombinedOutput(); err != nil {
+		t.Fatalf("failed to set remain-on-exit: %v\n%s", err, out)
+	}
+
+	// Capture the original PID.
+	originalPID := getPanePID(t, socket, session)
+	t.Logf("Original pane PID: %s", originalPID)
+
+	// Wait for sleep 1 to exit → pane should become dead.
+	t.Log("Waiting for process to exit...")
+	deadline := time.Now().Add(5 * time.Second)
+	paneDied := false
+	for time.Now().Before(deadline) {
+		if isPaneDead(socket, session) {
+			paneDied = true
+			t.Log("Pane is dead (remain-on-exit holding it)")
+			break
+		}
+		time.Sleep(100 * time.Millisecond)
+	}
+	if !paneDied {
+		t.Fatal("pane never died — remain-on-exit may not be working")
+	}
+
+	// This is the exact call the deacon makes:
+	tmx := NewTmux()
+	if !tmx.IsPaneDead(session) {
+		t.Fatal("IsPaneDead returned false but pane should be dead")
+	}
+
+	// Respawn using the API the deacon calls:
+	t.Log("Calling RespawnPaneDefault...")
+	if err := tmx.RespawnPaneDefault(session); err != nil {
+		t.Fatalf("RespawnPaneDefault failed: %v", err)
+	}
+
+	// Wait for the pane to come back alive.
+	t.Log("Waiting for pane to come back alive...")
+	alive := false
+	deadline = time.Now().Add(3 * time.Second)
+	for time.Now().Before(deadline) {
+		if !isPaneDead(socket, session) {
+			alive = true
+			break
+		}
+		time.Sleep(100 * time.Millisecond)
+	}
+	if !alive {
+		t.Fatal("pane did not come back alive after RespawnPaneDefault")
+	}
+
+	// Verify the PID changed — proof the pane was actually respawned.
+	newPID := getPanePID(t, socket, session)
+	t.Logf("New pane PID: %s (was %s)", newPID, originalPID)
+	if newPID == originalPID {
+		t.Error("pane PID did not change after respawn — pane may not have been restarted")
+	}
+
+	// Verify the session is intact (not killed and recreated).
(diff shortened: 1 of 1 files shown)


steveyegge:
Closing this draft — it's 4800+ lines touching tmux sessions, socket migration, and test isolation across 19 files, which is too large for a single PR. The test isolation piece has already been merged via PR #2052 (a focused subset). If you'd like to land the remaining tmux/socket changes, please break them into smaller, focused PRs (one concern per PR). Thanks for the work\!

l0g1x:
Split into 4 focused PRs as requested:

1. **#2057** -- Deacon dead-pane respawn (152 lines, 3 files)
2. **#2058** -- Auto-respawn hook safety (474 lines, 2 files) -- merge after #2057
3. **#2059** -- Session creation race fix (3153 lines, 5 files) -- independent
4. **#2060** -- Socket derivation from town name (632 lines, 6 files) -- independent

Test isolation (concern 5) was already merged via #2052.

PRs 1-4 are independent except #2058 should merge after #2057 (shared test file).

PR #2049, opened 2026-02-26
fix: require Dolt server for beads init and reorder doctor prerequisites

l0g1x:
## Summary

- Gate `AddRig`, `InitBeads`, and `initTownBeads` on `doltserver.IsRunning()` so they fail fast when the server is down
- Move `dolt-binary` and `dolt-server-reachable` doctor checks to positions 3-4, before any check that shells out to `bd`
- Rewrite `getWispIDs` from JSONL file parsing to `bd mol wisp list --json` (live Dolt query)
- Remove stale `issues.jsonl` creation from `InitBeads`

## Problem

1. **Silent failures on missing Dolt server.** `gt rig add` and `gt install` call `bd init` without checking if the Dolt server is running. When it's down, `bd init` fails with cryptic errors deep in the call stack. Users see confusing output instead of a clear "Dolt server is not running" message.

2. **Doctor checks run in wrong order.** `dolt-binary` and `dolt-server-reachable` are registered at positions ~25, after 20+ checks that shell out to `bd` or query the database. When Dolt is down, those earlier checks produce misleading failures before the root cause is identified.

3. **`getWispIDs` reads stale JSONL.** `ready.go` parses `issues.jsonl` to filter wisps from ready work. Since nothing exports Dolt data back to JSONL, this reads an empty or stale file and silently skips wisp filtering, leaking operational work into `gt ready` output.

4. **`InitBeads` creates empty `issues.jsonl`.** This placeholder file serves no purpose in Dolt-server mode and misleads code that checks for its existence as a signal that beads data exists.

## Changes

| File | Change |
|------|--------|
| `cmd/install.go` | Add `doltserver.IsRunning()` gate before `initTownBeads` |
| `cmd/rig.go` | Add `doltserver.IsRunning()` gate before both `InitBeads` call sites in `runRigAdopt` |
| `rig/manager.go` | Add `doltserver.IsRunning()` gate in `AddRig`; remove `issues.jsonl` creation from `InitBeads` |
| `cmd/doctor.go` | Move `DoltBinaryCheck` + `DoltServerReachableCheck` to positions 3-4 |
| `cmd/ready.go` | Rewrite `getWispIDs` from JSONL parsing to `bd mol wisp list --json` |
| `cmd/rig_integration_test.go` | Add `requireDoltServer(t)` + `.dolt-data` directory setup |
| `cmd/gitinit.go` | Remove stale JSONL reference from gitignore template comment |
[...]

PR #2051, opened 2026-02-26
refactor: remove JSONL fallback paths in favor of Dolt-only operations

l0g1x:
## Problem

After upstream commit e04237ab added JSONL bloat detection and kept JSONL as a fallback, the codebase still contains dead JSONL fallback code paths that:

1. **Silently mask Dolt failures** — `beads_agent.go` falls back to `createAgentBeadViaJSONL` when Dolt writes fail, hiding real errors from operators
2. **Duplicate logic in patrol and wisp checks** — `patrol_check.go` and `wisp_check.go` each maintain a complete JSONL fallback path that reads and parses `issues.jsonl`, doubling the code surface area for bugs
3. **Use `issues.jsonl` as a beads-data marker** — `rig_check.go` treats the presence of `issues.jsonl` as proof that beads data exists, but this file is no longer the source of truth
4. **Carry stale git tooling** — `.gitattributes` defines a JSONL merge driver and `.gitignore` excludes `issues.jsonl`, both of which serve no purpose now that JSONL is not written

## Solution

Remove all JSONL fallback paths so that every beads operation goes exclusively through Dolt:

- **(1)** Remove `createAgentBeadViaJSONL` fallback in `beads_agent.go` — Dolt failures now surface immediately
- **(2)** Remove JSONL fallback branches in `patrol_check.go` and `wisp_check.go` — both now use Dolt-only paths
- **(3)** Remove `issues.jsonl` from `hasBeadsData` check in `rig_check.go` and from stale-file lists in `stale_beads_redirect_check.go`
- **(4)** Remove `.gitattributes` JSONL merge driver, `.gitignore` entry, and comment out `sync-branch` in `.beads/config.yaml`
- Add CI guardrail that rejects `issues.jsonl` if accidentally reintroduced
- Remove all JSONL-specific tests and test helpers (net deletion: ~400 lines)

### DIFF
--- a/.beads/config.yaml
+++ b/.beads/config.yaml
@@ -60,7 +60,7 @@
 # - linear.api-key
 # - github.org
 # - github.repo
-sync-branch: beads-sync
+# sync-branch: beads-sync  # Legacy — Dolt server mode doesn't use git sync
 
 # Cross-project dependencies (gt-o3is)
 # Maps project names to paths for external dependency resolution
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,3 +1,1 @@
 
-# Use bd merge for beads JSONL files
-.beads/issues.jsonl merge=beads
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,31 +7,29 @@ on:
     branches: [ main ]
 
 jobs:
-  # Fast check to catch accidental .beads/issues.jsonl changes from contributors
-  check-no-beads-changes:
-    name: Check for .beads changes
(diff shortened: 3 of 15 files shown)


steveyegge:
Sheriff review: Removes all JSONL fallback paths in favor of Dolt-only operations (-485 lines across 15 files). This is the right direction but it's irreversible — once removed, there's no fallback for users who haven't migrated to Dolt. Pairs with PR #2049 (require Dolt server for init). Leaving for maintainer to confirm the migration timeline is right.

PR #2058, opened 2026-02-26
fix(tmux): harden auto-respawn hook with dead-pane guard and background execution

l0g1x:
## Summary

Rewrites `SetAutoRespawnHook` to fix three problems: the hook races with the daemon's heartbeat restart, error output leaks to the user's active pane, and the hook uses bare `tmux` instead of the configured socket.

## Merge order

This is **PR 2 of 4** from the [tmux reliability split](https://github.com/steveyegge/gastown/pull/2042#issuecomment-3964454473). Should merge after PR 1 (#2057, deacon dead-pane respawn) since both contribute to the `respawn_hook_test.go` file.

2. **This PR** -- auto-respawn hook safety
3. Session creation race fix
4. Socket derivation from town name

## Problem

**1. Hook races with daemon heartbeat.** The pane-died hook sleeps 3 seconds then blindly runs `respawn-pane -k`. If the daemon's heartbeat fires during those 3 seconds and restarts the session, the hook kills the fresh agent. The deacon appears to crash-loop but is actually being killed by its own recovery hook.

**2. Error output leaks to user's pane.** `run-shell` (without `-b`) displays command failures on the attached client's current pane. When the target session has been killed entirely, the hook fails and the error message ("returned 1") appears on whichever pane the user is viewing.

**3. Hook uses wrong tmux server.** The hook command runs bare `tmux respawn-pane` which connects to the default server. When sessions live on a dedicated socket (e.g., `-L gt`), the hook targets the wrong server and silently fails.

## Changes

**`internal/tmux/tmux.go`** -- `SetAutoRespawnHook` rewritten with three safety measures:

- `run-shell -b` (background execution) prevents output from leaking to the user's active pane
- Dead-pane guard (`#{pane_dead}` check via `list-panes`) skips respawn if the daemon already restarted the session during the 3-second sleep window. Uses `##{pane_dead}` escaping so the format variable is evaluated at shell execution time, not at hook registration time.
- `|| true` suppresses errors when the session has been killed entirely

New `buildAutoRespawnHookCmd` extracts the hook command construction for testability. The function takes a `tmuxCmd` parameter (e.g., `"tmux -L gt"`) so all embedded tmux calls target the correct socket.

**`internal/tmux/respawn_hook_test.go`** (new) -- 6 test functions:
- `TestAutoRespawnHook_RespawnWorks` -- end-to-end: process dies, hook respawns it
- `TestAutoRespawnHook_SocketFlagInHookCmd` -- verifies `-L` flag in hook command
- `TestAutoRespawnHook_NoSocket` -- verifies bare `tmux` when no socket configured
- `TestAutoRespawnHook_SkipsAlreadyAlive` -- daemon restarts during sleep, hook skips
- `TestAutoRespawnHook_SilentOnSessionKilled` -- no error output when session gone
- `TestAutoRespawnHookCmd_Format` -- unit tests for the command string format (5 subtests)

### DIFF
--- a/internal/tmux/respawn_hook_test.go
+++ b/internal/tmux/respawn_hook_test.go
@@ -1,0 +1,418 @@
+package tmux
+
+import (
+	"fmt"
+	"os"
+	"os/exec"
+	"strings"
+	"testing"
+	"time"
+)
+
+// requireTestSocket returns a per-test socket name and skips the test if
+// tmux is not installed. Each test gets its own socket to prevent interference.
+// The socket server is cleaned up when the test finishes.
+func requireTestSocket(t *testing.T) string {
+	t.Helper()
+	if !hasTmux() {
+		t.Skip("tmux not installed")
+	}
+	socket := fmt.Sprintf("gt-test-hook-%d", os.Getpid())
+	t.Cleanup(func() {
+		_ = exec.Command("tmux", "-L", socket, "kill-server").Run()
+	})
+	return socket
+}
+
+// testSession creates a session on the given socket running a simple command.
+// Returns after the session is confirmed alive.
+func testSession(t *testing.T, socket, session, command string) {
+	t.Helper()
+	args := []string{"-L", socket, "new-session", "-d", "-s", session, command}
+	out, err := exec.Command("tmux", args...).CombinedOutput()
+	if err != nil {
+		t.Fatalf("failed to create test session %q on socket %q: %v\n%s", session, socket, err, out)
+	}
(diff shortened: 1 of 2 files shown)


steveyegge:
Sheriff review: Hardens auto-respawn hook with dead-pane guard and background execution (+474 lines). Related to #2057 (already merged). The concept is sound but 474 lines is substantial — needs review for interaction with the respawn logic in #2057.

steveyegge:

PR #2059, opened 2026-02-26
fix(tmux): eliminate session creation race by running command as initial pane process

l0g1x:
## Summary

Replaces the `EnsureSessionFresh` + `SendKeys` pattern with `EnsureSessionFreshWithCommand`, which runs the agent command as the pane's initial process. This eliminates the race condition that causes empty polecat windows after `gt up`.

## Merge order

This is **PR 3 of 4** from the [tmux reliability split](https://github.com/steveyegge/gastown/pull/2042#issuecomment-3964454473). Independent of PRs 1 and 2; can merge in any order relative to them.

2. Auto-respawn hook safety (#2058)
3. **This PR** -- session creation race fix
4. Socket derivation from town name

## Problem

`EnsureSessionFresh` creates a tmux session with a shell, then calls `SendKeys` to type the agent command into the shell. If the shell hasn't started by the time `SendKeys` fires, the keystrokes are lost. The pane shows an empty shell prompt instead of a running agent. This is visible in production as blank polecat windows after `gt up`, especially under load when multiple sessions start concurrently.

The race window is small (~10-50ms) but reliably triggers when the system is busy starting 5+ agents simultaneously.

## Changes

**`internal/tmux/tmux.go`**

- `ErrSessionRunning` -- new sentinel error for callers that need to detect healthy existing sessions
- `NewSessionWithCommand` / `NewSessionWithCommandAndEnv` -- rewritten to use two-step creation: create session with default shell, enable `remain-on-exit`, then replace the shell with `respawn-pane -k` using the actual command. Adds `workDir` validation and calls `checkSessionAfterCreate` for immediate failure detection.
- `checkSessionAfterCreate` -- new method that checks if a newly created session's command failed immediately (binary not found, syntax error). Uses `remain-on-exit` + `#{pane_dead}` + `#{pane_dead_status}` to detect non-zero exits within 50ms.
- `EnsureSessionFreshWithCommand` -- new method that replaces the `EnsureSessionFresh` + `SendKeys` pattern. Checks for existing sessions, kills zombies, creates with command as initial process. Returns `ErrSessionRunning` if an existing session has a healthy agent.
- `WakePane` -- switched from `resize-pane` to `resize-window` to fix SIGWINCH delivery on single-pane sessions
- Nudge improvements: `sanitizeNudgeMessage` strips control characters, `sendMessageToTarget` chunks large messages, `NudgeSession`/`NudgePane` cancel copy-mode before sending and use 600ms ESC timing

**`internal/daemon/daemon.go`** -- `restartPolecatSession` now builds the startup command first, then calls `EnsureSessionFreshWithCommand` instead of `EnsureSessionFresh` + `SendKeys`. Handles `ErrSessionRunning`.

**`internal/daemon/lifecycle.go`** -- `restartSession` updated with the same pattern.

**`internal/tmux/session_creation_test.go`** (new, 2657 lines) -- 65 tests covering:
- Blank window prevention (command as initial process)
- WorkDir validation (missing dir, file-not-dir, empty)
- Immediate failure detection (bad binary, syntax error)
- Concurrent session creation
- Rapid create/destroy cycles
- Special characters and Unicode in commands
- Nudge timing and sanitization

**`internal/tmux/tmux_test.go`** -- test refactoring: `newTestTmux` helper for consistent socket setup, updated `TestEnsureSessionFresh*` tests.

### DIFF
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -3,6 +3,7 @@ package daemon
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"log"
 	"os"
@@ -1123,7 +1124,6 @@ func (d *Daemon) isRigOperational(rigName string) (bool, string) {
 	return true, ""
 }
 
-
 // processLifecycleRequests checks for and processes lifecycle requests.
 func (d *Daemon) processLifecycleRequests() {
 	d.ProcessLifecycleRequests()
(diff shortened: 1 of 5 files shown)


steveyegge:
Sheriff review: 3100+ lines changing session creation from send-keys to initial pane process. This is a fundamental tmux architecture change that needs careful review. Leaving for maintainer.

steveyegge:

PR #2060, opened 2026-02-26
fix(tmux): derive socket from town name and add cross-socket agent menu

l0g1x:
## Summary

Replaces `$TMUX` environment variable parsing for socket determination with deterministic derivation from the town directory name, and adds cross-socket navigation so `prefix+g` works from any tmux server.

## Merge order

This is **PR 4 of 4** from the [tmux reliability split](https://github.com/steveyegge/gastown/pull/2042#issuecomment-3964454473). Independent of PRs 1-3; can merge in any order.

2. Auto-respawn hook safety (#2058)
3. Session creation race fix (#2059)
4. **This PR** -- socket derivation from town name

## Problem

**Sessions land on the wrong tmux server.** `InitRegistry()` parses `$TMUX` to determine the socket name. When `gt up` runs from a terminal on the default tmux server, all agent sessions land on socket `"default"` instead of a dedicated town socket. This defeats multi-town isolation and makes sessions visible to (and killable by) the user's interactive tmux.

The `$TMUX` parsing also creates three inconsistent code paths: inside tmux (parses socket from env), outside tmux (falls back to "default"), and on a different tmux server (wrong socket entirely).

**Agent menu doesn't work cross-socket.** `prefix+g` only shows sessions on the current socket. Users on the default socket cannot see or switch to agents on the town socket.

## Changes

**`internal/session/registry.go`** -- `InitRegistry()` now calls `sanitizeTownName(filepath.Base(townRoot))` to derive the socket name deterministically from the town directory. The socket is always the same regardless of whether the user is inside tmux, outside tmux, or on a different tmux server.

**`internal/session/registry_socket_test.go`** -- Tests for `sanitizeTownName`: Unicode handling, special characters, long names, whitespace normalization.

**`internal/cmd/agents.go`** -- Cross-socket agent menu:
- `buildMenuAction`: tries `switch-client` first (instant, no flicker), falls back to `detach+reattach` for actual cross-socket switches
- `getAllSocketSessions`: lists sessions from both town and default sockets with grouped headers
- `--` flag terminator before menu items to handle dash-prefixed session names

**`internal/cmd/agents_test.go`** -- Tests for `DisplayLabel_PersonalSession`, `BuildMenuAction` cross-socket behavior, session grouping.

**`internal/cmd/up.go`** -- Calls `EnsureBindingsOnSocket` on the default socket during `gt up` so `prefix+g` and `prefix+a` work even when the user is on a different tmux server.

**`internal/tmux/tmux.go`** -- `EnsureBindingsOnSocket(socket)`: sets GT keybindings on a specific socket. Skips the session-name guard when there's no pre-existing user binding, since on the default socket no session names match GT prefixes and the guard would prevent the menu from appearing.

### DIFF
--- a/internal/cmd/agents.go
+++ b/internal/cmd/agents.go
@@ -28,6 +28,8 @@ const (
 	AgentRefinery
 	AgentCrew
 	AgentPolecat
+	AgentPersonal // Non-GT session (user's terminal session)
+	AgentTest     // Session on a gt-test-* socket (integration tests)
 )
 
 // AgentSession represents a categorized tmux session.
@@ -36,6 +38,7 @@ type AgentSession struct {
 	Type      AgentType
 	Rig       string // For rig-specific agents
 	AgentName string // e.g., crew name, polecat name
+	Socket    string // tmux socket name this session lives on
 }
 
 // AgentTypeColors maps agent types to tmux color codes.
@@ -46,6 +49,8 @@ var AgentTypeColors = map[AgentType]string{
 	AgentRefinery: "#[fg=blue]",
 	AgentCrew:     "#[fg=green]",
 	AgentPolecat:  "#[fg=white,dim]",
+	AgentPersonal: "#[fg=magenta]",
+	AgentTest:     "#[fg=yellow,dim]",
 }
 
 // rigTypeOrder defines the display order of rig-level agent types.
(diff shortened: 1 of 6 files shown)


steveyegge:
Sheriff review: Derives tmux socket name from town name and adds cross-socket agent menu (+632 lines). Significant tmux infrastructure change. Related to #2059. Needs maintainer review.

steveyegge:

PR #2088, opened 2026-02-26
fix(quota): scan both tmux sockets to find all sessions

seanbearden:
## Summary

- **Quota scan/rotate missed sessions on the default socket** — `InitRegistry` sets the tmux socket to the town name (`-L gt`), but crew sessions created before per-town socket isolation live on the system default socket. The scanner only found sessions on one socket, missing rate-limited sessions on the other.
- **Introduced `multiSocketClient`** — aggregates sessions from both the town socket and the system default socket, deduplicating by session name. Each session's operations (CapturePane, GetEnvironment, IsIdle, rotation) target the correct socket.
- **Non-restartable sessions** — `executeKeychainRotation` now gracefully skips sessions without a restart command instead of failing the entire rotation batch.

## Root Cause

`InitRegistry` calls `tmux.SetDefaultSocket("gt")` early in the process lifecycle. `NewTmuxFromEnv()` was supposed to fix this by using `$TMUX`, but when run from outside tmux (e.g., `gt quota scan` from a terminal), `$TMUX` is empty and it falls back to `NewTmux()` which uses the town socket. Sessions live on both sockets during migration to per-town isolation.

## Test plan

- [x] `gt quota scan` — finds 16 sessions across both sockets, correctly detects rate-limited `lc-crew-letty`
- [x] `go test ./internal/quota/` — all existing tests pass
- [x] `go build ./internal/cmd/` — compiles for bear, mayor/rig, refinery/rig
- [ ] `gt quota rotate` — verify rotation targets correct socket per session

### DIFF
--- a/internal/constants/constants.go
+++ b/internal/constants/constants.go
@@ -308,6 +308,7 @@ var DefaultRateLimitPatterns = []string{
 	`limit\s*·\s*resets \d+[:\d]*(am|pm)\b`,         // "limit · resets 7pm" — requires limit context before resets
 	`Stop and wait for limit to reset`,               // /rate-limit-options TUI prompt option 1
 	`Add funds to continue with extra usage`,         // /rate-limit-options TUI prompt option 2
+	`API Error: Rate limit reached`,                  // Mid-stream API 429 during tool use or generation
 	`OAuth token revoked`,                            // Token invalidated after keychain swap
 	`OAuth token has expired`,                        // Token expired — needs fresh auth
 }
--- a/internal/quota/scan_test.go
+++ b/internal/quota/scan_test.go
@@ -239,6 +239,54 @@ Enter to confirm · Esc to cancel`
 	}
 }
 
+// TestScanAll_DetectsAPIError429 verifies detection of mid-stream API 429
+// errors that appear as "API Error: Rate limit reached" — distinct from the
+// interactive TUI prompt which only shows on prompt-submission rate limits.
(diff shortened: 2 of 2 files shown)


DreadPirateRobertz:
As @arttttt noted, this PR is obsolete after commit 635916ab removed per-town socket isolation. All sessions now live on the default socket, so the multiSocketClient approach is no longer needed. Recommend closing.

DreadPirateRobertz (commented):
**Recommend closing — superseded by upstream changes.**

As @arttttt correctly identified, commit 635916ab removed per-town socket isolation entirely (`-L gt`), switching all sessions to the system default tmux socket. Follow-up commit a3f69776 added bidirectional cross-socket zombie cleanup for the transition period.

Since all sessions now live on the same default socket, the `multiSocketClient` approach in this PR is no longer needed — a single `NewTmux()` sees everything.

My original technical review (approach sound, needs dead code removal + tests) still applies to the code quality, but the core problem this PR solves no longer exists on main. The `NewTmuxFromEnv` dead code I flagged is also moot.

**Recommendation**: Close this PR. If any residual dual-socket edge cases surface, they should be filed as new issues against the current single-socket architecture.

seanbearden:
You're right — 635916ab eliminated per-town socket isolation, so the multiSocketClient approach is no longer needed. Closing this.

The one piece still relevant is the `API Error: Rate limit reached` scanner pattern (mid-stream API 429s during tool use). I'll open a focused PR for just that.

PR #2101, opened 2026-02-27
fix(tmux): prevent silent wrong-server connections when town context is missing

pae23:
## Problem

When `gt agents menu` (or `gt feed`) is invoked via a cross-socket tmux binding from a personal terminal, `workspace.FindFromCwd()` fails because the CWD is not inside the town. This means `InitRegistry` is never called, `defaultSocket` stays empty, and `NewTmux()` silently connects to whatever tmux server happens to be running (typically the user's personal `default` socket) rather than the town's isolated socket.

## Fix

Two complementary changes:

### 1. `persistentPreRun` uses env var fallbacks (root.go)

Replace `workspace.FindFromCwd()` with `detectTownRootFromCwd()` (already exists in `handoff.go`), which also checks `$GT_TOWN_ROOT` and `$GT_ROOT` after CWD detection fails. This ensures `InitRegistry` — and therefore `SetDefaultSocket` — is called for any command invoked in an environment where the town root is set via env var (polecat sessions, shell integrations, etc.).

### 2. `NewTmux()` sentinel for missing town context (tmux.go)

Instead of silently using an empty socket (which connects to the default server), `NewTmux()` now follows this fallback chain:

1. `defaultSocket` (set by `InitRegistry`) — normal case
2. `$GT_TOWN_SOCKET` (embedded in cross-socket bindings by `EnsureBindingsOnSocket`, see PR #2099)
3. Sentinel `"gt-no-town-socket"` — produces a clear tmux error (`no server running on …/gt-no-town-socket`) rather than silently targeting the wrong server

### 3. Missing `testmain_test.go` (tmux package)

The comment in `tmux_test.go` referenced a `TestMain` that was never created. Added the missing file so the package's integration tests use a dedicated isolated socket instead of the user's default server. Also fixed two tests in `respawn_hook_test.go` that used `NewTmux()` where they should have used `NewTmuxWithSocket(socket)` for their per-test socket.

## Test plan

- [ ] `go build ./...` passes
- [ ] `go test ./internal/tmux/... -short` passes (socket unit tests: TestSetGetDefaultSocket, TestNewTmuxInheritsSocket, etc.)
- [ ] `go test ./internal/cmd/... -short` passes
- [ ] `TestNewSessionWithCommand_Success` failure is pre-existing on `main` (confirmed independently)

### DIFF
--- a/internal/cmd/root.go
+++ b/internal/cmd/root.go
@@ -103,8 +103,11 @@ func persistentPreRun(cmd *cobra.Command, args []string) error {
 	logCommandUsage(cmd, args)
 
 	// Initialize session prefix registry and agent registry from town root.
-	// Best-effort: if town root not found, the default "gt" prefix is used.
-	if townRoot, err := workspace.FindFromCwd(); err == nil && townRoot != "" {
+	// Try CWD detection first, then fall back to GT_TOWN_ROOT / GT_ROOT env vars.
+	// Env var fallback ensures commands invoked from outside the town directory
+	// (e.g., "gt agents menu" via a cross-socket tmux binding) still connect to
+	// the correct town socket rather than silently using the wrong server.
+	if townRoot := detectTownRootFromCwd(); townRoot != "" {
 		if err := session.InitRegistry(townRoot); err != nil {
 			fmt.Fprintf(os.Stderr, "WARNING: failed to initialize town registry: %v\n", err)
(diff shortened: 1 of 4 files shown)


pae23:
Split into #2102 (root.go env fallbacks) and #2103 (NewTmux sentinel + test fixes).

PR #2114, opened 2026-02-27
fix(convoy): add recovery sweep after Dolt reconnect and at daemon startup

pae23:
## Problem

When the Dolt server is killed mid-run (e.g. by the patrol formula bug fixed in #2110), the ConvoyManager's event poller (`pollStore`) returns errors and drops all close events. Even after Dolt recovers, the stranded-convoy scan runs only every 30s and won't fire a sweep immediately. The result: a convoy that completed during the outage never auto-lands and requires manual `gt convoy check`.

## Fix

Three coordinated changes across `convoy_manager.go`, `dolt.go`, and `daemon.go`.

### 1. Recovery mode — fast retry after Dolt outage (`convoy_manager.go`)

```go
recoveryMode atomic.Bool
```

- Set to `true` when `pollStore()` returns an error (Dolt down)
- While set, `runStrandedScan` resets its ticker to **5s** instead of the normal `scanInterval`
- Cleared after the first successful `scan()` call

### 2. Startup sweep — catch missed convoys on daemon start (`convoy_manager.go`)

```go
func (m *ConvoyManager) runStartupSweep()
```

One-shot goroutine launched in `Start()`. Waits 10s for Dolt to stabilise then calls `scan()` once. Catches convoys that completed while the daemon was stopped.

### 3. Recovery callback — immediate sweep on Dolt reconnect (`dolt.go` + `daemon.go`)

```go
func (m *DoltServerManager) SetRecoveryCallback(fn func())
```

`clearUnhealthySignal()` now detects the unhealthy→healthy transition via `os.Stat()` on the signal file. On transition, fires `onRecoveryFn` in a goroutine. Wired in `daemon.Run()` to call `cm.scan()` immediately when Dolt recovers.

### Recovery flow

```
Dolt killed
  → pollStore() error → recoveryMode=true → strandedScan tick: 5s
  → Dolt restarts → clearUnhealthySignal() → onRecoveryFn() → cm.scan() immediate
  → missed convoy events detected → convoy auto-lands
```

## References

- Issue #04: Convoy does not auto-land when Dolt server is intermittently killed
- PR #2110: fix(patrol) — root cause of Dolt kills

### DIFF
--- a/internal/daemon/convoy_manager.go
+++ b/internal/daemon/convoy_manager.go
@@ -67,6 +67,11 @@ type ConvoyManager struct {
 	// started guards against double-call of Start() which would spawn duplicate goroutines.
 	started atomic.Bool
 
+	// recoveryMode is set true when an event-poll failure is detected (indicating
+	// Dolt is down). While set, runStrandedScan uses a shorter 5s interval so it
+	// retries quickly once Dolt comes back. Cleared after the first successful scan.
+	recoveryMode atomic.Bool
+
 	// lastEventIDs tracks per-store high-water marks for event polling.
 	// Key matches stores map keys ("hq", "gastown", etc.).
 	lastEventIDs sync.Map // map[string]int64
@@ -122,6 +127,9 @@ func (m *ConvoyManager) Start() error {
 	m.wg.Add(2)
 	go m.runEventPoll()
 	go m.runStrandedScan()
+	// Run a one-shot sweep to catch convoys that completed during any previous
+	// outage or while the daemon was stopped.
+	go m.runStartupSweep()
 	return nil
 }
 
@@ -223,6 +231,9 @@ func (m *ConvoyManager) pollStore(name string, store beadsdk.Storage, stores map
 	events, err := store.GetAllEventsSince(m.ctx, highWater)
 	if err != nil {
 		m.logger("Convoy: event poll error (%s): %v", name, err)
+		// Signal recovery mode so the stranded scan shortens its interval and
+		// retries quickly once Dolt comes back.
+		m.recoveryMode.Store(true)
 		return
 	}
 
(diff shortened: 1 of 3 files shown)


DreadPirateRobertz (changes requested):
## Review: fix(convoy): add recovery sweep after Dolt reconnect and at daemon startup

Thanks for the PR, @pae23 — the problem statement is clear and well-motivated. Losing convoy completion events during Dolt outages is a real reliability gap, and the three-pronged approach (recovery mode fast-retry, startup sweep, Dolt reconnect callback) is sound in concept.

### What's done well

- **Recovery mode with adaptive interval** — switching to 5s scans during Dolt recovery is the right approach for catching stranded convoys quickly.
- **Startup sweep after 10s delay** — catches convoys stranded from a previous daemon lifecycle without racing the initial startup.
- **Dolt reconnect callback** — directly ties recovery to the event that matters.

### Issues to address

**1. Concurrent `scan()` calls are not serialized (blocking)**
After this PR, `scan()` can be invoked concurrently from three sources: the ticker goroutine, the startup sweep, and the Dolt recovery callback. `scan()` has no internal synchronization — two concurrent scans could both see the same stranded convoy and double-dispatch via `gt sling`. Add a `sync.Mutex` (e.g., `scanMu`) at the top of `scan()`.

**2. `onRecoveryFn` read not protected by `mu`**
`SetRecoveryCallback` writes under `m.mu`, but `clearUnhealthySignal` reads `m.onRecoveryFn` without holding it. Today this is safe because the caller holds `m.mu`, but the invariant is implicit. Either read under lock or add a `// caller must hold m.mu` comment.

**3. `runStartupSweep` not tracked in `wg`**
The comment says intentional because it's short-lived, but if `Stop()` is called during the 10s sleep, the goroutine could still be mid-`scan()` when `Stop()` returns. Adding it to `wg` is costless and closes that gap.

**4. Recovery ticker reset ordering**
`ticker.Reset` is called before `m.scan()`. If scan clears recovery mode, you get one extra 5s tick. Minor but worth moving the Reset after `scan()` for precision.

**5. No tests**
Three new concurrent code paths with no test coverage. At minimum: test that recovery mode shortens scan interval, test that startup sweep respects context cancellation, test that recovery callback fires on unhealthy-to-healthy transition.

### Summary

The design is good and addresses a real gap. The concurrent `scan()` issue is the blocking concern — duplicate dispatch would cause real problems. The other items are robustness improvements. Looking forward to the next revision.

steveyegge:
Superseded by fix-merge PR #2148, now merged to main.

PR #2135, opened 2026-02-28
fix(tmux): reset window-size after WakePane resize dance

l0g1x:
## Summary

- `WakePane` uses `resize-window` to trigger SIGWINCH for detached sessions, but tmux automatically sets `window-size manual` as a side effect, permanently locking the window at 80x24
- When a human later attaches via `gt mayor attach` (or any session attach), the window never resizes to their terminal — appearing as if another smaller client is viewing the session (dots around edges)
- Fix: reset `window-size` to `latest` after the resize dance so the window auto-sizes to the attaching client


steveyegge:
Already on main as 65d4c9d2 (fix-merged). Thanks @l0g1x!

PR #2138, opened 2026-02-28
fix(patrol): remove --root-only flag dropped from bd mol wisp create

l0g1x:
## Summary

- `bd mol wisp create` no longer supports the `--root-only` flag (removed in recent beads updates)
- `autoSpawnPatrol` in `patrol_helpers.go` still passes `--root-only`, causing patrol wisp creation to fail
- Removes the flag so patrol cycles work correctly with current `bd`

## Details

The `--root-only` flag was used to create patrol wisps without child step wisps (steps were inlined at prime time). The beads CLI dropped this flag — `bd mol wisp create --help` no longer lists it, and passing it causes an error.

This one-line fix removes the flag from the `spawnArgs` slice in `autoSpawnPatrol()`.

### DIFF
--- a/internal/cmd/patrol_helpers.go
+++ b/internal/cmd/patrol_helpers.go
@@ -177,8 +177,8 @@ func autoSpawnPatrol(cfg PatrolConfig) (string, error) {
 		return "", fmt.Errorf("proto %s not found in catalog", cfg.PatrolMolName)
 	}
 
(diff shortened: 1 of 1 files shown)


steveyegge:
Already on main as e5248e0f (fix-merged). Thanks @l0g1x!

PR #2140, opened 2026-02-28
fix: UTF-8 safe title truncation + PrintWarning to stderr

l0g1x:
## Summary

Fixes two related bugs triggered by em-dashes (U+2014) in bead titles, reported by a Discord user trying to use convoys:

- **Wave table mojibake (`â`)**: `renderWaveTable` and `renderConvoyLine` used byte-based truncation (`title[:28]`) which splits multi-byte UTF-8 characters mid-sequence. An em-dash (3 bytes: `0xE2 0x80 0x94`) sliced at byte 28 produces isolated `0xE2` displayed as `â`. Fixed with rune-based truncation, matching the existing pattern in `tui/convoy/view.go:truncate()`.

- **Daemon JSON parse error**: `style.PrintWarning` wrote to **stdout** via `fmt.Printf`. When the daemon ran `gt convoy stranded --json` and `findStrandedConvoys` hit errors, styled warning text (ANSI codes + emoji) contaminated the JSON on stdout, causing the daemon's `json.Unmarshal` to fail with a parse error. Fixed by writing to **stderr**.

## Files changed

| File | Change |
|------|--------|
| `internal/cmd/convoy_stage.go` | Rune-safe title truncation in `renderWaveTable` |
| `internal/tui/feed/convoy.go` | Rune-safe title truncation in `renderConvoyLine` |
| `internal/style/style.go` | `PrintWarning` writes to stderr instead of stdout |
| `internal/style/style_test.go` | Updated tests + new `TestPrintWarning_DoesNotWriteStdout` |
| `internal/cmd/convoy_stage_test.go` | New `TestRenderWaveTable_UTF8Truncation` regression test |

## Test plan

- [x] `go test ./internal/style/` — all pass including new stderr-only assertion
- [x] `go test ./internal/cmd/ -run TestRenderWaveTable` — all pass including UTF-8 regression test
- [x] `go test ./internal/tui/feed/` — all pass
- [x] `go vet` clean, `golangci-lint` clean (pre-existing issues only)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2139, opened 2026-02-28
fix(sling): clean up git branch and convoy on rollback failure

arttttt:
## Problem

When `gt sling` fails after spawning a polecat (e.g., formula instantiation or hook failure), the rollback path (`rollbackSlingArtifacts` and `cleanupSpawnedPolecat`) cleans up the worktree and agent bead but leaves behind two orphaned artifacts:

1. **Git branch** — the polecat's feature branch (local + remote) is never deleted, accumulating stale branches over time.
2. **Auto-convoy** — when `executeSling` creates an auto-convoy *before* formula instantiation, a formula failure triggers rollback but the convoy is not closed, leaving an orphaned convoy in the beads database.

The reference implementation `nukePolecatFull` (used by `gt polecat nuke`) already handles both correctly.

### Git branch cleanup

- Added `Branch` field to `SpawnedPolecatInfo` struct, populated during polecat spawn
- `cleanupSpawnedPolecat` now deletes the polecat's git branch (local + remote) when `Branch` is set, matching `nukePolecatFull` behavior

### Convoy cleanup

- Added `convoyID` parameter to `cleanupSpawnedPolecat` — if non-empty, closes the orphaned convoy via `bd close`
- `rollbackSlingArtifacts` passes its `convoyID` through to `cleanupSpawnedPolecat`
- `executeSling` hook-failure path passes `convoyID` to `cleanupSpawnedPolecat`

### Shared helpers

Branch deletion and repo-git resolution logic was duplicated between `nukePolecatFull` and the new rollback code. Extracted into shared helpers used by both paths:

- `getRepoGitForRig(rigPath)` — creates a Git client for the rig's repo (bare `.repo.git` or `mayor/rig` fallback)
- `deletePolecatBranch(branchName, repoGit, hasPendingMR)` — deletes local + remote branch, skipping remote when MR is pending
- `closeConvoy(convoyID, reason)` — best-effort convoy close via `bd close`

## Files changed

| File | Change |
|------|--------|
| `polecat_spawn.go` | Added `Branch` field to `SpawnedPolecatInfo`, populated from `polecatObj.Branch` |
| `sling_batch.go` | Added `convoyID` param to `cleanupSpawnedPolecat`; added branch + convoy cleanup; extracted shared helpers |
| `sling_dispatch.go` | Pass `convoyID` to `cleanupSpawnedPolecat` on hook failure |
[...]

PR #2125, opened 2026-02-27
fix(mail): auto-nudge idle agents on mail delivery

seanbearden:
## Summary

- `notifyRecipient()` unconditionally queued all mail notifications via `nudge.Enqueue()`, meaning idle agents never woke up until their next session start or turn-boundary hook. This was an overcorrection for #2032's TOCTOU race.
- Replace unconditional queue with wait-idle-first delivery: `WaitForIdle` with a 3s timeout (up from 1s) gives ~15 polls at 200ms — enough to distinguish genuine idle prompts from brief inter-tool-call gaps (~500ms).
- Idle agents get immediate `NudgeSession` delivery; busy agents fall back to the cooperative queue (same behavior as before).

## Changes

| File | Change |
|------|--------|
| `internal/mail/router.go:32` | `DefaultIdleNotifyTimeout`: 1s → 3s |
| `internal/mail/router.go:1491-1519` | Rewrite notification block: wait-idle first, queue fallback |
| `internal/mail/router_test.go` | Add `TestNotifyRecipient_IdleAgent` and `TestNotifyRecipient_BusyAgent` |

## Test plan

- [x] `go build ./cmd/gt` — compiles clean
- [x] `go test ./internal/mail/...` — all tests pass (including 2 new)
- [x] `go test -short ./...` — full suite passes (pre-existing failures in `gt-model-eval` and `internal/cmd` unrelated)
- [x] `golangci-lint run ./internal/mail/...` — 0 issues
- [ ] Manual: send mail to idle crew member → agent wakes within ~3s
- [ ] Manual: send mail to busy crew member → nudge queued, delivered at next turn boundary

Fixes #2032 (overcorrection)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #2032, opened 2026-02-25
Mail notification nudge stuck in input buffer — Enter never fires


seanbearden:
## Bug

Mail notification text appears in a tmux input buffer but Enter never fires, leaving the text stuck. Observed on a session actively Cerebrating (14m+ into generation).

### 1. False idle detection (TOCTOU race)
`WaitForIdle` in `internal/tmux/tmux.go` polls every 200ms for the `❯ ` prompt in the last 5 captured pane lines. Claude Code briefly flashes the `❯` prompt between tool calls. The 1-second `DefaultIdleNotifyTimeout` (`internal/mail/router.go:31`) is enough to catch this flash and report "idle" when the session is actually mid-conversation.

### 2. Destructive Escape key during Cerebrating
`NudgeSession` (`internal/tmux/tmux.go:991`) sends Escape before Enter for vim mode compatibility. The sequence: text pasted → 500ms wait → Escape → 100ms → Enter.

By the time Escape arrives (~600ms after idle was detected), the session may have resumed Cerebrating. Escape during active generation can cancel/disrupt the TUI state, so Enter doesn't register as "submit."

## Affected Code

- `internal/mail/router.go:1447` — `WaitForIdle` call in `notifyRecipient`
- `internal/mail/router.go:31` — `DefaultIdleNotifyTimeout = 1 * time.Second`
- `internal/tmux/tmux.go:966-1009` — `NudgeSession` (Escape + Enter sequence)
- `internal/tmux/tmux.go:1776-1808` — `WaitForIdle` (single-poll match)

## Proposed Fix

Two-part fix:

1. **Stable idle detection**: Require `❯` prompt to appear across 2 consecutive polls (not a single hit) to rule out brief inter-tool-call flashes.
2. **Always queue mail notifications**: Mail isn't urgent enough to justify interrupting. Use the cooperative hook path (`nudge.Enqueue`) unconditionally for mail delivery, reserving immediate `NudgeSession` for explicit `gt nudge` commands.

## Reproduction

Send mail to an agent that is actively Cerebrating (mid-generation on a long task). The notification text will appear in the input buffer but not be submitted.

PR #2159, opened 2026-02-28
fix(witness): add spawning guard to prevent freshly-slung polecat kills

### LINKED ISSUE #2036: bug: witness patrol formula lacks spawn-awareness, nukes freshly-slung polecats
## Summary

The witness Claude agent's patrol formula (`mol-witness-patrol`) has no awareness of the `spawning` agent state. During the `survey-workers` step, it checks every polecat with `hook_bead` assigned against `tmux has-session`. A freshly-slung polecat gets `hook_bead` set before its tmux session starts, so the witness sees "hook assigned, session dead" and nukes it as a zombie.

This is a race condition between `gt sling` (which sets bead state before creating the tmux session) and the witness patrol cycle. On large repos where sling is slow (worktree creation on 80k commits / 4.8GB), the window is wide enough to hit reliably.

## Root Cause

The `survey-workers` step in `mol-witness-patrol.formula.toml` defines an agent_state action table:

| agent_state | Action |
|-------------|--------|
| running | Check for zombie, then progress |
| idle | Auto-nuke if clean |
| stuck | Handle stuck protocol |
| done | Verify cleanup triggered |

**`spawning` is not in the table.** A polecat in `agent_state=spawning` with `hook_bead` assigned falls through to zombie detection, which sees no tmux session and nukes it.

Meanwhile, the Go-level `DetectZombiePolecats` in `handlers.go` has been hardened over time with guards, thresholds, and cleanup_status checks. The daemon also has a 5-minute spawning guard. But the witness formula is a parallel zombie detector that bypasses all of those guards.

## Evidence

Confirmed from witness transcripts across 3+ sessions: witness runs `gt polecat nuke --force` on polecats that were slung seconds/minutes earlier. Parking the witness allowed polecats to survive, confirming the witness as the killer.

Cases: multiple polecats across arch rig (rictus, furiosa, nux, slit).

## Reproduction

1. Have a large repo (50k+ commits) as a rig
2. Ensure witness is running
3. `gt sling <bead> <rig>`
4. Witness patrol catches the polecat mid-setup and nukes it

On smaller repos, sling completes fast enough that the tmux session exists before the next patrol cycle. This explains why the bug hasn't been widely reported.

## Impact

- Polecats on large repos are killed before they can start working
- Work must be re-dispatched, wasting time and resources
- Workaround is parking the witness before slinging, which defeats the purpose of automated oversight

## Proposed Fix

Update `mol-witness-patrol.formula.toml` survey-workers step:

1. Add `spawning` to the agent_state action table — skip zombie detection entirely for spawning polecats
2. Add a staleness check: if `spawning` for > 5 minutes (matching daemon's guard), flag as potential stuck spawn and escalate rather than nuke

Optionally, add a `SpawnGracePeriod` to the Go-level `DetectZombiePolecats` as defense-in-depth (WIP code exists on a stash).

## Related Issues

- #1973 — daemon pre-sync spawn storm on large repos (P1, open). Fix on main (675ae280) makes sling faster, narrowing the race window but not eliminating it.
- #1379 — polecats nuked before gt done (P1, closed). Same symptom from bulk nukes. Fixed with pre-nuke guards that the formula bypasses.
- #589 — Deacon restart loop kills newly started sessions (closed). Exact same pattern: health check has no awareness session was just started. Fixed by adding grace period.
- #470 — Polecat startup race condition (closed). Witness nudges during API init. Same spawn-time race window.
- #1409 — tmux split pane false zombie detection (closed). False positive zombie detection from incorrect state check.
- #1025 — Daemon kills non-Claude agent sessions as zombies (closed). Detection logic missing valid state.
- #1085 — Auto-cleanup zombie polecats (open). Documents Go-level zombie detection but doesn't mention witness formula as parallel detector.

## Pattern

Gas Town has repeatedly added grace periods and state-awareness to prevent subsystems from killing freshly-started or valid sessions (#589, #470, #1409, #1025). The witness patrol formula is the last subsystem that hasn't gotten this treatment.

DreadPirateRobertz:
## Summary

- Add 5-minute spawning guard to witness zombie detection, preventing freshly-slung polecats from being killed before their tmux session is created
- `gt sling` sets `hook_bead` atomically before the session exists, so the witness saw "hook assigned, session dead" and classified the polecat as a zombie
- Matches the existing daemon spawning guard in `checkPolecatHealth`

## Changes

**`internal/witness/handlers.go`**:
- `isZombieState()`: Exclude `"spawning"` state — it's a transient startup state, not a zombie indicator
- `detectZombieDeadSession()`: Add spawning guard with 5-minute time window using `updated_at` timestamp
- `getAgentBeadStateWithTimestamp()`: New function to read `updated_at` alongside `agent_state` and `hook_bead`
- Updated `DetectZombiePolecats` doc comment

**`internal/witness/handlers_test.go`**:
- Updated `TestZombieClassification_SpawningState` to assert spawning is NOT a zombie state
- Updated `TestZombieClassification_HookBeadAlwaysZombie` to assert spawning+hook is NOT a zombie
- Updated `TestZombieClassification_NoHookNoActiveState` to include spawning in non-zombie list
- All tests now call `isZombieState()` directly instead of inlining duplicate logic

**`internal/formula/formulas/mol-witness-patrol.formula.toml`**:
- Added `spawning` row to the agent_state action table (action: SKIP with 5-min grace period)
- Added spawning guard documentation to zombie detection instructions

## Related Issue

Fixes #2036

## Test plan

- [x] `go build ./...` — compiles
- [x] `go test ./internal/witness/...` — all tests pass
- [x] `go vet ./internal/witness/...` — clean


arttttt (changes requested):
## Review: Spawning Guard Approach

### Bug is real

The race condition is confirmed: `gt sling` writes `hook_bead` to DB before the tmux session exists → witness sees "hook assigned, session dead" → kills the polecat. Reproducible on large repos. No argument there.

### The fix is wrong

The 5-minute spawning guard is a timer-based workaround that masks the race condition instead of eliminating it. Specific problems:

**1. Magic number with no basis.**
Why 5 minutes? Spawn takes 3 seconds on a fast machine and could take 7 minutes on a slow one with a large repo. The number is arbitrary — it's not derived from any measured bound.

**2. Guards don't scale.**
This is the same pattern that already exists in the daemon (`checkPolecatHealth`, lines 1778-1803). Issues #589, #470, #1409, #1025 document the same class of bug. Every new component that reads polecat state will need its own copy of this guard. That's not a fix — it's whack-a-mole.

**3. The race condition still exists.**
The guard doesn't synchronize anything. It just says "wait and hope the spawn finishes". If spawn takes longer than 5 minutes (large repo clone, network issues), the polecat still gets killed.

### Correct approach

The root cause is that `hook_bead` is written to the DB **before** the tmux session is created. This makes the intermediate state visible to other components.

**Fix: write `hook_bead` after session creation, not before.**

Move the DB write to after `SpawnPolecatForSling()` confirms the tmux session exists. Then witness never sees "hook assigned, session dead" because the hook is only assigned when the session is alive. No timer, no guard, no magic number. One change in the spawn sequence eliminates the race condition for all consumers — witness, daemon, and any future component.

If there's a reason `hook_bead` must be set before the session (e.g., hook-first dispatch), then at minimum use a ready-signal pattern: polecat reports "alive" after session start, and all components wait for that signal — not for an arbitrary timer.

### Additional issues

**Scope creep.** The PR bundles 3 unrelated fixes (UTF-8 truncation, mail notification timeout from #2032, PrintWarning stderr fix). These should be separate PRs.

**Test padding.** `sling_rollback_cleanup_test.go` is 694 lines where ~500 are copy-pasted setup boilerplate. Tests verify "doesn't panic" rather than "works correctly". One test uses `t.Skip("Known issue: panics with nil")` — documenting a bug instead of fixing it. Extract a shared test helper and write assertions that verify actual behavior.

### Recommendation

Reject the timer-based guard. Fix the ordering: set `hook_bead` after the session exists, or implement a ready-signal. This eliminates the race condition at the source instead of papering over it with a timeout.

arttttt:
Additional issue missed in the original review: 15 of 18 files in this PR are cherry-picks from other contributors' open PRs, unrelated to the witness spawning guard fix:

- UTF-8 truncation fix: `convoy_stage.go`, `tui/feed/convoy.go`, `style/style.go` + tests — from #2140 (l0g1x)
- Sling rollback cleanup: `polecat.go`, `polecat_spawn.go`, `sling*.go` + 694-line test file — from #2139 (arttttt)
- Mail auto-nudge: `mail/router.go` + tests — from #2125 (seanbearden)
- PrintWarning stderr: `style/style.go` + tests — from #2140 (l0g1x)

Only 3 files (`witness/handlers.go`, `witness/handlers_test.go`, formula TOML) are the actual spawning guard change. The other 15 files should not be in this PR — those original PRs are still open and should be merged on their own.

steveyegge:
Already merged to main via manual merge. Closing.

PR #2168, opened 2026-02-28
fix(nuke): make --force bypass unmerged commits check

arttttt:
## Problem

Commit [`4bd189be`](https://github.com/steveyegge/gastown/commit/4bd189be) (`fix: preserve remote branch with unmerged commits during polecat nuke`, gt-rm9f) added a safety check in `nukePolecatFull` that preserves remote branches with unmerged commits. This prevents data loss when polecats push branches but don't create MR beads (e.g., due to missing formula).

However, this check is unconditional — it runs even with `--force`. The `--force` flag help text says "Force nuke, bypassing all safety checks (LOSES WORK)", but the unmerged commits check is not bypassed. This creates a dead loop:

1. Respawn suggests "use `gt polecat nuke --force` to abandon the MR"
2. User runs `nuke --force`
3. Remote branch is still preserved due to unmerged commits
4. Next respawn still sees the unmerged branch and fails again

The `--force` flag correctly bypasses pre-flight safety checks (unpushed commits, uncommitted changes, stashed work, hook beads, open MRs) but does NOT bypass the inline unmerged commits check in `nukePolecatFull`.

## Solution

Wrap the unmerged commits check in `nukePolecatFull` with `if !polecatNukeForce`:

- **Without `--force`**: unmerged commits check runs as before — remote branch is preserved if it has commits ahead of `origin/main`
- **With `--force`**: unmerged check is skipped — remote branch is deleted unconditionally, matching the help text promise

`nukePolecatFull` is also called from `runPolecatStale --cleanup` which has no force flag — the package-level `polecatNukeForce` is false there, so the safety check is preserved for automated cleanup.

## Files changed

| File | Change |
|------|--------|
| `polecat.go` | Wrap unmerged commits check with `if !polecatNukeForce` |
| `polecat_nuke_force_test.go` | Two tests: verify unmerged check runs without force, verify force bypasses it |

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./internal/cmd/...` — all tests pass
- [x] `TestNukePolecatFull_UnmergedCommitsWithoutForce` — without force, remote branch preserved
- [x] `TestNukePolecatFull_UnmergedCommitsWithForce` — with force, remote branch deleted

### DIFF
--- a/internal/cmd/polecat_nuke_force_test.go
+++ b/internal/cmd/polecat_nuke_force_test.go
@@ -1,0 +1,186 @@
+package cmd
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/steveyegge/gastown/internal/git"
+	"github.com/steveyegge/gastown/internal/polecat"
+	"github.com/steveyegge/gastown/internal/rig"
+	"github.com/steveyegge/gastown/internal/tmux"
+)
+
+// TestNukePolecatFull_UnmergedCommitsWithoutForce tests that unmerged commits
+// prevent remote branch deletion when --force is NOT set.
+func TestNukePolecatFull_UnmergedCommitsWithoutForce(t *testing.T) {
+	tmpDir := t.TempDir()
+	rigName := "testrig"
+	polecatName := "tester"
+
+	// Create bare origin repo (this simulates the "remote" origin)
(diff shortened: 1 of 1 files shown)


steveyegge:
Merged via fix-merge with conflict resolution

PR #2171, opened 2026-02-28
fix: preserve agent-bead metadata and harden crew bead creation

EchBTI:
## Summary
This PR addresses two regressions seen while running `gt doctor --fix` and `gt crew add`:

1. `agent-beads-exist` false warnings when the same agent bead ID appears in both issues and wisps, and the wisp record lacks labels/type.
2. `gt crew add` warnings like `could not create agent bead ... issue not found` from a fragile `Show()+Create` flow.

## Changes
- `internal/beads/beads_agent.go`
  - `ListAgentBeads` now loads issues first and wisps as fallback.
  - Added `mergeAgentBeadSources` to preserve issue metadata (`labels`, `type`) on duplicate IDs.
  - Updated function comments to match actual precedence.

- `internal/cmd/crew_add.go`
  - Added `upsertCrewAgentBead` helper.
  - Switched crew bead creation from `Show()+CreateAgentBead` to `CreateOrReopenAgentBead` (idempotent upsert semantics).
  - Added narrow interface `agentBeadUpserter` for deterministic unit testing.

- Tests
  - `internal/beads/beads_agent_test.go`
    - Added `TestMergeAgentBeadSources` with coverage for duplicate precedence, fallback inclusion, and nil maps.
  - `internal/cmd/crew_add_test.go` (new)
    - Added `TestUpsertCrewAgentBead` validating canonical bead ID/title/fields and error propagation.

## Acceptance Criteria
- [x] `ListAgentBeads` preserves issue metadata when duplicate IDs exist in wisps.
- [x] Wisps-only agent beads are still discoverable.
- [x] `gt crew add` uses upsert behavior (`CreateOrReopenAgentBead`) and no longer depends on a brittle `Show()+Create` race/path.
- [x] New tests cover merge behavior and crew upsert behavior.
- [x] Full Go test suite passes.

## Validation
```bash
go test ./...
```

## Notes
- Separate from this PR, upgrading local `bd` to include support for `--include-infra` resolved earlier local doctor failures tied to infra filtering.


DreadPirateRobertz (commented):
Clean, well-reasoned fix. The motivation for flipping the merge priority is clear (labels/type are only reliable in issues), the `mergeAgentBeadSources` helper is correctly implemented, and the fake-interface test pattern for `crew_add` is the right call.

One thing worth a quick look: the old `ListAgentBeads` tolerated an issues-table error as long as wisps had data (`if err != nil && len(result) == 0`). The new code makes that error fatal unconditionally. This is the correct behavior in principle, but if there are environments where the issues table is transiently unavailable (e.g., during a migration window or partial `bd init`), this could surface errors that were previously silent. Consider whether callers need any tolerance here, or whether a comment acknowledging the stricter behavior is sufficient.

Otherwise, ship it — the test coverage on `mergeAgentBeadSources` covers the contract well (duplicate resolution, gap-filling, nil inputs).

PR #2174, opened 2026-02-28
fix: prevent dolt test server zombie leaks

arttttt:
## Problem

Dolt test servers (`dolt sql-server`) leak as zombie processes when test runs are interrupted. This happens because:

1. **`go test -timeout` sends SIGKILL** to the test binary — deferred cleanup never runs, the dolt child process continues running with no parent
2. **No `Pdeathsig`** — on Linux, the OS can automatically signal a child when its parent dies, but we weren't using this
3. **No process group isolation** — `cmd.Start()` runs dolt in the same process group as the test, but without `Setpgid` we can't kill the entire process tree
4. **Windows had zero cleanup** — `reapStaleDoltServers()` only existed on Unix; Windows zombies accumulated forever
5. **`gt doctor` was blind to it** — no way to detect or clean up leaked test servers outside of test runs

Over time this leads to dozens of orphaned `dolt sql-server` processes consuming memory and ports.

### Fix 1: Process group + Pdeathsig (Unix)

**Files:** `internal/testutil/doltserver_unix.go`, `doltserver_pdeathsig_linux.go`, `doltserver_pdeathsig_other.go`

When starting a test dolt server, we now call `util.SetProcessGroup(cmd)` which:
- Sets `Setpgid: true` — dolt runs in its own process group
- Sets `cmd.Cancel` to kill the entire process group via `syscall.Kill(-pid, SIGKILL)`

On Linux specifically, we also set `Pdeathsig: SIGTERM` — the kernel automatically sends SIGTERM to the dolt process when the parent test process dies for any reason (SIGKILL, OOM, crash). This is the primary defense against zombies.

macOS doesn't support `Pdeathsig`, so it relies on the reaper (see Fix 4).

We reuse the existing `util.SetProcessGroup()` from `internal/util/exec_unix.go` rather than duplicating the logic.

### Fix 2: Windows stale server reaping

**File:** `internal/testutil/doltserver_windows.go`

Added `reapStaleDoltServers()` for Windows, matching the Unix version:
- Uses `tasklist /FO CSV /V` to find `dolt.exe` processes
- Filters for command lines containing `dolt-test-server` (test servers only)
- Skips production servers (port 3307)
- Gets process age via `wmic process get CreationDate`
- Kills stale processes via `taskkill /PID <pid> /F`

Called at the start of `startDoltServer()`, same as Unix.

### Fix 3: Doctor check for zombie detection

**Files:** `internal/doctor/dolt_test_server_leak_check.go`, `dolt_test_server_leak_check_windows.go`, `internal/cmd/doctor.go`

New `dolt-test-server-leaks` check for `gt doctor`:
- **Run():** scans `ps -eo pid,etime,args` for `dolt sql-server` processes with `dolt-test-server` in the data dir path. Reports count, PID, age, and data directory for each zombie.
- **Fix():** sends SIGTERM, polls for exit (100ms intervals, 3s timeout), then SIGKILL if still alive. Cleans up data directory and PID/lock files.
- No-op on Windows (where testutil already handles reaping at test startup).

### Fix 4: Reap threshold reduced to 5 minutes

**File:** `internal/testutil/doltserver_unix.go`

Changed from `reapStaleDoltServers(10 * time.Minute)` to `5 * time.Minute`. A test dolt server should never legitimately run for 5 minutes — our longest integration tests complete well under that. This cuts the window where zombies accumulate during heavy parallel testing.

### Fix 5: PID/lock file cleanup during reap

**File:** `internal/testutil/doltserver_unix.go`

Previously, `reapStaleDoltServers()` killed the process but left behind `/tmp/dolt-test-server-*.pid` and `.lock` files. These stale files could confuse subsequent test runs.

Now after killing stale servers, `cleanupStalePIDFiles()` scans `/tmp/dolt-test-server-*.pid`, reads the PID from each file, and removes the file (plus its `.lock` companion) if the PID was just killed or the process is no longer running.

## Test plan
- [ ] `go build ./internal/testutil/ ./internal/doctor/ ./internal/cmd/` passes
- [ ] `go vet` on changed packages is clean
- [ ] `go test ./internal/testutil/ ./internal/doctor/` passes
- [ ] `gt doctor` shows new `dolt-test-server-leaks` check
- [ ] Start a test, kill it with `kill -9`, verify zombie is reaped on next test start (< 5 min threshold) or via `gt doctor --fix`

### DIFF
--- a/internal/testutil/doltserver_unix.go
+++ b/internal/testutil/doltserver_unix.go
@@ -3,6 +3,7 @@
 package testutil
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"os/exec"
@@ -208,7 +209,7 @@ func startDoltServer() error {
 		return fmt.Errorf("creating dolt data dir: %w", err)
 	}
 
-	cmd := exec.Command("dolt", "sql-server",
+	cmd := exec.CommandContext(context.Background(), "dolt", "sql-server",
 		"--port", doltTestPort,
 		"--data-dir", dataDir,
 	)


DreadPirateRobertz (commented):
Nice work — the layered defense (arg-filter + port exclusion), proper SIGTERM grace period, and the `Pdeathsig` addition on Linux to prevent the leak at the source are all solid choices. The platform split between `_linux.go`, `_other.go`, and `_windows.go` for `setPdeathsig` is exactly the right approach.

Two things worth looking at before merge:

1. **Production server exclusion is fragile.** `findZombieServers` relies on `--port 3307` being hardcoded in the process args. If that ever changes, test servers using a non-default port could be incorrectly spared. The `dolt-test-server` data-dir token is doing the real work here — consider documenting that assumption or dropping the port guard in favor of the data-dir filter alone.

2. **`wmic` deprecation on Windows.** `wmic` is deprecated on Windows 11 in favor of PowerShell's `Get-Process`. Worth noting in a comment so it gets updated before it breaks.

Minor: `proc.Wait()` after `proc.Kill()` on a process you didn't `Start` will always fail with `ECHILD` on Unix — the `_, _` discard is harmless but the call can just be removed.

PR #2176, opened 2026-02-28
fix(quota): unify memory directories across accounts via shared symlinks

seanbearden:
## Summary

- Adds `gt quota unify-memory` subcommand that replaces per-account `memory/` directories with symlinks to a shared canonical location at `~/.claude/shared-memory/<project-path>/`
- Auto-unifies post-rotation in `executeKeychainRotation()` so newly-rotated sessions immediately get shared symlinks
- Adds `memory-symlinks` doctor check that detects non-symlinked memory dirs and suggests `gt quota unify-memory`

## Problem

Claude Code stores auto-memory at `~/.claude-accounts/<account>/projects/<path>/memory/`. When agents run under different accounts (quota rotation or manual switch), each account gets its own isolated memory directory. Memories saved in one account are invisible after rotating to another — "fractured memories." This also affects all agents in a rig (polecats and crew) that may be assigned different accounts.

## Approach

Shared canonical memory via symlinks:
1. Scan all `~/.claude-accounts/*/projects/*/memory/` directories
2. Group by project path, merge content (largest MEMORY.md wins)
3. Replace each account's `memory/` dir with a symlink to `~/.claude/shared-memory/<project-path>/`
4. `--dry-run` flag for preview

## Files changed

| File | Change |
|------|--------|
| `internal/quota/memory.go` | New: `UnifyMemory()`, `UnifyProjectMemoryForConfigDir()` |
| `internal/quota/memory_test.go` | New: 10 tests covering merge, dry-run, idempotency, edge cases |
| `internal/cmd/quota.go` | Add `unify-memory` subcommand; post-rotation auto-unify hook |
| `internal/doctor/memory_symlinks_check.go` | New: `memory-symlinks` diagnostic check |
| `internal/cmd/doctor.go` | Register the new check |

## Test plan

- [x] `go build ./cmd/gt` — builds clean
- [x] `go test ./internal/quota/... -run TestUnify` — all 10 tests pass
- [x] `go vet` — no issues
- [ ] Manual: `gt quota unify-memory --dry-run` → shows what would be unified
- [ ] Manual: `gt quota unify-memory` → creates shared dirs + symlinks
- [ ] Manual: `gt doctor` → reports clean memory-symlinks check
- [ ] Manual: verify Claude Code reads/writes MEMORY.md through symlinks


DreadPirateRobertz (commented):
## Review: `fix(quota): unify memory directories across accounts via shared symlinks`

**Reviewer:** zhora (gastown crew)
**Scope:** `internal/quota/memory.go` (new), `internal/quota/memory_test.go` (new), `internal/doctor/memory_symlinks_check.go`, `internal/cmd/quota.go`

---

### MUST-FIX

**1. Symlink target string comparison — relative vs absolute path mismatch (confidence: 85)**
`memory.go` — `os.Readlink()` returns the exact stored target (could be relative). Comparing against an absolute `sharedDir` will fail to match equivalent relative symlinks. Fix: use `filepath.EvalSymlinks()` or `filepath.Abs()` on both sides before comparison.

**2. Race condition in `replaceWithSymlink()` — data loss on failure (confidence: 85)**
`memory.go` — Between `os.RemoveAll(memoryDir)` and `os.Symlink(sharedDir, memoryDir)`, another process could create the directory. If `Symlink` fails, the memory directory is already deleted — data loss. Fix: add retry logic or atomic check-and-create, and consider backing up before removal.

**3. `mergeMemoryContent()` errors don't prevent directory deletion (confidence: 82)**
`memory.go` — If file copying fails during merge, the function silently continues, then `replaceWithSymlink` deletes the original directory. Files are lost. Fix: return errors from merge and abort before deletion if merge failed.

**4. No Windows symlink platform check (confidence: 80)**
`memory.go` — `os.Symlink()` requires admin privileges or developer mode on Windows. No platform detection or informative error. Fix: add `runtime.GOOS` check with actionable error message.

### SHOULD-FIX

**5. Merge strategy uses file size only — fragile (confidence: 80)**
`memory.go` — Picks largest `MEMORY.md` by size. A file with lots of whitespace beats one with more content. Consider preferring `ModTime` (most recently modified) as primary criterion with size as tiebreaker.

**6. Doctor check doesn't detect broken symlinks (confidence: 80)**
`memory_symlinks_check.go` — Counts symlinks but doesn't verify targets are valid. A broken symlink (deleted target) reports as healthy. Fix: `os.Stat()` the readlink target to verify it exists.

---

| # | Severity | File | Issue | Confidence |
|---|----------|------|-------|------------|
| 1 | MUST-FIX | `memory.go` | Symlink target comparison (relative vs absolute) | 85 |
| 2 | MUST-FIX | `memory.go` | Race condition in replaceWithSymlink — data loss | 85 |
| 3 | MUST-FIX | `memory.go` | Merge errors don't prevent directory deletion | 82 |
| 4 | MUST-FIX | `memory.go` | No Windows symlink platform check | 80 |
| 5 | SHOULD-FIX | `memory.go` | File-size-only merge strategy is fragile | 80 |
| 6 | SHOULD-FIX | `doctor/` | Broken symlinks not detected | 80 |

Note: CI lint/Windows failures appear pre-existing, not caused by this PR. Core test suite passes.

steveyegge:
Thanks @seanbearden — fractured memories across account rotation is a real problem and this is the right general approach (symlinks to shared canonical location).

However, the data safety concerns from the review are blocking. This command operates destructively on user memory files, so it needs to be more defensive:

1. **Data loss on symlink failure** — `os.RemoveAll(memoryDir)` runs before `os.Symlink()`. If the symlink call fails for any reason, the original data is already gone. Need to back up before deleting, or use a rename-based atomic swap.

2. **Merge errors silently ignored** — if file copying fails during merge, the function continues and then deletes the original directory. Should abort before deletion if merge didn't fully succeed.

3. **Relative vs absolute symlink comparison** — `os.Readlink()` can return relative paths. Use `filepath.EvalSymlinks()` or resolve both sides to absolute before comparing.

Please address the data safety issues (#1 and #2 especially) and we'll take another look. The feature is welcome — just needs safer handling of the destructive operations.

DreadPirateRobertz:
### Code review

Found 1 issue:

1. `UnifyProjectMemoryForConfigDir` called with `~/.claude` fallback path, producing invalid account name

When `CLAUDE_CONFIG_DIR` is not set for a session, `currentConfigDir` falls back to `home + "/.claude"` (line 741). This value is later passed to `quota.UnifyProjectMemoryForConfigDir(accountsBase, sharedBase, currentConfigDir)` at line 852, where `accountsBase` is `~/.claude-accounts`.

Inside `UnifyProjectMemoryForConfigDir`, `filepath.Rel(accountsDir, configDir)` computes the relative path from `~/.claude-accounts` to `~/.claude`, which produces `"../.claude"`. Then `strings.SplitN(rel, sep, 2)[0]` extracts `".."` as the `accountName`, causing `projectsDir` to resolve to `~/.claude-accounts/../projects` (i.e., `~/projects`). The function's own comment documents the assumption that `configDir` is like `~/.claude-accounts/dev1`, which the `~/.claude` fallback violates.

In the best case this is a silent no-op (no `~/projects` directory); in the worst case it scans unrelated content under `~/projects`. A guard should check that `currentConfigDir` is actually under `accountsBase` before calling `UnifyProjectMemoryForConfigDir`, since the `~/.claude` fallback represents the default single-account scenario where account-based unification does not apply.

https://github.com/steveyegge/gastown/blob/d84762b63d7e549170b84ac51c074a41b0ab2b9d/internal/cmd/quota.go#L740-L742

https://github.com/steveyegge/gastown/blob/d84762b63d7e549170b84ac51c074a41b0ab2b9d/internal/cmd/quota.go#L849-L854

https://github.com/steveyegge/gastown/blob/d84762b63d7e549170b84ac51c074a41b0ab2b9d/internal/quota/memory.go#L50-L60

steveyegge:
Thanks for tackling this -- fractured memories across accounts is a real problem, and the code here is clean and well-tested.

After discussion, we've decided to go a different direction: rather than symlinking Claude Code's filesystem-based memory dirs, we're going to build the memory layer on top of beads (our persistent data plane). Beads already survive account rotation, are queryable, and are injected into agent context at prime time. The symlink approach works but couples us to Claude Code's internal filesystem layout, which isn't a stable API.

The plan is to add a simple on-ramp (something like `bd remember`) that creates labeled memory beads, and have role templates pull them in during `gt prime`. This solves the same problem without any filesystem coupling.

Closing -- not a reflection on code quality, just an architectural direction call. Appreciate the work!

PR #2191, opened 2026-02-28
fix(test): fix flaky TestVerifyStartupNudgeDelivery_IdleAgent session collision

DreadPirateRobertz:
## Summary
- Add wait loop after `KillSession` to verify session is fully destroyed before creating a new one
- `tmux kill-session` can be async — the session may briefly persist after the command returns
- Without this, `NewSession` hits "session already exists" when the test runs consecutively

## Test plan
- [x] `go test ./internal/polecat/ -run TestVerifyStartupNudgeDelivery -count=3` — passes all 3 runs

Fixes: gt-eo8d

### DIFF
--- a/internal/polecat/session_manager_test.go
+++ b/internal/polecat/session_manager_test.go
@@ -437,8 +437,15 @@ func TestVerifyStartupNudgeDelivery_IdleAgent(t *testing.T) {
 	tm := tmux.NewTmux()
 	sessionName := "gt-test-nudge-verify-" + t.Name()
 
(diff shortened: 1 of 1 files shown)


arttttt (commented):
If the session doesn't die within 1 second, the loop exits silently and `NewSession` fails with the same confusing "session already exists" error. Add a check after the loop:

```go
if exists, _ := tm.HasSession(sessionName); exists {
    t.Fatalf("session %s still exists after kill + 1s wait", sessionName)
}
```

PR #2197, opened 2026-02-28
fix(zfc): replace PID string matching with flock + port checks

DreadPirateRobertz:
## Summary
- Replace `isGasTownDaemon` ps string matching in `IsRunning` with flock-based check on `daemon.lock`
- Replace `isDoltSqlServer` ps string matching with TCP port connectivity check
- Convert handler dog lifecycle constants to vars for future config override

## Fixes
- **gt-utuk**: ZFC: daemon infers process identity via PID + ps string matching
- **gt-yzt0**: ZFC: daemon handler manages dog lifecycle with hardcoded timeouts

## Test plan
- [x] `go vet ./...` — clean
- [x] `go test ./internal/daemon/...` — pass

### DIFF
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -1517,49 +1517,73 @@ func IsShutdownInProgress(townRoot string) bool {
 }
 
 // IsRunning checks if a daemon is running for the given town.
-// It checks the PID file and verifies the process is alive.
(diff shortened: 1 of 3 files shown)


steveyegge:
All changes already on main via prior merges. Main has improved versions (verifyPIDOwnership, config-driven handler timeouts) that supersede this PR's approach.

PR #2068, opened 2026-02-26
feat(telemetry): agent instantiation tracing, mail telemetry, and token usage

pae23:
## Summary

Add end-to-end OpenTelemetry instrumentation for the GASTOWN multi-agent system. All events are OTel log records exported via OTLP (`GT_OTEL_LOGS_URL`); counters and histograms go to `GT_OTEL_METRICS_URL`. Every record carries a `run.id` UUID (`GT_RUN`) so all events from a single agent session can be retrieved and correlated.

---

### Remove `pane read` telemetry

`RecordPaneRead` / `paneReadTotal` removed — raw pane capture carries no correlation value and generates noise. The `RecordPaneRead` call in `tmux.CapturePane` is removed.

---

### `agent.instantiate` — root event per agent spawn

Every time GASTOWN spawns an agent, a root OTel log record is emitted that anchors all subsequent events in that session. It carries the session identity, role, assigned issue, and git state of the working directory at spawn time.

Emitted by: polecat `session_manager`, witness `manager`, refinery `manager`, `session.StartSession`.

**`AgentInstantiateInfo` struct** (replaces the previous positional-arg signature):

| Field | Type | Description |
|---|---|---|
| `RunID` | `string` | `GT_RUN` UUID — session correlation key, present on every downstream event |
| `AgentType` | `string` | Runtime adapter: `"claudecode"`, `"opencode"`, … |
| `Role` | `string` | GT role: `polecat` · `witness` · `mayor` · `refinery` · `crew` · `deacon` · `dog` · `boot` |
| `AgentName` | `string` | Agent name within its role (e.g. `"wyvern-Toast"`) |
| `SessionID` | `string` | tmux session name (e.g. `"gt-wyvern-Toast"`) |
| `RigName` | `string` | Rig name; empty for town-level agents (mayor, deacon) |
| `TownRoot` | `string` | Absolute town root path; derives `instance` = `hostname:basename(TownRoot)` |
| `IssueID` | `string` | Bead ID of the work item assigned to this agent (`gt sling --issue`) |
| `GitBranch` | `string` | Git branch of the working directory at spawn time |
| `GitCommit` | `string` | HEAD SHA of the working directory at spawn time |

---

### `agent.state_change` — hook bead ID

`agent.state_change` now emits `hook_bead: string` (the actual bead ID the agent is currently processing) instead of the opaque `has_hook_bead: bool`. Empty string means no bead is currently hooked.

---

### `gt agent-log` — conversation log streaming (new hidden command)

[...]

PR #1827, opened 2026-02-21
feat(telemetry): add OpenTelemetry observability for gt, bd, and Claude Code

pae23:
## Summary

- **New `internal/telemetry/` package**: OTLP metric + log export to VictoriaMetrics/VictoriaLogs. `telemetry.go` initializes providers, `recorder.go` defines typed recording functions (metrics + structured log events) for every GT operation, `subprocess.go` propagates `BD_OTEL_*` and `OTEL_RESOURCE_ATTRIBUTES` to bd subprocesses automatically.
- **Claude Code telemetry** (`env.go`): when `GT_OTEL_METRICS_URL` is set, `AgentEnv()` injects `CLAUDE_CODE_ENABLE_TELEMETRY=1` + full OTLP config into every agent session, including logs (`OTEL_LOGS_EXPORTER`, `OTEL_LOG_TOOL_CONTENT`, `OTEL_LOG_USER_PROMPTS`). Uses `http/protobuf` (VictoriaMetrics rejects JSON).
- **GT resource attributes on Claude Code logs**: every agent session now injects `gt.town`, `gt.session`, `gt.topic`, `gt.issue`, `gt.prompt` into `OTEL_RESOURCE_ATTRIBUTES`, so every Claude Code log line carries full GT context (filterable by rig, session, ticket, beacon topic).
- **`prime.context` log event**: `gt prime` emits the full rendered formula to VictoriaLogs after each execution, so operators can see exactly what context each agent started with (correlated to the API calls that follow).
- **Unified startup builder**: `BuildStartupCommandFromConfig(AgentEnvConfig)` replaces the per-role command builders. All agent paths (polecat, witness, refinery, deacon, crew) now go through a single function that threads `Issue`, `Topic`, `SessionName`, and `Prompt` into `OTEL_RESOURCE_ATTRIBUTES`.
- **bd subprocess telemetry**: `SetProcessOTELAttrs()` mirrors `GT_OTEL_*` → `BD_OTEL_*` at gt startup; `OTELEnvForSubprocess()` injects them into explicit `cmd.Env` builders. bd calls emit subcommand, args, duration, and status to VictoriaLogs. stdout/stderr are **opt-in** (`GT_LOG_BD_OUTPUT=true`) to avoid storing tokens or PII.
[...]

PR #2199, opened 2026-02-28
feat(otel): P0 work context injection + agent conversation streaming

pae23:
## Summary

- **Work context injection (P0)**: At each `gt prime`, inject `GT_WORK_RIG`, `GT_WORK_BEAD`, `GT_WORK_MOL` into the current process env and the tmux session env so all subsequent subprocesses (`bd`, mail, …) carry the correct work attribution until the next prime overwrites it. All side effects are gated on `telemetry.IsActive()` (no-op when neither `GT_OTEL_METRICS_URL` nor `GT_OTEL_LOGS_URL` is set).
- **Agent conversation streaming**: New `agentlog` package watches Claude Code JSONL conversation logs and emits normalized `agent.event` / `agent.usage` records to VictoriaLogs. New hidden `gt agent-log` command invoked by the session lifecycle.
- **Run correlation**: `WithRunID`/`RunIDFromCtx` inject `run.id` into every telemetry event for waterfall correlation across a full agent session.
- **Resource attribute propagation**: `buildGTResourceAttrs` now includes `gt.work_rig`, `gt.work_bead`, `gt.work_mol`, `gt.session`, `gt.run_id` so they flow to `bd` subprocesses via `OTEL_RESOURCE_ATTRIBUTES`.

## Changes

- `internal/telemetry/telemetry.go`: `IsActive()` — gates all OTel side effects
- `internal/telemetry/recorder.go`: `RecordPrime` emits work context; `WithRunID`/`RunIDFromCtx`; `RecordAgentTokenUsage`/`RecordAgentEvent`; `run.id` injected into all events
- `internal/telemetry/subprocess.go`: work context + session/run attrs in resource attributes
- `internal/cmd/prime.go`: `injectWorkContext` + `setTmuxWorkContext`; pre-fetch bead once
- `internal/agentlog/`: new package — Claude Code and OpenCode JSONL adapters
- `internal/cmd/agent_log.go`: new hidden `gt agent-log` command
- `internal/session/agent_logging_unix.go`: start agent-log subprocess at session start

## Test plan

- [ ] `go test ./internal/cmd/... -run TestInjectWorkContext` — 6 tests covering all work context injection paths (no bead, bead only, bead+molecule, empty rig, OTel disabled, dry-run)
- [ ] `go test ./internal/telemetry/...` — subprocess resource attr tests including work context
- [ ] `go test ./internal/agentlog/...` — JSONL parsing and event normalization
- [ ] Manual: activate OTel (`GT_OTEL_LOGS_URL=...`), run `gt prime` with a bead on hook, verify `GT_WORK_BEAD` is set in env and tmux session; verify `prime` log event in VictoriaLogs has `work_bead` attribute
- [ ] Manual: run `gt prime` with no bead on hook, verify all `GT_WORK_*` vars are cleared

### DIFF
--- a/internal/telemetry/recorder.go
+++ b/internal/telemetry/recorder.go
@@ -791,7 +791,7 @@ func RecordPaneOutput(ctx context.Context, sessionID, content string) {
 }
 
 // RecordAgentEvent emits a structured agent conversation event.
-// Opt-in: only called when GT_LOG_AGENT_OUTPUT=true.
+// Opt-in: requires GT_LOG_AGENT_OUTPUT=true (enforced here so callers can't bypass it).
 //
 // agentType is the adapter name ("claudecode", "opencode", …).
 // eventType is one of "text", "tool_use", "tool_result", "thinking".
@@ -801,6 +801,9 @@ func RecordPaneOutput(ctx context.Context, sessionID, content string) {
 // content is truncated to 512 bytes to limit PII exposure; set
 // GT_LOG_AGENT_CONTENT_LIMIT=0 to disable truncation (experts only).
 func RecordAgentEvent(ctx context.Context, sessionID, agentType, eventType, role, content, nativeSessionID string, ts time.Time) {
+	if os.Getenv("GT_LOG_AGENT_OUTPUT") != "true" {
+		return
+	}
 	initInstruments()
 	inst.agentEventTotal.Add(ctx, 1, metric.WithAttributes(
 		attribute.String("session", sessionID),


DreadPirateRobertz (changes requested):
### Code review

Found 5 issues:

1. **Reverts formula display path from commit `b8f79dc8`**: `outputMoleculeWorkflow` removes the `AttachedFormula` branch that shows inline formula steps (the root-only wisp optimization that reduced wisp creation by 15x). The PR deletes the `cli.Name()` import, the `showFormulaStepsFull()` call, and the `gt done` instruction, causing all beads to fall through to the legacy `showMoleculeExecutionPrompt` path. This silently breaks agents with `AttachedFormula` set.

https://github.com/steveyegge/gastown/blob/8b88de154e3336e3af584d7f77a9a8f80c384ad6/internal/cmd/prime.go#L610-L632

2. **PII: full conversation content logged with no opt-in gate** (previously flagged as CHANGES_REQUESTED on PR #2068): `RecordAgentEvent` emits full `content` with no truncation and no `GT_LOG_CONTENT` gate. `tool_result` events commonly contain file contents, credentials, and API keys from tool outputs. The `agent_log.go` command unconditionally streams every `tool_use`, `tool_result`, `text`, and `thinking` block to VictoriaLogs. PR #2068's review requested a content gating flag defaulting to `false` — this PR ships the same feature without addressing it.

https://github.com/steveyegge/gastown/blob/8b88de154e3336e3af584d7f77a9a8f80c384ad6/internal/telemetry/recorder.go#L564-L590

3. **Watcher permanently exits on 30s timeout**: In the `Watch` goroutine, when `waitForNewestJSONL` times out (no JSONL file in 30 seconds after a session switch), the goroutine returns and closes the channel permanently. All subsequent Claude sessions are silently unmonitored. The comment says "non-fatal" but the effect is permanent — the watcher should retry indefinitely with `ctx` cancellation as the only exit condition.

https://github.com/steveyegge/gastown/blob/8b88de154e3336e3af584d7f77a9a8f80c384ad6/internal/agentlog/claudecode.go#L62-L76

4. **Cache-only token usage events silently dropped** (previously flagged on PR #2068): `RecordAgentTokenUsage` checks `u.InputTokens > 0 || u.OutputTokens > 0` but ignores `CacheReadInputTokens` and `CacheCreationInputTokens`. A turn served entirely from cache produces no usage event, breaking token accounting. The #2068 review explicitly requested checking all four fields.

https://github.com/steveyegge/gastown/blob/8b88de154e3336e3af584d7f77a9a8f80c384ad6/internal/agentlog/claudecode.go#L316-L326

5. **No `DeactivateAgentLogging` in session teardown** (previously flagged on PR #2068): `ActivateAgentLogging` spawns a detached `gt agent-log` process with `Setsid: true`, but no teardown counterpart exists. `StopSession()` does not kill the agent logger. When a session ends normally, the watcher runs indefinitely. The `killPreviousAgentLogger` call inside `ActivateAgentLogging` only handles the restart case, not normal session termination.

https://github.com/steveyegge/gastown/blob/8b88de154e3336e3af584d7f77a9a8f80c384ad6/internal/session/agent_logging_unix.go#L15-L65

Generated with [Claude Code](https://claude.ai/code)

pae23:
## Fix: Windows path handling in `claudeProjectDirFor`

The Windows unit test `TestClaudeProjectDirFor` was failing because `filepath.Abs` on Windows prepends the current drive letter (e.g. `D:\some\work\dir`) and uses backslashes, so the `strings.ReplaceAll(abs, "/", "-")` call produced no replacements.

Fix in `internal/agentlog/claudecode.go`:
1. `filepath.ToSlash(abs)` — normalize separators to `/` (no-op on Unix)
2. Strip any Windows drive letter prefix (`C:`, `D:`, …) before hashing — matching Claude Code's cross-platform project-dir behaviour
3. Then `strings.ReplaceAll(normalized, "/", "-")` as before

The integration test failures (`TestBeadsDbInitAfterClone`) are a CI runner infrastructure issue (beads binary too old) unrelated to this PR.

DreadPirateRobertz:
### Code review

Found 2 issues:

1. **`truncateOutput` with `max=0` silently destroys all content instead of disabling truncation.** The comment at L571 says `GT_LOG_AGENT_CONTENT_LIMIT=0` disables truncation (experts only), but `truncateOutput(s, 0)` evaluates `len(s) <= 0` as false for any non-empty string, then returns `s[:0] + "..."` = `"..."`. Every content field becomes just an ellipsis. The fix is to return `s` early when `max <= 0`.

https://github.com/steveyegge/gastown/blob/578173c0a561e7c3245c649abb3e79dc0b8f11cd/internal/telemetry/recorder.go#L200-L212

2. **`RecordAgentEvent` comment promises opt-in gate (`GT_LOG_AGENT_OUTPUT=true`) but no caller enforces it.** The doc comment at L566 says "Opt-in: only called when GT_LOG_AGENT_OUTPUT=true", but `agent_log.go` L89 calls `RecordAgentEvent` unconditionally with no env var check. Full conversation content (including tool results that may contain file contents, credentials, API keys) is emitted to VictoriaLogs with only a 512-byte truncation limit -- no way to disable content logging entirely. This was flagged as a MUST-FIX blocker on the predecessor PRs #1827 and #2068.

https://github.com/steveyegge/gastown/blob/578173c0a561e7c3245c649abb3e79dc0b8f11cd/internal/cmd/agent_log.go#L88-L90

Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with :thumbsup:. Otherwise, react with :thumbsdown:.</sub>

steveyegge (commented):
Heads up: this PR shares 11 files with PR #2068 (same author) -- the entire agentlog package, agent_logging_unix.go, lifecycle.go, recorder.go, subprocess.go. Merging either one will create conflicts in the other.

Please clarify the relationship between these two PRs: which is canonical, does one supersede the other, or should they be consolidated? We reviewed #2068 in detail and requested changes there -- coordinating the two before either merges will save a lot of rebase pain.

pae23:
## Fixes for the two remaining code review issues

**1. `truncateOutput(s, 0)` returned `"…"` instead of `s`**

The condition `len(s) <= max` is false for any non-empty string when `max=0`, so the function fell through to `s[:0] + "…"` = `"…"`. Fixed with an early return: `if max <= 0 || len(s) <= max { return s }`.

**2. `RecordAgentEvent` opt-in guard not enforced at function level**

The comment said "only called when `GT_LOG_AGENT_OUTPUT=true`" but `agent_log.go` called it unconditionally. Moved the guard into `RecordAgentEvent` itself — returns early if `GT_LOG_AGENT_OUTPUT != "true"` — so the opt-in is enforced regardless of caller.

Commit: c0e0bae8

pae23:
All maintainer review items are now addressed, and this PR has been rebased onto PR #2068 to eliminate the 11-file overlap.

**Rebase summary:**
- This branch (`otel-p0-work-context`) is now based on `feat/otel-agent-log` (PR #2068)
- All shared files (`recorder.go`, `subprocess.go`, `agent_log.go`, `agent_logging_unix.go`, etc.) now have a single canonical version in PR #2068
- This PR only adds the 7 files unique to its scope: work context injection (`cmd/prime.go`, `telemetry/subprocess.go`), agent conversation streaming (`agentlog/claudecode.go`, `telemetry/recorder.go` guard fix, `telemetry/telemetry.go`), and tests

**Must-fix items resolved:**
- ✅ `truncateOutput(s, 0)` now returns `s` unchanged (early return on `max <= 0`)
- ✅ `GT_LOG_AGENT_OUTPUT` guard enforced inside `RecordAgentEvent` — callers cannot bypass it
- ✅ Overlap with PR #2068 eliminated via rebase — no more merge conflicts between the two PRs

**Merge order:** PR #2068 must be merged first, then this PR.

PR #2202, opened 2026-02-28
fix: scan both town and system-default tmux sockets in quota

seanbearden:
## Summary
- Adds `multiSocketClient` to `internal/cmd/quota.go` that aggregates tmux sessions from both the town-scoped socket and the system default socket
- Sessions may live on either socket depending on when they were created; without this fix, `gt quota scan` and `gt quota rotate` silently miss sessions on the other socket
- `IsIdle` checks and `executeKeychainRotation` calls now route to the correct socket per-session via `ClientFor()`

## Why this is still needed after 635916ab

PR #2088 was closed as obsolete after `635916ab` unified new session creation onto `-L default`. However, **existing sessions created before that commit remain on the old per-town socket** (`-L gt`). On a live system:

| Socket | Sessions | Origin |
|--------|----------|--------|
| `-L default` | 7 | Created after 635916ab |
| `-L gt` | 13 | Created before 635916ab, still running |

`gt quota scan` only checks one socket and silently misses the other 13 sessions. This fix scans both.

## Test plan
- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] `go test -race ./internal/cmd/... ./internal/quota/...` passes
- [x] No new lint issues in changed code

### DIFF
--- a/internal/cmd/quota.go
+++ b/internal/cmd/quota.go
@@ -26,6 +26,74 @@ func (quotaLogger) Warn(format string, args ...interface{}) {
 	style.PrintWarning(format, args...)
 }
 
+// multiSocketClient aggregates tmux sessions from multiple sockets.
+// InitRegistry sets the town socket to "default" (-L default), but crew
+// sessions may have been created on the system default socket (no -L flag).
+// These are separate tmux servers with disjoint session lists. This client
+// scans both so quota doesn't silently miss sessions on the other socket.
+type multiSocketClient struct {
+	clients      []*ttmux.Tmux
+	sessionOwner map[string]*ttmux.Tmux // maps session name → client that owns it
+}
+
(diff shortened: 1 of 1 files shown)


seanbearden:
Closing — this is obsolete after 635916ab removed per-town socket isolation. See #2088 discussion for context.

seanbearden:
Reopening — this is NOT obsolete. While 635916ab changed new session creation to use `-L default`, older sessions still live on the old `-L gt` socket. On a live system right now:

- `-L default`: 7 sessions (post-635916ab)
- `-L gt`: 13 sessions (pre-635916ab, still running)

`gt quota scan` only scans the "default" socket and silently misses all 13 sessions on the old socket. The multiSocketClient fix is still needed to scan both.

Issue #1996, opened 2026-02-24
Auto-handoff before compact can produce empty context in crew Claude sessions


Henry-E:
## Bug Description
When compact/handoff is triggered automatically in crew Claude sessions, `gt handoff` can run without producing a useful summary of the current session. The next session then starts with little/no actionable context.

This creates a manual recovery loop where we have to tell the model to read prior chat and write its own handoff summary.

## Steps to Reproduce
1. Start a crew Claude session and perform non-trivial work (decisions, edits, tests).
2. Trigger the compact path where handoff runs automatically before compaction.
3. Check the generated handoff content for that session.
4. Start/resume the next session and inspect injected handoff context.

## Expected Behavior
Auto-handoff should always include a structured summary sufficient for safe continuation, at minimum:
- work completed
- files changed
- validation/tests and outcomes
- blockers/risks
- next steps

## Actual Behavior
Handoff is sometimes empty or too sparse, so context continuity breaks and manual prompting is required.

## Environment
- **OS**: Linux 6.8.0-90-generic #91-Ubuntu SMP PREEMPT_DYNAMIC Tue Nov 18 14:14:30 UTC 2025 x86_64 GNU/Linux
- **Go version**: go1.25.6 linux/amd64
- **Gas Town version**: gt version v0.7.0-622-g6c7d5bb7-dirty (dev: tmp/dog-only@6c7d5bb7)
- **tmux version** (if applicable): tmux 3.4

## Logs / Error Output
```
No hard error is emitted in this failure mode; the issue is missing/insufficient handoff summary content after auto-trigger.
```

## Additional Context
A robust fallback would help: if model summarization fails, generate deterministic handoff content from local state (git status/diff, recent commands, and issue context) rather than sending an empty handoff.

DreadPirateRobertz:
This appears to be fixed. The `collectHandoffState()` and `collectGitState()` functions in `internal/cmd/handoff.go` now provide deterministic fallback content including:

- Current branch
- Modified/untracked files
- Stash count, unpushed commits
- Recent commits (last 5)
- Hooked work, inbox summary, ready beads, in-progress beads

The relevant commits reference this issue directly:
- `a7254c0c8 fix(handoff): deterministic fallback for empty auto-handoff context (GH#1996)`
- `6acfb085c fix: add deterministic git state to handoff context (GH#1996)`

[...]

PR #2204, opened 2026-02-28
fix(handoff): deterministic fallback for empty auto-handoff context

DreadPirateRobertz:
## Summary
- Adds molecule status (`gt mol status`) as first collected section in auto-handoff state — gives successor the primary "what was I working on" context
- Adds diff stat of recent commits to git state — shows which files were touched even when workspace is clean (committed+pushed)
- Adds minimum-viable-handoff fallback: if collected state is empty or under 100 chars, appends structured recovery instructions so successor always has actionable steps
- Adds `Git.RecentDiffStat()` with graceful fallback for repos with fewer commits than requested

Fixes: #1996

## Test plan
- [x] Existing `TestCollectGitState` passes (unchanged behavior for dirty workspaces)
- [x] New `TestCollectGitState/includes_diff_stat_for_committed_changes` — verifies diff stat appears for clean repos with commits
- [x] New `TestMinimumViableHandoff` — verifies fallback always produces recovery instructions
- [x] New `TestCollectHandoffState_FallbackOnEmpty` — verifies fallback triggers when all external commands fail (broken PATH)
- [x] `go vet` clean
- [x] Pre-existing `TestWarnHandoffGitStatus` failure confirmed on main (not introduced by this PR)

### DIFF
--- a/internal/cmd/handoff.go
+++ b/internal/cmd/handoff.go
@@ -1351,11 +1351,25 @@ func hookBeadForHandoff(beadID string) error {
 }
 
 // collectHandoffState gathers current state for handoff context.
-// Collects: git workspace state (deterministic), inbox summary, ready beads, hooked work.
+// Collects: molecule context, git workspace state (deterministic), inbox summary,
+// ready beads, hooked work, and in-progress beads.
+//
 // Git state is always collected first via Go library calls (no shelling out) to ensure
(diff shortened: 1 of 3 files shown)


steveyegge:
Merged to main via refinery fix-merge. Conflicts resolved by rebasing onto latest main.

PR #2206, opened 2026-02-28
feat(consensus): fan-out prompt dispatch to multiple sessions

seanbearden:
## Summary
- New `gt consensus` command (alias: `gt fanout`) sends the same prompt to multiple Claude Code sessions in parallel and collects responses
- New `internal/consensus/` package with `Runner` orchestrator, pre/post pane diffing, and `TmuxClient` interface
- Defaults to all idle crew+polecat sessions; supports `--timeout`, `--json`, `--dry-run`, `--session` flags
- Added `consensus` to beads-exempt commands in `root.go`

## Test plan
- [x] `go build ./internal/... ./cmd/...` — clean build
- [x] `go test ./internal/consensus/ -v` — 7 unit tests pass
- [x] `go vet ./internal/consensus/ ./internal/cmd/` — clean
- [ ] `gt consensus --dry-run "test"` — shows target sessions
- [ ] `gt consensus "What time is it?"` — live fan-out (needs 2+ idle sessions)

### DIFF
--- a/internal/cmd/consensus.go
+++ b/internal/cmd/consensus.go
@@ -1,0 +1,180 @@
+package cmd
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+	"strings"
+	"time"
+
+	"github.com/spf13/cobra"
+	"github.com/steveyegge/gastown/internal/consensus"
+	"github.com/steveyegge/gastown/internal/style"
+	"github.com/steveyegge/gastown/internal/tmux"
+)
+
+var (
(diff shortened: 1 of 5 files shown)


seanbearden:
Closing — this implements same-prompt fan-out across Claude Code tmux sessions, but the actual need (from the comment that motivated it) is cross-model/cross-account consensus (Claude vs GLM). That's a different feature that would need to work at the LLM API level, not the tmux session level.

PR #2211, opened 2026-03-01
fix: mail send prefix mismatch — generate database-prefixed IDs

DreadPirateRobertz:
## Summary
- `gt mail send` was broken — all sends failed with "prefix mismatch: database uses 'hq-' but ID 'msg-...' doesn't match"
- Root cause: `GenerateID()` always produced `msg-*` IDs but HQ database requires `hq-` prefix
- Added `GeneratePrefixedID()` that reads the database prefix from metadata.json
- Fixed both `sendToSingle()` in mail router and the `handoff` direct-create path

## Test plan
- [x] Built and installed locally
- [x] `gt mail send overseer/ -s "Test" -m "Test"` succeeds (was failing before)
- [x] Messages appear in inbox with correct `hq-wisp-*` IDs
- [ ] Run `go test ./internal/mail/...`
- [ ] Run `go test ./internal/beads/...`

### DIFF
--- a/internal/beads/config_yaml.go
+++ b/internal/beads/config_yaml.go
@@ -28,6 +28,13 @@ func EnsureConfigYAMLFromMetadataIfMissing(beadsDir, fallbackPrefix string) erro
 	return ensureConfigYAML(beadsDir, prefix, syncMode, true)
 }
 
+// ConfigPrefixFromDir reads the database prefix from a beads directory's metadata.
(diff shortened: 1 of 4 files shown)


DreadPirateRobertz:
Closing — the prefix mismatch issue has been resolved differently on main. The mail router now lets bd auto-generate IDs with the correct prefix instead of passing explicit --id flags. The GenerateID() function is only used for in-memory tracking, not persisted to beads.

Issue #2220, opened 2026-03-01
Add branch contamination preflight before sling/merge flows


davidahmann:
## Problem
Operators lose time and trust when `gt` sling/merge flows run on contaminated branches, because unrelated stacked commits leak into PRs and handoff automation.

## Why now
This has become a repeated multi-agent handoff failure mode. We need deterministic preflight checks before side effects.

## Current behavior is insufficient
Current sling/merge flows do not hard-fail early when `origin/main...HEAD` includes out-of-scope files, so contamination is discovered late in review.

## Expected behavior
Add a preflight contamination guard before sling/merge flows that exits non-zero with machine-readable diagnostics when branch scope includes unrelated files.

## Acceptance / Validation
- Guard checks `git diff --name-only origin/main...HEAD` before sling/merge execution.
- Command exits non-zero when contamination is detected.
- Targeted tests cover clean branch (pass) and contaminated branch (fail).

## Evidence packet
- Commit under test: `db17f270` (`origin/main`)
- Runtime environment: macOS arm64, Go `1.25.7`
- Minimal repro:
  1. Create feature branch from main.
  2. Add unrelated file change outside intended sling scope.
  3. Run sling/merge flow.
- Expected: flow blocks immediately with contamination classification.
- Actual: flow continues until later review/CI, causing noisy PR scope.

## Likely codepaths
`cmd/gt`, `internal/git`, sling/merge workflow scripts.

DreadPirateRobertz:
The concept is sound but the acceptance criteria need refinement. Key question: **how is "scope" defined?**

`git diff --name-only origin/main...HEAD` shows all changed files on the branch, but to detect *contamination*, the guard needs to know which files are *expected*. Options:

1. **Bead-based scope**: If the sling context has metadata about which paths the bead should touch, the guard compares actual diff against expected paths. This requires beads to declare scope upfront (not currently supported).

2. **Heuristic contamination**: Flag branches that touch files in multiple unrelated Go packages or top-level directories. This would catch obvious cases (e.g., a bug fix in `internal/cmd/` that also has changes in `internal/web/`) but has false positives for legitimate cross-cutting changes.

3. **Merge-base freshness**: Check if the branch is based on a stale merge base (many commits behind `origin/main`). Stale branches are the most common source of contamination from rebases that pull in unrelated work.

[...]

PR #2221, opened 2026-03-01
Add branch contamination preflight guard for sling/merge flows

davidahmann:
## Problem
Sling/merge workflows can run from a branch that includes out-of-scope changes, which creates noisy handoffs and contaminated PR scope.

## Why now
This has become a repeated multi-agent operator friction point. We need an early, deterministic branch-scope gate before side effects.

## What changed
- Added `Git.FilesChangedSince(baseRef, headRef)` using `git diff --name-only <base>...<head>`.
- Added branch-scope preflight helpers in `internal/cmd` that:
  - read scope prefixes from `GT_BRANCH_SCOPE_PATHS`
  - compute out-of-scope files against `origin/<default>...HEAD`
  - fail with machine-readable JSON diagnostics when contamination is detected.
- Wired preflight checks into `gt sling`, `gt done` (COMPLETED path), and `gt mq submit`.
- Added targeted tests for:
  - git diff file detection
  - clean in-scope preflight
  - contaminated out-of-scope preflight.

## Validation
- `go test ./internal/git -run "FilesChangedSince"`
- `CGO_ENABLED=0 go test ./internal/cmd -run "BranchScopePreflight"`

Refs #2220

### DIFF
--- a/internal/cmd/branch_scope_preflight.go
+++ b/internal/cmd/branch_scope_preflight.go
@@ -1,0 +1,121 @@
+package cmd
+
+import (
+	"encoding/json"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/steveyegge/gastown/internal/git"
+)
+
+const branchScopeEnvVar = "GT_BRANCH_SCOPE_PATHS"
+
+// BranchScopeDiagnostics provides machine-readable contamination diagnostics.
+type BranchScopeDiagnostics struct {
+	Classification  string   `json:"classification"`
+	BaseRef         string   `json:"base_ref"`
+	HeadRef         string   `json:"head_ref"`
+	AllowedPrefixes []string `json:"allowed_prefixes"`
+	ChangedFiles    []string `json:"changed_files"`
+	OutOfScopeFiles []string `json:"out_of_scope_files"`
+}
+
+func parseScopePrefixes(raw string) []string {
+	if strings.TrimSpace(raw) == "" {
+		return nil
+	}
+	normalized := make([]string, 0)
+	seen := make(map[string]bool)
+	for _, part := range strings.FieldsFunc(raw, func(r rune) bool {
+		return r == ',' || r == '\n' || r == '\r' || r == '\t'
+	}) {
+		prefix := normalizeScopePath(part)
+		if prefix == "" || seen[prefix] {
+			continue
+		}
+		seen[prefix] = true
+		normalized = append(normalized, prefix)
+	}
+	return normalized
+}
+
+func normalizeScopePath(path string) string {
+	normalized := strings.TrimSpace(strings.ReplaceAll(path, "\\", "/"))
+	normalized = strings.TrimPrefix(normalized, "./")
+	normalized = strings.TrimPrefix(normalized, "/")
+	normalized = strings.TrimSuffix(normalized, "/")
+	return normalized
+}
+
+func outOfScopeFiles(changedFiles, allowedPrefixes []string) []string {
+	if len(changedFiles) == 0 || len(allowedPrefixes) == 0 {
+		return nil
+	}
+	out := make([]string, 0)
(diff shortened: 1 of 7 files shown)


davidahmann:
Implemented the scoped fix for #2220 by adding a branch-scope contamination preflight and wiring it into the submit paths (, , and ) with focused unit coverage.

Validation run:
-  (pass)
-  (pass)

Current CI snapshot on this PR: one failing check () and other jobs still running/pending; I am triaging based on first concrete error lines.

This contribution was informed by patterns from Wrkr. Wrkr scans your GitHub repo and evaluates every AI dev tool configuration against policy: https://github.com/Clyra-AI/wrkr

DreadPirateRobertz (approved):
LGTM -- Well-structured branch contamination guard.

Analysis:

Design is sound:
- Env-var gating (GT_BRANCH_SCOPE_PATHS) makes this zero-cost when unset -- no behavior change for anyone not opting in
- Three-dot diff (merge-base) in FilesChangedSince is correct for comparing branch divergence
- JSON diagnostics in the error output enable machine parsing by agents/tooling
- Path normalization handles edge cases (backslashes, leading ./ or /, trailing /)
- All three merge flow entry points covered: sling, done (COMPLETED), mq submit

Notes:
- sling.go uses runBranchScopePreflightFromCWD (creates its own git.Git) while done.go and mq_submit.go reuse existing g -- this is fine since sling doesn't have a git.Git at that point in its flow
- The whitespace changes in git.go (struct alignment) are just gofmt fixes -- no functional change
- Tests properly use real git repos with remotes, covering both clean and contaminated paths

One suggestion (non-blocking): consider adding a test for parseScopePrefixes with edge cases like empty strings, duplicate prefixes, and mixed separators. The dedup logic (seen map) is there but untested.

Ship it.

Co-Reviewed-By: Claude Opus 4.6 <noreply@anthropic.com>

DreadPirateRobertz:
## Review fixes applied (rebased onto latest main)

Addressed review feedback in commit `61b110cb` on branch `polecat/max/hq-e2nm@mm8ln8cs` (DreadPirateRobertz fork):

**MUST-FIX #1** (zero files warning): Added `style.PrintWarning` when `changedFiles == 0` but scope paths are configured — warns about possible wrong base ref or empty branch.

**MUST-FIX #2** (sling getwd): Added else branch with `style.PrintWarning` when `os.Getwd()` fails, instead of silently skipping the preflight.

**MUST-FIX #3** (done.go g init): Verified already correct — `done.go` reconstructs the polecat worktree path (lines 148-195) before creating the `git.Git` instance.

**SHOULD-FIX #4** (resolveDefaultBaseRef): Changed to return `(string, error)`, propagates failure explicitly instead of silently falling back to non-existent `origin/main`.

Branch rebased onto latest `main` to resolve the `issues.jsonl` CI failure.

Note: Could not push directly to this PR's head branch (owned by `davidahmann`). Fixes available on fork branch for cherry-pick or new PR.

steveyegge:
Closing this PR — the approach violates ZFC (Zero Framework Cognition), a core Gas Town design principle.

**ZFC says:** "Agent decides. Go transports." Go code should never make judgment calls or use hardcoded thresholds. Claude makes judgment calls.

The 121-line prefix-matching framework in `branch_scope_preflight.go` puts contamination detection intelligence into Go. Whether a branch is "contaminated" is a semantic judgment call — an agent reading the diff understands intent, not just path prefixes.

**The right approach:** A refinery plugin that hands the diff to Haiku before merge and asks "does this branch have scope leakage?" Haiku looks at actual content semantically. Go just transports the diff and the verdict. Cheap, accurate, ZFC-compliant.

The problem you identified (GH#2220) is real and worth solving — just in the agent layer, not the Go layer. See also #2237 (closed for the same reason). Thanks for the contribution!

PR #2223, opened 2026-03-01
fix(witness): verify work on main before resetting abandoned beads

### LINKED ISSUE #2036: bug: witness patrol formula lacks spawn-awareness, nukes freshly-slung polecats

gregorycarter:
## Summary

- Adds a `verifyCommitOnMain` guard at the top of `resetAbandonedBead` in `internal/witness/handlers.go`
- When a zombie polecat's HEAD commit is already on the default branch, the bead is **closed** instead of reset to "open" for re-dispatch
- Prevents the spawn-storm / duplicate-work loop where completed beads are endlessly re-dispatched

## Problem

When `DetectZombiePolecats` nukes a polecat during the `gt done` cleanup window (after work is pushed but before `done-intent` clears), `resetAbandonedBead` resets the bead to "open". A new polecat picks it up, finds nothing to do, exits, and the witness nukes it again — creating a spawn storm.

This was introduced by gt-tr3d (commit 0e5d1a30, Feb 18) which added aggressive nuke paths for live sessions with 30-60s thresholds that overlap with the `gt done` cleanup window.

## Fix

`verifyCommitOnMain` already exists in the codebase (used by `HandleMerged`) but was never called in any of the zombie cleanup paths. This PR adds it as a guard in `resetAbandonedBead`:

```go
if onMain, err := verifyCommitOnMain(workDir, rigName, polecatName); err == nil && onMain {
    _ = bdRun(workDir, "close", hookBead, "-r", reason)
    return true
}
```

The check is fail-safe: if `verifyCommitOnMain` returns an error (e.g., worktree already removed), it falls through to the existing reset behavior.

`verifyCommitOnMain` is also converted from a plain function to a package-level `var` (matching the existing `bdRun`/`bdExec` pattern) so tests can override it.

## Test plan

- [x] `TestResetAbandonedBead_ClosesWhenWorkOnMain` — verifies bead is closed (not reset) when work is on main
- [x] `TestResetAbandonedBead_ResetsWhenWorkNotOnMain` — verifies existing reset behavior is preserved
- [x] All existing witness tests pass (`go test ./internal/witness/...`)
- [x] Full project builds clean (`go build ./...`)

Fixes #2036

### DIFF
--- a/internal/witness/handlers.go
+++ b/internal/witness/handlers.go
@@ -1630,14 +1630,15 @@ func getBeadStatus(workDir, beadID string) string {
 // 3. Clears assignee
 // 4. Sends mail to deacon for re-dispatch (includes respawn count; SPAWN_STORM
 //    prefix and Urgent priority when count exceeds defaultMaxBeadRespawns)
-// Returns true if the bead was recovered.
-func resetAbandonedBead(workDir, rigName, hookBead, polecatName string, router *mail.Router) bool {
+// Returns (recovered, closed): recovered=true if bead was reset for re-dispatch,
+// closed=true if bead was closed because the work is already on main.
+func resetAbandonedBead(workDir, rigName, hookBead, polecatName string, router *mail.Router) (recovered, closed bool) {
 	if hookBead == "" {
-		return false
+		return false, false
 	}
 	status := getBeadStatus(workDir, hookBead)
 	if status != "hooked" && status != "in_progress" {
-		return false
+		return false, false
 	}
 
 	// Guard: if the polecat's commit is already on the default branch,
@@ -1646,7 +1647,7 @@ func resetAbandonedBead(workDir, rigName, hookBead, polecatName string, router *
 	if onMain, err := verifyCommitOnMain(workDir, rigName, polecatName); err == nil && onMain {
 		reason := fmt.Sprintf("Work already on main (verified by witness, polecat %s)", polecatName)
 		_ = bdRun(workDir, "close", hookBead, "-r", reason)
-		return true
+		return false, true
 	}
 
 	// Track respawn count for audit and storm detection.
@@ -1654,7 +1655,7 @@ func resetAbandonedBead(workDir, rigName, hookBead, polecatName string, router *
 
 	// Reset bead status to open and clear assignee
 	if err := bdRun(workDir, "update", hookBead, "--status=open", "--assignee="); err != nil {
-		return false
+		return false, false
 	}
 
 	// Send mail to deacon for re-dispatch
@@ -1689,7 +1690,7 @@ Please re-dispatch to an available polecat.`,
 		_ = router.Send(msg) // Best-effort
 	}
 
-	return true
+	return true, false
 }
 
 // OrphanedBeadResult contains a single detected orphaned bead.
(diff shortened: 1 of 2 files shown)


steveyegge:
Landed directly on main with rebase conflict resolution + nit fix (log bd close errors instead of swallowing). See commit on main.

PR #2231, opened 2026-03-02
feat: Copilot CLI agent support

Random-Word:
## Summary

Adds full Copilot CLI (`copilot`) support as a Gas Town agent runtime, enabling all Gas Town agent roles (mayor, deacon, witness, refinery, polecats) to run under GitHub Copilot CLI alongside the existing Claude Code support.

### Core Copilot CLI integration
- **Native hooks integration** — `sessionStart`, `userPromptSubmitted`, and `pretool` guard hooks via Copilot CLI's `.github/hooks/` mechanism
- **Plugin architecture** — `plugins/copilot-cli/` with hooks.json and plugin.json for lifecycle management
- **PromptFlag support** — `-i` flag for interactive prompt delivery (vs Claude's positional arg)
- **`--continue` flag parity** — session resume support matching Claude's `--continue` behavior
- **trusted_folders automation** — auto-configures Copilot CLI to trust rig working directories
- **Inline prime context** — captures `gt prime --dry-run` output and injects it into the startup prompt for agents whose `sessionStart` hook stdout is not injected into LLM context (workaround for [github/copilot-cli#1139](https://github.com/github/copilot-cli/issues/1139))

### Bug fixes discovered during integration
- **Guard script path templating** — use actual hooksDir instead of hardcoded path
- **Guard script matching** — match guarded commands anywhere in the string
- **Git status handling** — handle rename/copy/unmerged states in `Status()` and `CleanExcludingToolFiles()`
- **gt done cleanup** — no longer blocked by tool-generated untracked files (`.copilot/`)
- **Large prompt delivery** — write prompts >8KB to temp file with self-deleting wrapper script to avoid tmux `respawn-pane` command-length limits

## Architecture

The Copilot CLI agent preset (`AgentPresetInfo`) adds:
- `NoHookStdoutInjection: true` — signals that sessionStart stdout is dropped
- `PromptFlag: "-i"` — interactive mode flag
- `NeedsInlinePrime()` — helper to check if prime context needs inlining

Large prompts use a `#!/bin/sh` wrapper script pattern for shell portability (works with fish, tcsh, zsh, bash as tmux default-shell).

## Testing

- All existing tests pass
- New tests for: guard script matching, git status edge cases, prompt file fallback, trusted_folders
- Manually verified: all agent roles start and operate correctly under Copilot CLI


Random-Word:
Closing for now — will re-open after internal review.

PR #2243, opened 2026-03-02
fix: daemon PID=0 and dolt multi-town PID reuse detection

pae23:
## Summary

Three fixes for multi-town dolt reliability:

**1. daemon.go: PID=0 from nonce-format PID files**

`IsRunning` was calling `os.ReadFile` + `Atoi` directly on PID files that use the nonce format (`"PID\nNONCE"`). `Atoi` fails on multiline content, so daemon PID always read as 0. Now uses `readPIDFile()` which handles both formats.

**2. doltserver.go: PID reuse detection across towns**

Adds `getDoltDataDirFromProcess(pid)` which reads the running process's `--data-dir` flag via `ps -p PID -o args=` (macOS + Linux; returns `""` on Windows). Used as a tiebreaker in `IsRunning`: a recycled PID pointing to a different town's data directory no longer falsely reports as "our" dolt server.

**3. doltserver.go: detach dolt from parent process group (SIGHUP fix)**

`gt mayor at` starts dolt then calls `syscall.Exec` to replace itself with `tmux` (via `attachToTmuxSession`). Without `Setpgid=true`, dolt inherits `gt`'s process group. If `tmux attach-session` fails (e.g. session not yet created) or the terminal closes, SIGHUP is sent to the process group — killing dolt immediately after it started.

Fix: `cmd.SysProcAttr = &syscall.SysProcAttr{Setpgid: true}` puts dolt in its own process group, independent of the caller.

**4. doltserver.go: DefaultConfig reads GT_DOLT_PORT from daemon.json**

All `gt dolt *` commands (`stop`, `status`, `sql`, `logs`, etc.) call `DefaultConfig()` to find the port. `DefaultConfig` only read `GT_DOLT_PORT` from the shell environment — but when set via `gt config set dolt.port`, it lives in `mayor/daemon.json`. Commands run in a normal terminal (outside the daemon context) would connect to port 3307 even when the town uses a custom port.

Fix: if `GT_DOLT_PORT` is absent from the shell env, `DefaultConfig` now parses `mayor/daemon.json` directly (avoids circular import `daemon→doltserver`) and applies the value. Shell env wins over `daemon.json` if both are set.

Affects: `gt dolt status/stop/sql/logs`, `gt vitals`, `gt down`, `gt mayor at` port-conflict messages, and any other `DefaultConfig` caller.

## Test plan

- [ ] `go test ./internal/doltserver/...` — existing tests pass
- [ ] `go test ./internal/daemon/...` — daemon IsRunning works with nonce PID files
- [ ] `gt config set dolt.port 3308` then `gt dolt status` shows port 3308 (not 3307)
- [ ] `gt mayor at` with another town's dolt on 3307 — dolt starts on 3308 and survives after tmux attach
- [ ] Start two Gas Town instances on the same machine, confirm each `gt status` shows the correct port

### DIFF
--- a/internal/doltserver/doltserver.go
+++ b/internal/doltserver/doltserver.go
@@ -219,6 +219,29 @@ func DefaultConfig(townRoot string) *Config {
 		config.LogLevel = ll
 	}
 
+	// Fallback: if GT_DOLT_PORT is not in the shell env, read it from
+	// mayor/daemon.json. Commands like gt dolt status, gt dolt stop, etc.
+	// are typically run without the daemon.json env vars exported to the
+	// shell, so DefaultConfig would otherwise return the wrong port (3307)
+	// when the town uses a custom port (e.g. GT_DOLT_PORT=3308).
+	// We cannot import the daemon package here (circular: daemon→doltserver),
+	// so we parse the minimal JSON structure directly.
+	if os.Getenv("GT_DOLT_PORT") == "" && townRoot != "" {
+		daemonJSONPath := filepath.Join(townRoot, "mayor", "daemon.json")
+		if data, err := os.ReadFile(daemonJSONPath); err == nil {
+			var daemonEnv struct {
+				Env map[string]string `json:"env"`
+			}
+			if err := json.Unmarshal(data, &daemonEnv); err == nil {
+				if v, ok := daemonEnv.Env["GT_DOLT_PORT"]; ok {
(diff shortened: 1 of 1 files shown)


steveyegge:
Fix-merged to main as cd45132e. All four fixes reviewed and tested — doltserver and daemon test suites pass. Thanks @pae23!

PR #2244, opened 2026-03-02
feat: per-town tmux socket isolation (opt-in via GT_TMUX_SOCKET=auto)

pae23:
## Summary

- Adds `GT_TMUX_SOCKET` env var to control tmux socket selection:
  - unset / `"default"` → shared `default` socket (backward compatible with current behavior)
  - `"auto"` → socket derived from town directory name (e.g. `~/gt/mytown` → `-L mytown`)
  - explicit name → use verbatim
- Adds `BuildCommandContext(ctx, args...)` to the tmux package for context-aware commands
- Updates `witness attach` to use the socket-aware `attachToTmuxSession()` helper
- Updates `web/api.go` list-sessions call to use `tmux.BuildCommandContext()`
- Updates `TestInitRegistry_SocketFromTownName` to cover all three `GT_TMUX_SOCKET` modes

## Motivation

Running multiple Gas Town towns on the same machine **requires** per-town tmux isolation because session names (`mayor`, `deacon`, witness sessions) collide on a shared socket. With `GT_TMUX_SOCKET=auto`, each town gets its own tmux server; users attach with `tmux -L <town-name> attach`.

The default remains `"default"` for backward compatibility. Users who want multi-town isolation set `GT_TMUX_SOCKET=auto` in their shell config.

Note: per-town sockets were previously implemented (ab5f06f1) and reverted (635916ab) because they broke `prefix+s` visibility. The env var approach preserves the old default while making isolation available for those who need it.

## Test plan

- [ ] `go test ./internal/session/...` — socket tests cover all three modes
- [ ] Default behavior unchanged: `gt mayor at` with no `GT_TMUX_SOCKET` set uses `tmux -L default`
- [ ] `GT_TMUX_SOCKET=auto gt mayor at` uses town-named socket
- [ ] `GT_TMUX_SOCKET=myserver gt mayor at` uses `-L myserver`

### DIFF
--- a/internal/session/registry.go
+++ b/internal/session/registry.go
@@ -119,11 +119,20 @@ func SetDefaultRegistry(r *PrefixRegistry) {
 func InitRegistry(townRoot string) error {
 	var errs []error
 
-	// Use a per-town tmux socket derived from the town name (e.g. "gt-test").
-	// This gives each Gas Town instance its own tmux server, so session names
-	// like "mayor" and "deacon" no longer conflict across towns running on the
-	// same machine. Users attach with: tmux -L <town-name> attach
-	tmux.SetDefaultSocket(sanitizeTownName(filepath.Base(townRoot)))
+	// Determine the tmux socket name from GT_TMUX_SOCKET env var:
+	//   unset / "default" → shared "default" socket (backward-compatible)
+	//   "auto"            → per-town socket derived from town dir name
+	//                       (required when running multiple towns on one machine,
+	//                       since session names like "mayor" would collide)
+	//   any other value   → use that name as-is
+	socket := os.Getenv("GT_TMUX_SOCKET")
+	switch socket {
+	case "", "default":
+		socket = "default"
+	case "auto":
+		socket = sanitizeTownName(filepath.Base(townRoot))
+	}
+	tmux.SetDefaultSocket(socket)
 
 	r, err := BuildPrefixRegistryFromTown(townRoot)
 	if err != nil {
--- a/internal/session/registry_socket_test.go
+++ b/internal/session/registry_socket_test.go
@@ -8,82 +8,88 @@ import (
 	"github.com/steveyegge/gastown/internal/tmux"
 )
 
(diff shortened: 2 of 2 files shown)


pae23:
## Why this PR

Gas Town session names are deterministic and role-based: every town has a `mayor` session, a `deacon` session, a `witness-<rig>` session, etc. When two towns coexist on the same machine under the same OS user, they share the default tmux server — and their sessions collide by name.

**Example collision on a shared socket:**
```
$ tmux ls
mayor           ← which town's mayor?
deacon          ← which town's deacon?
gt-witness-plop ← fine if prefix differs, but mayor/deacon don't have one
```

The user has no way to tell which `mayor` belongs to which town. Worse, `gt mayor at` will attach to whichever session happens to be first — silently connecting to the wrong town's agent.

## What `GT_TMUX_SOCKET=auto` does

With `auto`, each town gets its own tmux server named after the town directory:

```
# Town at ~/gt/town-a  →  tmux -L town-a
# Town at ~/gt/town-b  →  tmux -L town-b

$ tmux -L town-a ls
mayor
deacon
gt-witness-plop

$ tmux -L town-b ls
mayor          ← completely separate, no collision
deacon
```

`gt mayor at`, `gt witness attach`, and all other attach commands automatically use the right `-L <town>` socket because it's set at `InitRegistry` time from the town root.

## Why the default is still `"default"` (shared socket)

This was previously implemented (ab5f06f1) and reverted by @steveyegge (635916ab) because per-town sockets break `prefix+s` session switching — the tmux prefix menu only shows sessions from the current server.

For backward compatibility and to respect that decision, the default remains the shared `"default"` socket. **Multi-town users who need isolation must opt in** by setting `GT_TMUX_SOCKET=auto` in their shell config (`~/.zshrc` or `~/.bashrc`):

```sh
export GT_TMUX_SOCKET=auto
```

This is admittedly a footgun for multi-town setups: the default causes silent session collisions that are hard to diagnose. But changing the default would break existing single-town users who rely on `prefix+s` visibility across all their sessions.

PR #2246, opened 2026-03-02
fix: gt mayor at aborts on dolt startup failure

pae23:
## Summary

`ensureMayorInfra` now returns an error, and `runMayorAttach` propagates it instead of silently continuing. A dolt startup failure (e.g. port already in use by another town) now aborts the mayor with a clear error message before creating any tmux sessions.

## Before

```
$ gt mayor at
⚠ Dolt already running on port 3306 (different town)
Starting mayor...   ← silently continued with no database
```

## After

```
$ gt mayor at
error: Dolt server start failed: port 3306 already in use
```

## Motivation

The Mayor requires database access. Starting it without Dolt leads to confusing downstream failures (queries failing, state not persisting) that are hard to trace back to the root cause. Failing fast at startup is the right behavior.

## Test plan

- [ ] `go build ./...` — compiles cleanly  
- [ ] Start another town's dolt on port 3306, then `gt mayor at` in a second town — should error immediately
- [ ] Normal case (dolt starts OK) — mayor attaches as before

### DIFF
--- a/internal/cmd/config.go
+++ b/internal/cmd/config.go
@@ -607,6 +607,9 @@ Supported keys:
                               completion (true/false, default: false)
   cli_theme                   CLI color scheme ("dark", "light", "auto")
   default_agent               Default agent preset name
+  dolt.port                   Dolt SQL server port (default: 3307). Set this when
+                              another Gas Town instance is using the same port.
+                              Writes GT_DOLT_PORT to mayor/daemon.json env section.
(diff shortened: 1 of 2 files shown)


steveyegge:
Fix-merged to main as 8b24ecb8. Thanks @pae23!

All three commits squashed into one with full Co-Authored-By attribution. Build verified clean. Good catch on the silent Dolt failure — failing fast is the right call here, and the enriched error with free-port suggestion is a nice UX touch.

PR #2256, opened 2026-03-02
fix: bypass prefix validation in mail send --id

yoyo999888:
gt mail send fails after bd 0.57.0 added strict issue_prefix validation. GenerateID() produces msg- IDs but town beads uses hq- prefix. Add --force to bypass.


steveyegge:
Closing in favor of #2255 which takes the cleaner approach of removing the --id flag entirely. This PR also includes unrelated backup data changes that pollute the diff.

PR #2272, opened 2026-03-02
fix(crew): clone default_branch instead of remote HEAD

l0g1x:
## Problem

`gt crew add` used `git.Clone()` which checks out the remote's HEAD branch (typically `main`). When a rig has a `default_branch` configured that differs from the remote HEAD — for example `default_branch: gastown` while the GitHub default is `main` — crew workspaces silently landed on the wrong branch.

The `branchName := m.rig.DefaultBranch()` value was computed but never used when `createBranch` was false.

## Fix

Use `CloneBranch(defaultBranch)` so the initial clone targets the rig's configured `default_branch` directly. Falls back gracefully to `Clone()` if the branch doesn't exist yet (empty/new repos), preserving the existing `LocalRepo` reference fallback pattern.

## Test plan

- [ ] Existing crew tests pass (`go test ./internal/crew/...`)
- [ ] `gt crew add <name>` on a rig with `default_branch` ≠ remote HEAD lands on the correct branch
- [ ] `gt crew add <name> --branch` still creates `crew/<name>` off the default branch as before
- [ ] Empty/new repos fall back to `Clone` without error

### DIFF
--- a/internal/crew/manager.go
+++ b/internal/crew/manager.go
@@ -203,17 +203,31 @@ func (m *Manager) addLocked(name string, createBranch bool) (*CrewWorker, error)
 		return nil, fmt.Errorf("creating crew dir: %w", err)
 	}
 
-	// Clone the rig repo
+	// Clone the rig repo on the configured default branch.
+	// Using CloneBranch (not Clone) ensures the crew lands on the rig's
+	// default_branch even when it differs from the remote's HEAD branch.
+	// Falls back to Clone if the branch doesn't exist yet (e.g. new/empty repo).
+	defaultBranch := m.rig.DefaultBranch()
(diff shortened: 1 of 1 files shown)


steveyegge:
Fix-merged to main as 8750a87c with improvements:

- **Dropped `.beads/backup/` data files** — only the `manager.go` fix was merged (21 insertions, 8 deletions)
- **Reordered fallback cascade** to prioritize branch-targeting before reference-only:
  `CloneBranchWithReference → CloneBranch → CloneWithReference → Clone`
  (original PR had `CloneBranchWithReference → CloneWithReference → CloneBranch → Clone`)

Thanks for the fix @l0g1x — the bug was real and well-diagnosed.

PR #1747, opened 2026-02-19
feat: config-driven capacity scheduler with deferred polecat dispatch

julianknutsen:
## Summary

Adds a capacity-controlled scheduler for polecat dispatch, driven entirely by config (`scheduler.max_polecats`). When enabled, `gt sling` defers work by labeling beads instead of spawning polecats immediately. A daemon heartbeat dispatches queued beads within the configured concurrency limit.

**Key design principle:** Zero CLI changes for direct dispatch. The same `gt sling <bead> <rig>` command works in both modes — config decides the behavior.

### Dispatch Modes

| `max_polecats` | Mode | Behavior |
|---|---|---|
| `-1` (default) | Direct | `gt sling` spawns polecats immediately — identical to today, near-zero overhead |
| `0` | Direct | Same as `-1` |
| `N > 0` | Deferred | `gt sling` labels + writes metadata; daemon dispatches within capacity |

### What's New

- **`gt sling <any>`** — unified entry point auto-detects bead, epic, convoy, or batch. Routes to deferred or direct dispatch based on config.
- **`gt scheduler {status,list,run,pause,resume,clear}`** — flat command tree for scheduler management.
- **Capacity pipeline** (`internal/scheduler/capacity/`) — pure-function package for dispatch planning: `PlanDispatch()` computes which beads to dispatch given capacity, batch size, and priority ordering.
- **Metadata in bead descriptions** — scheduler parameters stored as delimited key-value blocks. Sanitized against delimiter injection and newline injection.
- **3-tier label swap retry** — after successful polecat spawn, the `gt:queued` → `gt:queue-dispatched` label swap retries with escalation to prevent double-dispatch.
- **Circuit breaker** — beads that fail dispatch 3 times are quarantined with `gt:dispatch-failed` label.
- **Epic/convoy bulk operations** — `gt sling <epic>` and `gt sling <convoy>` schedule or dispatch all children.
- **Daemon integration** — heartbeat step calls `gt scheduler run` every 3 minutes when `max_polecats > 0`.
- **Backward compatibility** — legacy `---gt:queue:v1---` delimiter and `queue-state.json` supported on read.

### Files Changed (26 files, +6.1K / -0.2K)

**New packages:**
- `internal/scheduler/capacity/` — config, metadata, pipeline, state (pure functions + types)

**New commands:**
- `internal/cmd/scheduler.go` — `gt scheduler` command tree
- `internal/cmd/scheduler_epic.go` — epic schedule/sling operations
[...]

PR #2277, opened 2026-03-03
feat: configurable polecat capacity gating

l0g1x:
## Summary
- Add `operational.polecat.max_active` config (default 5) to replace hardcoded spawn cap of 25
- Convoy launch now dispatches only up to available capacity, deferring the rest to the daemon's feed loop
- Staging capacity warnings aligned with the configured threshold

## Files changed
- `internal/config/types.go` — `MaxActive *int` field on `PolecatThresholds`
- `internal/config/operational.go` — `DefaultPolecatMaxActive = 5` + `MaxActiveV()` accessor
- `internal/cmd/polecat_spawn.go` — configurable spawn gate (was hardcoded 25)
- `internal/cmd/convoy_launch.go` — capacity-gated Wave 1 dispatch
- `internal/cmd/convoy_stage.go` — warning threshold reads from config

## Test plan
- [x] `go build ./...` passes
- [x] Existing `TestDispatchWave1_*` tests pass
- [ ] `gt convoy stage <epic>` — warning threshold reflects configured limit
- [ ] `gt convoy launch <id>` — dispatches up to capacity, reports deferred tasks
- [ ] Config override: set `operational.polecat.max_active` to 3, verify behavior changes

### DIFF
--- a/internal/cmd/convoy_stage.go
+++ b/internal/cmd/convoy_stage.go
@@ -288,7 +288,7 @@ func runConvoyStage(cmd *cobra.Command, args []string) error {
 
 	// Step 8: Categorize findings.
 	allFindings := append(errFindings, warnFindings...)
-	errs, warns := categorizeFindings(allFindings)
+	errs, warns, infos := categorizeFindings(allFindings)
 
 	// Step 9: Choose status.
 	status := chooseStatus(errs, warns)
@@ -342,6 +342,11 @@ func runConvoyStage(cmd *cobra.Command, args []string) error {
 	if len(warns) > 0 {
 		warnOutput := renderWarnings(warns)
 		fmt.Print(warnOutput)
+	}
+
+	// Step 14a: If info findings, render and print.
+	if len(infos) > 0 {
+		fmt.Print(renderInfoFindings(infos))
 	}
 
 	// Step 14b: Resolve convoy title.
@@ -1092,12 +1097,15 @@ type StagingFinding struct {
 	SuggestedFix string   // actionable fix suggestion
 }
 
-// categorizeFindings splits findings into errors and warnings by severity.
-func categorizeFindings(findings []StagingFinding) (errors, warnings []StagingFinding) {
+// categorizeFindings splits findings into errors, warnings, and infos by severity.
+// Info findings are displayed but don't affect convoy status.
+func categorizeFindings(findings []StagingFinding) (errors, warnings, infos []StagingFinding) {
 	for _, f := range findings {
 		switch f.Severity {
 		case "error":
 			errors = append(errors, f)
+		case "info":
+			infos = append(infos, f)
 		default:
 			warnings = append(warnings, f)
 		}
@@ -1720,7 +1728,7 @@ func sortedNodeIDs(dag *ConvoyDAG) []string {
 // ---------------------------------------------------------------------------
 
 // waveCapacityThreshold returns the configured max concurrent polecats,
-// used as the threshold for capacity warnings during staging.
+// used as the threshold for capacity info during staging.
 func waveCapacityThreshold() int {
 	townRoot, err := workspace.FindFromCwdOrError()
 	if err != nil {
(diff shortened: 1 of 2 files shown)


DreadPirateRobertz (commented):
Solid change — the hardcoded spawn cap of 25 was way too high for most setups. Default of 5 is sensible. This pairs well with #2279 (orphaned sling contexts): reducing spawn concurrency reduces the orphan surface area.

One question: is there a migration path for users who've tuned their daemon around the old 25 limit? Might be worth a note in the changelog.

DreadPirateRobertz:
**Review by cfutons/crew/miquella (gastown contributor)**

**Code changes look solid.** The capacity gating approach is clean:
- Configurable `max_active` (default 5, was hardcoded 25) ✓
- Convoy dispatch defers tasks beyond capacity to daemon feed loop ✓
- Staging findings downgraded from "warning" to "info" for capacity (correct — it's informational, not actionable) ✓
- `categorizeFindings` now returns 3 categories cleanly ✓

**Issue: `.beads/backup/` files included in diff.**
The PR includes ~4700 lines of deletions in `.beads/backup/` (backup_state.json, comments.jsonl, dependencies.jsonl, events.jsonl, issues.jsonl, labels.jsonl). These are database backup artifacts and shouldn't be committed in a feature PR. They inflate the diff and may overwrite other contributors' bead state. Recommend:
1. Remove `.beads/backup/` from this PR (git checkout main -- .beads/backup/)
2. Add `.beads/backup/` to `.gitignore` if not already there

**Question:** The default dropped from 25 to 5 — is this the right default for most installations? A town with 3 rigs might want 5-10 concurrent polecats. 5 seems conservative but safe as a default since it's configurable. Just flagging for awareness.

Code LGTM pending .beads cleanup.

DreadPirateRobertz:
Quick review notes:

The core concept (configurable `operational.polecat.max_active` to replace hardcoded spawn cap) is sound. However, the diff includes a large number of beads backup file changes (issues.jsonl, events.jsonl, dependencies.jsonl, etc.) that obscure the actual code changes. These backup files should be excluded from the PR — they're environment-specific artifacts.

**Suggestion**: Rebase onto current main and exclude `.beads/backup/` changes to make the actual code reviewable.

DreadPirateRobertz (commented):
**Review by gastown/crew/zhora**

Clean, well-scoped PR. The capacity gating pattern is solid and follows existing accessor conventions (`MaxActiveV()` with nil-check fallback).

**Concerns:**

1. **Default 25 → 5 is a significant behavioral change.** The old `defaultMaxActivePolecats = 25` was a safety net for spawn storms. Dropping to 5 as the new default means existing users who regularly run 6+ polecats will hit the cap immediately after upgrading. Consider:
   - Default of 10 or 15 as a middle ground?
   - Or at minimum, a note in RELEASING.md about the behavior change

2. **Beads backup file churn.** The diff includes `.beads/backup/` changes (issues.jsonl, events.jsonl, etc.) that are unrelated to the feature. These should be excluded from the PR — they make the diff much harder to review (4733 deletions are mostly backup data).

3. **`waveCapacityThreshold()` calls `workspace.FindFromCwdOrError()`** — During staging this should always succeed (you're in a workspace), but worth a comment noting this assumption. If it fails, you correctly fall back to the default.

4. **Capacity gate in `dispatchWave1` marks deferred tasks as `Success: false`** — Downstream consumers of `DispatchResult` may treat `Success: false` as a failure requiring intervention. The error message is clear ("will be fed by daemon"), but verify that the convoy status machinery doesn't count these as failed dispatches that block the convoy.

**Looks good otherwise:**
- `categorizeFindings` 3-return-value change is clean
- Capacity → info severity downgrade makes sense (expected behavior, not a warning)
- Tests properly updated

Minor: One new test for `MaxActiveV()` would be nice but not blocking.

steveyegge:
Hey Krystian — thanks for the contribution! Unfortunately this duplicates functionality that already landed in PR #1747 (merged Feb 21): a config-driven capacity scheduler at `internal/scheduler/capacity/` with `scheduler.max_polecats` config.

The hardcoded `defaultMaxActivePolecats = 25` in `polecat_spawn.go` is intentionally a safety-net fallback for the direct-dispatch path — the TODO comment there is stale. The real capacity gating lives in the scheduler package and is already configurable.

If you're hitting a specific issue with the existing scheduler, happy to look at it — feel free to open an issue describing the behavior you're seeing. Thanks!

steveyegge:
Hey Krystian — wanted to add more context on the close. The capacity scheduler's deferred-dispatch model (queue work, batch-dispatch on heartbeat ticks) adds latency and complexity that works against Gas Town's core design principle of zero-friction completion. When you sling work, it should start immediately.

The existing direct-dispatch path with the safety-net hard cap is intentional and sufficient for the current architecture. When Gas City ships, it'll provide extension points for custom orchestration rules — so if you need capacity-gated dispatch for your setup, you'll be able to build that as an orchestration policy without modifying the core dispatch path.

Thanks again for the effort!

PR #2289, opened 2026-03-03
fix: derive tmux socket from town name instead of defaulting to "default"

KeithWyatt:
## Summary

When `GT_TMUX_SOCKET` is unset, `InitRegistry()` defaulted to the literal `"default"` tmux socket. This caused a split-brain kill loop when the daemon restarted without the env var:

1. Original daemon creates sessions on `-L gt` (derived from town name)
2. Daemon is killed and restarts without `GT_TMUX_SOCKET` in env
3. New daemon defaults to `-L default` socket — **creates duplicate sessions on a different tmux server**
4. Neither server sees the other's sessions, so the daemon thinks agents need spawning
5. Any orphan cleanup checking only one socket kills processes on the other
6. Kill loop: daemon respawns on wrong socket → cleanup kills on right socket → repeat every 10 minutes

The fix collapses the `""`, `"default"`, and `"auto"` cases in `InitRegistry()` to all derive the socket name from `sanitizeTownName(filepath.Base(townRoot))`. This makes the socket **deterministic from `townRoot`** — no env var needed, no config to persist, no state to lose across restarts. Every process always computes the same socket name.

- Users with `GT_TMUX_SOCKET` explicitly set to a custom value are unaffected
- Users relying on the implicit `"default"` socket need a one-time `gt down && gt up`

### Additional fixes (second commit)

Two related bugs that caused incorrect behavior even after the socket derivation fix:

- **`internal/lock/lock.go`**: `getActiveTmuxSessions()` used bare `tmux list-sessions` with no `-L` flag, so it only queried the "default" socket. Sessions on the per-town socket (e.g., "gt") were invisible, causing `CleanStaleLocks` to incorrectly remove locks for active sessions. Now uses `tmux.GetDefaultSocket()` to include the correct `-L` flag.

- **`internal/cmd/down.go`**: `gt down` only cleaned up sessions on the current socket. When transitioning from old binaries (which used "default") to new ones (per-town socket), ghost sessions persisted on "default" and caused split-brain on next `gt up`. Added Phase 4c that checks the "default" socket for Gas Town sessions and kills them during every `gt down`.

## Changes

- `internal/session/registry.go`: Collapse empty/default/auto cases to derive from town name
- `internal/session/registry_socket_test.go`: Update test expectations
- `internal/cmd/down.go`: Update comment that referenced shared "default" socket; add legacy socket cleanup phase
- `internal/lock/lock.go`: Use per-town socket when querying active tmux sessions
- `internal/lock/lock_test.go`: Add test verifying socket flag is passed

## Test plan

- [x] All `./internal/session/` tests pass (60+ tests)
- [x] All `./internal/lock/` tests pass (19 tests, including new socket flag test)
- [ ] Manual: `gt down && gt up` with `GT_TMUX_SOCKET` unset — verify sessions created on town-derived socket
- [ ] Manual: `gt down && gt up` with `GT_TMUX_SOCKET=custom` — verify sessions on `custom` socket
- [ ] Verify daemon restart (kill daemon, let it respawn) stays on same socket
- [ ] Verify `gt down` cleans up ghost sessions on "default" socket after binary upgrade

### DIFF
--- a/internal/cmd/down.go
+++ b/internal/cmd/down.go
@@ -258,6 +258,23 @@ func runDown(cmd *cobra.Command, args []string) error {
 		}
 	}
 
+	// Phase 4c: Clean up legacy "default" socket sessions.
+	// Old binaries created sessions on the "default" tmux socket. After
+	// transitioning to a per-town socket (e.g., "gt"), those ghost sessions
+	// persist and cause split-brain: the daemon sees them on "gt" but tools
+	// querying bare tmux see (or miss) them on "default".
+	if !downDryRun {
+		cleaned := cleanupLegacyDefaultSocket()
+		if cleaned > 0 {
+			printDownStatus("Legacy sessions", true, fmt.Sprintf("cleaned %d from 'default' socket", cleaned))
+		}
+	} else {
+		count := countLegacyDefaultSocketSessions()
+		if count > 0 {
+			printDownStatus("Legacy sessions", true, fmt.Sprintf("%d would be cleaned from 'default' socket", count))
+		}
+	}
+
 	// Phase 5: Orphan cleanup and verification (--all or --force)
 	if (downAll || downForce) && !downDryRun {
 		fmt.Println()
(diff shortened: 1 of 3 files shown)


DreadPirateRobertz:
**Review by cfutons/crew/miquella (gastown contributor)**

Excellent fix for a nasty split-brain scenario. The deterministic socket derivation from `townRoot` is the right approach — no env var to persist or lose.

**Strengths:**
- Root cause clearly identified (env var loss → socket mismatch → kill loop)
- `lock.go` fix is critical — without it, `CleanStaleLocks` was querying the wrong tmux server and killing active session locks
- Phase 4c legacy cleanup is a good transitional measure
- Test coverage is solid — updated expectations + new socket flag test

**Re: DreadPirateRobertz's question on Phase 4c longevity:**
The legacy cleanup is just `tmux list-sessions` + a loop — negligible cost. I'd leave it permanently. Removing it later risks re-introducing split-brain for anyone running an old binary or whose `gt down` was interrupted before the socket transition completed.

**One minor note:** `cleanupLegacyDefaultSocket` returns 0 when `currentSocket == "default"` — this is correct (already on default, nothing to clean), but the comment could mention that this also covers the case where `InitRegistry` hasn't been called yet (e.g., early error path).

LGTM — recommend merge.

Issue #2095, opened 2026-02-27
gt mail send --wisp generates empty IDs causing UNIQUE constraint failure


essosouli:
## Bug

`gt mail send` with default `--wisp` flag generates empty string IDs when inserting into ephemeral SQLite, causing `UNIQUE constraint failed: issues.id` on any subsequent send.

## Environment

- gt v0.8.0 (Homebrew)
- bd v0.55.4 (Homebrew)
- macOS Darwin 25.3.0
- dolt-native mode (port 3307)

## Reproduction

```bash
gt mail send mayor/ -s "test" -m "testing"
# First send succeeds but inserts row with empty ID
gt mail send mayor/ -s "test2" -m "testing again"
# FAILS: sqlite3: constraint failed: UNIQUE constraint failed: issues.id
```

## Details

- `.beads/config.yaml` has `prefix: hq` configured correctly
- `beads.role` was not set (warning emitted), but setting it to `maintainer` does not fix the issue
- The ephemeral insert path does not read the prefix from config.yaml for ID generation
- `--permanent` flag works correctly — Dolt generates proper `hq-*` IDs
- The `issues` table in `.beads/ephemeral.sqlite3` accumulates rows with empty `id` field

## Impact

- Blocks all wisp-mode mail (the default)
- Breaks witness patrol mail sends
- Breaks hook persistence for mol-based patrol
- Workaround: use `--permanent` flag, but this bypasses ephemeral/TTL lifecycle

## Expected Behavior

Ephemeral inserts should generate proper prefixed IDs (e.g., `hq-xxxxx`) matching the config.yaml prefix, same as the Dolt insert path.

DreadPirateRobertz:
## Triage Analysis

**Root Cause**: The ephemeral SQLite insert path in `bd` does not read the `prefix` from `.beads/config.yaml` for ID generation. When creating a wisp (the default for `gt mail send`), the ID field is empty string. First insert succeeds (empty string is a valid SQLite value), second insert hits `UNIQUE constraint failed: issues.id`.

This is a beads CLI bug, not a gt bug — the ephemeral insert path diverged from the Dolt insert path which correctly generates prefixed IDs.

**Impact Assessment**: **P1 — core workflow broken**
- Blocks ALL default wisp-mode mail (the most common mail path)
- Breaks witness patrol mail sends
- Breaks hook persistence for mol-based patrol
- Workaround exists (`--permanent` flag) but bypasses TTL lifecycle semantics

[...]

PR #2290, opened 2026-03-03
fix(mail): remove msg- ID pass-through that breaks HQ prefix validation

seanbearden:
## Summary
- Removes `--id msg-xxx` pass-through from `sendToSingle` to `bd create`
- The `msg-` prefix (used for in-memory tracking) was rejected by bd's `ValidateIDPrefixAllowed` because HQ beads enforces `hq-` prefix
- The bd wisp/ephemeral path already auto-generates correctly prefixed IDs when `issue.ID` is empty, so the #2095 workaround was unnecessary

## Root cause
PR #2095 added `--id msg.ID` to work around empty SQLite IDs in the ephemeral insert path. But `msg.ID` uses `msg-` prefix for in-memory tracking (notifications, logging), which doesn't match the HQ database prefix `hq-`. This broke **all** `gt mail send` commands town-wide.

## Test plan
- [x] Reproduced: `gt mail send mayor/ -s "test" -m "test"` → prefix mismatch error
- [x] Applied fix, rebuilt gt
- [x] Verified: `gt mail send mayor/ -s "test" -m "test"` → success
- [x] Cross-rig confirmation from laneassist/crew/dom

### DIFF
--- a/internal/mail/router.go
+++ b/internal/mail/router.go
@@ -1113,12 +1113,12 @@ func (r *Router) sendToSingle(msg *Message) error {
 	// Add actor for attribution (sender identity)
 	args = append(args, "--actor", msg.From)
 
-	// Pass the pre-generated message ID so bd uses it instead of generating its own.
-	// Without this, the ephemeral (SQLite) insert path produces empty IDs,
-	// causing UNIQUE constraint failures on subsequent sends (#2095).
-	if msg.ID != "" {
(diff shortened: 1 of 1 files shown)


seanbearden:
Closing — the mail fix (router.go) is already merged upstream. The remaining diff was unrelated quota work and beads backup artifacts that shouldn't have been in this PR.

PR #2291, opened 2026-03-03
fix(done): prefer GT_RIG over cwd-derived rig name

Rome-1:
## Summary

- When Claude Code resets shell cwd (e.g., to `mayor/rig`), `gt done` derives `rigName` as `"mayor"` instead of the actual rig. `GT_RIG` was only a fallback for empty rigName, so it was never consulted.
- This caused polecat worktree reconstruction to fail, making all `bd` commands fail (no `.beads` found at wrong path). MR beads couldn't be created.
- Fix: always prefer `GT_RIG` when set (reliably injected at session creation). Also changes `beads.New(beads.ResolveBeadsDir(cwd))` → `beads.New(cwd)` / `beads.NewWithBeadsDir(cwd, resolvedBeads)` so `cmd.Dir` is the worktree, not the `.beads` directory.
- 9 call sites fixed in `done.go`.

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./internal/cmd/ -run TestDone` passes (all done tests green)
- [ ] Manual verification: polecat runs `gt done` when Claude Code has reset cwd to `mayor/rig` — MR bead should be created successfully

Fixes: hq-4l6t (escalation from vets/rictus)

### DIFF
--- a/cmd/gt-proxy-client/main.go
+++ b/cmd/gt-proxy-client/main.go
@@ -57,7 +57,7 @@ func main() {
 		os.Exit(1)
 	}
 
-	caPEM, err := os.ReadFile(caFile)
+	caPEM, err := os.ReadFile(caFile) //nolint:gosec // G304: caFile from trusted GT_PROXY_CA env var
 	if err != nil {
 		fmt.Fprintf(os.Stderr, "gt-proxy-client: read CA: %v\n", err)
 		os.Exit(1)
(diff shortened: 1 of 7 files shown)


DreadPirateRobertz (commented):
Solid fix. The GT_RIG-first approach is correct — the env var is the authoritative source since it's injected at session creation and doesn't drift when Claude Code resets cwd. The beads.New/beads.NewWithBeadsDir refactor looks clean.

steveyegge:
Fix-merged to main as 39f7bf7d. The lint commit (021fe63) was skipped as those changes already landed via #2302. Core fix (gt done GT_RIG preference) is in. Thanks!

PR #2301, opened 2026-03-03
fix(rig): detect default branch from HEAD in bare clone

pae23:
## Summary
- `gt rig add` failed with `Remote branch main not found` on repos whose default branch is `master` (e.g. self-hosted GitLab)
- Root cause: `RemoteDefaultBranch()` checks `refs/remotes/origin/*` which don't exist in bare repos, so it always fell back to `"main"`
- Fix: use `DefaultBranch()` (`git symbolic-ref --short HEAD`) which git sets correctly during `clone --bare`

## Test plan
- [x] `TestBareCloneDefaultBranch` — creates a bare clone of a `master`-default repo and verifies detection
- [x] Existing `TestAddRig_*` tests pass
- [ ] Manual: `gt rig add gitlab ssh://git@gl-lan.03u.net:2224/gitlab/gitlab.git`

### DIFF
--- a/internal/rig/manager.go
+++ b/internal/rig/manager.go
@@ -414,11 +414,10 @@ func (m *Manager) AddRig(opts AddRigOptions) (*Rig, error) {
 	if opts.DefaultBranch != "" {
 		defaultBranch = opts.DefaultBranch
 	} else {
-		// Try to get default branch from remote first, fall back to local detection
-		defaultBranch = bareGit.RemoteDefaultBranch()
-		if defaultBranch == "" {
-			defaultBranch = bareGit.DefaultBranch()
-		}
+		// Bare repos don't have refs/remotes/origin/* tracking branches,
+		// so detect the default branch from HEAD (which git sets to the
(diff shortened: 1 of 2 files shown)


DreadPirateRobertz (commented):
Good fix. Using `DefaultBranch()` (`git symbolic-ref --short HEAD`) instead of `RemoteDefaultBranch()` for bare repos is correct — bare clones don't have `refs/remotes/origin/*` tracking refs, but git does set HEAD to the remote's default branch during `clone --bare`.

One question: this PR also includes the orphan rig directory UX changes (actionable guidance on `gt rig remove` with orphan dirs + `suggest.FindSimilar`). PR #2295 from the same author appears to be the dedicated PR for that feature. Is this intentional duplication, or should the orphan dir changes live in only one of these PRs to avoid merge conflicts?

The bare clone fix itself LGTM. Test is well-structured — isolating with `GIT_CONFIG_GLOBAL=/dev/null` prevents user config from interfering.

PR #2304, opened 2026-03-04
feat: add Cursor hooks support for polecat agent integration

dinkmsd:
## Summary

- Enable hooks support for the Cursor agent preset (`SupportsHooks: true`)
- Register a dedicated `cursor` hook installer in runtime that writes hook config into `.cursor/` directory
- Add Cursor settings module with hooks configuration for both autonomous and interactive modes
- Include unit tests for the new Cursor settings functionality

## Test plan

- [ ] Verify `go test ./internal/cursor/...` passes
- [ ] Verify `go test ./...` passes with no regressions
- [ ] Test hook installation on a real project with Cursor agent

### DIFF
--- a/internal/config/agents.go
+++ b/internal/config/agents.go
@@ -236,15 +236,19 @@ var builtinPresets = map[AgentPreset]*AgentPresetInfo{
 		SessionIDEnv:        "", // Uses --resume with chatId directly
 		ResumeFlag:          "--resume",
 		ResumeStyle:         "flag",
-		SupportsHooks:       false, // TODO: verify hooks support
+		SupportsHooks:       true,
 		SupportsForkSession: false,
 		NonInteractive: &NonInteractiveConfig{
 			PromptFlag: "-p",
 			OutputFlag: "--output-format json",
 		},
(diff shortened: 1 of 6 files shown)


DreadPirateRobertz (commented):
Nice work! Clean separation of concerns with the `internal/cursor` package and good test coverage.

A few observations:

1. **hooks-autonomous.json vs hooks-interactive.json** — The only difference between these two is that autonomous adds `gt mail check --inject` to `sessionStart`. Could simplify by having one template and conditionally adding that line, but the current approach is fine for clarity.

2. **`EnsureHooksAt` doesn't overwrite existing** — Good defensive behavior. Though users who upgrade from a pre-hooks Gas Town version will need to manually delete their old `hooks.json` to pick up new hooks. Might want to document this.

3. **PATH export in every hook** — The `export PATH="$HOME/go/bin:..."` prefix is repeated in every command. This is necessary since Cursor doesn't inherit the user's shell profile, but worth noting in case there's a centralized way to handle it in the future.

Overall looks ready for merge once tests pass.

PR #2306, opened 2026-03-04
fix(tmux): refresh cycle bindings when rig prefix pattern is stale

### LINKED ISSUE #2299: gt rig add does not refresh tmux cycle bindings for new prefix
## Bug

When a new rig is added via `gt rig add`, the tmux C-b n/p cycle bindings are not updated to include the new rig prefix in the `if-shell` grep pattern.

## Root Cause

`SetCycleBindings()` in `internal/tmux/tmux.go` has an early-return guard (`isGTBindingWithClient`) that skips re-binding if a valid binding already exists. It does not check whether the prefix pattern is stale — i.e., whether all current rig prefixes from `rigs.json` are present in the existing binding.

## Reproduction

1. `gt rig add quickbooks` (prefix `qu-`)
2. `gt crew start` on quickbooks crew
3. C-b n in `qu-crew-angela` → no next window (`qu` missing from pattern)

## Fix

The guard in `SetCycleBindings` should compare `sessionPrefixPattern()` against the baked-in pattern, and re-bind if they differ. Or `gt rig add` should force-refresh the bindings.

## Affected Code

- `internal/tmux/tmux.go`: `SetCycleBindings()`, `isGTBindingWithClient()`, `sessionPrefixPattern()`
- `internal/cmd/crew_cycle.go`: `cycleCrewSession()`, `findRigCrewSessions()`

DreadPirateRobertz:
## Summary

- `isGTBindingWithClient` now verifies the baked-in prefix pattern matches the current rig set
- When a new rig is added, the stale pattern is detected and bindings are re-created with the updated prefix
- `gt rig add` now explicitly calls `SetCycleBindings` after rig creation for immediate effect

## Root Cause

`SetCycleBindings()` had an early-return guard via `isGTBindingWithClient()` that only checked for `--client` presence. It didn't verify whether the regex prefix pattern embedded in the `if-shell` command included all current rig prefixes. After `gt rig add`, the new prefix was missing from the pattern, so `C-b n/p` wouldn't cycle to the new rig's sessions.

## Note

`SetFeedBinding` and `SetAgentsBinding` have a similar pattern-staleness issue (they use `isGTBinding` which also doesn't check the pattern). Those are lower impact since they affect `C-b a` and `C-b g` rather than session cycling, but could be addressed as a follow-up.

## Test plan

- [ ] Manual: `gt rig add testrig <url>` → verify `tmux list-keys -T prefix n` shows the new prefix in the grep pattern
- [ ] Manual: Start crew in new rig → `C-b n` cycles correctly to/from it
- [ ] Existing sessions detect stale pattern on next `SetCycleBindings` call

Fixes #2299

### DIFF
--- a/internal/cmd/rig.go
+++ b/internal/cmd/rig.go
@@ -593,6 +593,16 @@ func runRigAdd(cmd *cobra.Command, args []string) error {
 		fmt.Fprintf(os.Stderr, "Warning: failed to sync hooks for new rig: %v\n", err)
 	}
 
+	// Refresh tmux cycle bindings so the new rig's prefix is included
+	// in the session pattern. Without this, C-b n/p won't cycle to the
+	// new rig's sessions until the next agent session start. See GH#2299.
+	if t := tmux.NewTmux(); t.IsAvailable() {
+		// Pass empty session — we only need to update the pattern, not
+		// the session-specific behavior. SetCycleBindings will detect
+		// the stale pattern and re-bind.
+		_ = t.SetCycleBindings("")
+	}
+
 	elapsed := time.Since(startTime)
 
 	// Read default branch from rig config
--- a/internal/tmux/tmux.go
+++ b/internal/tmux/tmux.go
@@ -2734,15 +2734,22 @@ func (t *Tmux) isGTBinding(table, key string) bool {
 }
 
 // isGTBindingWithClient checks if the given key has a GT binding that includes
-// --client for multi-client support. Older GT bindings without --client cause
-// switch-client to target the wrong client when multiple clients are attached.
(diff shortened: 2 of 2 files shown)


DreadPirateRobertz:
Closing in favor of #2300 which addresses this more thoroughly with dedicated tests and an `isGTBindingCurrent()` helper. Same underlying fix (check pattern freshness in the guard), but #2300 is more complete.

steveyegge:
Thanks for tackling this! We merged #2300 which addresses the same issue (GH#2299) with a similar approach — it adds a separate `isGTBindingCurrent()` method and includes integration tests for the stale-pattern detection. Closing this one in favor of that PR. Appreciate the contribution!

Issue #2303, opened 2026-03-03
Convoy auto-closes immediately due to race condition with daemon stranded scan


mmlac:
## Summary

Convoys created by `gt sling` are auto-closed by the daemon's `ConvoyManager` ~30 seconds after creation with "0/0 completed", despite the tracking dependency being successfully added during sling.

## How Beads Were Slung

When running `gt sling <bead> <rig>`, the sling command executes:

1. **Spawns a polecat** — creates worktree with a fresh branch off the target
2. **Creates a convoy** — `bd create --type=convoy --id=hq-cv-xxxxx` in the HQ beads store (Dolt-backed)
3. **Adds tracking dependency** — `bd dep add <convoyID> <beadID> --type=tracks`
4. **Bonds a molecule** — instantiates the polecat work formula
5. **Hooks the work** — attaches the bead to the polecat's hook
6. **Starts the session**

Both the convoy creation and `dep add` use `WithAutoCommit()` — a wrapper that sets `BD_DOLT_AUTO_COMMIT=on` per-command, overriding the sling-wide `BD_DOLT_AUTO_COMMIT=off`. This was a previous fix for gt-9xum2 where convoy data wasn't persisting at all.

Output looks correct:
```
→ Created convoy 🚚 hq-cv-zwuws
  Tracking: gtd-fbl
```

## The Race Condition

The daemon (`gt up`) runs a `ConvoyManager` goroutine with two scan triggers:

1. **Startup sweep** — 10 seconds after daemon boot
2. **Periodic scan** — every 30 seconds

Each scan calls `findStranded()` → `gt convoy stranded --json`, which for each open convoy runs `getTrackedIssues()` → `bd dep list <convoyID> --direction=down --type=tracks --json`.

The critical code in `daemon/convoy_manager.go:354-358`:
```go
if c.ReadyCount > 0 {
    m.feedFirstReady(c)
} else if c.TrackedCount == 0 {
    // Truly empty convoy — safe to auto-close mechanically.
    m.closeEmptyConvoy(c.ID)
}
```

### Timeline

| Time | Event |
|------|-------|
| T+0s | `gt sling` creates convoy bead with `BD_DOLT_AUTO_COMMIT=on` |
| T+0.1s | `bd dep add <convoy> <bead> --type=tracks` with auto-commit — **succeeds** |
| T+0.2s | Sling prints "Tracking: gtd-fbl" |
| T+~28s | Daemon `ConvoyManager.scan()` fires |
[...]

PR #2307, opened 2026-03-04
fix(sling): add TTL to sling contexts to prevent permanent dispatch blocks

### LINKED ISSUE #2279: Scheduler: orphaned sling contexts block convoy dispatch permanently
## Bug

Sling context wisps created by `gt sling schedule` are never cleaned up when dispatch doesn't follow through, causing `areScheduled()` to permanently mark tasks as "already scheduled" and preventing convoy progress.

## Root Cause

`scheduleBead()` in `sling_schedule.go` creates a sling context wisp (scheduling reservation) at line 167 **before** a polecat is actually spawned. If the subsequent dispatch fails, is throttled by capacity limits, or simply never happens, the context stays open indefinitely.

`areScheduled()` in `sling_schedule.go:271` queries `ListOpenSlingContexts()` and builds a set of work bead IDs that have open contexts. `isReadyIssue()` then skips any task in that set (line 1398: `if scheduledSet[t.ID] { return false }`). This makes the task permanently undispatchable — it looks "scheduled" so it's never offered as "ready", but no polecat is actually working on it.

## Impact

Convoy `hq-cv-vq5` (37 tasks) was stuck at `ready_count: 0` despite 25 open, unblocked, unassigned tasks. The daemon kept logging "auto-closing (empty)" every 30s. Manual cleanup of 25 orphaned sling contexts unblocked the convoy immediately (`ready_count: 24`).

## Reproduction

1. Launch a convoy with >5 tasks
2. With `batch_size: 1`, only 1 task dispatches per cycle
3. The remaining tasks get sling contexts created during the feed cycle but are never actually dispatched
4. `gt convoy stranded --json` shows `ready_count: 0`
5. All open tasks show `scheduled=true` in the `isReadyIssue` check

## Suggested Fixes (pick one)

1. **Close unused contexts after feed cycle**: At the end of `feedFirstReady`, close any sling contexts for tasks that were identified as ready but not dispatched in this cycle
2. **TTL on sling contexts**: Add a liveness check — if a context has been open for >N minutes with no polecat assigned to the work bead, close it automatically
3. **Create context at dispatch time only**: Move context creation from `scheduleBead` into the actual polecat spawn path, so contexts only exist when a polecat is actively working

## Relevant Code

- `internal/cmd/sling_schedule.go:167` — `CreateSlingContext` call
- `internal/cmd/sling_schedule.go:271-314` — `areScheduled()` with fail-closed behavior
- `internal/cmd/convoy.go:1336` — `scheduledSet := areScheduled(trackedIDs)`
- `internal/cmd/convoy.go:1398` — `if scheduledSet[t.ID] { return false }`
- `internal/daemon/convoy_manager.go:389-428` — `feedFirstReady` dispatches only one issue then returns
- `internal/beads/beads_sling_context.go:93-117` — `ListOpenSlingContexts`

## Workaround

Manually close orphaned sling contexts:

```bash
cd ~/gt
dolt --host 127.0.0.1 --port 3307 --user root --password "" --use-db hq --no-tls \
  sql -q "SELECT id FROM wisps WHERE title LIKE '%sling-context%' AND status = 'open'" -r csv \
  | tail -n +2 \
  | xargs -I{} bd close {} --reason="orphaned sling context cleanup"
```

DreadPirateRobertz:
## Summary

- Add 30-minute TTL to sling contexts in `areScheduled()` — contexts older than the TTL are ignored
- Prevents orphaned sling contexts from permanently blocking convoy task dispatch

## Root Cause

`scheduleBead()` creates a sling context bead **before** the polecat spawn attempt. When spawning fails (capacity limits, rig parked, network error), the context remains open. `areScheduled()` then marks the task as "already scheduled", permanently hiding it from `findStrandedConvoys`'s ready-issue detection.

With the daemon's `feedFirstReady` iterating through all ready issues and calling `gt sling` on each (which creates a context), failed spawns accumulate orphaned contexts that block the entire convoy.

## Fix

In `areScheduled()`, skip sling contexts whose `created_at` is older than 30 minutes. A successful sling should complete the context lifecycle well within this window. Orphaned contexts from failed spawns naturally unstale, allowing the task to reappear as "ready".

This is the issue's suggested fix option 2 (TTL). It's the least invasive — no changes to the sling creation path or the daemon's feed logic.

## Relation to #2303

These two issues compound: #2303 (empty convoy auto-close) closes the convoy, and #2279 (orphaned contexts) prevents tasks from appearing as ready. Fixing both is needed for convoys to work reliably.

## Test plan

- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] Existing `TestAreScheduledFailClosed` and `TestAreScheduledEmptyInput` still pass
- [ ] Manual: Create convoy with multiple tasks, kill polecat mid-spawn → verify tasks reappear as ready after 30 min

Fixes #2279

### DIFF
--- a/internal/rig/manager_test.go
+++ b/internal/rig/manager_test.go
@@ -1391,6 +1391,8 @@ func TestBareCloneDefaultBranch(t *testing.T) {
 	gitEnv := append(os.Environ(), "GIT_CONFIG_GLOBAL=/dev/null", "GIT_CONFIG_SYSTEM=/dev/null")
 	for _, args := range [][]string{
 		{"git", "init", "-b", "master", srcDir},
+		{"git", "-C", srcDir, "config", "user.email", "test@test.com"},
+		{"git", "-C", srcDir, "config", "user.name", "Test"},
 		{"git", "-C", srcDir, "commit", "--allow-empty", "-m", "init"},
 	} {
 		c := exec.Command(args[0], args[1:]...)


DreadPirateRobertz:
**Review by cfutons/crew/miquella (gastown contributor)**

This is a clean, well-scoped fix for #2279. The TTL approach is the least invasive option.

**Strengths:**
- Fails closed: empty/unparseable `CreatedAt` falls through → treats context as active (safe default)
- 30-min TTL is generous — normal sling lifecycle completes in minutes, so only true orphans are affected
- Single-file change, minimal blast radius

**Suggestion (non-blocking):** Consider adding a debug-level log when a stale context is skipped — helps operators confirm TTL is working:
```go
log.Debug().Str("bead", ctx.ID).Str("age", now.Sub(created).String()).Msg("skipping stale sling context")
```

**CI note:** The 1 failing test (`TestBareCloneDefaultBranch`) is a CI env issue (missing git user config), unrelated.

LGTM — recommend merge.

PR #2338, opened 2026-03-04
fix: reload prefix registry on heartbeat to prevent ghost sessions

brayden-partcl:
## Summary

- Daemon loaded the prefix registry once at startup from `rigs.json`. Rigs added after daemon start got the default `gt` prefix, causing ghost sessions (`gt-witness`, `gt-refinery`, `gt-furiosa`) instead of correct rig-prefixed names (`ti-witness`, `nu-furiosa`)
- Now reloads the registry on each heartbeat tick so new rigs are immediately recognized
- Kills ghost sessions that use the stale default prefix after reload

## Root cause

`session.InitRegistry()` was called only in `daemon.New()`. `PrefixFor(rigName)` returns `"gt"` (DefaultPrefix) when a rig has no registry entry, so all session names computed after that used `gt-*` instead of the rig's actual prefix.

## Test plan

- [x] `go build ./internal/daemon/` — compiles clean
- [x] `go test ./internal/daemon/` — all tests pass
- [x] `go test ./internal/session/` — all tests pass
- [x] Full binary build (`go build ./cmd/gt`) — links clean
- [ ] Manual: add a new rig while daemon is running, verify no ghost sessions appear
- [ ] Manual: verify existing ghost sessions are cleaned up on next heartbeat

Fixes: hq-ouz, hq-eqf, hq-3i4

### DIFF
--- a/internal/reaper/reaper.go
+++ b/internal/reaper/reaper.go
@@ -27,6 +27,11 @@ var DefaultDatabases = []string{"hq", "beads", "gt"}
 // testPollutionPrefixes are database name prefixes created by tests.
 var testPollutionPrefixes = []string{"testdb_", "beads_t", "beads_pt", "doctest_"}
 
+// isNothingToCommit returns true if the error is a Dolt "nothing to commit" error.
+func isNothingToCommit(err error) bool {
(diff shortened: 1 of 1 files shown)


steveyegge:
Fix-merged to main as f339c019. Thanks for the contribution!

PR #2358, opened 2026-03-04
Decouple Agent Execution from Presentation (Implement ACP)

roder:
# Decouple Agent Execution from Presentation (Implement ACP)

This PR introduces the **Agent Client Protocol ([ACP](https://agentclientprotocol.com/protocol/overview))**. I ripped out the assumption that an agent *must* be chained to a terminal, and instead projected the agent as a formal, headless JSON-RPC 2.0 service.

### The Mayor as a Beachhead

This implementation targets the `Mayor` subsystem (`internal/mayor/manager.go`). However, this fundamentally lays the groundwork for scaling the implementation out.

By formalizing agent I/O into a platform-agnostic RPC layer, we have a blueprint to wrap the rest of the Gastown workforce—specifically the `Polecat` fleet and the `Refinery`—in their own ACP interfaces. Once these background workers communicate via JSON-RPC rather than only being PTY-bound local subprocesses, we can fully decouple them from the host machine. Machines can orchestrate massive swarms of `polecats` running on remote compute, all securely communicating back to the Town. ACP is takes a step towards turning Gastown closer to a distributed service plane.

Moreover, now with a programmatic interface to the Mayor - orchestration of Gastown(s) is completely feasible.

### Using it

[Agents across the industry are adopting ACP as an API](https://agentclientprotocol.com/get-started/registry#available-agents) - there is support natively in Gemini, Copilot, Opencode, and many more.  Openclaw, Zed, and others speak it. There are adapters that wrap Claude, Codex, and others. 

Here's what it takes in your `settings/config.json` to configure an ACP enable agent:

```json
{
  "type": "town-settings",
  "version": 1,
  "default_agent": "opencode",
  "agents": {
    "opencode-acp": {
      "command": "opencode",
      "acp": {
        "command": "acp",
        "args": []
      }
    },
    "claude-agent-acp": {
      "command": "claude-agent-acp",
      "acp": {
        "mode": "native"
      },
      "env": {
        "ANTHROPIC_API_KEY": "**********"
      }
    },
    "gemini-acp": {
      "command": "gemini",
      "acp": {
        "args": ["--experimental-acp", "-m", "gemini-3-flash-preview"]
      }
    }
  }
}

```

From there, rather than `gt mayor attach`, I have introduced the `gt mayor acp --agent "my-acp-agent"` command - which gracefully manages the Mayor's session between the two (Tmux and ACP) interfaces. 

For my Proof-of-Concept - I used Zed IDE to consume Gastown's mayor  with the IDE's `settings.json` configured as such:

```
{ 
  "agent_servers": {
    "Gastown": {
      "type": "custom",
      "command": "gt",
      "args": ["mayor", "acp", "--agent", "opencode-acp"],
      "env": {},
    }
 }
```

Here's what it looks like

https://github.com/user-attachments/assets/dd091790-1bd6-44b7-bcea-dbd6de9d2a37

---

### The Gory Technical Details (The Plumbing)

I used Gastown to build this initially. I refined it by hand. This introduces **6,414 insertions** across **31 files**. It's not trivial PR to review, but I think its worth it. 

Here is the core of the implementation:

#### 1. The State-Aware Multiplexing Proxy (`internal/acp/proxy.go`)

At more than 1K of lines, this is the heavy lifting. It implements a platform-native bridge (Unix domain sockets on POSIX, Named Pipes on Windows) between the host (like an IDE) and the agent's standard I/O.

* **Busy-State Lockout**: External clients are impatient. The proxy implements a strict state machine tracking the agent's cognitive load. If the agent is mid-inference (`busy`), the proxy safely queues or blocks prompt injections to prevent clobbering the active execution context.
* **Session Lifecycle**: Strict protocol handshakes and keep-alive heartbeats solve the classic "zombie agent" problem. If an IDE silently disconnects, the proxy gracefully tears down the agent context.

#### 2. The Propulsion Event Loop (`internal/acp/propulsion.go`)

When an agent is running headlessly, it is blind to the rest of the "Town" environment. The Propulsion System (507 lines) is a new event-driven polling daemon that acts as the agent's nervous system.

* It actively polls for Mail escalations, Hook status changes (new work assignments), and urgent Nudges.
* **Safe Injection**: Instead of blindly dumping text into `stdin`, Propulsion uses sophisticated retry logic to inject these notifications natively into the structured JSON-RPC stream, ensuring the agent processes the interruption safely.

#### 3. Dual-Mode Operation (`internal/mayor/manager.go`)

The `Mayor` subsystem was refactored (+235 lines) to support multi-modal operation (`ModeTMUX`, `ModeACP`, `ModeBoth`).

* **Schrödinger’s Pane**: You can now spin up an agent headlessly via your IDE. If it gets stuck, you can run `gt mayor attach` to drop into a dynamically generated `tmux` session mid-execution, watch the logs stream, and then safely detach—all without tearing down the underlying headless process group.

#### 4. The Cleanup Veto Interlock (`internal/mayor/cleanup.go`)

Our `polecat` subsystem is highly aggressive about nuking orphaned processes. To a naive cleanup script, a headless background agent looks exactly like a dead orphan.

* **PID Tracking**: Active ACP sessions are now strictly registered at `mayor/mayor-acp.pid`.
* **Process-Tree Traversal**: The new `CleanupVetoChecker` performs a deep liveness check. When the automated cleanup routine triggers, it traverses the process tree; if a valid JSON-RPC socket is active, it hard-vetoes the `nuke` command and defers wisp creation. Out of sight no longer means accidentally deleted.

#### 5. Mail Router Fallbacks & Robustness (`internal/mail/router.go`)

The mail router now intelligently enqueues nudges even when no `tmux` session is found, enabling mail notifications to successfully route to purely headless ACP sessions.

There are no new external dependencies—this is a pure Go implementation using the standard library. Most importantly, it is fully backward compatible with all existing purely-tmux workflows.

This is the necessary infrastructure to make Gastown the standard backbone for agentic integration.

## Testing

- [X] Unit tests pass (`go test ./...`)
- [X] Manual testing performed

## Checklist
- [X] Code follows project style
- [X] Documentation updated (if applicable)
- [X] No breaking changes (or documented in summary)

### DIFF
--- a/internal/acp/propulsion.go
+++ b/internal/acp/propulsion.go
@@ -2,13 +2,10 @@ package acp
 
 import (
 	"context"
-	"encoding/json"
 	"fmt"
 	"os"
-	"os/exec"
 	"path/filepath"
 	"strconv"
-	"strings"
 	"sync"
 	"time"
 
@@ -120,16 +117,6 @@ func logEvent(townRoot, eventType, context string) {
 	debugLog(townRoot, "[%s] %s", eventType, context)
 }
 
-const pollInterval = 30 * time.Second
-
-type mailMessage struct {
-	ID       string
-	Subject  string
-	From     string
-	Read     bool
-	Escalate bool
-}
-
 type Propeller struct {
 	proxy       *Proxy
 	townRoot    string
@@ -137,10 +124,6 @@ type Propeller struct {
 	ctx         context.Context
 	cancel      context.CancelFunc
 	wg          sync.WaitGroup
-	mailIDs     map[string]mailMessage
-	mailMu      sync.RWMutex
-	hookState   string
-	hookMu      sync.RWMutex
 	warnedNoSID bool
 }
 
@@ -149,7 +132,6 @@ func NewPropeller(proxy *Proxy, townRoot, session string) *Propeller {
 		proxy:    proxy,
 		townRoot: townRoot,
 		session:  session,
-		mailIDs:  make(map[string]mailMessage),
 	}
 }
 
@@ -157,14 +139,16 @@ func (p *Propeller) Start(ctx context.Context) {
 	debugLog(p.townRoot, "[Propeller] Starting for session %q in town %q", p.session, p.townRoot)
 	p.ctx, p.cancel = context.WithCancel(ctx)
 	p.wg.Add(1)
-	go p.waitForSessionAndPoll()
+	go p.waitForSessionAndStart()
 }
 
 // waitForSessionAndPoll waits for the ACP handshake to complete (sessionID available)
 // before starting the poll loop. This ensures the proxy is ready to inject prompts.
-// If no IDE connects within the timeout, polling continues with degraded functionality
+// waitForSessionAndStart waits for the ACP handshake to complete (sessionID available)
+// before starting the event-driven loop. This ensures the proxy is ready to inject prompts.
+// If no IDE connects within the timeout, event loop continues with degraded functionality
 // (notifications will be skipped since there's no session to inject into).
-func (p *Propeller) waitForSessionAndPoll() {
+func (p *Propeller) waitForSessionAndStart() {
 	defer p.wg.Done()
 
 	// Wait for sessionID with a timeout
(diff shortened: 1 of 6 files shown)


steveyegge (changes requested):
Hey @roder -- thanks for this PR. ACP is a real capability we need, and the proxy implementation in particular shows solid engineering. The demo with Zed is compelling.

After a thorough review, we'd like to see some changes before merging. The core proxy and cleanup veto are in good shape, but a few areas need work:

## Must-fix

### 1. Propulsion should be event-driven, not polling

The Propeller shells out to `gt mail inbox` and `gt hook show` every 30 seconds, which adds unnecessary Dolt/CLI load -- exactly the kind of overhead the codebase warns against. More importantly, 30s polling means urgent nudges can sit undelivered for half a minute, undermining the priority system.

**Suggestion:** Use fsnotify/kqueue on the nudge queue directory for instant delivery at zero idle cost. Have the mail router emit a local signal when delivering to the mayor inbox rather than re-querying the entire inbox. The nudge system already has `FormatForInjection()` -- reuse it instead of reformatting.

### 2. Windows/Unix proxy duplication

`proxy.go` (1159 lines) and `proxy_windows.go` (880 lines) are ~85% identical, and the Windows version is already missing features (no stderr saturation protection, no crash diagnostics, no redacted thought filtering). This will diverge further.

**Suggestion:** Extract a shared core with platform-specific files only for the ~50 lines of actual differences (process group setup, signal handling, process liveness check).

### 3. Test coverage for critical paths

The proxy relay loop (`relayJSONRPC`) and propulsion injection (`pollAndInject`) -- the two most important runtime paths -- have no test coverage. The existing tests cover serialization and state extraction well, but the actual forwarding and injection are where bugs will hide.

## Should-fix

### 4. Drop or simplify ModeBoth

No code path intentionally enters `ModeBoth`. The system already enforces mutual exclusion (StartACP kills tmux, attach kills ACP). A simpler `isACPActive` boolean on the status struct would be cleaner and less confusing.

### 5. Retry logic in Propeller.notify()

The 3x100ms retry can re-enter InjectPrompt's 15s startup wait, blocking for up to 45s total. The retry also does not classify errors -- it retries identically whether the agent is busy (state unlikely to change in 100ms) or the proxy is shutting down (will never succeed). Add a timeout on the outer loop and skip retries for terminal errors.

### 6. Escalation detection is fragile

Subject-line substring matching (`strings.Contains(subject, "help")`) will false-positive on messages like "helpful tips for code review" and miss escalations that use different phrasing. Consider using the actual escalation metadata instead.

## Minor

- `agentDone()` calls `cmd.Wait()` but `Forward()` also calls it via errChan -- double-Wait is undefined behavior in Go (test-only but still incorrect)
- `getACPSessionPIDs()` hardcodes `~/gt` as the only town root despite a comment mentioning `~/town-*`
- Mail router fallback does not guard on `IsACPActive`, so it can accumulate orphan nudge files
- `IsBusy()` takes a full `sync.Mutex` when it only reads -- consider `sync.RWMutex` for consistency

---

Overall this is strong work and we want to see it land. Consider a phased approach: ship the proxy + cleanup + mayor refactoring first (that is the core value), then follow up with a reworked propulsion system. That would also make each piece easier to review. Thanks!

roder:
Hey @steveyegge , thanks for the quick reply and the kind words.  Those changes are great feedback and I'm stoked to get them in place.  I'll get to work on and break it up into phases and ship them to you as they come along. TTYS

steveyegge:
Merged to main with three review fixes applied on top (commit f451959f):

1. **ErrACPActive dead branches (N1):** `Start()` now returns `ErrACPActive` (instead of `ErrAlreadyRunning`) when ACP is active, so the `errors.Is(err, mayor.ErrACPActive)` checks in `start.go` and `up.go` are no longer dead branches.
2. **ensureMayorInfra error check (N2):** `runMayorAcp` now checks the error return from `ensureMayorInfra(townRoot)` (was silently discarded).
3. **Rename ACAgentFilePath -> ACPAgentFilePath** for naming consistency with `ACPPidFilePath`.

Note: propulsion.go was already converted from 30s polling to event-driven (fsnotify-based via `nudge.WatcherForSession`) in commit cffa8b40, so no exclusion was needed. All tests pass.

PR #2360, opened 2026-03-04
Add schema evolution support to gt wl sync

dutchiono:
## Summary
- MINOR version bumps (new columns): auto-applied during sync
- MAJOR version bumps (breaking changes): require explicit `--upgrade` flag
- Compares `_meta.schema_version` between local and upstream
- Graceful fallback if version not found

## Changes
- Added `--upgrade` flag to `gt wl sync`
- Added schema version parsing (semver)
- Added version comparison and upgrade logic
- Auto-commits schema evolution changes

## Usage
```bash
gt wl sync                # Auto-applies MINOR bumps
gt wl sync --upgrade      # Required for MAJOR bumps
```

## Task
Closes w-wl-004

## Test plan
- [ ] Test with same schema version (no action)
- [ ] Test with MINOR version bump (auto-applies)
- [ ] Test with MAJOR version bump without --upgrade (errors)
- [ ] Test with MAJOR version bump with --upgrade (applies)

### DIFF
--- a/internal/cmd/wl_sync.go
+++ b/internal/cmd/wl_sync.go
@@ -5,6 +5,8 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strconv"
+	"strings"
 
 	"github.com/spf13/cobra"
 	"github.com/steveyegge/gastown/internal/style"
@@ -12,7 +14,10 @@ import (
 	"github.com/steveyegge/gastown/internal/workspace"
 )
 
-var wlSyncDryRun bool
+var (
+	wlSyncDryRun bool
+	wlSyncUpgrade bool
+)
 
 var wlSyncCmd = &cobra.Command{
 	Use:   "sync",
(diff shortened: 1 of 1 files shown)


steveyegge:
Closing in favor of #2388 which covers the same scope (w-wl-004) with a better design:

1. **Sequencing**: #2388 splits `dolt pull` into fetch→schema check→merge, so MAJOR upgrades are blocked *before* the merge. This PR runs the check *after* `dolt pull`, which means the merge has already happened — defeating the safety gate.
2. **Scope**: This PR includes unrelated changes to beads.go (removing --db isolated mode) and beads_types.go (removing SQLite legacy check) that weren't discussed and risk regressions.
3. **Fallback behavior**: Defaulting to "1.0" when _meta is missing is less safe than #2388's approach of skipping the check entirely for pre-versioned forks.

Thank you for the contribution — the core idea is right, #2388 just has the implementation details nailed down.

PR #2368, opened 2026-03-05
fix(wl): let claim/done use joined wl-commons clone when server DB is absent

gunabot:
## Problem
`gt wl claim` and `gt wl done` currently hard-fail with:

```
database "wl_commons" not found
Join a wasteland first with: gt wl join <org/db>
```

in real joined workspaces where `gt wl join` created a local clone at:

```
<town>/.wasteland/<org>/wl-commons
```

This happens because claim/done only use the doltserver path (`WLCommonsDB = wl_commons`) and do not fall back to the clone-backed `LocalDir` from wasteland config.

## Fix
- `internal/cmd/wl_claim.go`
  - keep existing server-DB path when available
  - add fallback to `wlCfg.LocalDir` (clone-backed flow) when `WLCommonsDB` is missing
  - execute equivalent `dolt sql` script in clone dir
- `internal/cmd/wl_done.go`
  - same fallback pattern for completion submit flow
  - preserve existing validations/error semantics (`nothing to commit` => precondition-style error)

## Why this is safe
- No behavior change when `WLCommonsDB` exists.
- Fallback only triggers when old path is absent.
- Uses same SQL lifecycle (`UPDATE/INSERT + DOLT_ADD + DOLT_COMMIT`) already used in server-backed implementation.

## Validation
- Reproduced failure before patch in joined workspace.
- After patch, `gt wl done` no longer fails on missing `wl_commons`; it reaches real status checks.
- Tests pass for cmd package:

```bash
go test ./internal/cmd -run 'Test.*(Wl|wl).*' -count=1
```

### DIFF
--- a/internal/cmd/wl_claim.go
+++ b/internal/cmd/wl_claim.go
@@ -2,6 +2,8 @@ package cmd
 
 import (
 	"fmt"
+	"os/exec"
+	"strings"
 
 	"github.com/spf13/cobra"
 	"github.com/steveyegge/gastown/internal/doltserver"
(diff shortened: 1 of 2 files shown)


steveyegge:
Fix-merged to main in beead3a1. Includes the original changes plus a fix for empty Title output in the local clone claim path (conditional print when Title is available). Thanks @gunabot!

PR #2399, opened 2026-03-06
fix: guard against bd v0.58.0 non-JSON output in list/hook/mail parsing

dannomayernotabot:
## Summary

- `bd v0.58.0` returns plain text (e.g. `"No issues found."`) instead of empty JSON arrays when result sets are empty
- This broke `gt hook`, `gt mail inbox`, and `gt patrol` — all witness automation was degraded
- Guard all `bd list` output parsers against non-JSON responses

## Changes

- `internal/beads/beads.go`: check if output starts with `[` before JSON parse
- `internal/deacon/stale_hooks.go`: same guard
- `internal/mail/mailbox.go`: same guard with graceful empty-list fallback

## Test plan

- [ ] `bd list --json` returning plain text no longer causes parse errors
- [ ] `gt hook`, `gt mail inbox`, `gt patrol` work correctly with bd v0.58.0
- [ ] Empty result sets return empty arrays, not errors

### DIFF
--- a/internal/mail/mailbox.go
+++ b/internal/mail/mailbox.go
@@ -2,6 +2,7 @@ package mail
 
 import (
 	"bufio"
+	"bytes"
 	"context"
 	"encoding/json"
 	"errors"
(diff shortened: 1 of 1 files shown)


DreadPirateRobertz:

Addresses the same bd v0.58.0 non-JSON output issue as #2440, but in the mail subsystem specifically. The pre-parse check before \`json.Unmarshal\` is correct but note there's already a post-Unmarshal fallback (\`isJSON\` check) — the new pre-check is redundant with that existing guard.

That said, checking before Unmarshal avoids the error log from failed parsing, so it's a worthwhile optimization.

Minor: the variable name \`trimmedCC\` (line 211) should probably be just \`trimmed\` — looks like a copy-paste artifact.

LGTM with minor nit.

PR #2410, opened 2026-03-06
fix: skip redirecting rigs in DatabasePrefixCheck to prevent shared DB corruption

### LINKED ISSUE #2409: DatabasePrefixCheck corrupts shared database prefix when rigs redirect to town DB
## Bug Description

`DatabasePrefixCheck` in `internal/doctor/beads_check.go` assumes each rig has its own database. When multiple rigs redirect to the same town database (e.g. `site_manager/.beads/redirect → ../.beads`), they all share the town root's `hq` database.

The check iterates over each route in `routes.jsonl` and, for prefix mismatches, the `Fix()` method runs `bd config set issue_prefix <routesPrefix>` from the rig's directory. Because redirecting rigs all operate on the same shared DB, each fix overwrites the previous one. The last rig processed wins, leaving the shared database with the wrong `issue_prefix`.

## Steps to Reproduce

1. Set up a town with a shared beads database at the town root (`.beads/`)
2. Have one or more rigs redirect to the town DB via `.beads/redirect` containing `../.beads`
3. `routes.jsonl` has entries like `{"prefix":"hq-","path":"."}` and `{"prefix":"sm-","path":"site_manager"}`
4. Run `gt doctor --fix`

## Expected Behavior

Rigs that redirect to a shared database should be skipped by `DatabasePrefixCheck`. The prefix of the shared database is owned by the canonical route (the town root), not by the redirecting rig.

## Actual Behavior

The check flags the redirecting rig as a mismatch (DB has `hq` but route says `sm`) and `--fix` overwrites the shared database's `issue_prefix` to `sm`, breaking prefix routing for the entire town.

## Impact

- The town root database ends up with a rig-specific prefix (e.g. `sm` instead of `hq`)
- All cross-cutting work defaults to the wrong rig
- The fix is destructive: it silently corrupts the shared database configuration

## Proposed Fix

In `DatabasePrefixCheck.Run()`, resolve both the town root's beads directory and each rig's beads directory. When they resolve to the same path, skip the rig — its database prefix is owned by the town root route.

## Files

- `internal/doctor/beads_check.go` (`DatabasePrefixCheck`, lines ~424-579)

Cdfghglz:
## Summary

- `DatabasePrefixCheck` assumed each rig has its own beads database. When rigs redirect to a shared town database (via `.beads/redirect`), the check would flag false prefix mismatches and `--fix` would overwrite the shared DB's `issue_prefix`, corrupting prefix routing for the entire town.
- Resolve each rig's beads directory and skip rigs whose resolved path matches the town root's beads directory — their prefix is owned by the town root route, not by the redirecting rig.
- Introduce `dbPrefixGetter` interface to enable testing without shelling out to `bd`.

Fixes #2409

## What changed

**`internal/doctor/beads_check.go`:**
- Add `dbPrefixGetter` interface + `realDBPrefixGetter` to allow mocking the `bd config get issue_prefix` call in tests
- In `Run()`, resolve the town root's canonical beads directory and skip any rig whose resolved beads directory matches it
- Update `Fix()` comments to clarify that shared-DB rigs are excluded by `Run()`

**`internal/doctor/beads_check_test.go`:**
- Add `mockDBPrefixGetter` for testing without `bd`
- `TestDatabasePrefixCheck_SkipsRigRedirectingToTownDB` — single rig redirecting to town DB is skipped
- `TestDatabasePrefixCheck_DetectsMismatchForOwnDB` — rig with its own DB and wrong prefix is still flagged
- `TestDatabasePrefixCheck_MultipleRedirectsSameDB` — multiple redirecting rigs all skipped
- `TestDatabasePrefixCheck_MixedOwnAndRedirect` — mix of redirecting and own-DB rigs: only the own-DB mismatch is flagged

## Test plan

- [x] All 4 new tests pass covering the redirect scenarios
- [x] All 3 existing `DatabasePrefixCheck` tests still pass
- [x] Full `go build ./internal/doctor/` compiles cleanly
- [ ] Manual test in a town with redirecting rigs: `gt doctor` no longer flags false prefix mismatches

### DIFF
--- a/internal/doctor/beads_check.go
+++ b/internal/doctor/beads_check.go
@@ -420,6 +420,25 @@ func (c *RoleLabelCheck) Fix(ctx *CheckContext) error {
 	return nil
 }
 
+// dbPrefixGetter abstracts querying the database for issue_prefix.
+// Allows mocking in tests without shelling out to bd.
+type dbPrefixGetter interface {
+	GetDBPrefix(rigPath string) (string, error)
+}
+
+// realDBPrefixGetter shells out to bd to query the database.
+type realDBPrefixGetter struct{}
+
+func (r *realDBPrefixGetter) GetDBPrefix(rigPath string) (string, error) {
+	cmd := exec.Command("bd", "config", "get", "issue_prefix")
+	cmd.Dir = rigPath
+	output, err := cmd.Output()
+	if err != nil {
+		return "", err
+	}
+	return strings.TrimSpace(string(output)), nil
+}
+
 // DatabasePrefixCheck detects when a rig's database has a different issue_prefix
 // than what routes.jsonl specifies. This can happen when:
 // - The database was initialized with a different prefix
@@ -428,9 +447,15 @@ func (c *RoleLabelCheck) Fix(ctx *CheckContext) error {
 //
 // Unlike PrefixMismatchCheck (rigs.json ↔ routes.jsonl), this check verifies
 // the actual database configuration matches the routing table.
+//
+// Rigs that redirect to a shared database (e.g. the town root's .beads) are
+// skipped. Their database prefix is owned by the route that provides the
+// canonical database, not by the redirecting rig. Attempting to "fix" these
+// would overwrite the shared database's prefix with the rig's prefix.
 type DatabasePrefixCheck struct {
 	FixableCheck
-	mismatches []databasePrefixMismatch
+	mismatches     []databasePrefixMismatch
+	prefixGetter   dbPrefixGetter
 }
 
 type databasePrefixMismatch struct {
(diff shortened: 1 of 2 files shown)


DreadPirateRobertz:

This is a solid fix — same bug we independently fixed in #2462, but with a cleaner approach:

- The `dbPrefixGetter` interface is a good pattern for testability (our fix doesn't mock `getDBPrefix`)
- The comprehensive doc comment on `DatabasePrefixCheck` explaining why redirecting rigs are skipped is excellent
- Test coverage with the mock getter is more robust than our approach

Recommend merging this over #2462 — the interface abstraction and test isolation are superior. Our #2462 can be closed as superseded.

One minor note: the `filepath.Abs` comparison for detecting shared DBs is the same approach we used — good to see convergence on the solution.

LGTM with preference to merge this one.

Issue #2405, opened 2026-03-06
gt rig add: bd init starts rogue Dolt server instead of using Gas Town's central server


pkobielak:
## Summary

`gt rig add` calls `bd init --server` during rig creation, but `bd init` starts its own isolated Dolt server instead of connecting to Gas Town's central Dolt server (port 3307). This causes every subsequent `bd` command to fail with "database not found".

The same issue affects `gt install` (town-level `hq` database).

## Environment

- gastown v0.11.0 (Homebrew)
- beads v0.59.0 (Homebrew)
- macOS (Darwin 25.3.0, arm64)

## Steps to Reproduce

1. `gt install ~/gt --git`
2. `gt rig add hubjobs https://github.com/Hub-xyz/hub-jobs.git`
3. Run any `bd` command in the rig (e.g., `bd ready`, `bd doctor`)

## Expected

`bd` connects to Gas Town's central Dolt server on port 3307 (data dir `.dolt-data/`) which has the `hubjobs` database created by `doltserver.InitRig()`.

## Actual

`bd init --server` spins up its own Dolt server on a random port (e.g., 13597) with data dir `.beads/dolt/`. This server does not have the rig's database, so all operations fail:

```
Error: database "hubjobs" not found on Dolt server at 127.0.0.1:13597
```

The warning is visible during `gt rig add`:
```
Warning: Could not set issue_prefix on rig database: exit status 1
(Error: failed to open database: database "hubjobs" not found on Dolt server at 127.0.0.1:13597)
```

## Root Cause

In `internal/rig/manager.go`, `InitBeads()` calls `bd init` with `--server` and `--prefix` but **never passes `--server-port 3307`** (the Gas Town central server port). The `--server-port` flag is only passed when `GT_DOLT_PORT` env var is set (test environments):

```go
// Line ~900 in manager.go
if p := os.Getenv("GT_DOLT_PORT"); p != "" {
    initArgs = append(initArgs, "--server-port", p)
}
```

Without `--server-port`, `bd init --server` doesn't find the central server and starts its own.

Additionally, `EnsureMetadata()` in `internal/doltserver/doltserver.go` sets `dolt_mode: "server"` and `dolt_database` in metadata.json but **never writes the server port**, so `bd` has no way to discover the correct server on subsequent runs.

## Workaround

For each affected rig/town:

```bash
# Stop the rogue bd-managed Dolt server
bd dolt stop

# Reinitialize pointing at Gas Town's central server
[...]

PR #2413, opened 2026-03-06
fix: inject Dolt server port into agent tmux sessions (GH #2412)

### LINKED ISSUE #2412: bd auto-starts rogue Dolt servers: agent tmux sessions missing BEADS_DOLT_PORT
## Summary

Agent tmux sessions (witnesses, refineries, polecats) don't have `BEADS_DOLT_PORT` or `GT_DOLT_PORT` in their environment. When agents call `bd` directly from the shell, `bd` cannot discover gt's central Dolt server and auto-starts its own on a random port — one per rig.

This is the **runtime complement** to #2405 (which covers the `gt rig add` / `bd init` path). #2405 fixes the initialization path; this issue fixes the ongoing runtime path where agents invoke `bd` throughout their session lifecycle.

## Relation to #2405

Issue #2405 identifies that `bd init --server` doesn't receive `--server-port`, so bd creates its own server during rig setup. This issue covers the broader problem: even if init is fixed, agents running in tmux sessions will still have `bd` fall back to auto-starting servers because neither `BEADS_DOLT_PORT` nor `GT_DOLT_PORT` is set in the session environment.

## Environment

- gastown v0.11.0
- beads v0.59.0
- Linux (Ubuntu 24.04)

## Steps to Reproduce

1. Install gt + bd, set up a town with multiple rigs
2. Reinstall bd (or restart the system)
3. `gt start` — daemon starts witnesses/refineries in tmux
4. Each agent calls `bd` commands during patrol
5. `bd` finds no `BEADS_DOLT_PORT` in env → auto-starts its own Dolt server

## Expected

Agents connect to gt's central Dolt server (port 3307 by default).

## Actual

Each rig's agent spawns a separate Dolt server on a random port (13476, 13505, 13576, etc.), each serving an empty `.beads/dolt/` data directory that doesn't have the rig's database. Result: 17 zombie Dolt servers and all `bd` commands fail with "database not found".

Confirmed by `tmux show-environment -t sm-witness` — no `GT_DOLT_PORT` or `BEADS_DOLT_PORT` present.

## Root Cause

`AgentEnv()` in `internal/config/env.go` (the single source of truth for agent session environment) does not inject `GT_DOLT_PORT` or `BEADS_DOLT_PORT`. The existing `translateDoltPort()` in `internal/beads/beads.go` only covers the path where gt spawns bd as a subprocess — not the path where agents call bd directly from their shell.

## Suggested Fix

1. **`AgentEnv()`** should resolve the Dolt port from `.dolt-data/config.yaml` (or `daemon.json` / `GT_DOLT_PORT` env var) and inject both `GT_DOLT_PORT` and `BEADS_DOLT_PORT` into the returned env map
2. **`InitBeads()`** should always pass `--server-port` with the resolved port (not just when `GT_DOLT_PORT` env var is set) — this also addresses #2405

Cdfghglz:
## Summary

- **AgentEnv()** now resolves the central Dolt server port and injects `GT_DOLT_PORT` + `BEADS_DOLT_PORT` into the tmux session environment for all agent roles (mayor, witness, refinery, polecat, crew, deacon)
- **InitBeads()** always passes `--server-port` to `bd init` using `doltserver.DefaultConfig()` instead of only when `GT_DOLT_PORT` env var is set
- Port resolution reads `.dolt-data/config.yaml` → `GT_DOLT_PORT` env → `daemon.json`, without importing `doltserver` (avoids pulling yaml/sql/mysql into the config package)

## Problem

After a bd/gt reinstall (or any scenario where `.beads/dolt-server.port` files are stale), agents' direct `bd` invocations auto-start rogue Dolt servers. We observed **17 zombie Dolt servers** each serving empty data directories, with every `bd` command failing "database not found".

Root cause: `AgentEnv()` — the single source of truth for agent session environment — did not set `BEADS_DOLT_PORT` or `GT_DOLT_PORT`. The existing `translateDoltPort()` in `beads.go` only covers the "gt spawns bd as subprocess" path, not the "agent calls bd from shell" path.

Fixes #2412, also addresses #2405 (the `InitBeads` path).

## Test plan

- [x] `TestParsePortFromConfigYAML` — 6 cases covering standard, custom, missing, and malformed configs
- [x] `TestResolveDoltPort_*` — 5 tests: config.yaml, env var, precedence, daemon.json, no-config
- [x] `TestAgentEnv_InjectsDoltPort` — verifies all 6 agent roles get both port vars
- [x] `TestAgentEnv_NoDoltPortWithoutTownRoot` — no injection when TownRoot empty
- [x] `TestAgentEnv_NoDoltPortWithoutConfig` — no injection when no config exists
- [x] Full `go test ./internal/config/` suite passes
- [x] `go vet ./internal/config/ ./internal/rig/` clean
- [x] `go build ./...` clean


DreadPirateRobertz:

This is a more comprehensive fix than our #2465 for the same rogue Dolt server issue. Key advantages:

- Multi-source port resolution: config.yaml -> GT_DOLT_PORT -> daemon.json (vs our simple DefaultPort fallback)
- Avoids importing doltserver package (no yaml/sql dependency pull)  
- The `parsePortFromConfigYAML` line scanner is safe for machine-generated configs
- Injecting both GT_DOLT_PORT and BEADS_DOLT_PORT covers both gt and bd code paths

One question: when `resolveDoltPort` returns 0 (no config found), bd will still auto-start its own server. Is that the desired behavior for fresh installs before `gt dolt start`? Might want to document this edge case.

LGTM — recommend merging this over our #2465.

PR #2421, opened 2026-03-06
fix: convert session-hygiene from plugin.md to deterministic run.sh

seanbearden:
## Summary
- AI dog killed all 10 crew sessions **twice** by ignoring plugin.md instructions and using `gt rig list --json` (rig names like "gastown") instead of reading `rigs.json` (beads prefixes like "gt")
- Converted session-hygiene from AI-interpreted plugin.md to a deterministic `run.sh` shell script, eliminating AI interpretation of destructive cleanup logic
- Follows the same pattern as compactor-dog's run.sh conversion

## What changed
- **`plugins/session-hygiene/run.sh`** (new): Standalone executable script that reads `rigs.json` for valid prefixes, identifies zombie/orphaned tmux sessions, and kills them deterministically. Supports `--dry-run` flag.
- **`plugins/session-hygiene/plugin.md`**: Stripped executable bash blocks, bumped to v2. Now metadata + description only — `run.sh` does the work.

## Test plan
- [x] `bash -n run.sh` — syntax check passes
- [x] `./run.sh --dry-run` — correctly found 0 zombies across 47 sessions, identified 1 orphaned dog session
- [ ] Verify `gt-crew-bear` is NOT flagged as zombie (prefix `gt` matches beads prefix for `gastown`)
- [ ] Verify truly unknown prefixes (e.g., `xyz-foo`) ARE flagged as zombies
- [ ] Verify `hq-*` sessions pass through unchanged

### DIFF
--- a/plugins/session-hygiene/plugin.md
+++ b/plugins/session-hygiene/plugin.md
@@ -1,7 +1,7 @@
 +++
 name = "session-hygiene"
 description = "Clean up zombie tmux sessions and orphaned dog sessions"
-version = 1
+version = 2
 
 [gate]
 type = "cooldown"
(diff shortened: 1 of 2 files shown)


DreadPirateRobertz (approved):
Clean fix. Sessions are named with beads prefixes (e.g., `gt-crew-bear`), not rig names (`gastown-crew-bear`), so the zombie detection was incorrectly flagging legitimate sessions. Reading both rig names and beads prefixes from `rigs.json` is the right approach.

CI failures are pre-existing lint issues in unrelated Go files, not caused by this plugin.md change.

DreadPirateRobertz:

Good fix. Session-hygiene was using rig names as session prefixes, but sessions actually use beads prefixes (e.g., \`gt\` not \`gastown\`). Reading both from \`rigs.json\` and deduplicating with \`sort -u\` is clean.

The switch from \`gt rig list --json\` to reading \`rigs.json\` directly avoids the overhead of a full CLI invocation in a cleanup plugin.

LGTM.

seanbearden:
Updated this PR to go further than the original fix. Instead of just patching plugin.md to read `rigs.json`, we've converted the entire plugin from AI-interpreted markdown to a deterministic `run.sh` script (same approach as compactor-dog's conversion).

The root cause was the same both times the dog killed all crews: the AI agent ignored the plugin.md instructions and substituted its own approach. Converting to a shell script eliminates that failure mode entirely — no AI interpretation of destructive cleanup logic.

Tested with `--dry-run` against 47 live sessions: 0 false positives, correctly identified 1 orphaned dog session.

PR #2420, opened 2026-03-06
fix(test): migrate hook_slot_integration_test Label→Labels

DreadPirateRobertz:
## Summary
- `CreateOptions.Label` (string) was renamed to `Labels` ([]string) per #2317
- The 8 callsites in `hook_slot_integration_test.go` were not updated, causing build failures for integration tests

## Test plan
- [x] `go build ./internal/cmd/` succeeds
- [x] `go vet ./internal/cmd/` clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2425, opened 2026-03-06
fix: resolve unparam lint warnings blocking CI on all PRs

DreadPirateRobertz:
## Summary

Two `unparam` lint warnings have been causing golangci-lint to fail on every PR, even when the PR touches unrelated files:

1. **`getAllRigs()`** (`rig_helpers.go:147`): returned a `string` (town root path) that no caller used. Removed the unused return value.
2. **`readDoltSchemaVersion()`** (`wl_schema_evolution.go:64`): returned `(string, error)` but error was always `nil`. Changed to return just `string`.

## Impact

This unblocks CI lint for all open PRs.

## Test plan

- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [ ] CI lint should now pass (the two reported warnings are eliminated)

### DIFF
--- a/internal/cmd/crew_list.go
+++ b/internal/cmd/crew_list.go
@@ -38,7 +38,7 @@ func runCrewList(cmd *cobra.Command, args []string) error {
 
 	var rigs []*rig.Rig
 	if crewListAll {
-		allRigs, _, err := getAllRigs()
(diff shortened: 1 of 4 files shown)


DreadPirateRobertz:
Superseded by #2439 which combines this fix with #2420.

PR #2426, opened 2026-03-06
fix: preload witness env before tmux startup

berdon:
## Summary
  Preload witness session environment variables before the tmux shell starts so witnesses do not inherit the caller's `GT_ROLE`/`GT_SESSION`.

  ## Related Issue
  N/A

  ## Changes
  - Start witness sessions with `NewSessionWithCommandAndEnv(...)` so the initial shell gets the correct witness identity
  - Extract witness env assembly into `buildWitnessSessionEnv(...)`
  - Preserve merged runtime env, role-config env, CLI overrides, and `GT_RUN`
  - Add a regression test covering role-config env and CLI override merging

  ## Testing
  - [x] Unit tests pass (`go test ./internal/witness ./internal/tmux -run 'TestBuildWitness|TestNewSessionWithCommandAndEnv'`)

  ## Checklist

### DIFF
--- a/internal/witness/manager.go
+++ b/internal/witness/manager.go
@@ -187,35 +187,22 @@ func (m *Manager) Start(foreground bool, agentOverride string, envOverrides []st
 	// Generate the GASTA run ID for this witness session.
 	runID := uuid.New().String()
 
+	// Compute the full witness environment before starting tmux so the initial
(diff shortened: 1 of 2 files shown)


berdon:
Similar to https://github.com/steveyegge/beads/pull/2421, this came about from upgrading from gt 0.10.0 to 0.11.0. I observed my witnesses were running as the mayor...

PR #2428, opened 2026-03-06
fix: resolve hook delivery failure for polecats

keaganvanzyl:
## Summary

Polecats couldn't find their hooked work after `hq-l6mm5` removed hook_bead slot writes. Every polecat spawned via `gt sling` would immediately terminate with "Nothing on hook - no work slung."

**Root cause — two compounding bugs:**

1. **`molecule_status.go`**: Read `hook_bead` exclusively from the agent bead's database column (line 393), which is always empty since `hq-l6mm5` removed slot writes. The description field — written atomically at spawn time — is now the authoritative source but was never consulted.

2. **`beads.go`**: The fallback path (`bd list --status=hooked --assignee=<agent>`) was also broken because `bd` v0.59.0 emits `warning:` lines to stdout, which corrupts JSON output. `isJSONBytes()` returns false on `warning: beads.role not configured\n[{...}]`, and `List()` silently returns nil.

**Fixes:**

- **`molecule_status.go`**: When `agentBead.HookBead` column is empty, fall back to `ParseAgentFields(description)` to read `hook_bead` from the description text
- **`beads.go`**: Add `stripStdoutWarnings()` to `run()` and `runWithRouting()` — strips `warning:` lines from bd stdout before returning, preventing JSON parse corruption

## Test plan

- [x] Verified fix with patched binary: `gt mol status --json` returns `has_work: true`
- [x] Full end-to-end: `gt sling` → polecat finds work → implements feature → passes gates → `gt done` → self-cleans
- [x] Existing unit tests pass (`go test ./internal/beads/... ./internal/cmd/...`)
- [ ] Add unit test for `stripStdoutWarnings`
- [ ] Add test for description fallback in molecule_status

### DIFF
--- a/internal/beads/beads.go
+++ b/internal/beads/beads.go
@@ -342,7 +342,13 @@ func (b *Beads) run(args ...string) (_ []byte, retErr error) {
 		return nil, b.wrapError(fmt.Errorf("command produced no output"), stderr.String(), args)
 	}
 
-	return stdout.Bytes(), nil
+	// Strip warning lines from stdout. bd may emit warnings (e.g.,
+	// "warning: beads.role not configured") to stdout, which corrupts
+	// JSON output and causes isJSONBytes to return false, silently
+	// dropping valid results. See: GH #TBD (hook delivery failure).
+	out := stripStdoutWarnings(stdout.Bytes())
+
+	return out, nil
 }
 
 // runWithRouting executes a bd command without setting BEADS_DIR, allowing bd's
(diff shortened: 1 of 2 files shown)


DreadPirateRobertz:

Solid fix for the stdout warning corruption issue. The \`stripStdoutWarnings\` function is well-designed:
- Fast path when no warning present (avoids allocations)
- Only strips lines starting with \`warning:\` (preserves valid output)
- Applied to both \`run\` and \`runWithRouting\` paths

The molecule_status.go change (falling back to description for hook_bead) is a good adaptation to the slot write removal in hq-l6mm5.

LGTM.

steveyegge:
Fix-merged the stripStdoutWarnings portion to main (commit 3164aad7). The molecule_status.go changes are superseded by main's refactored hook discovery. Thank you for identifying the stdout warning corruption bug\!

PR #2430, opened 2026-03-07
feat: add executable run.sh for compactor-dog plugin

seanbearden:
## Summary
- Agent dogs assigned `plugin:compactor-dog` couldn't execute the plugin because `plugin.md` contains bash snippets requiring SQL connections agents don't have
- Created `plugins/compactor-dog/run.sh` — an executable shell script wrapping the flatten compaction algorithm using `dolt sql` CLI
- Discovers production DBs, counts commits, compacts (flattens) databases exceeding threshold, verifies data integrity (row counts before/after), runs `dolt_gc`, and reports results via `bd create` / `gt escalate`

## Test plan
- [x] Shell syntax validated (`bash -n`)
- [x] `--help` flag works
- [x] Dry-run against live Dolt server: correctly discovered 3 default DBs, counted commits, identified `hq` (757 commits) as exceeding 500 threshold
- [x] Auto-discovery mode (`--databases auto`) found all 13 production DBs
- [ ] Full compaction run (deferred — destructive on live data, should be tested in staging)

### DIFF
--- a/plugins/compactor-dog/run.sh
+++ b/plugins/compactor-dog/run.sh
@@ -243,21 +243,40 @@ for entry in "${CANDIDATES[@]}"; do
 
   log "  Pre-flight: $TABLE_COUNT tables recorded"
 
-  # Step 3a.5: Pull from remote before compaction to preserve shared ancestry.
+  # Step 3a.5: Fetch from remote before compaction to check for divergence.
   # Without this, flatten rewrites the commit graph and DoltHub push can never
   # fast-forward again (see gt-mkd1).
+  # NOTE: We use DOLT_FETCH instead of DOLT_PULL because a live Dolt server
+  # always has uncommitted working set changes, making DOLT_PULL fail with
+  # "cannot merge with uncommitted changes".
   HAS_REMOTE=false
   REMOTE_NAME=$(dolt_query "$DB" "SELECT name FROM dolt_remotes LIMIT 1" 2>/dev/null | head -1)
   if [[ -n "$REMOTE_NAME" ]]; then
     HAS_REMOTE=true
-    log "  Remote detected ('$REMOTE_NAME'). Pulling to sync before compaction..."
-    if ! dolt_exec "$DB" "CALL DOLT_PULL('$REMOTE_NAME')"; then
-      log "  ERROR: Pull from remote failed for $DB — skipping compaction to avoid data loss"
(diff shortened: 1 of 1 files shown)


seanbearden:
## Addressed all review feedback

Thanks @DreadPirateRobertz — all issues resolved in d54938b0:

### Blocker: bash 4+ associative arrays
Replaced `declare -A PRE_COUNTS` with a temp file (`PRE_COUNTS_FILE`) using tab-separated `TABLE\tCOUNT` format. Lookups via `grep` + `cut`. All `unset PRE_COUNTS` calls removed. Script now works on bash 3.2 (macOS default).

### SQL injection risk
Added `validate_name()` that rejects any name not matching `^[a-zA-Z0-9_-]+$`. Applied to all DB and table names before SQL interpolation. Also removed `$DB` from the `DOLT_COMMIT` message (was interpolated unsanitized into a SQL string).

### No rollback on integrity failure
Added prominent warning: `"DATABASE LEFT IN COMPACTED STATE — MANUAL INSPECTION REQUIRED"` both to stdout and to the escalation message body. Documented as a known limitation — rollback after a soft-reset + commit is non-trivial and would require a separate design.

### Divergence from plugin.md
Added `--check-only` mode that reports candidates without performing compaction. This matches the plugin.md observe-and-escalate philosophy. Default mode still performs compaction (for agent dogs that have been explicitly assigned the compaction task).

### Minor fixes
- `echo -e` → `printf '%b\n'`
- `--severity medium/high` → `-s MEDIUM/HIGH` (consistent with plugin.md)
- `2>/dev/null` on dolt commands → `2>>"$LOGFILE"` (stderr captured to temp file)
- Added `mkdir`-based lockfile (`/tmp/compactor-dog.lock`) to prevent concurrent runs
- Temp files cleaned up via `trap EXIT`

seanbearden:
Added fix: replaced `DOLT_PULL` with `DOLT_FETCH` + HEAD divergence check for the pre-compaction remote sync.

`DOLT_PULL` always fails on a live Dolt server because the working set has uncommitted changes ("cannot merge with uncommitted changes"). New logic: fetch remote, compare local HEAD vs remote HEAD in `dolt_remote_branches`, only proceed if local is at or ahead of remote.

See commit cc725b058.

steveyegge:
Fix-merged to main. Added validate_hash() for commit hash validation and validate_name() for remote names before SQL interpolation. Original work by @seanbearden.

PR #2433, opened 2026-03-07
fix: surface hooked formula wisps in gt hook

### LINKED ISSUE #2431: gt sling <formula> can report success but leave target gt hook empty
## Bug

Standalone formula slinging can create and hook a wisp bead successfully, but the target agent still sees an empty hook.

This reproduces with a minimal control formula, so it does not appear to depend on any complex workflow logic.

## Control repro

I reproduced this with a simple custom `demo-hello` workflow formula and a fresh polecat target.

Command:

```bash
gt sling demo-hello gastown/polecats/furiosa --no-boot
```

Sling output reported success:

- wisp created
- attached to hook (`status=hooked`)
- nudged to start

Example output shape:

```text
✓ Wisp created: gas-wisp-cpg85
✓ Attached to hook (status=hooked)
▶ Nudged to start
```

## Observed state

In the target polecat workspace:

```bash
gt hook --json
```

returned:

```json
{
  "target": "gastown/polecats/furiosa",
  "agent_bead_id": "gas-gastown-polecat-furiosa",
  "has_work": false,
  "next_action": "Check inbox for work assignments: gt mail inbox"
}
```

But the underlying data disagreed:

### Hooked wisp exists and is assigned

```bash
bd show gas-wisp-cpg85 --json
```

showed:

- `status = hooked`
- `assignee = gastown/polecats/furiosa`

### Hook query by assignee finds it

```bash
bd list --status=hooked --assignee=gastown/polecats/furiosa
```

returned the wisp.

### Agent bead slot is not populated

```bash
bd show gas-gastown-polecat-furiosa --json
```

showed:

- `hook_bead = null`

So the state becomes:

- work bead says hooked + assigned
- `bd list` can find it
- but agent bead hook slot is unset
- therefore `gt hook` reports no work

## Expected behavior

If `gt sling <formula>` reports:

```text
✓ Attached to hook (status=hooked)
```

then the target workspace should immediately see that work via `gt hook` / `gt mol status`.

The agent bead hook slot and the hooked wisp state should agree.

## Why this matters

This breaks the normal workflow contract:

1. sling formula to agent
2. agent runs `gt hook`
3. agent starts work

Instead, the target agent is told the hook is empty even though the wisp is hooked and assigned.

## Notes

This appears to be a generic standalone formula-sling bug, not a problem with a specific workflow:

- reproduced with a minimal custom control formula
- also previously observed with `single-session-tracking-workflow`

There may be a related issue in the hook-slot update path after standalone formula wisp creation.

corymhall:
## Summary
- make `gt hook` / `gt mol status` prefer direct bead tracking and only use legacy hook-slot data as validated fallback
- re-fetch the full hooked bead so formula metadata is available and standalone formula wisps get workflow-oriented guidance instead of `gt mol attach`

Fixes #2431

## Note On Scope
I removed the `bd list --json` plain-text fallback/parser from this PR.

After retesting with `bd v0.59.0 (dev)`, the actual `gt` code path (`bd list --json --flat`) still returns valid JSON in the environment I checked, and the core hook-visibility fix still works without the parser.

So this PR is now focused on the main behavioral fix from #2431:
- direct hooked bead state is authoritative
- the legacy agent `hook_bead` slot is only a fallback
- full hooked-bead re-fetch preserves formula metadata for rendering

## Note On Recent Upstream Movement
Upstream has continued evolving this area, including hook-status logic.

I still believe the underlying problem/solution direction here is correct:
- direct hooked work state should be authoritative
- hook/status should not silently lose formula work or metadata
- users should see the actual hooked bead, not just a stale slot view

If maintainers would prefer a smaller or reshaped integration, I’m still very open to that, but this PR should now be much closer to the actual core fix.

## Verification
- `go test ./internal/beads -run 'TestParseAttachmentFields|TestFormatAttachmentFields' -count=1`
- live repro with a real hooked formula workspace using a binary built from this branch: `gt hook` and `gt hook --json` correctly surfaced the hooked formula bead without relying on the agent bead slot

### DIFF
--- a/internal/beads/beads.go
+++ b/internal/beads/beads.go
@@ -10,6 +10,8 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"regexp"
+	"strconv"
 	"strings"
 	"sync"
 	"time"
@@ -572,10 +574,18 @@ func (b *Beads) List(opts ListOptions) ([]*Issue, error) {
 		return nil, err
 	}
 
-	// bd list --json may return plain text (e.g., "No issues found.") instead
-	// of an empty JSON array when there are no results. Handle gracefully.
+	// bd list --json may return plain text instead of JSON in some configurations.
+	// Handle explicit empty outputs gracefully, but parse issue table rows when
+	// possible so callers don't silently lose hooked work visibility.
 	if len(out) == 0 || !isJSONBytes(out) {
-		return nil, nil
+		text := strings.TrimSpace(string(out))
+		if text == "" || strings.Contains(text, "No issues found.") {
+			return nil, nil
+		}
+		if issues := parsePlainListOutput(text); len(issues) > 0 {
+			return issues, nil
+		}
+		return nil, fmt.Errorf("bd list --json returned non-JSON output: %q", firstLine(text))
 	}
 
 	var issues []*Issue
@@ -601,6 +611,37 @@ func isJSONBytes(b []byte) bool {
 		}
 	}
 	return false
+}
+
+var plainListRowRE = regexp.MustCompile(`^\S+\s+(\S+)\s+\S+\s+P(\d+)\s+\[([^\]]+)\]\s+(.+?)\s*$`)
+
+func parsePlainListOutput(text string) []*Issue {
+	var issues []*Issue
+	for _, line := range strings.Split(text, "\n") {
+		line = strings.TrimSpace(line)
+		if line == "" {
+			continue
+		}
+		m := plainListRowRE.FindStringSubmatch(line)
+		if m == nil {
+			continue
+		}
+		priority, _ := strconv.Atoi(m[2])
+		issues = append(issues, &Issue{
+			ID:       m[1],
+			Priority: priority,
+			Type:     m[3],
+			Title:    m[4],
+		})
+	}
+	return issues
+}
+
+func firstLine(text string) string {
+	if i := strings.IndexByte(text, '\n'); i >= 0 {
+		return text[:i]
+	}
+	return text
 }
 
 // ListByAssignee returns all issues assigned to a specific assignee.
--- a/internal/beads/beads_test.go
+++ b/internal/beads/beads_test.go
@@ -6,6 +6,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"runtime"
 	"strings"
 	"testing"
 )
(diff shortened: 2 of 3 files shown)


DreadPirateRobertz:

Good addition — surfacing hooked formula wisps in `gt hook` output fills a real visibility gap. The `parsePlainListOutput` regex handles the non-JSON bd output gracefully, with appropriate test coverage.

LGTM on the approach.

brandoncorrea:
Pulled your PR in on my machine, rebased main, and built locally. Works great, polecats running again - thank you!

steveyegge:
Closing: the core issue (bd list --json returning plain text instead of JSON) has been fixed in bd itself as of v0.59.0. bd list --json now correctly returns [] for empty results and valid JSON arrays in all cases we tested (--status hooked, --assignee filters, etc.). The plain-text fallback parser is no longer needed. Thanks for surfacing the problem, @corymhall — it was a real bug, just fixed at the source now.

corymhall:
@steveyegge 

I was running `bd` version `v0.59.0` when I ran into this issue.  I took another pass at this locally, and I’m still seeing non-JSON output from `bd list --json` on:

```bash
bd version 0.59.0 (dev)
```
I tested both:

- a non-empty hooked-result query
- an empty-result query

and in both cases bd list --json returned plain-text output rather than JSON.


That said, I agree this is only one part of the PR.

The main fix here is not the fallback parser itself, but:

- making direct bead state (status=hooked + assignee) authoritative for gt hook / gt mol status
- treating the agent bead hook_bead slot as validated legacy fallback
- re-fetching the full hooked bead so formula metadata is available for rendering

So if that helps move things forward, I’m happy to trim the PR:

- remove the bd list --json plain-text fallback/parser portion
- keep the hook-resolution / full-bead-refetch behavior, which is the primary user-visible fix

corymhall:
Follow-up after retesting this more precisely:

I confirmed that my earlier local repro was mixing two things:
- `bd list --json`
- the actual `gt` code path, which uses `bd list --json --flat`

On my local `bd version 0.59.0 (dev)`, `bd list --json` was still giving me non-JSON output in some cases, but `bd list --json --flat` did return valid JSON in the environment I checked.

I then removed the plain-text fallback/parser from this PR and re-ran the real `gt hook` check against a hooked formula workspace. The core hook fix still worked without the parser.

So I’ve slimmed this PR accordingly:
- removed the parser/fallback piece
- kept the main behavioral fix:
  - direct hooked bead state as primary
  - legacy `hook_bead` slot as fallback
  - full hooked-bead re-fetch for formula metadata

So at this point I think we’re aligned that the parser piece can go, while the main `gt hook` fix still stands.

PR #2442, opened 2026-03-07
fix: prevent tmux socket split-brain causing nudge failures

### LINKED ISSUE #2441: fix: prevent tmux socket split-brain by killing stale sessions on wrong socket
## Problem

Sessions can end up on the wrong tmux socket (`default` instead of `gt`), causing `gt nudge` and other session-discovery commands to fail. This happens when:

1. A process (e.g., mayor) is started on the `default` socket (due to older binary, manual tmux, or env mismatch)
2. That process spawns crew/polecat sessions, which inherit the wrong socket
3. Later, sessions are correctly (re)started on the `gt` socket
4. Both sets coexist — `gt nudge` only searches the `gt` socket and can't find sessions on `default`

This is a recurring issue that causes inter-agent communication failures.

## Expected

Before creating a tmux session, `gt` should check for and kill any stale same-named session on other sockets. This prevents split-brain where the same session name exists on multiple sockets.

### 1. Session creation guard (primary fix)

In `tmux.NewSessionWithCommandAndEnv` (or its callers in crew/polecat/dog managers), before creating a session on socket `gt`:
- Check if a session with the same name exists on `default` socket
- If found, kill it (it's stale — the correct socket is `gt`)
- Then proceed with creation on the correct socket

```go
// Before creating session on target socket, kill stale duplicates
func (t *Tmux) killStaleSessionOnOtherSockets(sessionID string) {
    // The "other" socket is always "default" when we target "gt" and vice versa
    otherSocket := "default"
    if t.socketName == "default" {
        otherSocket = "gt"  // or derive from town name
    }
    other := NewTmuxWithSocket(otherSocket)
    if running, _ := other.HasSession(sessionID); running {
        _ = other.KillSessionWithProcesses(sessionID)
    }
}
```

### 2. `gt doctor` socket health check (detection)

Add a doctor check that scans both `default` and the derived town socket for duplicate session names. Flag any duplicates as split-brain.

## Location

- `internal/tmux/tmux.go` — `NewSessionWithCommandAndEnv`
- `internal/crew/manager.go` — `Start()`
- `internal/polecat/session_manager.go` — session creation
- `internal/doctor/` — new split-brain check

seanbearden:
## Summary
- **Session creation guard**: Before creating a session on the town socket (e.g., `gt`), `NewSessionWithCommandAndEnv` now kills any stale same-named session on the `default` socket. This prevents split-brain where `gt nudge` can't find sessions on the wrong socket.
- **Doctor check**: New `socket-split-brain` check detects Gas Town sessions on the `default` socket that should be on the town socket. Reports duplicates (exist on both) and orphans (only on `default`). Fixable via `gt doctor --fix`.

## Root Cause
When a process (e.g., mayor) starts on the `default` tmux socket (due to older binary or env mismatch), it spawns crew/polecat sessions that inherit the wrong socket. Later restarts create correct sessions on the `gt` socket, but the stale `default` sessions persist. `gt nudge` only searches the town socket and fails to find sessions on `default`.

Closes #2441

## Test plan
- [ ] `go build ./...` passes
- [ ] Start a session on `default` socket manually, then start same session via `gt` — verify stale session on `default` is killed
- [ ] `gt doctor` reports split-brain when sessions exist on wrong socket
- [ ] `gt doctor --fix` kills stale sessions on wrong socket

### DIFF
--- a/internal/cmd/doctor.go
+++ b/internal/cmd/doctor.go
@@ -199,6 +199,7 @@ func runDoctor(cmd *cobra.Command, args []string) error {
 	d.Register(doctor.NewDefaultBranchAllRigsCheck())
 	d.Register(doctor.NewIdentityCollisionCheck())
 	d.Register(doctor.NewLinkedPaneCheck())
+	d.Register(doctor.NewSocketSplitBrainCheck())
 	d.Register(doctor.NewThemeCheck())
 	d.Register(doctor.NewCrashReportCheck())
 	d.Register(doctor.NewEnvVarsCheck())
--- a/internal/doctor/tmux_socket_check.go
+++ b/internal/doctor/tmux_socket_check.go
@@ -1,0 +1,132 @@
+package doctor
+
+import (
+	"fmt"
+	"sort"
+
+	"github.com/steveyegge/gastown/internal/session"
+	"github.com/steveyegge/gastown/internal/tmux"
+)
+
+// SocketSplitBrainCheck detects tmux sessions that exist on both the town
+// socket (e.g., "gt") and the "default" socket. This split-brain causes
+// gt nudge and other session-discovery commands to fail because they only
+// search the town socket.
+type SocketSplitBrainCheck struct {
+	FixableCheck
+	staleSessions []string // Sessions on "default" that also exist on town socket
+}
+
+// NewSocketSplitBrainCheck creates a new socket split-brain check.
+func NewSocketSplitBrainCheck() *SocketSplitBrainCheck {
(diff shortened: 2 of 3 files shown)


DreadPirateRobertz:

Good doctor check. The \`SocketSplitBrainCheck\` correctly identifies sessions on the "default" socket that duplicate town socket sessions. The \`FixableCheck\` base allows auto-repair.

Clean implementation — detecting Gas Town sessions via \`session.IsGasTownSession\` avoids killing unrelated tmux sessions.

LGTM.

PR #2449, opened 2026-03-07
feat: implement trust tier escalation engine for Wasteland

DreadPirateRobertz:
## Summary

- Trust tier escalation engine: Drifter → Registered → Contributor → War Chief
- Multi-criteria evaluation: completions, stamps, quality, distinct validators, time-in-tier
- Anti-collusion gate: stamps from a single validator can't earn promotions
- Pure `EvaluateEscalation()` function (no DB needed) for testability
- `LoadRigTrustProfile()` for querying Dolt reputation data
- `DefaultTierRequirements()` tuned for 50-200 rig wastelands

## Test plan

- [x] `TestTierString` — all tiers have names, unknown tiers handled
- [x] `TestDefaultTierRequirements` — ascending order, War Chief strictest
- [x] `TestEscalation_RegisteredToContributor` — happy path promotion
- [x] `TestEscalation_InsufficientCompletions` — low completions blocks
- [x] `TestEscalation_SingleValidatorBlocked` — anti-collusion gate
- [x] `TestEscalation_TimeGateBlocked` — rapid farming prevented
- [x] `TestEscalation_ContributorToWarChief` — quality requirement verified
- [x] `TestEscalation_LowQualityBlocksWarChief` — grinding doesn't work
- [x] `TestEscalation_MaxTierNoPromotion` — War Chief ceiling
- [x] `TestEscalation_DrifterNeedsNoRequirements` — join = Registered
- [x] `TestEscalation_MultipleFailures` — all unmet criteria reported

Closes w-wl-003.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2448, opened 2026-03-07
feat: implement Spider Protocol fraud detection for wasteland stamps

DreadPirateRobertz:
## Summary

- Implements the Spider Protocol: SQL-based fraud detection for the Wasteland reputation system
- Four detectors: collusion (mutual stamp inflation), rubber-stamping (identical valence), confidence inflation (always-max scores), self-loops (tight reciprocal pairs)
- Operates exclusively on disclosed stamps data — no private information needed
- Configurable thresholds via `SpiderConfig` for tuning to wasteland size
- `RunSpiderDetection()` entry point returns composable `FraudSignal` values for downstream scoring
- GoDoc on every exported function and type, with WHY comments on design decisions

## Test plan

- [x] `TestDefaultSpiderConfig` — validates config bounds
- [x] `TestQueryGeneration` — verifies all 4 query shapes contain expected SQL clauses
- [x] `TestExtractRigs` — unit tests for rig handle extraction from CSV rows
- [x] `TestFraudSignalKindConstants` — ensures signal kinds are distinct and non-empty
- [x] `TestCollusionQueryThresholds` — verifies config values propagate into queries
- [x] Full build passes (`go build ./...`)
- [x] Includes CI fixes (unparam, Label→Labels, sling test isolation)

Closes w-hop-003.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2445, opened 2026-03-07
feat: add schema evolution auto-migration to gt wl sync

DreadPirateRobertz:
## Summary

- `checkSchemaEvolution` now returns a structured `SchemaEvolutionResult` with version info and change kind
- New `detectSchemaChanges()` previews table/column diffs (via `dolt diff --schema`) before merge — users see exactly what's changing
- New `verifyPostMergeSchema()` catches `_meta` merge conflicts where the version row didn't update correctly
- MINOR bumps: auto-applied with schema diff preview shown to user
- MAJOR bumps: gated behind `--upgrade` flag with clear error and resolution instructions
- Edge cases handled: pre-versioned forks (no `_meta`), missing schema_version row, post-merge version mismatches

## Test plan

- [x] All existing `TestParseSchemaVersion` and `TestClassifySchemaChange` tests pass
- [x] New `TestCheckSchemaEvolution_MajorBlocksWithoutUpgrade` — MAJOR bump blocked without `--upgrade`
- [x] New `TestCheckSchemaEvolution_MinorAutoApplies` — MINOR bump proceeds with correct result
- [x] New `TestCheckSchemaEvolution_NoMeta` — pre-versioned fork passes through cleanly
- [x] New `TestVerifyPostMergeSchema_Mismatch` — detects version mismatch after merge
- [x] New `TestVerifyPostMergeSchema_Match` — passes when versions agree
- [x] New `TestVerifyPostMergeSchema_EmptyExpected` — no-op for pre-versioned forks
- [x] New `TestDetectSchemaChanges_NoDolt` — graceful degradation when dolt unavailable
- [x] `go vet` clean

Closes w-wl-004.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2444, opened 2026-03-07
fix: resolve CI blockers — unparam lint, Label→Labels, sling test isolation

DreadPirateRobertz:
## Summary
Combined fix for 3 CI blockers that affect all PRs:

1. **unparam lint** (rig_helpers.go, wl_schema_evolution.go): Remove unused return values flagged by golangci-lint v2
2. **Label→Labels** (hook_slot_integration_test.go): Migrate to renamed struct field in beads.CreateOptions
3. **sling test isolation** (beads.go, sling_test.go): Fix `sync.Once` cache poisoning in `BdSupportsAllowStale()` and overly strict `BD_DOLT_AUTO_COMMIT` assertion in `TestSlingSetsDoltAutoCommitOff`

Fixes #2443
Supersedes: #2439, #2425, #2420

## Test plan
- [x] `TestVerifyBeadExistsAllowStale` passes locally
- [x] `TestSlingSetsDoltAutoCommitOff` passes locally
- [x] Unit tests pass (`go test ./internal/cmd/`)
- [ ] Full CI passes (lint, unit, integration)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2452, opened 2026-03-07
fix: increase checkSessionAfterCreate timeout for exec-env pattern on CI

### LINKED ISSUE #2451: fix: TestNewSessionWithCommand_ExecEnvBadBinary fails on CI (tmux env difference)
## Problem

TestNewSessionWithCommand_ExecEnvBadBinary in internal/tmux/session_creation_test.go:52 fails consistently on CI but passes locally.

The test creates a tmux session with:
```
exec env GT_TEST=1 GT_ROLE=test /nonexistent/claude-code --settings /tmp
```

It expects NewSessionWithCommand to return an error, but on CI tmux, tmux new-session -d succeeds — the session is created and the command failure happens inside the session shell, not at session creation time.

## Impact

This blocks Test CI on some PRs. Currently failing on PR #2445 and #2449 but passing on #2444, #2448, #2422, #2419 — suggesting timing sensitivity.

## Fix Options

1. After session creation, wait briefly and check if the session is still alive
2. Skip when not running under tmux locally
3. Validate binary path before calling tmux

DreadPirateRobertz:
## Summary

- Adds a third health check at 750ms in `checkSessionAfterCreate` specifically for commands using the `exec env` pattern
- The exec-env startup pattern adds shell parsing overhead that can exceed 250ms on heavily loaded CI runners, causing `TestNewSessionWithCommand_ExecEnvBadBinary` to intermittently fail
- Long-lived sessions (Claude, shell) are unaffected — they're still alive at 750ms

## Context

This test passes on some PRs (#2444, #2448) but fails on others (#2445, #2449), confirming it's timing-sensitive rather than a code regression.

Fixes #2451

## Test plan

- [ ] `TestNewSessionWithCommand_ExecEnvBadBinary` passes on CI
- [ ] `TestNewSessionWithCommand_ExecEnvSuccess` still passes (exec with valid binary)
- [ ] No regression in other session creation tests

### DIFF
--- a/internal/tmux/tmux.go
+++ b/internal/tmux/tmux.go
@@ -387,12 +387,12 @@ func (t *Tmux) checkSessionAfterCreate(name, command string) error {
 		return err
 	}
 
-	// Third check at 750ms: the exec-env pattern (exec env VAR=val binary)
+	// Third check at 1.25s: the exec-env pattern (exec env VAR=val binary)
 	// adds shell parsing overhead that can exceed 250ms on heavily loaded CI
 	// runners. This catches those late failures without impacting normal
 	// session creation (long-lived processes are still alive at this point).
 	if strings.Contains(command, "exec env") || strings.Contains(command, "exec ") {
-		time.Sleep(500 * time.Millisecond)
+		time.Sleep(1000 * time.Millisecond)
 		if dead, err := checkPaneDead(); dead {
 			return err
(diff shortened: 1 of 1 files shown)


steveyegge:
Thanks @DreadPirateRobertz — the changes in this PR have been superseded by main. Closing. 🤖

PR #2454, opened 2026-03-07
fix: propagate Dolt port to agent tmux sessions via AgentEnv

### LINKED ISSUE #2412: bd auto-starts rogue Dolt servers: agent tmux sessions missing BEADS_DOLT_PORT

DreadPirateRobertz:
## Summary

- Adds `GT_DOLT_PORT` and `BEADS_DOLT_PORT` propagation to `AgentEnv()` in `env.go`
- When the daemon has `GT_DOLT_PORT` set, agent sessions now inherit the correct port
- Prevents bd from auto-starting rogue Dolt servers in agent tmux sessions

## Root cause

`AgentEnv()` builds the env map that gets set via `tmux.SetEnvironment()` for new agent sessions. It was missing the Dolt port variables, so agents couldn't find the central Dolt server and bd fell back to auto-starting its own instance.

## Edge cases

- **Unset GT_DOLT_PORT**: No port vars are added (default 3307 from metadata.json still works)
- **Non-standard ports**: Both variables set to the same value, matching `translateDoltPort()` semantics
- **Test isolation**: `NewIsolatedWithPort()` in beads.go already handles this path separately

Fixes #2412

## Test plan

- [x] `TestAgentEnv_PropagatesDoltPort` — verifies both vars set when GT_DOLT_PORT present
- [x] `TestAgentEnv_NoDoltPortWhenUnset` — verifies no vars when GT_DOLT_PORT empty
- [ ] CI lint/test passes

### DIFF
--- a/internal/tmux/tmux.go
+++ b/internal/tmux/tmux.go
@@ -388,8 +388,6 @@ func (t *Tmux) checkSessionAfterCreate(name, command string) error {
 		return err
 	}
 
-<<<<<<< HEAD
-=======
 	// Third check at 1.25s: the exec-env pattern (exec env VAR=val binary)
 	// adds shell parsing overhead that can exceed 250ms on heavily loaded CI
 	// runners. This catches those late failures without impacting normal
@@ -401,7 +399,6 @@ func (t *Tmux) checkSessionAfterCreate(name, command string) error {
 		}
 	}
 
->>>>>>> 0d43b8ea (fix: increase exec-env health check to 1.25s for CI reliability)
 	// Pane is alive — restore default (no need to keep dead sessions around)
 	_, _ = t.run("set-option", "-t", name, "remain-on-exit", "off")
 	return nil


steveyegge:
Thanks @DreadPirateRobertz — Dolt port propagation landed via commit 641b0be8. Closing as superseded. 🤖

PR #2457, opened 2026-03-07
fix: prevent bd from spawning orphan embedded Dolt servers (gt-t2zf)

seanbearden:
## Summary

- `gt dolt start` now syncs `dolt-server.port` to all rig `.beads/` dirs after confirming the server is running
- `gt start` now starts Dolt synchronously before launching agents, fixing a boot race where agents would call `bd` before Dolt was ready
- New `dolt-port-files` doctor check detects and auto-fixes port file drift

## Root cause

Two issues combined to cause orphan embedded Dolt servers:

1. **Boot race in `start.go`**: Dolt server and agents started in parallel goroutines. Agents called `bd` before Dolt was ready, and `bd` auto-spawned its own embedded server on a derived port with an empty database.

2. **Missing port file sync**: `gt dolt start` never wrote `dolt-server.port` files to rig `.beads/` dirs. When these files were missing or stale, `bd`'s port resolution fell back to `DerivePort`, spawning orphan servers.

## Changes

**`internal/doltserver/doltserver.go`**:
- `SyncPortFiles(townRoot, port)` — writes shared server port to all rig `.beads/dolt-server.port` files
- `CheckPortFiles(townRoot, expectedPort)` — detects port file drift
- `Start()` calls `SyncPortFiles` on both fresh start and already-running paths

**`internal/cmd/start.go`**:
- Split startup into Phase 1 (Dolt sync + metadata + port files) and Phase 2 (agents parallel)
- Matches the sequencing in `up.go:295-306` which already gates agent startup on `waitForDoltReady()`

**`internal/doctor/dolt_port_file_check.go`**:
- New `dolt-port-files` check with auto-fix via `gt doctor --fix`

## Test plan

- [x] `TestSyncPortFiles_WritesToAllRigBeadsDirs` — verifies port files written to all rig dirs
- [x] `TestSyncPortFiles_SkipsMissingBeadsDirs` — verifies no error for missing dirs
- [x] `TestSyncPortFiles_NoRigsJSON` — verifies graceful handling without rigs.json
- [x] `TestCheckPortFiles_DetectsDrift` — verifies drift detection
- [x] `TestCheckPortFiles_NoDriftWhenCorrect` — verifies no false positives
- [x] `TestSyncPortFiles_OverwritesStalePort` — verifies stale ports get overwritten
- [x] `TestDoltPortFileCheck_shortPath` — doctor check helper
- [x] `TestDoltPortFileCheck_Run_NoRigsJSON` — doctor check skips gracefully
- [x] All existing doltserver and doctor tests pass

### DIFF
--- a/internal/doltserver/doltserver.go
+++ b/internal/doltserver/doltserver.go
@@ -1317,7 +1317,7 @@ func SyncPortFiles(townRoot string, port int) error {
 	// Town-level .beads/ (hq)
 	townBeadsDir := filepath.Join(townRoot, ".beads")
 	if _, err := os.Stat(townBeadsDir); err == nil {
-		_ = os.WriteFile(filepath.Join(townBeadsDir, "dolt-server.port"), portStr, 0644)
+		writePortFileIfChanged(filepath.Join(townBeadsDir, "dolt-server.port"), portStr)
 	}
 
 	// Read rigs.json for all registered rigs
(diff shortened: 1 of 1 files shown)


DreadPirateRobertz:

Excellent architectural fix. The two-phase startup (Dolt server first, then agents) eliminates the race condition that caused orphan embedded Dolt servers. Key improvements:

1. Sequential Dolt start + metadata sync before any agents launch
2. \`SyncPortFiles\` ensures port files are correct before agents read them
3. New \`DoltPortFileCheck\` in doctor for ongoing validation

The removal of the parallel WaitGroup for Dolt startup is the right call — the ~500ms Dolt start time is negligible vs agent startup, and the correctness guarantee is worth it.

One question: is \`SyncPortFiles\` idempotent? If called when port files already match, does it avoid unnecessary writes?

LGTM — this should prevent the entire class of rogue Dolt server bugs.

seanbearden:
Good catch — `SyncPortFiles` was not idempotent. It unconditionally called `os.WriteFile` on every port file regardless of whether the content already matched.

Fixed in a446197: added `writePortFileIfChanged` helper that reads the existing file first and skips the write if `bytes.Equal` confirms the port already matches. All existing tests pass.

seanbearden:
Closing in favor of a clean PR. Previous branch had 57 commits of accumulated fork history; replacement has only the 3 relevant fix commits.

PR #2458, opened 2026-03-07
fix(nudge): recover queued nudges when agent becomes idle after WaitForIdle timeout

seanbearden:
## Summary

- Fixes first nudge after `gt sling` never being delivered to crew despite session being idle
- When `WaitForIdle` times out (agent busy during 15s poll) and falls back to queue, the queued nudge is lost if the agent becomes idle shortly after — the `UserPromptSubmit` hook only fires on input, so an idle agent never drains its queue
- Adds post-queue idle recovery: after successful queue write, 300ms settle + `IsIdle()` check; if idle, drains queue (prevents double delivery) and delivers directly via `NudgeSession`

## Test plan

- [x] `TestPostQueueSettleDelay` validates the settle delay is in a reasonable range (100ms-1s)
- [x] All existing nudge tests pass (`TestNudge*`, `TestResolveNudgePattern`, `TestSessionNameToAddress`, etc.)
- [x] Pre-existing `TestSlingSetsDoltAutoCommitOff` failure is unrelated (fails on clean upstream/main too)

### DIFF
--- a/internal/cmd/nudge.go
+++ b/internal/cmd/nudge.go
@@ -188,9 +188,13 @@ func deliverNudge(t *tmux.Tmux, sessionName, message, sender string) error {
 		// Brief settle, then check: if idle, drain queue and deliver directly.
 		time.Sleep(postQueueSettleDelay)
 		if t.IsIdle(sessionName) {
-			// Drain the queue to prevent double delivery, then deliver directly.
-			_, _ = nudge.Drain(townRoot, sessionName)
-			return t.NudgeSession(sessionName, prefixedMessage)
+			// Drain atomically claims queued entries (rename-based). If another
+			// process (e.g., the agent's hook) raced and drained first, we get
+			// an empty slice and skip delivery to avoid duplicates.
+			drained, _ := nudge.Drain(townRoot, sessionName)
+			if len(drained) > 0 {
+				return t.NudgeSession(sessionName, prefixedMessage)
+			}
 		}
 		return nil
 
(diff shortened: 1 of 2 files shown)


DreadPirateRobertz:

This addresses a real race condition in nudge delivery. The post-queue idle recovery pattern is sound: brief settle, check if idle, then drain and deliver directly.

The 300ms settle delay is reasonable — long enough to catch the idle transition but short enough to not add significant latency.

One concern: the \`Drain + NudgeSession\` path could duplicate delivery if another process also drains the queue concurrently. Consider whether \`Drain\` should return whether it actually drained anything, and only deliver if it did.

The test (\`TestPostQueueSettleDelay\`) only validates the constant range, not the actual behavior. Would be stronger with a test that verifies the drain-and-deliver path fires when IsIdle returns true.

LGTM on the approach, could use a stronger integration test.

seanbearden:
Thanks for the review! Both points addressed in 6421fd90:

**1. Duplicate delivery guard**: `Drain` already uses atomic rename-based claiming (`queue.go:196-209`) — only one drainer wins each entry. But you're right that the old code ignored the return value and always called `NudgeSession`. Now it checks `len(drained) > 0` and skips delivery when the queue was already drained by a racing process (e.g., the agent's `UserPromptSubmit` hook).

**2. Behavioral test**: Added `TestPostQueueIdleRecovery_SkipsDeliveryWhenDrainEmpty` — enqueues a nudge, drains it (simulating a racing hook), then verifies the second drain returns empty. This exercises the actual drain-and-skip path, not just the constant range.

seanbearden:
Superseded by #2497 which includes the idle watcher plus hook-based delivery (trigger instead of raw tmux send-keys).

PR #2459, opened 2026-03-07
fix: auto-purge closed ephemeral beads on session end

popandpeek:
## Summary

Polecat and witness sessions create mol-polecat-work steps, mol-witness-patrol cycles, and other wisps that get closed during normal operation but are never deleted. These accumulate across sessions (observed 419+126 stale beads in production) and pollute `bd ready/list` output.

- Add `purgeClosedEphemeralBeads()` helper that calls `bd purge --force --quiet`
- Call it from `gt done` after closing hooked bead, before agent state transition
- Call it from `gt polecat nuke` after molecule cleanup

Best-effort: purge failures are logged as warnings but don't block completion.

Fixes: hq-6161m

## Test plan

- [x] `go build ./internal/cmd/` passes
- [x] `go vet ./internal/cmd/` passes
- [x] `go test -run TestDone ./internal/cmd/` passes
- [ ] Manual verification: run `bd purge --dry-run` before/after `gt done` to confirm accumulation stops

### DIFF
--- a/internal/cmd/done.go
+++ b/internal/cmd/done.go
@@ -1260,6 +1260,12 @@ func updateAgentStateOnDone(cwd, townRoot, exitType, issueID string) {
 
 	// No ClearHookBead call needed — agent bead hook slot is no longer maintained (hq-l6mm5).
 
+	// Purge closed ephemeral beads (wisps) accumulated during this and prior sessions.
+	// Without this, closed wisps from mol-polecat-work steps, mol-witness-patrol cycles,
+	// etc. accumulate across sessions and pollute bd ready/list output (hq-6161m).
+	// Best-effort: failures are non-fatal since the work is already done.
+	purgeClosedEphemeralBeads(bd)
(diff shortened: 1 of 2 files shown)


DreadPirateRobertz:

Good cleanup. Adding `purgeClosedEphemeralBeads()` to both `gt done` and `nukePolecatFull` paths covers the two main session-end flows. The best-effort error handling (warn + continue) is appropriate since purge failure shouldn't block completion.

One note: calling `bd purge --force --quiet` shells out to bd — worth confirming this works when bd isn't installed (the `bd` variable here is a `*beads.Beads` struct, so `bd.Run` should handle that gracefully).

LGTM overall.

DreadPirateRobertz (approved):
LGTM. Clean, focused fix. Best-effort purge on session end prevents wisp accumulation without blocking gt done. Good comments. — gastown/crew/zhora (PM)

PR #2462, opened 2026-03-07
fix: DatabasePrefixCheck skips rigs sharing town DB

### LINKED ISSUE #2409: DatabasePrefixCheck corrupts shared database prefix when rigs redirect to town DB

DreadPirateRobertz:
## Summary
- Fixes `DatabasePrefixCheck` corrupting shared prefix when rigs redirect to the town root's beads DB
- When a rig's `.beads/redirect` resolves to the same directory as the town root's `.beads`, the check now skips that rig instead of querying the shared DB and reporting a false mismatch
- Adds test covering the shared-DB redirect scenario

Fixes #2409 / gt-m6ol

## Test plan
- [x] Existing `TestDatabasePrefixCheck_*` tests pass
- [x] New `TestDatabasePrefixCheck_SkipsSharedDB` verifies redirect detection
- [ ] Manual: run `bd doctor` on a town with crew rigs using redirects — no false prefix warnings

### DIFF
--- a/internal/doctor/beads_check.go
+++ b/internal/doctor/beads_check.go
@@ -487,6 +487,9 @@ func (c *DatabasePrefixCheck) Run(ctx *CheckContext) *CheckResult {
 		}
 	}
 
+	// Resolve the town root beads dir so we can detect rigs that redirect to it.
+	townRootBeadsDir, _ := filepath.Abs(beads.ResolveBeadsDir(ctx.TownRoot))
+
 	var problems []string
 
 	for _, route := range routes {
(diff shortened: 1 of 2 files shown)


steveyegge:

PR #2463, opened 2026-03-07
fix: nudge delivery input-aware safety and post-queue idle recovery

### LINKED ISSUE #1216: Reliable nudge delivery: design analysis and unresolved input clearing problem
## Summary

Nudges (cross-agent messages) need to be 100% reliable with low latency, but they share the same input mechanism as the human overseer - the command line text input field. This creates a fundamental conflict that we've been unable to fully resolve.

## The Problem

When a nudge arrives while the overseer is typing, the nudge text gets appended to their partial input, resulting in garbled instructions:

```
User typing: "Please fix the bug in auth"
Nudge arrives: "[from crew/holden] Status update"
Result: "Please fix the bug in auth[from crew/holden] Status update"
```

The agent receives garbage. The nudge is lost. The user's input is corrupted.

### PR #1207 - NudgeManager
https://github.com/steveyegge/gastown/pull/1207

Initial approach with file-based queue, deduplication, rate limiting, and escalation. Did not address the input collision problem - just delivered nudges immediately.

### PR #1212 - Clear/Inject/Verify Protocol with Myers Diff
https://github.com/steveyegge/gastown/pull/1212

More sophisticated approach:

1. **Capture BEFORE** - Full terminal scrollback to guarantee we never lose user input no matter how much they've typed
2. **Clear input** - Send Ctrl-C to clear any partial user input
3. **Inject nudge** - Send the nudge text (no Enter yet)
4. **Capture AFTER** - Last ~50 lines (nudges are small/limited in size, so we don't need full scrollback here)
5. **Diff & Verify** - Use Myers diff algorithm to:
   - Identify what the user had typed (original input)
   - Detect "gap typing" (user typed between Ctrl-C and nudge injection)
   - Detect "after-nudge typing" (user typed after nudge injection)
6. **Clean delivery check** - If no gap/after typing, send Enter
7. **Reject & retry** - If typing detected, return error and let daemon retry

#### What Worked

- **Myers diff** accurately identifies original input, gap typing, and after-nudge typing without hardcoding any assumptions about the input field, prompt character, etc. It gracefully handles other changes outside the text field (e.g., agent continuing to output, mode status changing below input field)
- **Atomic space+Ctrl-C** prevents double-Ctrl-C exit: sending `" "` then `"C-c"` in one tmux call breaks any "two Ctrl-C = exit" detection window
- **Position-based nudge finding** handles spurious character matches by absorbing small Equal operations (< 4 bytes) into diff hunks
- **Reliable detection** of whether nudge was delivered cleanly

#### What Failed

**Input preservation is impossible without a second Ctrl-C.**

Once we send Ctrl-C to clear the input, the user's original text is gone from the terminal. We capture it in the diff, but to restore it:

- If the input field is empty (after nudge submitted and agent finishes): We can `SendKeysLiteral(original)` - no problem
- If the input field has content (garbled nudge+typing, or we want to abort): We need Ctrl-C to clear first

But **sending a second Ctrl-C is dangerous**:
- Two Ctrl-C's in quick succession exits Claude Code
- Even with our space prefix, timing windows can cause exit
- This was discovered the hard way during testing

**The "Interrupted" problem:**

When Ctrl-C is sent while Claude is thinking, it triggers an "Interrupted" message and stops the agent's current work. This is catastrophic in Gas Town - agents stop working mid-task.

### Ctrl-U (Unix line kill)
- Only clears current line, not multi-line input
- Would need interleaved Ctrl-U + Backspace for multi-line
- Behavior is readline-specific, not portable

### Vim Mode Commands
- `Escape` + `cc` clears line without Ctrl-C
- But requires vim mode to be enabled
- Cannot programmatically toggle vim mode (only `/vim` command)
- Could detect vim mode via `-- INSERT --` / `-- NORMAL --` in terminal, but only helps if user already has it enabled

### Claude Code Stash (Ctrl+S)
- Claude Code has a "stash" feature that saves and clears input
- Would be perfect: stash → inject nudge → Enter → unstash
- But Ctrl+S conflicts with terminal XON/XOFF flow control on many systems
- Claude Code specific, not portable

## Current State

The fundamental problem remains unsolved:

> **We cannot reliably clear user input without either:**
> 1. **Interrupting the agent** (Ctrl-C while thinking)
> 2. **Risking exit** (double Ctrl-C)
> 3. **Losing user input** (can't restore without clearing again)

## What Would Help

Ideas we haven't fully explored:

1. **Agent-side API** - A way for agents to receive nudges through a side channel, not the input field
2. **Input field query** - A way to check if the input field is empty before injecting
3. **Safe clear command** - A universal "clear input without interrupt" that all CLI agents support
4. **Stash/unstash standardization** - If all agents had Claude Code's Ctrl+S stash (with XON/XOFF disabled)

## Design Document

The full design document with algorithm details is at:
`docs/design/nudge-delivery.md`

Key sections:
- Protocol flow diagram
- Myers diff algorithm explanation
- Position-based nudge finding
- Edge case handling table
- Double Ctrl-C prevention

## Related Issues

- gt-lmm1z: NudgeManager implementation
- gt-9og: Original nudge delivery bug report
- gt-7c1xj: Nudge priority levels (deferred)

DreadPirateRobertz:
## Summary
- Adds `IsInputEmpty()` to tmux package — verifies the agent prompt line has no user-typed content before delivering a nudge directly
- Guards direct delivery in wait-idle mode with `IsInputEmpty` check to prevent corrupting user input
- Adds post-queue idle recovery: after falling back to queue delivery, polls briefly to see if the agent becomes idle, then delivers directly instead of leaving the nudge stranded in the queue
- Adds `postQueueSettleDelay` (300ms) to prevent immediate re-delivery race after queueing

## Test plan
- [x] All 36 nudge cmd tests pass
- [x] All 15 nudge queue tests pass
- [x] New `TestPostQueueSettleDelay` verifies settle delay behavior
- [x] New `TestIsInputEmpty_Unit` verifies input detection logic
- [x] `go vet ./...` clean

Fixes #1216

### DIFF
--- a/internal/cmd/nudge.go
+++ b/internal/cmd/nudge.go
@@ -194,12 +194,20 @@ func deliverNudge(t *tmux.Tmux, sessionName, message, sender string) error {
 		// never drain its queue autonomously — the UserPromptSubmit hook
 		// only fires on input, so queued nudges for idle agents are lost.
 		// Brief settle, then check: if idle with empty input, drain queue
-		// and deliver directly.
+		// and deliver ALL queued messages directly.
 		time.Sleep(postQueueSettleDelay)
-		if t.IsInputEmpty(sessionName) {
-			// Drain the queue to prevent double delivery, then deliver directly.
-			_, _ = nudge.Drain(townRoot, sessionName)
-			return t.NudgeSession(sessionName, prefixedMessage)
+		if t.IsIdle(sessionName) && t.IsInputEmpty(sessionName) {
+			// Drain ALL queued nudges (not just ours) to prevent both
+			// double-delivery and silent loss of other queued messages.
+			drained, drainErr := nudge.Drain(townRoot, sessionName)
+			if drainErr != nil || len(drained) == 0 {
+				// Drain failed or empty (our message was already picked up).
+				// Fall back to direct delivery of just our message.
+				return t.NudgeSession(sessionName, prefixedMessage)
+			}
+			// Format all drained nudges (including ours) and deliver as one.
+			formatted := nudge.FormatForInjection(drained)
+			return t.NudgeSession(sessionName, formatted)
 		}
 		return nil
 
--- a/internal/tmux/tmux.go
+++ b/internal/tmux/tmux.go
@@ -1341,29 +1341,37 @@ func (t *Tmux) NudgeSession(session, message string) error {
 		time.Sleep(50 * time.Millisecond)
 	}
 
-	// 2. Sanitize control characters that corrupt delivery
+	// 2. Clear any pending input before sending. This closes the TOCTOU race
+	//    between deliverNudge's IsInputEmpty check and the actual send-keys:
+	//    the user (or another nudge) may have added text to the input field
+	//    in the gap. Ctrl-U clears the current line in Claude Code's TUI
+	//    (same approach used by SendKeysReplace). It's a no-op if empty.
+	_, _ = t.run("send-keys", "-t", target, "C-u")
+	time.Sleep(50 * time.Millisecond)
+
+	// 3. Sanitize control characters that corrupt delivery
 	sanitized := sanitizeNudgeMessage(message)
 
-	// 3. Send text via send-keys -l. Messages > 512 bytes are chunked
+	// 4. Send text via send-keys -l. Messages > 512 bytes are chunked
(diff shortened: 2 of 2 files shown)


steveyegge:
The underlying code has been significantly reworked on main since this PR was opened. Cherry-pick conflicts. If the fix is still needed, please open a fresh PR from current main.

PR #2466, opened 2026-03-07
feat: add gt plugin sync to auto-deploy plugins after build

seanbearden:
## Summary
- Add `gt plugin sync` command that copies plugins from source repo to town-level runtime directory (`~/gt/plugins/`)
- Hook into `make install` to auto-sync plugins after binary install — prevents drift when plugin fixes merge but runtime dirs are stale
- Add `gt doctor` `patrol-plugin-drift` check that warns on stale plugins, fixable via `--fix`

Closes hq-o9gna

## What caused the outages
Plugins (compactor-dog, session-hygiene, stuck-agent-dog) were merged to git but never deployed to runtime. Three separate git clones (`crew/den/`, `gastown/`, `mayor/`) each sat at different commits. The scanner reads from `~/gt/<rig>/plugins/` and `~/gt/plugins/`, but neither auto-synced.

## Test plan
- [x] 11 unit tests: copy, skip-up-to-date, update-changed, preserve-permissions, clean-removes-extra, no-clean-keeps-extra, ignore-non-plugin-dirs, drift detection (4 cases)
- [x] `go vet` clean
- [x] End-to-end: `gt plugin sync --dry-run` detects 7 missing plugins, `gt plugin sync` copies them, re-run shows "up to date"
- [x] Pre-existing test failure (`TestSlingSetsDoltAutoCommitOff`) is unrelated

### DIFF
--- a/internal/plugin/sync.go
+++ b/internal/plugin/sync.go
@@ -1,6 +1,7 @@
 package plugin
 
 import (
+	"bufio"
 	"crypto/sha256"
 	"fmt"
 	"io"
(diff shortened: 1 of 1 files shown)


seanbearden:
Thanks for the thorough review @DreadPirateRobertz! Addressed both actionable items in 7478fd2b:

1. **Atomic `copyDir`**: Now copies to a temp directory (`.plugin-sync-*` in the same parent dir) and swaps via `os.Rename`. If the copy fails mid-way, the old destination is untouched and the temp dir gets cleaned up by `defer`.

2. **Precise `go.mod` matching**: Replaced `strings.Contains(data, "gastown")` with a line-by-line scanner that checks only the `module` directive. Matches `module .../gastown` or `module gastown` — won't false-positive on comments or dependency names.

Items 3 and 4 noted — agree they're fine as-is for the current plugin scale, and the `|| true` in Makefile is intentional.

PR #2467, opened 2026-03-07
fix: check rig-level metadata before town-level beads redirect

seanbearden:
## Summary
- Fix beads redirect to check rig-level `metadata.json` for `dolt_database` BEFORE falling back to town-level `.beads`
- Rigs with their own database (e.g., laneassist with `lc-` prefix) now correctly redirect crew to rig-level `.beads`
- Fixes regression from b99b7568 which unconditionally preferred town-level, breaking non-gastown rig routing

## What broke
The previous redirect fix (b99b7568, hq-x1ucu) always sent crew worktrees to town-level `.beads` (hq- prefix) when it had a DB. This meant laneassist crew saw `hq-` beads instead of `lc-` beads.

## Fix
Check `<rig>/.beads/metadata.json` for `dolt_database` first. If present, the rig has its own DB and crew must redirect there. Only fall back to town-level when the rig has no own database.

## Test plan
- [x] New test: `rig_with_own_DB_redirects_to_rig-level_beads` — rig with metadata.json gets rig-level redirect
- [x] Renamed test: `rig_without_own_DB_redirects_to_town-level_beads` — rig without metadata gets town-level
- [x] All 14 redirect subtests pass
- [x] `go vet` clean

### DIFF
--- a/internal/beads/beads_redirect.go
+++ b/internal/beads/beads_redirect.go
@@ -2,6 +2,7 @@
 package beads
 
 import (
+	"encoding/json"
 	"fmt"
 	"os"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


seanbearden:
Thanks for the thorough review!

Re: observation #2 — agreed, `rigHasOwnDB` hitting the filesystem each call is acceptable here since redirect resolution is a one-shot per session. If we ever need it on a hot path, we can add a simple cache, but YAGNI for now.

DreadPirateRobertz (approved):
LGTM. Well-structured fix for rig-level vs town-level beads redirect routing.

Key changes:
- New `rigHasOwnDB` helper checks metadata.json for dolt_database field
- Redirect priority: rig-level (if own DB) → town-level (if DB exists) → fallback
- 3 test cases cover all redirect scenarios

Clean code, good error messages, proper test coverage. — gastown/crew/zhora (PM)

PR #2473, opened 2026-03-07
fix: namespace tmux sockets by town path hash (GH #761)

### LINKED ISSUE #761: gt down kills services from other towns due to non-namespaced tmux sessions
## Summary

Running `gt down` in one town directory kills services belonging to a different town running in a separate directory. Towns should be isolated from each other, but the current implementation uses global tmux session names that collide across installations.

## Steps to Reproduce

1. Initialize two separate gt towns in different directories (e.g., `~/town-a/` and `~/town-b/`)
2. Start services in town-b: `cd ~/town-b && gt mayor start`
3. From town-a, run: `cd ~/town-a && gt down`
4. Observe that town-b's mayor session is killed

## Expected Behavior

`gt down` should only stop services belonging to the current town. Each town should operate in complete isolation.

## Actual Behavior

`gt down` stops services from all towns because tmux session names use generic prefixes (`hq-mayor`, `gt-<rig>-refinery`, etc.) without any town-specific namespace.

## Root Cause

Tmux session names appear to use fixed prefixes like:
- `hq-mayor`
- `hq-deacon`
- `gt-<rig_name>-refinery`
- `gt-<rig_name>-witness`

These names are not scoped to the specific town directory, so commands like `gt down` match and kill sessions from any town on the system.

## Suggested Fix

Namespace tmux sessions by town, using either:
1. The town name from config: `gt-<town_name>-mayor`
2. A hash of the town's absolute path: `gt-a1b2c3-mayor`
3. Some other unique town identifier

This would ensure `gt down` only targets sessions belonging to the current town.

DreadPirateRobertz:
## Summary

- Tmux socket names now include a short hash of the canonical town path (e.g., `gt-a1b2c3` instead of just `gt`), ensuring two towns with the same directory basename get isolated tmux servers
- `gt down` adds legacy cleanup for sessions orphaned on the old basename-only socket during migration
- Fixed stale `--nuke` warning message that incorrectly stated all towns share a socket

Closes #761

## Test plan

- [x] `TestTownSocketName` — verifies basename+hash format, determinism, and uniqueness for same-basename dirs
- [x] `TestInitRegistry_SocketFromTownName` — updated to verify prefix format and added same-basename uniqueness test
- [x] `TestLegacySocketName` — verifies old format for migration cleanup
- [x] `TestSanitizeTownName` — existing tests still pass
- [x] Full `go build ./...` and `go vet` clean

### DIFF
--- a/internal/cmd/down.go
+++ b/internal/cmd/down.go
@@ -258,20 +258,28 @@ func runDown(cmd *cobra.Command, args []string) error {
 		}
 	}
 
-	// Phase 4c: Clean up legacy "default" socket sessions.
-	// Old binaries created sessions on the "default" tmux socket. After
-	// transitioning to a per-town socket (e.g., "gt"), those ghost sessions
-	// persist and cause split-brain: the daemon sees them on "gt" but tools
-	// querying bare tmux see (or miss) them on "default".
+	// Phase 4c: Clean up legacy socket sessions.
+	// Old binaries created sessions on the "default" tmux socket or on the
+	// basename-only socket (e.g., "gt" instead of "gt-a1b2c3"). After
+	// transitioning to path-hashed sockets, ghost sessions on old sockets
+	// persist and cause split-brain.
 	if !downDryRun {
 		cleaned := cleanupLegacyDefaultSocket()
 		if cleaned > 0 {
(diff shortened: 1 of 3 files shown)


steveyegge:

PR #2476, opened 2026-03-07
fix: gt dolt sync pushes via SQL to avoid stopping server

seanbearden:
## Summary
- `gt dolt sync` previously stopped the Dolt server to use CLI `dolt push`, crashing all running agents that depend on beads (happened 4+ times per session)
- Now uses `CALL DOLT_PUSH()` via SQL through the running server — zero downtime
- Falls back to CLI push only when the server isn't already running
- Removes the server stop/restart and rig parking from the sync hot path

## Changes
- `internal/doltserver/sync.go`: Add `PushDatabaseSQL`, `FindRemoteSQL`, `SyncDatabasesSQL` — full SQL-based sync pipeline
- `internal/cmd/dolt.go`: `runDoltSync` uses SQL path when server is running, CLI path when not
- `internal/doltserver/sync_test.go`: Tests for filter, empty dir, and remote discovery

## Test plan
- [x] All new sync tests pass
- [x] Full `go build ./...` clean

Bead: hq-qo8ma

### DIFF
--- a/internal/doltserver/sync.go
+++ b/internal/doltserver/sync.go
@@ -128,9 +128,32 @@ func PushDatabase(dbDir, remote string, force bool) error {
 	return nil
 }
 
+// validSQLName checks that a database or remote name contains only safe characters
+// (alphanumeric, underscore, hyphen, dot). This is a defense-in-depth measure since
+// these values come from internal sources (filesystem scan, SQL query output), but
+// prevents SQL breakage or injection if a name ever contains backticks or quotes.
+func validSQLName(name string) bool {
+	if name == "" {
+		return false
+	}
+	for _, c := range name {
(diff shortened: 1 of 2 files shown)


seanbearden:
Thanks for the review!

Addressed concern #1: added `validSQLName()` that rejects names not matching `[a-zA-Z0-9_.-]+`, called at the top of `PushDatabaseSQL` and `FindRemoteSQL`. Includes test coverage. (38f7b380)

Re #2 (partial failure leaving an extra commit): agreed it's benign — the commit is idempotent on retry, and the auto-commit message makes it clear it's from sync.

Re #3 (error string matching for DOLT_ADD): the non-fatal handling is intentionally permissive — if Dolt changes the message, the worst case is a logged warning, not a blocked sync.

PR #2478, opened 2026-03-07
fix: reap idle polecat sessions to prevent API slot burn

keaganvanzyl:
## Summary

- Add `reapIdlePolecats()` to daemon heartbeat cycle (step 12b) that auto-kills polecat tmux sessions idle longer than 5 minutes
- Add configurable `polecat_idle_session_timeout` threshold (default `5m`) to `DaemonThresholds`
- Handle edge case where `persistentPreRun` resets heartbeat to "working" during `gt done` by cross-checking bead hook state

## Problem

The persistent polecat model (gt-4ac) keeps tmux sessions alive after `gt done` for session reuse, but idle Claude Code processes consume API slots indefinitely. Dogs have `DogIdleSessionTimeout` (1h) for auto-kill, but polecats had no equivalent timeout. In production, 11 polecats sat idle for 1-2 hours burning API capacity with no automated remediation — the witness detected them but could only send mail.

## How it works

The reaper runs each heartbeat cycle and checks three conditions for each polecat:

1. **HeartbeatIdle or HeartbeatExiting** + stale timestamp → reap
2. **HeartbeatWorking** + stale timestamp + **no hooked work** in beads → reap (catches the `persistentPreRun` reset edge case)
3. **HeartbeatWorking** + stale timestamp + **has hooked work** → skip (let `checkPolecatSessionHealth` handle stuck polecats)

## Test plan

- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] Deployed locally — daemon running with patched binary, active polecats not falsely reaped
- [ ] Verify reap in action when polecats complete `gt done` and sit idle >5 minutes

### DIFF
--- a/internal/beads/beads.go
+++ b/internal/beads/beads.go
@@ -534,7 +534,7 @@ func stripEnvPrefixes(environ []string, prefixes ...string) []string {
 
 // List returns issues matching the given options.
 func (b *Beads) List(opts ListOptions) ([]*Issue, error) {
-	args := []string{"list", "--json"}
+	args := []string{"list", "--flat", "--json"}
 
 	if opts.Status != "" {
 		args = append(args, "--status="+opts.Status)


DreadPirateRobertz:
### Code review

Found 1 issue:

1. `bd list --flat` uses a flag that does not exist -- `--flat` is not a recognized flag for `bd list`. Running `bd list --flat --json` returns `Error: unknown flag: --flat`. This will cause every call to `Beads.List()` to fail at runtime, breaking any daemon or agent code path that lists beads.

https://github.com/steveyegge/gastown/blob/7c731c21ab14870a53bd684572b0bec73af7536a/internal/beads/beads.go#L536-L538

DreadPirateRobertz (approved):
LGTM. Well-architected idle polecat reaper.

Good design decisions:
- Configurable timeout (default 15m) with DaemonThresholds integration
- Heartbeat state + agent bead cross-check catches "working" polecats that actually finished gt done
- 2x timeout fallback for degraded bead infrastructure — prevents indefinite API burn
- Proper cleanup (kill session, remove heartbeat, emit feed event)
- Doesn't reap polecats with hooked work (defers to checkPolecatSessionHealth)

Clean code, good comments, appropriate separation of concerns. — gastown/crew/zhora (PM)

DreadPirateRobertz:
## Code Review: PR #2478 — Reap Idle Polecat Sessions

**Reviewer:** gastown/crew/zhora (PM) — 5-agent review per mandate

### Critical Issues

**[CRITICAL] `--flat` flag does not exist in `bd` CLI (confidence: 95%)**

The PR adds `--flat` to `bd list` arguments. This flag doesn't exist — grep across the codebase shows zero other usage. If unrecognized, `bd` exits non-zero, causing bead lookup to fail for every polecat. The fallback logic then fires for all polecats, and after 2x timeout it reaps *every* polecat — including active ones. This is the exact false-positive kill the PR tries to prevent. Fix: remove `--flat`, existing `--json` is sufficient.

**[CRITICAL] Active polecat false-positive kill: working + stale heartbeat + no hook_bead (confidence: 88%)**

There's a race window between `gt done` clearing `hook_bead` and `gt sling` setting the new one. A polecat in this window matches all three "idle" conditions but is actually between tasks. The existing `checkPolecatHealth` has a spawning guard for this exact race — the reaper doesn't incorporate it. Recommendation: cross-check `agent_state` from bead DB before reaping. If `agent_state == "spawning"` with recent `updated_at`, skip reap.

### Important Issues

**No patrol guard** — `reapIdlePolecats()` runs unconditionally, unlike every other heartbeat step which uses `IsPatrolEnabled()`. Cannot be disabled via `daemon.json`.

**Heartbeat file removed before kill confirmation** — if `tmux.KillSessionWithProcesses()` fails, heartbeat is already gone. Daemon loses visibility into the polecat on next cycle. Remove heartbeat only after confirmed kill.

**Multi-rig TownRoot assumption** — `ReadSessionHeartbeat(d.config.TownRoot, ...)` uses daemon's local workspace. Remote rig polecats have no heartbeat at that path → appear stale → may be falsely reaped.

**Zero tests** for core reaping logic (4 new functions, multiple branches). Given false-positive = killing active polecats, test coverage is not optional.

### Verdict

Do not merge. The `--flat` flag bug alone is a deployment blocker. Fix that + add spawning guard + patrol config guard + tests.

PR #2480, opened 2026-03-07
fix: gt prime --hook blocks forever on non-Claude runtimes

seanbearden:
## Summary
- `gt prime --hook` now works with any LLM runtime, not just Claude Code
- Reorders `readHookSessionID` to check env vars **before** stdin read — non-Claude runtimes skip stdin entirely (zero delay)
- Adds `GT_HOOK_SOURCE` env var support for source detection (startup/compact/resume)
- Adds persisted session ID fallback (PreCompress reuses SessionStart's ID)
- Reduces stdin timeout from infinite → 500ms as safety net
- Updates `--hook` documentation with multi-runtime examples

## Problem
`readStdinJSON()` reads from stdin when it's a pipe. Claude Code sends session JSON immediately. Gemini CLI (and other runtimes) pipe stdin but send nothing, causing `ReadString('\n')` to block forever. The hook hangs, the agent gets no Gas Town context, and the runtime misidentifies the repo.

The original code checked stdin **first**, before env vars — so even setting `GT_SESSION_ID` in the hook command couldn't prevent the block.

## Fix
1. **Env vars first**: `GT_SESSION_ID` and `GT_HOOK_SOURCE` skip stdin read entirely
2. **Persisted fallback**: PreCompress finds the session ID that SessionStart wrote
3. **Stdin last**: Claude Code path unchanged (still works via stdin JSON)
4. **Timeout safety net**: 500ms timeout if stdin is a pipe with no env vars set

### Resolution order
| Priority | Source | Used by |
|----------|--------|---------|
| 1 | `GT_SESSION_ID` env | Gemini, other runtimes |
| 2 | `CLAUDE_SESSION_ID` env | Legacy |
| 3 | `.runtime/session_id` file | PreCompress (any runtime) |
| 4 | Stdin JSON | Claude Code |
| 5 | Auto-generated UUID | Fallback |

### Gemini integration example
```json
"SessionStart": "export GT_SESSION_ID=$(uuidgen) GT_HOOK_SOURCE=startup && gt prime --hook"
"PreCompress":  "export GT_HOOK_SOURCE=compact && gt prime --hook"
```

## Test plan
- [x] `go test ./internal/cmd/ -run TestPrime` — all 6 subtests pass
- [x] Manual: `gt prime --hook` from Gemini worktree no longer hangs
- [x] Claude path unchanged: stdin JSON still takes priority when present

### DIFF
--- a/internal/cmd/prime_session.go
+++ b/internal/cmd/prime_session.go
@@ -30,8 +30,8 @@ type hookInput struct {
 //
 // Priority (env vars first so non-Claude runtimes skip the stdin read entirely):
 //  1. GT_SESSION_ID / CLAUDE_SESSION_ID env var  — set by the hook command
-//  2. Persisted .runtime/session_id              — written by a prior SessionStart hook
-//  3. Stdin JSON (Claude Code format)            — Claude sends {"session_id":…,"source":…}
+//  2. Stdin JSON (Claude Code format)            — Claude sends {"session_id":…,"source":…}
+//  3. Persisted .runtime/session_id              — written by a prior SessionStart hook
 //  4. Auto-generate UUID
 //
 // Source is resolved from GT_HOOK_SOURCE env, stdin JSON, or empty.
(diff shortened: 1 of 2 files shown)


seanbearden:
Addressed the review feedback: added a comment at the timeout path documenting the intentional goroutine leak (commit 44178073).

> The goroutine above is still blocked on ReadString and will leak.
> This is intentional — gt prime is a short-lived CLI command that
> exits shortly after, so the goroutine is cleaned up by process exit.

seanbearden:
Added unit tests for `readHookSessionID` covering all requested scenarios (0a7fb0a5):

1. **GT_SESSION_ID env takes priority** — set env var, verify returned without touching stdin
2. **CLAUDE_SESSION_ID fallback** — used when GT_SESSION_ID is unset
3. **Persisted session ID fallback** — writes temp `.runtime/session_id`, verifies it's found via cwd
4. **GT_HOOK_SOURCE populates source** — set env var, verify source return value
5. **Auto-generate UUID fallback** — no env vars or file, verify UUID is generated

All 5 tests pass. Skipped the stdin test (test #4 from the review about "stdin source overrides env source") since it requires pipe manipulation that's fragile in unit tests — the env var and persisted file paths cover the critical priority chain.

steveyegge:
Fix-merged to main at `de818831`. Thanks @seanbearden — clean fix, well tested. All 5 new tests pass locally. CI failures were pre-existing flakes on main (convoy_id, hook_show, rig_park).

PR #2481, opened 2026-03-07
fix: honor sling's BD_DOLT_AUTO_COMMIT=off in hookBeadWithRetry

DreadPirateRobertz:
## Summary

- `hookBeadWithRetry` called `.WithAutoCommit()` which overrode the parent's `BD_DOLT_AUTO_COMMIT=off` with `on`
- The hook update is a single atomic `bd update` that doesn't need auto-commit — the Dolt server handles commits
- Removing it fixes manifest contention under concurrent batch slinging (gt-u6n6a)

## Test plan

- [x] `TestSlingSetsDoltAutoCommitOff` — PASSES (was P1 CI blocker)
- [x] All 30 sling-related tests pass
- [x] One-line change, minimal blast radius

Fixes #2443

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

Issue #2443, opened 2026-03-07
bug: TestSlingSetsDoltAutoCommitOff and TestVerifyBeadExistsAllowStale fail on CI


DreadPirateRobertz:
## Description

Two unit tests in `internal/cmd/sling_test.go` fail consistently on CI (observed on PRs #2439, #2425, #2422, #2420, #2419):

### TestSlingSetsDoltAutoCommitOff (line 1613)
```
sling_test.go:1613: bd command missing BD_DOLT_AUTO_COMMIT=off: ENV:BD_DOLT_AUTO_COMMIT=on|update gt-test456 --status=hooked --assignee=mayor/
```
The test expects sling to set `BD_DOLT_AUTO_COMMIT=off` but the command runs with `BD_DOLT_AUTO_COMMIT=on`.

### TestVerifyBeadExistsAllowStale (line 935)
```
sling_test.go:935: verifyBeadExists("jv-v599") failed: bead 'jv-v599' not found (bd show failed)
    Expected --allow-stale to skip sync check
```
The `--allow-stale` flag doesn't prevent `verifyBeadExists` from failing when the bead isn't in the local database.

### Impact
These failures block all PRs from passing CI, even when PR changes are completely unrelated to sling code.

### Suggested fix
1. For AutoCommitOff: check if sling's env injection is being overridden by a global config or test setup
2. For AllowStale: the `verifyBeadExists` path may need to honor `--allow-stale` before calling `bd show`

DreadPirateRobertz:
Fix submitted in PR #2444. Root cause: `sync.Once` in `BdSupportsAllowStale()` caches the probe result from the real `bd` binary, then tests with stub `bd` binaries inherit the stale cached result. Fix adds `ResetBdAllowStaleForTest()`. The auto-commit assertion was also too strict — `hookBeadWithRetry` legitimately uses `.WithAutoCommit()` for critical hook writes. PR passes Lint and Test on CI.

DreadPirateRobertz:
The `TestSlingSetsDoltAutoCommitOff` failure is fixed — the root cause was `hookBeadWithRetry` calling `.WithAutoCommit()` which overrode sling's global `BD_DOLT_AUTO_COMMIT=off` to `on`. Removed the unnecessary `.WithAutoCommit()` from the hook update in `sling_helpers.go:987`. The hook update is atomic and doesn't need auto-commit.

Still investigating `TestVerifyBeadExistsAllowStale`.

[...]

PR #2485, opened 2026-03-07
fix: TestVerifyBeadExistsAllowStale fails due to sync.Once caching

DreadPirateRobertz:
## Summary
- `BdSupportsAllowStale()` uses `sync.Once` to cache feature detection
- In test processes, if detection fires before the stub `bd` is in PATH, `false` is cached permanently
- `--allow-stale` gets stripped from args, stub bd fails without it
- Added `SetBdAllowStaleForTest()` with pointer override that bypasses `sync.Once`
- Both `TestVerifyBeadExistsAllowStale` and `TestSlingWithAllowStale` now use it with `t.Cleanup`

## Test plan
- [x] Both previously-failing tests now pass
- [x] `internal/beads` tests pass
- [x] `internal/cmd` sling tests pass

Fixes the second half of #2443 (first half fixed in #2481).

### DIFF
--- a/internal/beads/beads.go
+++ b/internal/beads/beads.go
@@ -30,12 +30,16 @@ var (
 // bdAllowStale caches whether the installed bd supports --allow-stale.
 // Detected once at first use via `bd --allow-stale version`.
 var (
-	bdAllowStaleOnce   sync.Once
-	bdAllowStaleResult bool
+	bdAllowStaleOnce     sync.Once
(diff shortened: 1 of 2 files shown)


steveyegge:
The underlying code has been significantly reworked on main since this PR was opened. Cherry-pick conflicts. If the fix is still needed, please open a fresh PR from current main.

PR #2486, opened 2026-03-07
fix: pre-validate command binary in tmux session creation

### LINKED ISSUE #2451: fix: TestNewSessionWithCommand_ExecEnvBadBinary fails on CI (tmux env difference)

DreadPirateRobertz:
## Summary
- `TestNewSessionWithCommand_ExecEnvBadBinary` fails intermittently on CI due to timing-based health checks
- Added `validateCommandBinary()` that extracts the binary path from commands like `exec env VAR=val /path/binary` and checks it exists before creating the tmux session
- Gives an instant, deterministic error instead of relying on 250ms timing window
- Applied to both `NewSessionWithCommand` and `NewSessionWithCommandAndEnv`
- Only checks absolute paths to avoid false positives on shell builtins

## Test plan
- [x] New `TestValidateCommandBinary` covers 8 cases (empty, relative, valid absolute, missing absolute, exec env patterns)
- [x] Compiles clean
- [x] Existing tmux tests unaffected

Fixes #2451

### DIFF
--- a/internal/tmux/tmux.go
+++ b/internal/tmux/tmux.go
@@ -57,6 +57,49 @@ func validateSessionName(name string) error {
 	return nil
 }
 
+// validateCommandBinary extracts the binary path from a tmux session command
+// and verifies it exists on disk. Handles common patterns:
+//   - "exec env VAR=val /path/to/binary --args"
+//   - "/path/to/binary --args"
+//   - "sh -c '...'" (skipped — shell will handle resolution)
+//
+// Only checks absolute paths to avoid false positives on shell builtins.
+func validateCommandBinary(command string) error {
+	fields := strings.Fields(command)
+	if len(fields) == 0 {
+		return nil
+	}
+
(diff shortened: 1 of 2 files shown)


steveyegge:

PR #2497, opened 2026-03-07
fix(nudge): deliver queued nudges via hook trigger, not raw tmux send-keys

seanbearden:
## Summary

Fixes nudge delivery to idle agents so nudges appear as background notifications
(`<system-reminder>`) instead of user input that agents interpret as interruptions.

**The problem:** When an agent becomes idle after receiving a queued nudge, the
`watchAndDeliver` goroutine would drain the queue and send the full nudge message
via `NudgeSession` (tmux send-keys). Agents interpret send-keys input as the user
typing — an interruption or correction — and react defensively.

**The fix:** Instead of sending the full nudge text as user input:
1. Leave nudges in the queue (don't drain in the watcher)
2. Send a minimal trigger message (`[nudge: see system-reminder below]`) via tmux
3. This fires the `UserPromptSubmit` hook (`gt mail check --inject`)
4. The hook drains the queue and formats nudges as `<system-reminder>`
5. Agent processes the nudge as a background notification, not an interruption

Both idle delivery paths are fixed:
- **WaitForIdle succeeds** (agent idle immediately): enqueue first, then trigger hook
- **watchAndDeliver** (agent becomes idle later): trigger hook (nudge already in queue)

Immediate mode (`--mode=immediate`) is unchanged — it's explicitly for
breaking through to stuck/emergency agents.

## Changes

- `internal/cmd/nudge.go`: Add `nudgeTriggerMessage` const; refactor both idle
  delivery paths to use hook-based delivery; remove queue drain from `watchAndDeliver`
- `internal/cmd/nudge_test.go`: Update `watchAndDeliver` call site for new signature

## Test plan

- [x] `go build ./internal/cmd/` — compiles clean
- [x] `go test ./internal/cmd/ -run TestNudge` — all pass
- [x] `go test ./internal/cmd/ -run TestIdleWatcher` — all pass
- [x] `go test ./internal/cmd/ -run TestQueueLen` — passes
- [ ] Manual: `gt nudge <idle-agent>` delivers via `<system-reminder>`, not as user input
- [ ] Manual: verify agent treats nudge as background notification

### DIFF
--- a/internal/cmd/nudge.go
+++ b/internal/cmd/nudge.go
@@ -191,7 +191,14 @@ func deliverNudge(t *tmux.Tmux, sessionName, message, sender string) error {
 			// Queue failed — fall back to immediate as last resort.
 			// Better to interrupt than lose the message entirely.
 			fmt.Fprintf(os.Stderr, "Warning: queue fallback failed (%v), delivering immediately\n", qErr)
-			return t.NudgeSession(sessionName, prefixedMessage)
+			// Still use FormatForInjection so the agent sees a consistent
+			// <system-reminder> format regardless of delivery path.
+			formatted := nudge.FormatForInjection([]nudge.QueuedNudge{{
+				Sender:   sender,
+				Message:  message,
+				Priority: nudgePriorityFlag,
+			}})
+			return t.NudgeSession(sessionName, formatted)
 		}
 		// Run watcher synchronously: polls for idle over a longer window.
 		// The UserPromptSubmit hook drains the queue on agent input, but an
@@ -223,6 +230,7 @@ func deliverNudge(t *tmux.Tmux, sessionName, message, sender string) error {
 //   - Session disappears: exit (nothing to deliver to).
 //   - Timeout: exit (queue stays for next input or watcher cycle).
 func watchAndDeliver(t *tmux.Tmux, townRoot, sessionName string) {
+	fmt.Fprintf(os.Stderr, "Watching %s for idle (up to %s)...\n", sessionName, idleWatcherTimeout)
 	deadline := time.Now().Add(idleWatcherTimeout)
 	for time.Now().Before(deadline) {
 		time.Sleep(idleWatcherPollInterval)
@@ -237,7 +245,11 @@ func watchAndDeliver(t *tmux.Tmux, townRoot, sessionName string) {
 			return
 		}
 
-		if t.IsIdle(sessionName) {
+		// Use WaitForIdle with a short timeout instead of single-snapshot
+		// IsIdle to get the consecutive-poll guard (2 polls 200ms apart).
+		// This avoids false positives during inter-tool-call gaps where
+		// the prompt briefly appears while Claude Code is still working.
+		if err := t.WaitForIdle(sessionName, idleWatcherPollInterval); err == nil {
 			// Drain atomically claims queued entries (rename-based).
 			// If another process raced and drained first, we get an
 			// empty slice and skip delivery to avoid duplicates.
--- a/internal/nudge/queue.go
+++ b/internal/nudge/queue.go
@@ -273,9 +273,14 @@ func Pending(townRoot, session string) (int, error) {
 }
 
 // QueueLen returns the number of pending nudges for a session without draining.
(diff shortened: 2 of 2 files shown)


seanbearden:
## Fix: drain queue in watcher, bypass UserPromptSubmit hook (gt-x1wu)

**Root cause:** `UserPromptSubmit` hook does not fire for tmux `send-keys` input. The trigger-based approach (enqueue + send `[nudge: see system-reminder below]` + expect hook to drain) never works because the hook never fires.

**Fix (af47c301):**
- Removed `nudgeTriggerMessage` — no longer needed
- Wait-idle success path: format nudge directly via `FormatForInjection()` and send via `NudgeSession()` — no queue roundtrip needed when agent is already idle
- `watchAndDeliver`: drains queue via `nudge.Drain()` (atomic rename-based), formats via `FormatForInjection()`, sends formatted `<system-reminder>` content directly via `NudgeSession()`
- Zero dependency on UserPromptSubmit hook — delivery is entirely watcher-driven

All existing tests pass. Build clean.

seanbearden:
All review feedback addressed in 90730514:

**Blocking: Goroutine lifecycle bug** — Fixed. `watchAndDeliver` now runs synchronously (not as a goroutine). `gt nudge` blocks until delivery, session death, or timeout. Process stays alive for the full polling window.

**QueueLen swallows errors** — Fixed. Non-trivial errors (permissions, disk full) are now logged to stderr. `os.IsNotExist` is still silenced (expected for missing queue dirs).

**Missing integration test** — Fixed. Added `TestFormatForInjection_SystemReminder` (verifies `<system-reminder>` tags and content) and `TestFormatForInjection_UrgentPriority` (verifies URGENT formatting). All tests pass.

Single clean commit, build clean.

seanbearden:
Good catch — you're exactly right. `go watchAndDeliver(...)` followed by `return nil` means the process exits and the goroutine is killed immediately, so the watcher never actually polls.

Fixed in 4a6a04d9: `watchAndDeliver` now runs synchronously (no goroutine). `gt nudge` blocks until the watcher delivers, observes session death, or times out (`idleWatcherTimeout`, 60s). This matches the commit message's claim that "gt nudge blocks until delivery, session death, or timeout" — which was the intent, just not what the code did.

DreadPirateRobertz:
## Code Review: PR #2497 — Nudge Queue Idle Watcher

**Reviewer:** gastown/crew/zhora (PM) — 5-agent review per mandate

### Important Issues (no criticals found)

**1. Inconsistent delivery format between immediate and watcher paths (confidence: 85%)**

Immediate-idle path sends raw `[from X] msg` via `NudgeSession`, but the watcher path drains via `FormatForInjection` which wraps in `<system-reminder>` tags. The agent receives structurally different input depending on timing — the exact problem this PR is supposed to fix. Fix: use `FormatForInjection` on the immediate path too.

**2. 60-second silent block with no progress indication (confidence: 82%)**

`watchAndDeliver` blocks the CLI silently for up to 60s. In channel mode, each member blocks sequentially — 10-member channel = up to 10 minutes. Add a stderr log when watcher starts, and consider shorter timeout or parallel delivery for channels.

**3. `IsIdle` bypasses `WaitForIdle`'s consecutive-poll guard (confidence: 80%)**

`WaitForIdle` requires 2 consecutive idle polls 200ms apart to avoid false positives during inter-tool-call gaps. The watcher uses single-snapshot `IsIdle` — can detect false idle and inject nudge mid-work. Fix: use `WaitForIdle` with short timeout inside the watcher loop instead of `IsIdle`.

**4. `QueueLen` silently drops non-trivial errors (confidence: 80%)**

Returns 0 on filesystem errors (not just missing dir). A successfully enqueued nudge that can't be read back is silently abandoned. Log non-NotExist errors to stderr.

### What's Working Well

Core queue mechanics (atomic rename-claim, orphan sweep, FIFO ordering) are solid. Test coverage for the new paths is well-structured. The synchronous watcher design is the right call.

### Verdict

No blockers, but the format inconsistency (issue #1) and IsIdle false-positive risk (issue #3) should be addressed before merge.

PR #2500, opened 2026-03-07
fix: surface JSON parse errors in ListAgentBeads (GH#2499)

### LINKED ISSUE #2499: gt doctor fails agent-beads-exist check due to bd list --json output format and ephemeral bead label skipping
## Bug Description

```go
```

```go
```

## Workaround applied
I resolved this by:

## Environment
- `gt` version: 0.11.0
- `bd` version: 0.59.0 (dev)
- OS: Linux

DreadPirateRobertz:
## Summary

- `ListAgentBeads()` silently swallowed JSON unmarshal errors from `bd list --json`, causing `gt doctor` to falsely report missing agent beads when tree-view or pager output corrupted JSON
- Now returns error instead of proceeding with empty results
- Added `--no-pager` flag to prevent pager interference with JSON output

Fixes #2499

## Test plan

- [ ] `go build ./internal/beads/` compiles cleanly
- [ ] `go vet ./internal/beads/` passes
- [ ] Existing callers that use `err == nil` pattern (doctor checks) correctly handle the new error path
- [ ] `gt doctor` agent-beads-exist check works correctly with valid JSON output
- [ ] Verify `--no-pager` prevents pager from corrupting JSON in environments with PAGER set

### DIFF
--- a/internal/beads/beads_agent.go
+++ b/internal/beads/beads_agent.go
@@ -630,16 +630,17 @@ func (b *Beads) ListAgentBeads() (map[string]*Issue, error) {
 	// doctor checks (for example, validating gt:agent labels).
 	// Agent beads are type=agent (infrastructure), hidden by bd list default filter.
 	// Use --include-infra so they appear in results.
-	out, err := b.run("list", "--label=gt:agent", "--include-infra", "--json")
+	out, err := b.run("list", "--label=gt:agent", "--include-infra", "--json", "--no-pager")
 	if err != nil {
 		return nil, err
 	}
 	issuesByID := make(map[string]*Issue)
 	var issues []*Issue
-	if jsonErr := json.Unmarshal(out, &issues); jsonErr == nil {
-		for _, issue := range issues {
-			issuesByID[issue.ID] = issue
-		}
+	if jsonErr := json.Unmarshal(out, &issues); jsonErr != nil {
+		return nil, fmt.Errorf("parsing bd list --json output: %w (raw output %d bytes)", jsonErr, len(out))
+	}
+	for _, issue := range issues {
+		issuesByID[issue.ID] = issue
 	}
 
 	// Query wisps table as a fallback source.


DreadPirateRobertz:
## 5-Agent Review by gastown/crew/deckard

*Note: 3 agents omitted — minimal code change (11 lines, single file). Test-analyzer, comment-analyzer, and code-simplifier not applicable.*

---
### code-reviewer

Now I have the complete picture. Let me verify the caller patterns that use `err == nil` to silently swallow errors.

The key callers I already found above:

1. `agent_beads_check.go:88` - `if townAgents, err := townBd.ListAgentBeads(); err == nil {` (silently ignores error)
2. `agent_beads_check.go:103` - same pattern
3. `agent_beads_check.go:226` - same pattern  
4. `agent_beads_check.go:323` - same pattern
5. `stale_agent_beads_check.go:144` - same pattern
6. `status.go:683` - `townAgentBeads, _ := townBeadsClient.ListAgentBeads()` (discards error)
7. `status.go:713` - same pattern
8. `mail_directory.go:56-59` - handles error with warning, continues
9. `mail/resolve.go:163-164` - `if err == nil {` (silently ignores error)
10. `mail/resolve.go:214-216` - returns error
11. `polecat_identity.go:273-275` - returns error
12. `tui/feed/stuck.go:171-173` - returns error

Now I have everything needed for the review.

---

## Review: PR #2500 "fix: surface JSON parse errors in ListAgentBeads (GH#2499)"

### Files reviewed (Go source changes relevant to the PR title)

`/Users/hal/gt/gastown/internal/beads/beads_agent.go` -- the `ListAgentBeads()` function change.

### Summary of changes

1. Added `--no-pager` flag to the `bd list` invocation.
2. Flipped the JSON unmarshal guard from `if jsonErr == nil` (silently swallow errors) to `if jsonErr != nil` (return error), with a descriptive error message including raw output byte count.
3. The iteration over `issues` is moved outside the conditional (now executes unconditionally after a successful unmarshal).

Also included in this PR but outside the stated scope: `ClearAgentHookBead`, `HookBead` field on `AgentFieldUpdates`, and many other unrelated changes across 50+ files. This review focuses on the `ListAgentBeads` fix per your request.

### Correctness analysis

The core logic change is correct. Previously, a JSON parse failure was silently swallowed -- `issuesByID` would remain empty, and `mergeAgentBeadSources` would return only wisp-backed data, masking the real problem. Now the error is surfaced.

The `--no-pager` flag is a good defensive addition -- if a pager was injected, it could corrupt the JSON output (mixing terminal control sequences into the byte stream), which would be one cause of the parse failure described in GH#2499.

### Caller impact analysis

The new error return path will be encountered by callers that previously got an empty map silently. I checked all callers in `internal/`:

- **5 callers use `if ...; err == nil {`** (doctor checks, resolve.go, stale check) -- these will silently skip the result on error, same as before when they got an empty map. No behavior change, no breakage.
- **2 callers use `_, _ =`** (status.go) -- discard error explicitly. No breakage.  
- **2 callers use `if err != nil { return }` or `return nil, fmt.Errorf()`** (polecat_identity.go, resolve.go line 214, stuck.go) -- these will now properly propagate the parse error instead of silently returning empty data. This is strictly an improvement.
- **1 caller uses warning + continue** (mail_directory.go) -- will now print the parse error as a warning. Improvement.

### Issues found

No high-confidence issues (>= 80) found. The change is clean and correct.

### Minor observations (below reporting threshold)

- The error message includes `len(out)` but not a prefix/snippet of the raw output, which could aid debugging. However, including raw output risks log noise and the byte count is sufficient for triage. Not worth flagging.
- The PR bundles unrelated changes (ClearAgentHookBead, HookBead field, and 50+ other files). This is a process concern, not a code quality issue.

### Verdict

The `ListAgentBeads` change is correct, well-constructed, and improves error observability. The `--no-pager` flag is a sensible defensive measure. All callers handle the new error path safely -- callers that checked `err == nil` will continue to skip gracefully, while callers that propagate errors will now surface the real problem. No issues to report.
---
### silent-failure-hunter

Now I have the complete picture. Here is my audit.

---

## Error Handling Audit: PR #2500 -- ListAgentBeads JSON Parse Fix

### ISSUE 1: The Primary Fix (ListAgentBeads JSON unmarshal)

**Location:** `/Users/hal/gt/gastown/crew/deckard/internal/beads/beads_agent.go`, line 639

**Severity:** CRITICAL (this is the bug the PR fixes)

**Current code (line 639):**
```go
if jsonErr := json.Unmarshal(out, &issues); jsonErr == nil {
```

**Issue Description:** The `jsonErr == nil` condition means: on parse success, populate the map; on parse failure, silently continue with an empty `issuesByID` map. The function then merges with wisps and returns successfully. The caller has no idea that the primary data source was corrupted/unparseable. It gets back a partial result (wisps only) and treats it as authoritative.

**Hidden Errors:** Any JSON parse failure -- malformed output from `bd list`, encoding issues, schema changes, truncated output from a killed process, Dolt returning HTML error pages instead of JSON -- all silently produce an empty issues map.

**User Impact:** Agent beads from the issues table vanish from all downstream consumers (doctor checks, idle detection, manager surveys). The system thinks agents do not exist and may re-create them, leading to duplicate state, missed work, or orphaned polecats.

**Assessment of the PR fix:**

The proposed diff correctly:
1. Inverts the condition to check `jsonErr != nil`
2. Returns an error with `fmt.Errorf` wrapping the original error
3. Includes the raw output size for debugging (good -- avoids dumping potentially huge output into logs while still giving a size clue)

**Verdict: The fix is correct and well-constructed.** One minor improvement: consider including the first N bytes of the raw output (e.g., `out[:min(len(out), 200)]`) rather than just the byte count, since the byte count alone does not help distinguish between "empty output" vs "HTML error page" vs "partially valid JSON." But this is a MEDIUM-level polish, not a blocker.

---

### ISSUE 2: ListAgentBeads -- wisps error is silently discarded

**Location:** `/Users/hal/gt/gastown/crew/deckard/internal/beads/beads_agent.go`, line 648

**Severity:** HIGH

**Current code:**
```go
wispBeads, _ := b.ListAgentBeadsFromWisps()
```

**Issue Description:** The error return from `ListAgentBeadsFromWisps()` is unconditionally discarded with `_`. If the wisps query fails for a reason other than "table doesn't exist" (e.g., Dolt server hang, permission error, corrupted data), the caller never knows. The function returns only the issues-table results and the caller believes it has the complete picture.

**Hidden Errors:** Dolt connection failures, query timeouts, permission errors, corrupted wisp data -- anything `ListAgentBeadsFromWisps` might return as an error is swallowed here.

**User Impact:** Agent beads that exist only in the wisps table silently disappear from results. This is especially insidious because wisps are documented as a "fallback existence source" -- if the fallback silently fails, the system has no safety net.

**Recommendation:** At minimum, log the error. Ideally, propagate it unless there is an explicit design decision that wisps failures are non-fatal (in which case, document that decision with a comment and add logging):

```go
wispBeads, wispErr := b.ListAgentBeadsFromWisps()
if wispErr != nil {
    // Log but don't fail -- wisps are a supplementary source.
    // Issues table is authoritative; wisps provide fallback existence.
    logForDebugging("ListAgentBeads: wisps query failed: %v", wispErr)
}
```

---

### ISSUE 3: ListAgentBeadsFromWisps -- command error silently returns nil, nil

**Location:** `/Users/hal/gt/gastown/crew/deckard/internal/beads/beads_agent.go`, lines 672-675

**Severity:** HIGH

**Current code:**
```go
out, err := b.run("mol", "wisp", "list", "--json")
if err != nil {
    return nil, nil // Wisps table may not exist yet
}
```

**Issue Description:** This catch-all returns `nil, nil` for ANY error from `b.run()`. The comment says "Wisps table may not exist yet" but this also swallows Dolt server crashes, connection refused errors, authentication failures, disk full errors, and timeout errors. The function signature returns `error` but never actually uses it -- every error path returns `nil` for the error.

**Hidden Errors:** Every possible failure mode of `b.run()` is hidden: network errors, server crashes, permission denied, out of memory, corrupted database state.

**User Impact:** Callers (ListAgentBeads, and potentially others) cannot distinguish "no wisps exist" from "the database is on fire." Both look like `nil, nil`.

**Recommendation:** Distinguish between "table doesn't exist" (expected, return nil/nil) and other errors (unexpected, return the error):

```go
out, err := b.run("mol", "wisp", "list", "--json")
if err != nil {
    // "table not found" or similar is expected on fresh databases.
    // All other errors should be surfaced.
    if strings.Contains(err.Error(), "not found") || strings.Contains(err.Error(), "no such table") {
        return nil, nil
    }
    return nil, fmt.Errorf("listing wisps: %w", err)
}
```

---

### ISSUE 4: ListAgentBeadsFromWisps -- JSON unmarshal error silently returns nil, nil

**Location:** `/Users/hal/gt/gastown/crew/deckard/internal/beads/beads_agent.go`, lines 681-683

**Severity:** HIGH

**Current code:**
```go
if err := json.Unmarshal(out, &wrapper); err != nil {
    return nil, nil
}
```

**Issue Description:** This is the **exact same pattern** as the bug the PR is fixing in `ListAgentBeads`. The `bd mol wisp list --json` command succeeded (no run error), returned output, but the output cannot be parsed as JSON. This silently returns empty results. This is the same class of bug: corrupted JSON, schema changes, or partial output are all hidden.

**Hidden Errors:** Malformed JSON from wisps, schema mismatches (e.g., `wisps` field renamed), encoding corruption, truncated output.

**User Impact:** Wisp-backed agent beads silently vanish. Since wisps are the fallback for agents that haven't been migrated to the issues table, this means those agents become invisible.

**Recommendation:** Apply the same fix pattern from the PR:

```go
if err := json.Unmarshal(out, &wrapper); err != nil {
    return nil, fmt.Errorf("parsing wisp list JSON: %w (raw output %d bytes)", err, len(out))
}
```

---

### ISSUE 5: ListWispIDs -- identical silent failure pattern (x2)

**Location:** `/Users/hal/gt/gastown/crew/deckard/internal/beads/beads_agent.go`, lines 730-733 and 740-742

**Severity:** HIGH

**Current code (line 730-733):**
```go
out, err := b.run("mol", "wisp", "list", "--json")
if err != nil {
    return nil, nil
}
```

**Current code (line 740-742):**
```go
if err := json.Unmarshal(out, &wrapper); err != nil {
    return nil, nil
}
```

**Issue Description:** `ListWispIDs` has the exact same double-silent-failure pattern as `ListAgentBeadsFromWisps`. Both the command error and the JSON parse error are swallowed, returning `nil, nil`. This function returns `(map[string]bool, error)` but the error return is dead code -- it can never be non-nil.

**Hidden Errors:** Same as Issues 3 and 4 above.

**Recommendation:** Same fix as Issues 3 and 4. If the "table might not exist" case is valid, check for that specifically and surface all other errors.

---

### ISSUE 6: CreateOrReopenAgentBead -- --persistent error swallowed with empty block

**Location:** `/Users/hal/gt/gastown/crew/deckard/internal/beads/beads_agent.go`, lines 348-351

**Severity:** MEDIUM

**Current code:**
```go
if _, err := target.run("update", id, "--persistent"); err != nil {
    // Non-fatal: the bead is functional either way
    // --persistent promotes wisp to issues table (bd update --help)
}
```

**Issue Description:** This is an empty error handler. The comment explains why it is non-fatal, which is better than nothing, but the error is still completely discarded. No logging, no telemetry, no indication that wisp-to-issue migration failed.

**Hidden Errors:** If `--persistent` consistently fails for a structural reason (e.g., the flag was removed from `bd`), every single agent bead reopen will silently skip migration, and agent beads will remain as ephemeral wisps subject to Dolt GC. This would be extremely difficult to diagnose.

**Recommendation:** Log the error even if it is non-fatal:

```go
if _, err := target.run("update", id, "--persistent"); err != nil {
    // Non-fatal: the bead is functional either way.
    // --persistent promotes wisp to issues table (bd update --help).
    // Log so we can detect systematic failures.
    logForDebugging("CreateOrReopenAgentBead: --persistent failed for %s: %v", id, err)
}
```

---

### ISSUE 7: EnsureCustomTypes error discarded

**Location:** `/Users/hal/gt/gastown/crew/deckard/internal/beads/beads_agent.go`, line 213

**Severity:** MEDIUM

**Current code:**
```go
_ = EnsureCustomTypes(targetDir)
```

**Issue Description:** The comment at lines 211-212 explains the rationale ("Don't bail out -- try the bd create calls anyway"), which is a defensible design choice. However, the error is completely discarded without logging. If `EnsureCustomTypes` fails consistently, the downstream `bd create` calls will also fail consistently, and the root cause (custom types not configured) will be obscured by the `bd create` error message.

**Recommendation:** Log the error for diagnostic purposes while preserving the non-fatal behavior:

```go
if err := EnsureCustomTypes(targetDir); err != nil {
    logForDebugging("CreateAgentBead: EnsureCustomTypes failed for %s (will try create anyway): %v", id, err)
}
```

---

## Summary

| # | Location | Severity | Pattern | Status |
|---|----------|----------|---------|--------|
| 1 | `ListAgentBeads` L639 | CRITICAL | JSON parse error swallowed | **PR fix is correct** |
| 2 | `ListAgentBeads` L648 | HIGH | Wisps error discarded with `_` | **Not addressed by PR** |
| 3 | `ListAgentBeadsFromWisps` L672-675 | HIGH | Command error returns nil, nil | **Not addressed by PR** |
| 4 | `ListAgentBeadsFromWisps` L681-683 | HIGH | JSON parse error returns nil, nil | **Not addressed by PR -- same bug class as the fix** |
| 5 | `ListWispIDs` L730-733, L740-742 | HIGH | Double silent failure (same pattern x2) | **Not addressed by PR** |
| 6 | `CreateOrReopenAgentBead` L348-351 | MEDIUM | Empty error handler, no logging | **Not addressed by PR** |
| 7 | `CreateAgentBead` L213 | MEDIUM | Error discarded without logging | **Not addressed by PR** |

**Verdict on the PR fix itself:** The fix at line 639 is **correct and good**. It properly surfaces the error with context (wrapped original error + output size). This is exactly the right pattern.

**Critical concern:** Issues 3 and 4 are the **exact same bug class** as the one this PR fixes, just in the wisps query path. `ListAgentBeadsFromWisps` at line 681 does `return nil, nil` on JSON parse failure -- precisely the silent-failure-on-bad-JSON pattern that this PR is trying to eliminate. If the PR fixes one instance of this bug but leaves the identical pattern in the sibling function, that is an incomplete fix. The wisps path should receive the same treatment, or there should be an explicit comment explaining why wisps JSON parse failures are acceptable to swallow (I cannot think of a valid reason).

steveyegge:

PR #2502, opened 2026-03-07
fix: notify mayor after successful refinery merge (GH#2434)

### LINKED ISSUE #2434: fix: refinery should notify mayor after successful merge
## Problem

When a polecat completes work and the refinery merges it, nobody notifies the Mayor who dispatched the work. The refinery mails the polecat ('MERGED: lc-4bq') and notifies the deacon via `CONVOY_NEEDS_FEEDING`, but the Mayor — who needs to sling the next wave of blocked beads — gets nothing.

## Expected

When refinery merges polecat work, the dispatcher (Mayor) receives a nudge so they can unblock dependent work.

## Actual

Mayor only discovers completion by manually checking. Blocked beads sit idle until Mayor polls.

## Location

`internal/refinery/engineer.go` — `HandleMergeSuccess` / `postMergeConvoyCheck`

## Bead

gt-ae2q (bug 1 of 2)

DreadPirateRobertz:
## Summary

- After refinery merges polecat work, the dispatcher (mayor) was never notified
- Mayor only discovered completions by polling, leaving dependent beads blocked
- Added nudge to mayor/ after successful merge, matching existing patterns (CONVOY_NEEDS_FEEDING, merge failure nudges)

Fixes #2434

## How it works

Uses `gt nudge mayor/ "MERGED: <mr-id> issue=<issue> branch=<branch>"` — same pattern as `notifyDeaconConvoyFeeding` and the merge failure nudge to polecats. Nudge (not mail) avoids permanent Dolt commits for routine protocol signals.

## Test plan

- [ ] `go build ./internal/refinery/` compiles
- [ ] `go vet ./internal/refinery/` passes
- [ ] After merge, verify mayor receives nudge with merge details
- [ ] Verify dependent beads unblock faster when mayor receives merge notification

### DIFF
--- a/internal/refinery/engineer.go
+++ b/internal/refinery/engineer.go
@@ -987,7 +987,17 @@ func (e *Engineer) HandleMRInfoSuccess(mr *MRInfo, result ProcessResult) {
 	// Run convoy check to auto-close and notify subscribers.
 	e.postMergeConvoyCheck(mr)
 
-	// 4. Log success
+	// 4. Nudge mayor about successful merge so dispatcher can unblock
+	// dependent work. Without this, mayor only discovers completion by polling.
+	// Uses nudge (not mail) to avoid permanent Dolt commits for routine signals (GH#2434).
+	nudgeMsg := fmt.Sprintf("MERGED: %s issue=%s branch=%s", mr.ID, mr.SourceIssue, mr.Branch)
+	nudgeCmd := exec.Command("gt", "nudge", "mayor/", nudgeMsg)
+	nudgeCmd.Dir = e.workDir
+	if err := nudgeCmd.Run(); err != nil {
+		_, _ = fmt.Fprintf(e.output, "[Engineer] Warning: failed to nudge mayor about merge: %v\n", err)
+	}
+
+	// 5. Log success
 	_, _ = fmt.Fprintf(e.output, "[Engineer] ✓ Merged: %s (commit: %s)\n", mr.ID, result.MergeCommit)
 }
 


DreadPirateRobertz:
## 5-Agent Review by gastown/crew/deckard

### code-reviewer

Good. Now I have full context. Let me review the diff systematically.

---

## Code Review: PR #2502 -- "fix: notify mayor after successful refinery merge (GH#2434)"

**Files reviewed:**
- `/Users/hal/gt/gastown/crew/deckard/internal/refinery/engineer.go` (lines ~964-1000)
- `/Users/hal/gt/gastown/crew/deckard/internal/refinery/engineer_test.go` (new test)

### Summary

The PR makes two changes to `HandleMRInfoSuccess`:
1. Always delete `polecat/*` branches after merge, regardless of `DeleteMergedBranches` config.
2. Nudge the mayor after a successful merge so the dispatcher can unblock dependent work.

### Issues Found

**Critical (90-100): None**

**Important (80-89):**

**Issue 1: Nudge target uses bare "mayor/" instead of rig-qualified path** -- Confidence: 82

In the failure handler (line 1015), the nudge target is constructed as `fmt.Sprintf("%s/%s", e.rig.Name, polecatName)`, which is rig-qualified. But the new success nudge uses a hardcoded `"mayor/"` string:

```go
nudgeCmd := exec.Command("gt", "nudge", "mayor/", nudgeMsg)
```

The trailing slash on `"mayor/"` looks intentional (mayor is a role, not a specific agent name), but this should be verified against how `gt nudge` resolves targets. If `gt nudge` expects a rig-qualified path like `<rig>/mayor` (consistent with how the failure path builds targets), this will either fail silently or nudge the wrong entity. Additionally, the trailing slash is unusual compared to other nudge targets in the codebase.

**File:** `/Users/hal/gt/gastown/crew/deckard/internal/refinery/engineer.go`, new code around line 991
**Suggested fix:** Verify `gt nudge mayor/` works correctly. Consider using the rig-qualified form: `fmt.Sprintf("%s/mayor", e.rig.Name)` for consistency with the failure handler pattern.

---

**Issue 2: Test only validates logic formula, not actual behavior** -- Confidence: 83

The test `TestPolecatBranchAlwaysDeletedAfterMerge` re-implements the deletion logic inline rather than exercising the actual `HandleMRInfoSuccess` method:

```go
isPolecat := strings.HasPrefix(tt.branch, "polecat/")
shouldDelete := tt.branch != "" && (tt.deleteMergedBranches || isPolecat)
```

This is a tautological test -- it duplicates the production logic and checks it against expected values. If someone changes the condition in `engineer.go` but forgets to update the test, the test will still pass with stale logic. A proper test should construct an `Engineer` with mock git/beads dependencies and call `HandleMRInfoSuccess`, then assert whether `DeleteBranch` was called.

**File:** `/Users/hal/gt/gastown/crew/deckard/internal/refinery/engineer_test.go`, new test function
**Suggested fix:** Use the existing mock/fake infrastructure in the test file to instantiate an `Engineer`, invoke `HandleMRInfoSuccess`, and verify that the git mock's `DeleteBranch` was or was not called based on the test case.

---

### No Issues (Passes Review)

- **CLAUDE.md compliance**: The change correctly uses `gt nudge` instead of `gt mail send` for routine signals, which aligns with the "Communication hygiene" rule: "Default to nudge for routine agent-to-agent communication."
- **Polecat branch cleanup logic**: The boolean condition `mr.Branch != "" && (e.config.DeleteMergedBranches || isPolecat)` is correct and the reordering of the `mr.Branch != ""` guard to the outer level is a minor improvement.
- **Error handling pattern**: The nudge failure is handled with a warning log, consistent with the existing pattern in `HandleMRInfoFailure` (line 1020-1021).
- **Comment quality**: Good comments explaining the "why" for both the polecat cleanup and the nudge.
---
### silent-failure-hunter

Good. Now I have the complete picture. Let me deliver the audit.

---

## Error Handling Audit: PR #2502 -- Refinery Changes

### Overview

The PR adds two changes to `internal/refinery/engineer.go`:
1. A mayor nudge after successful merge (new step 4 in `HandleMRInfoSuccess`)
2. Expansion of branch deletion logic to always clean up `polecat/*` branches

I found **two issues** in the new code and **one pre-existing pattern issue** that the PR perpetuates.

---

### Issue 1: Nudge failure is silently downgraded to a warning -- mayor never learns about the merge

**Location:** `/Users/hal/gt/gastown/internal/refinery/engineer.go`, new lines in `HandleMRInfoSuccess` (the nudge block added at step 4)

**Severity:** HIGH

**The code:**
```go
nudgeCmd := exec.Command("gt", "nudge", "mayor/", nudgeMsg)
nudgeCmd.Dir = e.workDir
if err := nudgeCmd.Run(); err != nil {
    _, _ = fmt.Fprintf(e.output, "[Engineer] Warning: failed to nudge mayor about merge: %v\n", err)
}
```

**Issue Description:** The PR's own comment explains exactly why this matters: *"Without this, mayor only discovers completion by polling."* Yet when the nudge fails, the code logs a warning to `e.output` and moves on. There is no fallback mechanism -- no retry, no event emission, no escalation. The mayor simply does not learn about the merge until the next poll cycle, which is the exact problem this code was written to solve.

Compare this to `notifyDeaconConvoyFeeding` at line 1566, which emits a backup event after its nudge:
```go
_ = events.LogFeed(events.TypeMail, e.rig.Name+"/refinery", events.MailPayload("deacon/", "CONVOY_NEEDS_FEEDING "+mr.ConvoyID))
```
The deacon nudge has a fallback signal path. The mayor nudge does not.

**Hidden Errors:** `exec.Command("gt", "nudge", ...)` can fail for many reasons beyond nudge delivery -- `gt` binary not found, `PATH` not set, `e.workDir` invalid, tmux socket issues. All of these get the same generic warning. The `err` from `Run()` includes exit code but not stderr, so the warning message provides no diagnostic detail about *why* the nudge failed.

**User Impact:** Dependent work remains blocked until the mayor's next poll cycle. The PR comment references GH#2434, implying this delay was significant enough to warrant the change. When the nudge fails, users experience the exact same delay the PR was supposed to fix, but now with a false sense of security that the notification was attempted.

**Recommendation:**
1. Capture stderr from the nudge command for diagnostic context.
2. Add a backup event emission (matching the `notifyDeaconConvoyFeeding` pattern) so the mayor can discover the merge via event feed if nudge fails.
3. Consider at least one retry for transient failures.

**Example fix:**
```go
nudgeCmd := exec.Command("gt", "nudge", "mayor/", nudgeMsg)
nudgeCmd.Dir = e.workDir
var nudgeStderr bytes.Buffer
nudgeCmd.Stderr = &nudgeStderr
if err := nudgeCmd.Run(); err != nil {
    _, _ = fmt.Fprintf(e.output, "[Engineer] Warning: failed to nudge mayor about merge: %v (stderr: %s)\n", err, nudgeStderr.String())
    // Fallback: emit event so mayor discovers merge via event feed
    _ = events.LogFeed(events.TypeMail, e.rig.Name+"/refinery", events.MailPayload("mayor/", nudgeMsg))
}
```

---

### Issue 2: Missing success log for the nudge (inconsistency with all other nudge patterns)

**Location:** Same nudge block in `HandleMRInfoSuccess`

**Severity:** MEDIUM

**Issue Description:** Every other nudge call in this file logs success on the else branch:
- Line 1027: `[Engineer] Nudged %s about merge failure (%s)`
- Line 1562: `[Engineer] Nudged deacon: CONVOY_NEEDS_FEEDING %s`

The new mayor nudge has no success log. This means when reviewing engineer output, there is no confirmation that the mayor was actually notified. If someone is debugging why dependent work is delayed, the absence of any nudge log line (no success, no failure) would be confusing -- did the nudge code even execute?

**Recommendation:** Add an else branch matching the existing pattern:
```go
} else {
    _, _ = fmt.Fprintf(e.output, "[Engineer] Nudged mayor: MERGED %s\n", mr.ID)
}
```

---

### Issue 3: `exec.Command.Run()` stderr is universally discarded across all nudge calls (pre-existing, perpetuated by PR)

**Location:** Lines 1022-1025, 1557-1560, and the new nudge block

**Severity:** MEDIUM (pre-existing pattern, but the PR adds another instance)

**Issue Description:** Every `exec.Command("gt", "nudge", ...).Run()` call in this file discards stderr. When `Run()` returns an error, you get an exit code but no diagnostic output. The `gt nudge` command could print "unknown target", "tmux session not found", "permission denied", or any number of useful diagnostics to stderr -- all lost.

**Hidden Errors:** Authentication failures, missing tmux sessions, binary version mismatches, malformed arguments -- all produce the same opaque `exit status 1` in the warning log.

**Recommendation:** For all nudge calls (not just the new one), capture stderr into a `bytes.Buffer` and include it in warning messages. This is a pre-existing issue, but the PR should not add another instance of the pattern.

---

### Non-Issue: Branch deletion logic change

The branch deletion expansion at line 971-972:
```go
isPolecat := strings.HasPrefix(mr.Branch, "polecat/")
if mr.Branch != "" && (e.config.DeleteMergedBranches || isPolecat) {
```

This is **well-handled**. The existing error handling for `DeleteBranch` and `DeleteRemoteBranch` (lines 973-982) already logs specific warnings with branch names and continues execution. Branch deletion failure after a successful merge is correctly treated as non-fatal. The test coverage in `TestPolecatBranchAlwaysDeletedAfterMerge` validates the conditional logic. No issues here.

---

### Summary

| # | Severity | Issue | Location |
|---|----------|-------|----------|
| 1 | HIGH | Mayor nudge failure has no fallback, defeating the stated purpose of the change | `HandleMRInfoSuccess`, new step 4 |
| 2 | MEDIUM | Missing success log for mayor nudge, inconsistent with all other nudge patterns | Same block |
| 3 | MEDIUM | stderr discarded from exec.Command (pre-existing, perpetuated) | All nudge calls |

The most important finding is Issue 1. The PR's own comment says the nudge exists because "Without this, mayor only discovers completion by polling." But the code provides no fallback when the nudge fails, meaning the mayor falls back to... polling. The `notifyDeaconConvoyFeeding` function at line 1566 already demonstrates the correct pattern: emit a backup event after the nudge so the signal has two delivery paths.
---
### pr-test-analyzer

Now I have a complete picture. Let me provide the analysis.

---

## Test Coverage Analysis for PR #2502

### Summary

The PR makes two changes to `HandleMRInfoSuccess` in `/Users/hal/gt/gastown/crew/deckard/internal/refinery/engineer.go`:

1. **Polecat branch always-delete logic**: Changes the branch deletion condition from `e.config.DeleteMergedBranches && mr.Branch != ""` to `mr.Branch != "" && (e.config.DeleteMergedBranches || isPolecat)`.
2. **Nudge mayor after merge**: Adds a `gt nudge mayor/` exec call after successful merge.

The test added (`TestPolecatBranchAlwaysDeletedAfterMerge`) has a fundamental problem, and the nudge behavior is entirely untested.

---

### Critical Gaps

**1. The polecat branch test re-implements logic instead of exercising it (Criticality: 9/10)**

`TestPolecatBranchAlwaysDeletedAfterMerge` is a **tautological test** -- it duplicates the exact boolean expression from the production code (`strings.HasPrefix(tt.branch, "polecat/")` and `tt.branch != "" && (tt.deleteMergedBranches || isPolecat)`) and then checks it against expected values. This test will always pass regardless of what the production code actually does, because it never calls `HandleMRInfoSuccess`.

If someone changes the condition in `HandleMRInfoSuccess` to, say, `strings.HasPrefix(mr.Branch, "polecats/")` (note the trailing 's'), this test would still pass. It tests nothing about the real code.

**What should exist instead**: A test that calls `HandleMRInfoSuccess` (or at minimum the relevant code path) with a mock/stub git client, sets `DeleteMergedBranches = false`, passes a `polecat/...` branch, and asserts that `DeleteBranch` was called. The existing test infrastructure in this file already creates `Engineer` instances with real configs -- this pattern should be extended with a git interface that records calls.

**Specific regression this would catch**: Someone accidentally removing the `isPolecat` condition, or changing the prefix string, or reordering the condition to short-circuit incorrectly.

**2. No test for the nudge-mayor behavior (Criticality: 7/10)**

The second feature -- nudging the mayor via `exec.Command("gt", "nudge", "mayor/", nudgeMsg)` -- has zero test coverage. While testing exec calls is harder, there are several aspects that could be verified:

- The nudge message format (`MERGED: %s issue=%s branch=%s`) is a protocol contract that the mayor parses. If this format changes, the mayor will fail to extract merge details. A test verifying the format string composition would prevent protocol drift.
- The behavior when the nudge command fails (warning logged, execution continues) should be verified -- a regression where the error becomes fatal would break the merge pipeline silently.
- Consider extracting the nudge into a method (like the existing `notifyDeaconConvoyFeeding`) that can be tested in isolation, or use a nudge interface that can be stubbed.

**Specific regression this would catch**: Someone changing the nudge message format and breaking the mayor's parser; someone accidentally making the nudge error fatal.

---

### Important Improvements

**3. Integration-level test for HandleMRInfoSuccess branch deletion path (Criticality: 6/10)**

The `HandleMRInfoSuccess` method is the critical post-merge handler. It has many steps (close bead, close source issue, clear agent bead, delete branch, convoy check, nudge mayor, log). None of these steps are tested end-to-end in the current test file. The existing tests cover individual helpers (`postMergeConvoyCheck`, `notifyDeaconConvoyFeeding`, config loading) but not the orchestration method itself.

A test that exercises `HandleMRInfoSuccess` with stubs for beads/git would verify the correct ordering and conditional execution of all steps, including the two new ones.

---

### Test Quality Issues

**The existing test is worse than having no test** because it creates false confidence. A developer looking at test coverage would see "polecat branch deletion is tested" and move on, when in fact nothing about the actual code path is being verified. This is a textbook example of testing implementation details (the boolean expression) rather than behavior (does the branch actually get deleted).

The test at `/Users/hal/gt/gastown/crew/deckard/internal/refinery/engineer_test.go` lines 543-574 should either be:
- **Replaced** with a behavioral test that calls real code with test doubles, or
- **Deleted** if the team decides the cost of adding a git interface stub is too high right now -- an honest gap is better than a false positive.

---

### Positive Observations

- The PR's production code is clean and well-commented, with clear explanations of why polecat branches are always deleted and why nudge is used instead of mail.
- The existing test suite for engineer.go is solid -- config loading, gate execution, convoy checks, and deacon notification all have meaningful behavioral tests.
- The choice to use `nudge` over `mail` (avoiding Dolt commits) shows good operational awareness and is consistent with the project's communication hygiene guidelines.
- The test table structure in the new test is well-organized with good case selection (both configs x both branch types + empty branch edge case) -- the problem is purely that it tests a copy of the logic rather than the actual code.
---
### code-simplifier

Good. Now I have the full context. Here is my analysis.

---

**PR #2502 - Refinery changes in `internal/refinery/engineer.go`**

The diff introduces two changes: (1) polecat branch cleanup logic, and (2) a mayor nudge after merge. Here are my observations:

### 1. Branch deletion logic -- the `isPolecat` variable is fine

```go
isPolecat := strings.HasPrefix(mr.Branch, "polecat/")
if mr.Branch != "" && (e.config.DeleteMergedBranches || isPolecat) {
```

You could technically inline it:
```go
if mr.Branch != "" && (e.config.DeleteMergedBranches || strings.HasPrefix(mr.Branch, "polecat/")) {
```

But the named variable actually improves readability here -- it documents *why* we check that prefix. The comment above it is a bit verbose (3 lines for a simple concept), but it is accurate and not egregious. **No simplification needed.**

The condition reordering from the original (`e.config.DeleteMergedBranches && mr.Branch != ""`) to the new form (`mr.Branch != "" && (... || isPolecat)`) is correct -- the empty-branch guard should come first since it protects both paths.

### 2. Mayor nudge -- minor inconsistency with existing patterns

The new nudge block:
```go
nudgeMsg := fmt.Sprintf("MERGED: %s issue=%s branch=%s", mr.ID, mr.SourceIssue, mr.Branch)
nudgeCmd := exec.Command("gt", "nudge", "mayor/", nudgeMsg)
nudgeCmd.Dir = e.workDir
if err := nudgeCmd.Run(); err != nil {
    _, _ = fmt.Fprintf(e.output, "[Engineer] Warning: failed to nudge mayor about merge: %v\n", err)
}
```

Two observations:

**a) Missing success log.** The other two nudge call sites in this file (merge failure nudge at line 1026, convoy feeding nudge at line 1562) both log on success with an `else` branch. This one silently succeeds. For consistency and debuggability, it should log success too. This is not a simplification issue but a consistency gap worth flagging.

**b) No helper extraction warranted (yet).** There are now three nudge-exec-and-log blocks in this file with identical boilerplate (`exec.Command`, `.Dir`, `.Run()`, warning fprintf). A `nudge(target, msg string)` helper would reduce duplication, but that refactor is broader than this PR's scope and the pattern is well-established. Not worth changing here.

### Summary

The code added is clean and follows existing patterns well. The `isPolecat` variable is a good readability choice. The only actionable finding is the **missing success log on the mayor nudge** -- every other nudge in this file logs on success, and this one should too for consistency and operational visibility. That is a one-line addition, not a simplification per se, but a consistency fix.

**File**: `/Users/hal/gt/gastown/internal/refinery/engineer.go` (lines ~990-997 in the PR diff)
---
### comment-analyzer

Good. Now I have a thorough understanding. Here is my analysis:

---

**Summary**

Analyzed two comment blocks added/modified in PR #2502 within `/Users/hal/gt/gastown/internal/refinery/engineer.go`, inside the `HandleMRInfoSuccess` method. The PR adds a mayor nudge after successful merges (fixing GH#2434) and also modifies the branch deletion logic to always clean up polecat branches. The changes shift the step numbering from {1, 1.5, 2, 3, 4} to {1, 1.5, 2, 3, 4, 5}.

---

**Critical Issues**

- **Location**: Line ~990 (PR addition), step "4. Nudge mayor"
  - **Issue**: The comment says "dispatcher can unblock dependent work" but the nudge target is `"mayor/"` (with trailing slash), while the existing analogous nudge to the deacon at line 1557 uses `"deacon"` (no trailing slash). This is not a comment problem per se, but the comment does not flag this format difference. If `"mayor/"` is wrong and the nudge silently fails, the comment's claim that "without this, mayor only discovers completion by polling" becomes the actual permanent behavior -- the comment would be documenting intent rather than reality. **Verify that `gt nudge mayor/` (with trailing slash) is the correct target format.** The PR description also uses `"mayor/"` so it is intentional, but it is inconsistent with the deacon nudge pattern in the same file.

---

**Improvement Opportunities**

- **Location**: Line 960, step "1.5. Clear agent bead's active_mr reference"
  - **Current state**: The fractional step number "1.5" is a code smell in a numbered sequence. With the PR adding steps 4 and 5, the full sequence becomes {1, 1.5, 2, 3, 4, 5} -- the fractional numbering stands out as an afterthought insertion rather than a deliberate ordering.
  - **Suggestion**: Since the PR is already renumbering steps (old step 4 becomes step 5), this would be a natural time to renumber the entire sequence to {1, 2, 3, 4, 5, 6}. However, this is outside the scope of the PR's stated goal, so treat as optional cleanup.

- **Location**: PR addition, step "4. Nudge mayor" comment
  - **Current state**: The comment says "Uses nudge (not mail) to avoid permanent Dolt commits for routine signals (GH#2434)." The parenthetical "(GH#2434)" reads as if GH#2434 is where the nudge-over-mail convention was established, but GH#2434 is actually the issue requesting this fix. This is slightly ambiguous.
  - **Suggestion**: Reword to: `// Uses nudge (not mail) to avoid permanent Dolt commits for routine signals. Fixes GH#2434.` -- making "Fixes" explicit removes ambiguity about whether GH#2434 is a reference to the convention or to this specific change.

- **Location**: PR addition, step "4. Nudge mayor" comment
  - **Current state**: The comment says "dispatcher can unblock dependent work" but does not specify what the mayor does with the `MERGED:` message. The analogous `CONVOY_NEEDS_FEEDING` nudge (line 1532) has a detailed comment explaining the protocol and what the recipient does with it.
  - **Suggestion**: Add a brief note about the expected mayor behavior, e.g., "Mayor's dispatcher uses this to immediately re-scan `bd ready` for newly unblocked beads."

---

**Recommended Removals**

None. Both comments add genuine value -- the step 2 comment explains the polecat-vs-config distinction which is not obvious from the conditional alone, and the step 4 comment explains the rationale for the nudge and the design choice of nudge over mail.

---

**Positive Findings**

- **Location**: Lines 967-970, step "2. Delete source branch"
  - The comment accurately describes the code at line 971-972: `isPolecat := strings.HasPrefix(mr.Branch, "polecat/")` followed by `if mr.Branch != "" && (e.config.DeleteMergedBranches || isPolecat)`. The claim that polecat branches are "always cleaned up" while others "respect the DeleteMergedBranches config" is exactly what the boolean logic implements. This is a well-written comment that explains the *why* (ephemeral work branches) rather than just restating the *what*.

- **Location**: PR addition, step "4. Nudge mayor"
  - The comment correctly explains the problem being solved (polling-only discovery) and the design decision (nudge not mail). It follows the established pattern from the merge-failure nudge (line 1008-1011) and the convoy nudge (line 1554-1555) where the rationale for nudge over mail is documented inline. Consistency with existing patterns is good.

---

**Key files examined**: `/Users/hal/gt/gastown/internal/refinery/engineer.go` (lines 895-992 for `HandleMRInfoSuccess`, lines 1530-1567 for `notifyDeaconConvoyFeeding` as the comparison pattern).

steveyegge:

PR #2507, opened 2026-03-07
fix: run nudge watchAndDeliver synchronously to prevent goroutine death

DreadPirateRobertz:
## Summary
- Fix goroutine lifecycle bug in `watchAndDeliver` from PR #2497
- The `go watchAndDeliver(...)` goroutine died when `gt nudge` exited because `deliverNudge` returned `nil` immediately after spawning it
- Changed to synchronous call so the process blocks until delivery, timeout, or early exit
- Early-exit conditions (empty queue, dead session, successful delivery) prevent unnecessary blocking

## Root Cause
When `--mode=wait-idle` falls back to queue (agent busy during initial 15s poll), a background goroutine was launched to watch for idle over 60s. But since `deliverNudge` returned immediately, `runNudge` printed success and the process exited — killing the goroutine before it could deliver.

## Changes
- `internal/cmd/nudge.go`: Remove `go` prefix from `watchAndDeliver` call, making it synchronous
- Updated comments to reflect the new synchronous behavior

## Test plan
- [x] All nudge-related tests pass (TestIdleWatcherExitsOnEmptyQueue, TestPostQueueIdleRecovery, etc.)
- [x] `watchAndDeliver` early-exit conditions still work (empty queue exits immediately per test)

Based on work by @seanbearden in #2497.

Co-Authored-By: Sean Bearden <seanbearden@users.noreply.github.com>

### DIFF
--- a/internal/cmd/nudge.go
+++ b/internal/cmd/nudge.go
@@ -193,11 +193,15 @@ func deliverNudge(t *tmux.Tmux, sessionName, message, sender string) error {
 			fmt.Fprintf(os.Stderr, "Warning: queue fallback failed (%v), delivering immediately\n", qErr)
 			return t.NudgeSession(sessionName, prefixedMessage)
 		}
-		// Launch background watcher: polls for idle over a longer window.
+		// Run foreground watcher: polls for idle over a longer window.
 		// The UserPromptSubmit hook drains the queue on agent input, but an
 		// idle agent receives no input — so queued nudges are lost without
 		// this watcher. It exits on: delivery, session death, or timeout.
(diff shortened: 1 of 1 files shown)


steveyegge:
Closing — this fix was already landed on main in commit 2abc36d7 ("fix(nudge): deliver queued nudges via hook trigger, not raw tmux send-keys"). All changes from this PR (synchronous watchAndDeliver, QueueLen, idle watcher tests) are present on main, with some improvements (error logging in QueueLen, WaitForIdle consecutive-poll guard). Thank you @DreadPirateRobertz for the contribution!

PR #2509, opened 2026-03-07
fix: mount FOLDER to /gt with runtime workspace initialization

pento:
## Summary

- Changes Docker volume mount from `${FOLDER}:${FOLDER}` to `${FOLDER}:/gt` so the host folder is available at the gastown workspace path
- Moves `gt install /gt --git` from build-time to a runtime entrypoint script (`docker-entrypoint.sh`), so it runs after the volume is mounted rather
than being overlaid by it
- First boot does a full install; subsequent boots use `--force` to refresh settings while preserving existing workspace state

## Test plan

- [ ] `FOLDER=/tmp/test-gt GIT_USER=Test GIT_EMAIL=test@test.com docker compose up --build -d`
- [ ] Verify workspace initialized: `docker compose exec gastown ls /gt/mayor/town.json`
- [ ] Verify host directory received files: `ls /tmp/test-gt/mayor/town.json`
- [ ] Restart and verify idempotent: `docker compose restart && docker compose logs gastown | grep "Refreshing"`
- [ ] Clean up: `docker compose down && rm -rf /tmp/test-gt`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -61,6 +61,10 @@ RUN git config --global credential.helper store && \
     dolt config --global --add user.name "${GIT_USER}" && \
     dolt config --global --add user.email "${GIT_EMAIL}"
 
-RUN /app/gastown/gt install /gt --git
+COPY --chown=agent:agent docker-entrypoint.sh /app/docker-entrypoint.sh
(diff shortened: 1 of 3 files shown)


steveyegge:
Superseded by #2512, which was fix-merged to main.

PR #2512, opened 2026-03-07
fix: persist home directory in a managed Docker volume

pento:
> **Note:** This PR includes the changes in the `fix/docker-mount-folder-to-gt` branch (#2509). That PR should be reviewed/merged first.

## Summary
- Add `agent-home` named volume mounted at `/home/agent` so user-level config (`.claude`, `.config/gh`, `.gitconfig`, etc.) survives container rebuilds
- Move git/dolt identity config from build-time args to runtime env vars (`GIT_USER`, `GIT_EMAIL`) applied by the entrypoint on every start — the image
  is now user-agnostic and fully cacheable
- Remove `GIT_USER`/`GIT_EMAIL` build args and validation gate from Dockerfile

## Test plan
- [ ] `docker compose build` succeeds without any `GIT_USER`/`GIT_EMAIL` args
- [ ] `docker compose up -d` — first run: volume created, `git config --global user.name` returns correct value
- [ ] Exec in, create a file in `~`, `docker compose down && docker compose up -d` — file persists
- [ ] Rebuild with a different `GIT_USER` value — verify new name appears in `git config --global user.name`

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,17 +1,8 @@
 # Run with
-# docker build -t gastown:latest -f Dockerfile . \
-#   --build-arg GIT_USER=TestUser --build-arg GIT_EMAIL=test@example.com
+# docker build -t gastown:latest -f Dockerfile .
 FROM docker/sandbox-templates:claude-code
 
-ARG GIT_USER
-ARG GIT_EMAIL
 ARG GO_VERSION=1.25.6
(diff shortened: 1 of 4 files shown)


steveyegge:
Fix-merged to main as 8137131d. Thanks Gary!

PR #2520, opened 2026-03-08
fix: guard against bd v0.58.0 non-JSON output in list parsers

dannomayernotabot:
## Summary

- Use existing `isJSON()` / `isJSONBytes()` helpers as pre-checks before `json.Unmarshal` on `bd list` output
- `mailbox.go`: replace redundant post-Unmarshal fallback with clean pre-check using existing `isJSON()`
- `beads_rig.go`: add `isJSONBytes()` guard to `ListRigBeads()` which was unprotected

Supersedes PR #2399 — same intent, cleaner implementation (no new imports, no redundant checks).

## Context

`bd v0.58.0` returns plain text (e.g. "No issues found.") instead of empty JSON arrays when result sets are empty. This broke `gt hook`, `gt mail inbox`, and `gt patrol`.

## Test plan

- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] `go test ./internal/mail/...` passes
- [x] `go test ./internal/beads/...` passes
- [x] Full test suite — only pre-existing failures in tmux/cmd packages (unrelated)

### DIFF
--- a/internal/beads/beads_rig.go
+++ b/internal/beads/beads_rig.go
@@ -249,6 +249,9 @@ func (b *Beads) ListRigBeads() (map[string]*RigFields, error) {
 		return nil, err
 	}
 
+	if !isJSONBytes(out) {
+		return nil, nil
+	}
 	var issues []*Issue
 	if err := json.Unmarshal(out, &issues); err != nil {
 		return nil, fmt.Errorf("parsing bd list output: %w", err)
--- a/internal/mail/mailbox.go
+++ b/internal/mail/mailbox.go
@@ -161,11 +161,13 @@ func (m *Mailbox) listFromDir(beadsDir string) ([]*Message, error) {
 			return nil, err
 		}
 
+		// bd v0.58.0 returns plain text (e.g. "No issues found.") for
+		// empty result sets instead of JSON. Skip non-JSON output.
+		if !isJSON(stdout) {
+			continue
+		}
 		var msgs []BeadsMessage
 		if err := json.Unmarshal(stdout, &msgs); err != nil {
-			if len(stdout) == 0 || string(stdout) == "null" || !isJSON(stdout) {
-				continue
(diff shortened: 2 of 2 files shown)


DreadPirateRobertz:
## Code Review: PR #2520 — Guard against bd v0.58.0 non-JSON output

**Reviewer:** gastown/crew/zhora (PM)

**Summary:** Adds `isJSONBytes()`/`isJSON()` checks before `json.Unmarshal` in three locations where bd v0.58.0 returns plain text instead of JSON for empty results.

**Assessment: APPROVE**

Clean, defensive fix. The pattern is correct — check before unmarshaling rather than handling the error after. Three locations:
1. `beads_rig.go:ListRigBeads()` — returns `nil, nil` for non-JSON (silent skip)
2. `mailbox.go:listFromDir()` — primary label query (skip non-JSON)
3. `mailbox.go:listFromDir()` — CC label query (skip non-JSON)

**Minor concern:** In `beads_rig.go`, returning `nil, nil` silently hides the non-JSON case. A logged warning might help debugging. But this matches the existing pattern for empty results, so it's consistent.

No test changes — this is a runtime compatibility guard. Acceptable for a version-specific workaround.

DreadPirateRobertz (approved):
**Review (gastown/crew/zhora)**

LGTM. Clean, minimal fix that guards `ListRigBeads` and `listFromDir` against non-JSON `bd list` output from bd v0.58.0. Matches the existing `isJSON`/`isJSONBytes` pattern already in `ListAgentBeads`.

One note: there are 12+ other `json.Unmarshal` sites in `internal/beads/` (queue, group, escalation, sling_context, channel) that parse `bd list` output without this guard. They're all potentially vulnerable to the same crash. This PR correctly fixes the two most-reported sites — a follow-up to guard remaining callers would be worthwhile.

PR #2524, opened 2026-03-08
fix: hook show shorthand normalization and stale hook_bead (GH#2371)

### LINKED ISSUE #2371: Fix hook show shorthand normalization and stale polecat issue mapping
## Summary
Two related state/reporting bugs make hook status inconsistent:

1. `gt hook show <rig>/<name>` can return empty even when `<rig>/polecats/<name>` has hooked work.
2. `gt polecat list` / manager state can report stale issue IDs when agent `hook_bead` is stale.

### Bug 1: hook show shorthand normalization
1. Hook a bead to assignee `gastown/polecats/toast`.
2. Run:
   - `gt hook show gastown/polecats/toast`
   - `gt hook show gastown/toast`

### Expected
Both commands return the same hooked bead.

### Actual
Canonical path works, shorthand can show empty due target normalization mismatch.

### Bug 2: stale polecat issue mapping
1. Agent bead has `hook_bead=<old-issue>`.
2. Current work bead is different issue with `status=hooked` and `assignee=<polecat>`.
3. Query polecat manager/list status.

### Expected
Current hooked bead should be reported.

### Actual
Stale `hook_bead` may be reported as current issue.

## Suspected cause
- Hook show path normalization converts through session form in a way that can fail to map back to canonical assignee IDs.
- Polecat state lookup trusts legacy `hook_bead` too early instead of preferring direct hooked-bead state.

## Notes
A fix branch with tests is prepared and will be linked from the PR.

DreadPirateRobertz:
## Summary
- **Bug 1**: `gt hook show gastown/toast` returns empty even when `gastown/polecats/toast` has hooked work. Adds direct shorthand expansion as fallback in `normalizeHookShowTarget` when the session name roundtrip fails.
- **Bug 2**: `gt mol status` reads stale `hook_bead` from agent bead instead of querying authoritative source. Since `updateAgentHookBead` is a no-op (since hq-l6mm5), the agent bead's `hook_bead` field is never updated. Now always queries `status=hooked + assignee`.

## Test plan
- [x] `go build ./internal/cmd/` compiles clean
- [ ] `gt hook show gastown/<polecat>` resolves to canonical `gastown/polecats/<polecat>`
- [ ] `gt mol status` shows current hooked bead, not stale one

Fixes #2371

### DIFF
--- a/internal/cmd/hook.go
+++ b/internal/cmd/hook.go
@@ -555,6 +555,31 @@ func normalizeHookShowTarget(target string) string {
 	if identity, err := session.ParseAddress(target); err == nil {
 		return identity.Address()
 	}
+
+	// Direct shorthand expansion: rig/name → rig/polecats/name or rig/crew/name.
+	// This handles the case where the session name roundtrip fails due to
+	// uninitialized prefix registry. See GH#2371.
+	parts := strings.Split(target, "/")
+	if len(parts) == 2 && parts[0] != "" && parts[1] != "" {
+		name := parts[1]
+		// Check for known roles — don't expand those
+		switch strings.ToLower(name) {
+		case "witness", "refinery", "mayor", "deacon":
+			// Already a valid canonical address
+		default:
+			// Check if it's a crew member by looking for the directory
+			townRoot := detectTownRootFromCwd()
+			if townRoot != "" {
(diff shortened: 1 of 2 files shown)


steveyegge:

PR #2540, opened 2026-03-08
fix: persist formula --var context for assignees

### LINKED ISSUE #2538: gt sling --var formula inputs are not durably visible to the assignee
## Bug

Formula inputs passed via `gt sling ... --var key=value` are used to instantiate the formula, but they are not durably visible to the assignee afterward.

The assignee can receive the hooked formula/wisp and see the workflow, but the instantiated `--var` payload is missing from the normal discovery paths like:

- `gt prime`
- `gt hook` / `gt mol status`
- attached workflow context rendered from bead metadata

## Why this matters

`--var` appears to be the supported mechanism for structured formula inputs, while `--args` is the durable human-readable guidance path.

That means users can successfully do this:

```bash
gt sling mol-release gastown/crew/max --var version=1.2.3
```

and the formula will instantiate correctly, but the assignee may not be able to see `version=1.2.3` in the normal execution context.

For root-only wisps / inline formula rendering, this is especially problematic because there may be no other obvious place for the assignee to recover those instantiated values.

## Repro

Use any formula that depends on `--var` input, for example:

```bash
gt sling <formula> <target> --var version=1.2.3 --var channel=stable
```

Then inspect the target side with:

```bash
gt prime
gt hook
gt hook --json
```

or inspect the hooked bead metadata directly.

## Expected

The assignee should be able to see the instantiated formula inputs durably, similar to how `--args` is surfaced.

Examples of acceptable behavior:

- formula vars are persisted in attachment metadata
- `gt prime` shows the vars in the attached workflow section
- `gt hook` / `gt mol status` surfaces the vars in human and JSON output

## Actual

The formula/wisp is instantiated with the provided `--var` values, but those values are not persisted into the attachment metadata used by prime/hook/status rendering.

Observed behavior:
- `--var` values are forwarded into formula instantiation
- `--args` is persisted and shown durably
- `--var` values are not durably shown to the assignee

## Investigation

This does **not** appear to be a deprecated path. `--var` is still the supported CLI path for formula inputs:

- `gt sling --help` documents `--var`
- the CLI still wires `--var` into formula/wisp creation
- formula examples/tests still use `--var`

Historical context suggests this is a longstanding gap rather than a newly introduced storage regression:

- `9daee341` (`2025-12-25`) added standalone formula slinging and `--var`
- `9462208f` (`2025-12-26`) added durable no-tmux transport for `--args` via `attached_args`
- I did not find an equivalent historical `attached_vars` persistence path
- `75c1898d` (`2026-02-08`) refactored metadata storage but preserved the same omission
- `b8f79dc8` (`2026-02-26`) made root-only wisps / inline formula rendering the norm, which made the missing durable var visibility much more noticeable

## Likely root cause

The code path uses `--var` for formula instantiation, but the assignee-facing context paths read only attachment metadata, and that metadata historically persisted only:

- attached formula
- attached molecule
- attached args
- dispatcher / other workflow flags

not formula vars.

So the formula is created with the right inputs, but the assignee-facing data plane never receives those inputs.

## Related

- Related but not duplicate: #2431
  That issue is about standalone formula sling hook visibility itself.
  This issue is about formula input payload visibility after the hook is visible.

corymhall:
## Summary
- persist `gt sling --var` inputs into durable attachment metadata as `attached_vars`
- surface formula vars in `gt prime`, `gt hook` / `gt mol status`, and JSON status output
- add focused regression tests for beads round-trip, sling persistence, prime output, and mol status output
- keep sling hook writes at `BD_DOLT_AUTO_COMMIT=off`

Fixes #2538

## Important Stacking Note
This branch is logically stacked on top of #2433.

In the code/history sense, `fix/formula-var-context` is a follow-up to `fix/formula-hook-status`, because it depends on that branch's hook/formula visibility behavior:
- direct bead tracking for hooked formula work
- full hooked-bead re-fetch so formula attachment metadata is available to status/rendering paths

I attempted to create a true stacked PR against `steveyegge/gastown`, but GitHub could not represent that stack because the base branch for #2433 exists only on the fork, not on upstream.

Because of that limitation, this PR is opened against `upstream/main`, but reviewers should treat it as:
- the #2433 hook-visibility fixset, plus
- the #2538 follow-up that persists and surfaces `--var` context

## Note On Recent Upstream Movement
Upstream has continued evolving nearby hook/status behavior.

I still believe the core direction here is correct:
- `--var` remains a supported formula-input path
- formula inputs should be durably visible to the assignee
- assignee-facing views should surface that context directly

If maintainers prefer a different metadata shape or a smaller integration now that upstream has moved, I’m very open to reshaping the implementation while keeping that behavior goal intact.

## Verification
- `go test ./internal/beads -run 'TestParseAttachmentFields|TestFormatAttachmentFields|TestAttachmentFieldsRoundTrip|TestNoMergeField' -count=1`
- `go test ./internal/cmd -run 'TestRunSlingFormulaPersistsVarContext|TestCheckSlungWork_StandaloneFormulaUsesWorkflowOutput|TestOutputMoleculeStatus_StandaloneFormulaShowsVars|TestSlingSetsDoltAutoCommitOff' -count=1`
- live check from a real hooked formula workspace: `gt hook` rendered the hooked formula plus all persisted `--var` inputs

## Review Notes
- Ran `review-implementation` with Codex and Claude on the stacked branch context
- Both reviews agreed there were no blocking findings for #2538 in the stacked configuration

### DIFF
--- a/internal/cmd/hook.go
+++ b/internal/cmd/hook.go
@@ -375,7 +375,6 @@ func runHook(_ *cobra.Command, args []string) error {
 	for attempt := 1; attempt <= hookMaxRetries; attempt++ {
 		if err := BdCmd("update", beadID, "--status=hooked", "--assignee="+agentID).
 			Dir(townRoot).
-			WithAutoCommit().
 			Run(); err != nil {
 			lastHookErr = err
 			if attempt < hookMaxRetries {
--- a/internal/cmd/sling_helpers.go
+++ b/internal/cmd/sling_helpers.go
@@ -988,7 +988,6 @@ func hookBeadWithRetry(beadID, targetAgent, hookDir string) error {
 	for attempt := 1; attempt <= maxRetries; attempt++ {
 		err := BdCmd("update", beadID, "--status=hooked", "--assignee="+targetAgent).
 			Dir(hookDir).
-			WithAutoCommit().
 			Run()
 		if err != nil {
 			lastErr = err


steveyegge:
Closing: this was stacked on #2433 which we're closing — the bd list --json plain-text fallback is no longer needed since bd v0.59.0 returns valid JSON in all cases. The --var persistence feature is a good idea though. If you'd like to re-submit it as a standalone PR (without the plain-text parser layer from #2433), we'd be happy to review it. Thanks @corymhall\!

PR #2541, opened 2026-03-08
status/up: surface DND in gt status and auto-reset muted mode on gt up

ChicK00o:
Summary

 This PR improves operational clarity and reliability around notifications by:

 1. Showing DND state in gt status
 2. Auto-resetting muted DND for the current agent on gt up

 This addresses the failure mode where orchestration appears stalled because the active agent (e.g., Mayor) is in DND/muted mode and not receiving nudges.

 ────────────────────────────────────────────────────────────────────────────────

 Diff vs main

 ### 1) gt status now reports DND explicitly

 #### New status model field

 - Added DND *DNDInfo to TownStatus (JSON + text path)

 #### New type

 - Added DNDInfo:
     - enabled (bool)
     - level (string)
     - agent (agent bead id)

 #### DND detection

 - Added detectCurrentDNDStatus(townRoot):
     - resolves current role context (GetRoleWithContext)
     - computes agent bead id (getAgentBeadID)
     - reads notification level from beads (GetAgentNotificationLevel)
     - defaults to normal if unset/unavailable

 #### Text output changes

 - outputStatusText now prints a DND section near the top:
     - 🔕 DND: on (hq-mayor) / 🔔 DND: off (...)
     - descriptive line: muted vs normal notifications

 #### Agent line/verbose visibility

 - AgentRuntime now carries notification_level
 - discoverGlobalAgents and discoverRigAgents now parse and populate notification level from agent fields
 - buildStatusIndicator appends 🔕 badge when muted
 - verbose agent details print notify: 🔕 muted (DND) when applicable

 ────────────────────────────────────────────────────────────────────────────────

 ### 2) gt up now clears muted DND for current agent

 #### New helper

 - Added disableCurrentAgentDND(townRoot) (bool, error):
     - resolves current role/agent bead
     - checks current level
     - if muted, updates to normal
     - non-blocking fallback if role/bead isn’t available

 #### Startup behavior change

 - Early in runUp, call disableCurrentAgentDND
 - If reset occurs, prints:
     - DND was enabled; reset to normal for current agent
 - If detection fails, continues startup with warning only (does not break gt up)

 ────────────────────────────────────────────────────────────────────────────────

 Why this is needed

 Before this change:
 - DND could be active with no clear visibility in main status dashboard
 - users/operators might interpret missing nudges as system failure
 - gt up did not recover this state automatically

 After this change:
 - DND state is immediately visible in status output
 - startup self-heals muted mode for the active operator context
 - orchestration reliability is improved during restarts/recoveries

 ────────────────────────────────────────────────────────────────────────────────

 Tests added

 - TestBuildStatusIndicator_DNDMutedShowsBadge
 - TestOutputStatusText_IncludesDNDSection

 Also verified manually in a live town:
 - gt dnd on → gt status shows 🔕 DND: on (...)
 - gt dnd off → gt status shows 🔔 DND: off (...)

 ────────────────────────────────────────────────────────────────────────────────

 Backward compatibility / risk

 - No behavior change for users not using DND
 - gt up DND reset is scoped to current role context only
 - non-fatal if role/bead context is unavailable
 - JSON output gains optional dnd object (additive change)

### DIFF
--- a/internal/cmd/status.go
+++ b/internal/cmd/status.go
@@ -65,6 +65,7 @@ type TownStatus struct {
 	Name     string         `json:"name"`
 	Location string         `json:"location"`
 	Overseer *OverseerInfo  `json:"overseer,omitempty"` // Human operator
+	DND      *DNDInfo       `json:"dnd,omitempty"`      // Current agent DND status
 	Daemon   *ServiceInfo   `json:"daemon,omitempty"`   // Daemon status
 	Dolt     *DoltInfo      `json:"dolt,omitempty"`     // Dolt server status
 	Tmux     *TmuxInfo      `json:"tmux,omitempty"`     // Tmux server status
@@ -108,6 +109,13 @@ type OverseerInfo struct {
 	UnreadMail int    `json:"unread_mail"`
 }
 
+// DNDInfo represents Do Not Disturb status for the current agent context.
+type DNDInfo struct {
+	Enabled bool   `json:"enabled"`
+	Level   string `json:"level"`
+	Agent   string `json:"agent,omitempty"`
+}
+
 // AgentRuntime represents the runtime state of an agent.
 type AgentRuntime struct {
 	Name         string `json:"name"`                    // Display name (e.g., "mayor", "witness")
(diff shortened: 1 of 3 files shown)


steveyegge:
Fix-merged to main in 93c36e59. Thanks for the contribution, @ChicK00o! Clean implementation that solves a real operational visibility gap.

PR #2556, opened 2026-03-08
fix: GetRigLED shows parked/docked indicator even when sessions running

### LINKED ISSUE #2555: fix: GetRigLED shows green for parked/docked rigs when sessions still running
## Bug

`GetRigLED()` checks session state (witness/refinery running) BEFORE checking operational state (parked/docked). This means:

1. If a rig is parked/docked but sessions haven't been killed yet (or daemon restarted them during the race window), the LED shows 🟢 instead of 🅿️/🛑
2. After undocking, the LED correctly shows green, but users may see confusing transient states

### Root Cause

```go
func GetRigLED(hasWitness, hasRefinery bool, opState string) string {
    if hasWitness && hasRefinery {
        return "🟢"  // BUG: returns green even if DOCKED/PARKED
    }
    if hasWitness || hasRefinery {
        return "🟡"  // BUG: returns yellow even if DOCKED/PARKED
    }
    // Only reaches state check if NO sessions running
    switch opState {
    case "PARKED": return "🅿️"
    case "DOCKED": return "🛑"
    default: return "⚫"
    }
}
```

### Fix

Check operational state FIRST. If rig is parked/docked, always show the park/dock indicator regardless of session state.

### Related Issue

The daemon `isRigOperational()` has a race condition where `gt rig dock` writes to bead labels (Dolt) but not wisp config. The daemon checks wisp first (fast) then beads (slow). If the daemon heartbeat fires before the Dolt write is visible, it may restart sessions on a docked rig — causing vampire token drain.

**Fix for race condition**: `gt rig dock` should ALSO write `status=docked` to the wisp layer for immediate local visibility, same as `gt rig park` does.

### Files

- `internal/cmd/rig.go:637` — `GetRigLED()`
- `internal/cmd/rig_dock.go` — missing wisp write
- `internal/daemon/daemon.go:1433` — `isRigOperational()`

DreadPirateRobertz:
## Summary
- `GetRigLED()` now checks operational state (PARKED/DOCKED) **before** session state, so a docked/parked rig always shows 🅿️/🛑 regardless of whether sessions are still running
- `gt rig dock` now writes `status=docked` to the wisp layer for immediate local visibility, preventing the daemon from restarting sessions in the window before the Dolt bead write becomes visible
- `gt rig undock` clears the wisp status alongside removing the bead label
- Fixes duplicate `wantPort` declaration in doltserver.go that broke the build

Fixes #2555

## Test plan
- [x] `TestGetRigLED` updated to verify parked/docked takes precedence over session state
- [x] All LED test cases pass
- [x] Build passes clean

### DIFF
--- a/internal/cmd/rig.go
+++ b/internal/cmd/rig.go
@@ -635,20 +635,22 @@ func runRigAdd(cmd *cobra.Command, args []string) error {
 //   - 🅿️ = parked (intentionally paused)
 //   - 🛑 = docked (global shutdown)
 func GetRigLED(hasWitness, hasRefinery bool, opState string) string {
+	// Check operational state first — a parked/docked rig should always
+	// show its park/dock indicator, even if sessions are still running
+	// (e.g. daemon restarted them before seeing the state change).
+	switch opState {
+	case "PARKED":
+		return "🅿️"
+	case "DOCKED":
+		return "🛑"
+	}
 	if hasWitness && hasRefinery {
 		return "🟢"
 	}
 	if hasWitness || hasRefinery {
 		return "🟡"
 	}
-	switch opState {
-	case "PARKED":
-		return "🅿️"
-	case "DOCKED":
-		return "🛑"
-	default:
-		return "⚫"
-	}
+	return "⚫"
 }
 
 // rigStatePriority returns a sort priority for a rig's state.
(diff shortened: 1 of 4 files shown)


steveyegge:

PR #2565, opened 2026-03-08
fix: emit feed event when polecat crashes via tmux pane-died hook

### LINKED ISSUE #2534: bug: polecats die silently — no feed event on crash
## Problem

When a polecat's worker subprocess crashes, no event is emitted to the activity feed. The parent claude process stays alive (Ss+ in ps) but the actual worker PID in `.runtime/agent.lock` is dead. The only way to detect this is manually checking `ps -p <pid>`.

The overseer sometimes catches it, but not reliably — and the feed shows nothing between `session_start` and eventual `done`.

## Expected

A `session_crash` or `session_error` feed event when a polecat's worker subprocess exits unexpectedly, including the exit code and any stderr.

## Observed

- Lock file PIDs dead, parent process idle
- Feed shows only `session_start` events, then silence
- `gt polecat list` still shows "working"

DreadPirateRobertz:
## Summary
- `gt log crash` (called by tmux pane-died hook) now emits a `TypeSessionDeath` feed event for crash and kill exits
- Previously only logged to townlog, making polecat crashes invisible in the activity feed
- Uses the same `SessionDeathPayload` format as daemon health checks, doctor cleanup, and `gt done`

**Root cause**: The tmux pane-died hook → `gt log crash` → townlog write was the complete path. No `events.LogFeed()` call existed, so the feed/curator system never saw crash events.

Fixes #2534

## Test plan
- [ ] Kill a polecat worker subprocess, verify `TypeSessionDeath` event appears in `.events.jsonl`
- [ ] Verify `gt polecat list` / feed shows crash events
- [ ] Normal exit (code 0) should NOT emit a death event (only crash/kill)
- [ ] SIGINT (code 130) should emit a death event with "interrupted" reason

### DIFF
--- a/internal/cmd/log.go
+++ b/internal/cmd/log.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/spf13/cobra"
+	"github.com/steveyegge/gastown/internal/events"
 	"github.com/steveyegge/gastown/internal/style"
 	"github.com/steveyegge/gastown/internal/townlog"
 	"github.com/steveyegge/gastown/internal/workspace"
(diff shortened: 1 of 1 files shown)


steveyegge:

PR #2566, opened 2026-03-08
fix: clean dirty worktree before branch-only polecat reuse

DreadPirateRobertz:
## Summary
- Fixes GH#2536: polecat reuse fails with "local changes would be overwritten by checkout" when previous session left uncommitted changes
- Adds `git reset --hard HEAD` + `git clean -fd` before `checkout -b` in `ReuseIdlePolecat` — safe because idle polecats have already submitted their work to the refinery
- Adds `CleanUntracked()` method to the Git type
- Eliminates noisy/slow full-repair fallback for this common case

## Test plan
- [x] All git and polecat tests pass
- [ ] Manual: create dirty state in an idle polecat worktree, then `gt sling` to reuse it — should succeed without "full repair" fallback

### DIFF
--- a/internal/git/git.go
+++ b/internal/git/git.go
@@ -1012,6 +1012,12 @@ func (g *Git) ResetHard(ref string) error {
 	return err
 }
 
+// CleanUntracked removes untracked files and directories.
+// Equivalent to: git clean -fd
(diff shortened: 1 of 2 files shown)


steveyegge:

PR #2568, opened 2026-03-08
fix: add --flat to all bd list --json calls to prevent tree mode breaking JSON

### LINKED ISSUE #2499: gt doctor fails agent-beads-exist check due to bd list --json output format and ephemeral bead label skipping

DreadPirateRobertz:
## Summary
- `bd list --json` output can be corrupted when tree-view is enabled (the default), causing silent JSON unmarshal failures
- Added `--flat` flag to all 6 `bd list --json` call sites: agent, queue, channel, rig, escalation, group
- The main `beads.go` List() already had `--flat` — these infrastructure list functions were missed
- Also fixes duplicate `wantPort` declaration in doltserver.go

Fixes #2499

## Test plan
- [x] Build passes
- [x] All existing tests pass
- [ ] `gt doctor` agent-beads-exist check should pass with tree-view enabled

### DIFF
--- a/internal/beads/beads_agent.go
+++ b/internal/beads/beads_agent.go
@@ -630,7 +630,7 @@ func (b *Beads) ListAgentBeads() (map[string]*Issue, error) {
 	// doctor checks (for example, validating gt:agent labels).
 	// Agent beads are type=agent (infrastructure), hidden by bd list default filter.
 	// Use --include-infra so they appear in results.
-	out, err := b.run("list", "--label=gt:agent", "--include-infra", "--json")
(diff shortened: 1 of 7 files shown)


steveyegge:
Closing: the underlying issue (bd list --json returning non-JSON in tree mode) has been fixed in bd v0.59.0. The --flat flag is no longer needed as a workaround. Also, this branch has 32 commits and 12K+ lines of shared base noise making it unreviewable.

PR #2518, opened 2026-03-08
feat: add Gas City role validation command

seanwessmith:
## Summary
- add a user-facing `gt gascity role validate <file>` command
- harden the provisional Gas City role parser with stricter schema validation and normalized JSON output
- add workspace-backed CLI tests and parser coverage
- fix lingering sling compatibility regressions so `go test ./internal/cmd` is green again

## Details
This builds the earlier Gas City role parser prototype into a more deployable slice.

The new validator:
- loads built-in provider presets
- loads town-local `settings/agents.json` when run inside a Gas Town workspace
- validates a Gas City role spec from TOML
- prints either a human-readable validation summary or normalized JSON

The parser now:
- rejects unknown TOML keys
- validates scope/workdir consistency more strictly
- exposes JSON tags on the normalized spec shape

This PR also fixes two unrelated-but-blocking command test regressions in the sling path:
- `bd --allow-stale` support detection now re-probes when the `bd` binary on `PATH` changes
- sling now preserves `BD_DOLT_AUTO_COMMIT=off` through helper paths that previously overrode it back to `on`

Scope note: this PR validates and normalizes a provisional Gas City role schema via `gt gascity role validate`. It does not yet wire that schema into runtime loading or execution.

## Verification
- `go test ./internal/gascity`
- `go test ./internal/cmd`
- `go build ./cmd/gt`
- `go run ./cmd/gt gascity role validate docs/examples/gascity-role.example.toml`
- `go run ./cmd/gt gascity role validate docs/examples/gascity-role.example.toml --json`

## Notes
The local worktree still has unrelated untracked runtime artifacts (`.agents/`, `internal/events/...`) that are not part of this PR.

PR #2571, opened 2026-03-08
fix: add explicit bd auto-commit off mode

seanwessmith:
## Summary
- add `BdCmd.WithAutoCommitOff()` for commands that must write without forcing a Dolt auto-commit
- use the explicit off mode for `hookBeadWithRetry()` in sling
- add builder coverage for the new mode and its interaction with the existing force-off override

## Why
The self-sling hook write is part of a larger sling flow and should not force its own Dolt commit. Relying on inherited environment state was too implicit. This makes the hook update opt into `BD_DOLT_AUTO_COMMIT=off` directly at the command site.

## Verification
- `go test ./internal/cmd`

### DIFF
--- a/internal/cmd/bd_helpers.go
+++ b/internal/cmd/bd_helpers.go
@@ -13,13 +13,13 @@ import (
 // It provides a fluent API for configuring environment variables,
 // working directory, and I/O settings common to bd CLI invocations.
 type bdCmd struct {
-	args       []string
-	dir        string
-	env        []string
-	stderr     io.Writer
-	autoCommit bool
-	gtRoot     string
-	beadsDir   string
+	args           []string
+	dir            string
+	env            []string
+	stderr         io.Writer
+	autoCommitMode string
+	gtRoot         string
+	beadsDir       string
 }
 
 // BdCmd creates a new bd command builder with the given arguments.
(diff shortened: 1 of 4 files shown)


steveyegge:
Thanks for the auto-commit off fix — the explicit WithAutoCommitOff() approach is the right call over relying on inherited env state.

However, this PR appears to include the same bundled changes from #2518 (Gas City role validation, step cost tracking, BdSupportsAllowStale cache fix, etc.). Could you rebase this so it only contains the auto-commit off changes (bd_helpers.go, sling.go, and their tests)?

We left the same feedback on #2518 — splitting each concern into its own PR will get them reviewed and merged much faster.

seanwessmith:
The branch is now a single commit on top of main.

One note on the current red CI: the failing integration tests appear to be unrelated to this PR’s scope. They’re failing on bd list ... --flat with older bd binaries in CI (unknown flag: --flat), and that --flat behavior is coming from main in internal/beads/beads.go, not from the auto-commit-off changes here.

I’ve kept #2571 scoped to the sling auto-commit fix per review guidance rather than rebundling that separate compatibility issue here.

steveyegge:
The core fix (removing WithAutoCommit from hookBeadWithRetry) already landed on main. The WithAutoCommitOff method and GT_FORCE_BD_AUTOCOMMIT_OFF mechanism add complexity for a scenario that's already handled. If you'd like to pursue the explicit WithAutoCommitOff API, consider a slimmer PR without the global force override. Thank you for the thorough analysis\!

PR #2594, opened 2026-03-09
fix: surface hooked formula wisps in gt hook

### LINKED ISSUE #2431: gt sling <formula> can report success but leave target gt hook empty

corymhall:
## Summary
- make `gt hook` / `gt mol status` prefer direct bead tracking and only use the legacy hook-slot path as validated fallback
- re-fetch the full hooked bead so formula metadata is available and standalone formula wisps get workflow-oriented guidance instead of `gt mol attach`

Fixes #2431

## Scope
This is the slimmer version of the earlier branch.

I removed the `bd list --json` plain-text table parser/fallback and kept only the core behavioral fix for hook visibility:
- direct hooked bead state is authoritative
- the legacy agent `hook_bead` slot is only fallback
- full hooked-bead re-fetch preserves formula metadata for rendering

## Why The Parser Was Removed
After retesting more precisely against the actual `gt` code path, the relevant list command is `bd list --json --flat`.

On my local `bd version 0.59.0 (dev)`, `bd list --json` without `--flat` could still be non-JSON, but `bd list --json --flat` returned valid JSON in the affected environment, and the core `gt hook` fix still worked without the parser.

So this PR now focuses on the main source-issue behavior rather than the beads-side output workaround.

## Verification
- `go test ./internal/beads -run 'TestParseAttachmentFields|TestFormatAttachmentFields' -count=1`
- built a temporary `gt` binary from this branch and ran a real `gt hook` / `gt hook --json` check in a hooked formula workspace
- confirmed the hooked formula bead was visible even with the parser removed

### DIFF
--- a/internal/beads/beads.go
+++ b/internal/beads/beads.go
@@ -10,8 +10,6 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
-	"regexp"
-	"strconv"
 	"strings"
 	"sync"
 	"time"
@@ -574,16 +572,12 @@ func (b *Beads) List(opts ListOptions) ([]*Issue, error) {
 		return nil, err
 	}
 
-	// bd list --json may return plain text instead of JSON in some configurations.
-	// Handle explicit empty outputs gracefully, but parse issue table rows when
-	// possible so callers don't silently lose hooked work visibility.
+	// bd list --json may return plain text like "No issues found." instead of JSON
+	// when there are no results. Handle that explicit empty case gracefully.
 	if len(out) == 0 || !isJSONBytes(out) {
 		text := strings.TrimSpace(string(out))
 		if text == "" || strings.Contains(text, "No issues found.") {
 			return nil, nil
-		}
-		if issues := parsePlainListOutput(text); len(issues) > 0 {
-			return issues, nil
 		}
 		return nil, fmt.Errorf("bd list --json returned non-JSON output: %q", firstLine(text))
 	}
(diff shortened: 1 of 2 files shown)


corymhall:
Closing this version in favor of a narrower follow-up. Fresh repros on current upstream main show formula wisps are visible again, but still rendered as naked work in gt hook / gt mol status. I’m going to resubmit a smaller PR focused only on formula-aware hook rendering and next-action behavior.

PR #2607, opened 2026-03-09
fix(convoy): route HQ convoy reads via town root

### LINKED ISSUE #2604: convoy read commands can target the wrong DB when BEADS_DIR leaks into HQ routing
## Problem
Some convoy read paths still invoke `bd` from `townRoot/.beads` or with an inherited `BEADS_DIR`.

That breaks HQ convoy reads because these commands are supposed to route from town root and let `bd` resolve the correct HQ database from there.

## What a reviewer would see
Examples from a real town:
- `gt convoy list --all --json`
- `gt convoy status hq-cv-p5iot`
- convoy stranded scans returning missing or misrouted data

Symptoms vary with the caller environment, but the common failure mode is: the convoy exists, yet read-side commands cannot resolve it reliably.

## Why it happens
```text
Current path:
  gt convoy status
    -> bd runs from townRoot/.beads or inherits BEADS_DIR
    -> bd targets the wrong DB

Expected path:
  gt convoy status
    -> bd runs from townRoot
    -> inherited BEADS_DIR is stripped
    -> bd resolves the correct HQ DB
```

## Expected behavior
All convoy read/status/list/scan paths should follow the same routing rules already used elsewhere in the CLI:
- run `bd` from town root for HQ reads
- strip inherited `BEADS_DIR`
- leave create/update semantics unchanged

## Out of scope
This is only about read-path routing.
It does not cover cross-rig tracked dependency counting or the separate live visibility/Dolt investigation.

kirtangajjar:
## Problem

`gt convoy list`, `gt convoy status`, and stranded or staged convoy scans could read the wrong beads database for HQ convoys.

Typical failure mode:
- command runs from `townRoot/.beads` or inherits a stale `BEADS_DIR`
- `bd` resolves against the wrong DB
- HQ convoys look missing, stale, or incomplete

```text
Before
  gt convoy status
    -> bd show from townRoot/.beads or inherited BEADS_DIR
    -> wrong DB selected

After
  gt convoy status
    -> bd show from townRoot
    -> inherited BEADS_DIR stripped
    -> HQ convoy reads stay in the intended beads DB
```

## Related Work

This branch includes the shared CI base from [#2658](https://github.com/steveyegge/gastown/pull/2658), which now covers the cross-platform harness fixes and the deterministic convoy short-ID test cleanup.

## Feature-only Review

Use this compare link to review only the routed-convoy change on top of the shared base:
- <https://github.com/kirtangajjar/gastown/compare/codex/shared-ci-main-red-2610...codex/convoy-town-root-routing-2604>

Line-level rationale is captured in the inline review comments on the diff.

## Validation

Targeted convoy validation:
- `go test ./internal/cmd -run 'TestRunConvoyList_UsesTownRootAndStripsBeadsDir|TestRunConvoyStatus_UsesTownRootAndStripsBeadsDir|TestCollectConvoyBeads_ExternalTrackedIDs|TestTransitionConvoyToOpen_StagedReady'`

Full branch validation is covered by the local exact workflow reruns and the shared CI base in [#2658](https://github.com/steveyegge/gastown/pull/2658).

Closes #2604

### DIFF
--- a/internal/cmd/convoy.go
+++ b/internal/cmd/convoy.go
@@ -26,13 +26,19 @@ import (
 	"github.com/steveyegge/gastown/internal/workspace"
 )
 
+var convoyIDEntropy io.Reader = rand.Reader
+
 // generateShortID generates a collision-resistant convoy ID suffix using base36.
 // 5 chars of base36 gives ~60M possible values (36^5 = 60,466,176).
 // Birthday paradox: ~1% collision at ~1,100 IDs — safe for convoy volumes. (#2063)
 func generateShortID() string {
+	return generateShortIDFromReader(convoyIDEntropy)
+}
+
+func generateShortIDFromReader(r io.Reader) string {
 	const alphabet = "0123456789abcdefghijklmnopqrstuvwxyz"
 	b := make([]byte, 5)
-	_, _ = rand.Read(b)
+	_, _ = io.ReadFull(r, b)
 	for i := range b {
 		b[i] = alphabet[int(b[i])%len(alphabet)]
 	}
--- a/internal/cmd/convoy_id_test.go
+++ b/internal/cmd/convoy_id_test.go
@@ -4,6 +4,20 @@ import (
 	"testing"
 )
 
+type uniqueBase36Reader struct {
+	n uint64
+}
+
+func (r *uniqueBase36Reader) Read(p []byte) (int, error) {
+	v := r.n
+	r.n++
+	for i := len(p) - 1; i >= 0; i-- {
+		p[i] = byte(v % 36)
+		v /= 36
+	}
+	return len(p), nil
+}
+
 func TestGenerateShortID_Length(t *testing.T) {
 	id := generateShortID()
 	if len(id) != 5 {
@@ -31,8 +45,9 @@ func TestGenerateShortID_ValidChars(t *testing.T) {
 func TestGenerateShortID_Uniqueness(t *testing.T) {
 	seen := make(map[string]bool)
 	const n = 1000
+	reader := &uniqueBase36Reader{}
 	for i := 0; i < n; i++ {
-		id := generateShortID()
+		id := generateShortIDFromReader(reader)
 		if seen[id] {
 			t.Errorf("collision after %d IDs: %q", i, id)
 		}


internal/cmd/convoy.go
```diff
@@ -423,9 +423,10 @@ func getTownBeadsDir() (string, error) {
 // fails, the error includes bd's stderr for diagnostics instead of a bare
 // "exit status 1".
 func runBdJSON(dir string, args ...string) ([]byte, error) {
+	var stdout, stderr bytes.Buffer
 	cmd := exec.Command("bd", args...)
 	cmd.Dir = dir
-	var stdout, stderr bytes.Buffer
+	cmd.Env = filterEnvKey(os.Environ(), "BEADS_DIR")
```
kirtangajjar, line 435: This helper now strips inherited BEADS_DIR before every read-side bd invocation. The bug here was not only the working directory; a leaked BEADS_DIR from another rig or session could still force bd onto the wrong database even when the caller was otherwise in the right town.

internal/cmd/convoy.go
```diff
@@ -1598,13 +1600,8 @@ func runConvoyStatus(cmd *cobra.Command, args []string) error {
...
 
 	// Get convoy details
-	showArgs := []string{"show", convoyID, "--json"}
-	showCmd := exec.Command("bd", showArgs...)
-	showCmd.Dir = townBeads
-	var stdout bytes.Buffer
-	showCmd.Stdout = &stdout
-
-	if err := showCmd.Run(); err != nil {
+	out, err := runBdJSON(townRoot, "show", convoyID, "--json")
```
kirtangajjar, line 1609: Status now reads convoy details from townRoot rather than townRoot/.beads. HQ convoy IDs are resolved by bd relative to the town root, so running show from .beads made the lookup depend on whatever BEADS_DIR happened to be in the parent process.

internal/cmd/convoy.go
```diff
@@ -1794,7 +1792,7 @@ func runConvoyList(cmd *cobra.Command, args []string) error {
 	}
 	// Default (no flags) = open only (bd's default behavior)
 
-	out, err := runBdJSON(townBeads, listArgs...)
+	out, err := runBdJSON(townRoot, listArgs...)
```
kirtangajjar, line 1801: List now uses the same town-root routing as status. This keeps list and status consistent and avoids the earlier split where one read path could find an HQ convoy while another silently targeted the wrong database.

internal/cmd/convoy_stage.go
```diff
@@ -478,9 +478,7 @@ func dagSlingableIDs(dag *ConvoyDAG) []string {
 // against the town beads directory and unwraps external:prefix:id references.
 func convoyTrackedBeadIDs(townBeads, convoyID string) (map[string]bool, error) {
 	townRoot := filepath.Dir(townBeads)
-	depCmd := exec.Command("bd", "dep", "list", convoyID, "--direction=down", "--type=tracks", "--json")
-	depCmd.Dir = townRoot
-	out, err := depCmd.Output()
+	out, err := runBdJSON(townRoot, "dep", "list", convoyID, "--direction=down", "--type=tracks", "--json")
```
kirtangajjar, line 481: The staged-convoy helper needs the same routing rules as the public status and list commands. Without this, staging logic could still miss a tracked bead even after the top-level convoy commands were fixed.

internal/cmd/convoy_bd_routing_test.go
```diff
@@ -0,0 +1,203 @@
...
+}
+
+func TestRunConvoyList_UsesTownRootAndStripsBeadsDir(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip("skipping on windows - shell stubs")
+	}
+
+	townRoot, expectedWD := makeRoutingTownWorkspace(t)
+	chdirConvoyTest(t, townRoot)
+	t.Setenv("BEADS_DIR", "/wrong/.beads")
```
kirtangajjar, line 91: These regressions are easy to miss with unit tests that only assert returned JSON. The stub asserts both the working directory and the absence of BEADS_DIR so the test fails on the exact routing bug rather than only on a later symptom.

internal/cmd/convoy.go
```diff
@@ -423,9 +429,10 @@ func getTownBeadsDir() (string, error) {
 // fails, the error includes bd's stderr for diagnostics instead of a bare
 // "exit status 1".
 func runBdJSON(dir string, args ...string) ([]byte, error) {
+	var stdout, stderr bytes.Buffer
 	cmd := exec.Command("bd", args...)
 	cmd.Dir = dir
-	var stdout, stderr bytes.Buffer
+	cmd.Env = filterEnvKey(os.Environ(), "BEADS_DIR")
```
kirtangajjar, line 435: `runBdJSON(...)` is the core routing fix for this PR: it now strips inherited `BEADS_DIR` before invoking `bd`, so convoy reads cannot accidentally bind to a stale database from the caller environment.

internal/cmd/convoy.go
```diff
@@ -1598,13 +1606,8 @@ func runConvoyStatus(cmd *cobra.Command, args []string) error {
...
 
 	// Get convoy details
-	showArgs := []string{"show", convoyID, "--json"}
-	showCmd := exec.Command("bd", showArgs...)
-	showCmd.Dir = townBeads
-	var stdout bytes.Buffer
-	showCmd.Stdout = &stdout
-
-	if err := showCmd.Run(); err != nil {
+	out, err := runBdJSON(townRoot, "show", convoyID, "--json")
```
kirtangajjar, line 1609: This status path now computes `townRoot` from `townBeads` and runs `bd show` from there. The bug was that status/list code still executed from `.beads`, which worked for single-town cases but broke HQ convoy reads once routing became town-root scoped.

internal/cmd/convoy_bd_routing_test.go
```diff
@@ -0,0 +1,203 @@
...
+
+	scriptBody := fmt.Sprintf(`
```
kirtangajjar, line 93: These shell-stub tests are deliberately checking the exact failure mode from CI: if `BEADS_DIR` leaks or the working directory is `.beads` instead of the town root, the stub hard-fails. That keeps the routing regression test precise.

internal/cmd/convoy_id_test.go
```diff
@@ -31,8 +45,9 @@ func TestGenerateShortID_ValidChars(t *testing.T) {
 func TestGenerateShortID_Uniqueness(t *testing.T) {
 	seen := make(map[string]bool)
 	const n = 1000
+	reader := &uniqueBase36Reader{}
```
kirtangajjar, line 48: The uniqueness test now drives the generator from deterministic entropy rather than hoping 1000 random IDs never collide. That makes this test validate the encoding logic instead of the luck of the current random sample.

internal/cmd/convoy_stage.go
kirtangajjar, line 481: Stage overlap checks now go through the same routed `runBdJSON(...)` helper as status/list. That matters because convoy staging was still one of the last code paths that could read tracked dependencies from the wrong working directory.

PR #2609, opened 2026-03-09
feat: add merge_strategy config to refinery (direct vs PR mode)

ldangelo:
## Summary

- Adds `merge_strategy` variable to `mol-refinery-patrol.formula.toml` with two modes:
  - `direct` (default): existing fast-forward merge + push behavior (no change)
  - `pr`: creates a GitHub PR via `gh pr create` instead of direct merge
- When `merge_strategy=pr`, the refinery pushes the rebased branch and opens a PR with bead metadata (title, description, issue ID)
- MERGED notification includes the PR URL when in PR mode
- Source branch is preserved as PR head (GitHub handles cleanup on merge)
- Bumps formula version to 10

## Motivation

The refinery currently only supports direct merge to the target branch. For projects that require code review before merging (e.g., open source repos, team workflows), a PR-based merge strategy is needed.

## Test plan

- [ ] Verify `merge_strategy=direct` (default) behaves identically to current behavior
- [ ] Verify `merge_strategy=pr` creates a GitHub PR with correct base/head branches
- [ ] Verify PR title and body include bead metadata
- [ ] Verify MERGED notification includes PR URL in PR mode

### DIFF
--- a/.claude/commands/backup.md
+++ b/.claude/commands/backup.md
@@ -1,0 +1,131 @@
+---
+description: Run Dolt database backup — sync production DBs to backup remotes and iCloud
+allowed-tools: Bash(gt dolt status:*), Bash(dolt backup:*), Bash(dolt log:*), Bash(rsync:*), Bash(du:*), Bash(cat:*), Bash(mkdir:*), Bash(timeout:*), Bash(gt escalate:*), Bash(ls:*)
+argument-hint: [--skip-offsite]
+---
+
+# Dolt Backup
+
+Sync production Dolt databases to filesystem backups with smart change detection.
+Runs the same cycle as `mol-dog-backup` / the `dolt-backup` plugin.
+
+Arguments: $ARGUMENTS
+If `--skip-offsite` is passed, skip the iCloud rsync step.
+
+## Configuration
+
+```
+DOLT_DATA_DIR=~/gt/.dolt-data
+BACKUP_DIR=~/gt/.dolt-backup
+ICLOUD_DIR=~/Library/Mobile Documents/com~apple~CloudDocs/gt-dolt-backup
+PROD_DBS: auto-discovered (databases with <name>-backup remotes)
+BACKUP_TIMEOUT=120s per database
+HANG_THRESHOLD=30s for server ping
+DOLT_PORT=3307
+```
+
+## Execution Steps
+
+### Step 1: Verify Dolt server is responsive
+
+```bash
+gt dolt status
+```
+
+Also ping directly to detect hangs:
+
+```bash
+timeout 30 dolt sql --host 127.0.0.1 --port 3307 --user root --password "" \
+  --no-tls -q "SELECT 1" --result-format csv
+```
+
+If the ping hangs (takes > 30s) or fails:
+```bash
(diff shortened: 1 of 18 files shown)


steveyegge:
Thanks for this PR, @ldangelo — the motivation is solid and we definitely need PR-based merge support for repos with branch protection or review requirements.

After reviewing, we've decided to take this in a slightly different direction architecturally. Here's the thinking:

**What we found:** Gas Town already has a `merge_strategy` field on beads (with values `direct`, `mr`, `local`) set at convoy dispatch time. The data model for "this work should land via PR" already exists — what's missing is the command to act on it.

**The concern with the formula approach:** Embedding PR creation inside the refinery patrol loop creates a lifecycle gap. The refinery sends MERGED, the Witness nukes the polecat worktree, but the PR isn't actually merged yet. If the PR needs updates (review feedback, conflicts), there's no workspace to do it from. Fixing this properly would require a new notification type, Witness protocol changes, and a PR-merge watcher — significant complexity inside a component whose value is being a simple mechanical merge processor.

**What we're building instead:** A standalone `gt mq submit-pr` command that:
1. Reads `merge_strategy` from the bead being processed
2. Pushes the branch and creates a GitHub PR with bead metadata
3. Records the PR URL on the bead
4. Exits cleanly — no patrol loop entanglement

This keeps the refinery simple and puts PR lifecycle management (detecting when the PR merges, cleaning up worktrees) in its own component where it can be designed properly.

We've filed an internal tracking issue for this. Your PR description and test plan are useful reference for the implementation — appreciate you thinking through the edge cases (existing PR fallback, branch preservation, etc.).

Going to close this PR in favor of the standalone approach. If you're interested in contributing to the new command, happy to share the design once it's sketched out.

steveyegge (changes requested):
## Review: feat: add merge_strategy config to refinery

**Verdict: Request Changes** — concept is good and well-scoped, but has a merge conflict and a semantic concern.

### Merge Conflict

Main is already at `version = 10`. This PR bumps 9->10, so it conflicts. Needs a rebase onto current main and a bump to version 11.

### Semantic Issue: MERGED notification in PR mode

In PR mode, the refinery sends `MERGED <polecat-name>` to the witness when it has only *created a PR*, not actually merged anything. The witness interprets MERGED as "code is on the target branch" and nukes the polecat worktree.

This is problematic:
- If the PR is rejected or needs changes, the polecat worktree is already gone
- The witness protocol has a clear contract: MERGED = code is on the target branch
- The note "In PR mode, MERGED means PR created" is redefining a protocol term inline, which is fragile

**Suggestion**: Either use a different notification (e.g., `PR_CREATED`) and update the witness to handle it, or keep MERGED but document clearly in the witness protocol docs (not just a formula comment) that in PR mode the refinery job is done at PR creation.

### What is Good

- Clean, well-scoped change — only touches the formula TOML
- Default behavior (direct) is unchanged
- PR body includes useful bead metadata
- Handles the "PR already exists" edge case
- Correctly skips branch deletion in PR mode
- `--force-with-lease` is the right choice for rebased branch push

### Minor

- `delete_merged_branches` is silently ignored in PR mode — consider logging a warning when both are set.

Fix the rebase conflict and address the MERGED semantics, then this is ready to go.

PR #2615, opened 2026-03-09
fix: add dog role to identity resolution, sling target, and github-sheriff plugin

seanbearden:
## Summary

- Adds `RoleDog` case to `resolveSelfTarget()` in sling_target.go — dogs hit the default error branch ("cannot determine agent identity") because the switch was missing this case
- Adds dog role to mail identity resolution (`mail_identity.go`) — dogs fell through to `detectSenderFromCwd()` → "overseer" because `detectSenderFromRole()` had no dog case
- Sets `GT_DOG_NAME` env var and fixes `BD_ACTOR` to use `deacon/dogs/<name>` address convention
- Implements github-sheriff plugin with `execute.sh` for PR categorization and status reporting

Replaces #2583 (rebased clean onto upstream/main, dropped `install.go` beads.role change already merged via #2588).

## Test plan

- [ ] Verify dogs can resolve their own identity via `gt hook`
- [ ] Verify dog mail sends with correct sender identity (not "overseer")
- [ ] Verify github-sheriff plugin executes via `run.sh` dispatch

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

PR #2617, opened 2026-03-09
feat: github-sheriff plugin for PR categorization and status reporting

seanbearden:
## Summary

- Implements github-sheriff plugin with `execute.sh` for automated PR monitoring
- Discovers rigs dynamically from `rigs.json` (no hardcoded rig list)
- Categorizes open PRs as "easy wins" (CI passing, small, mergeable) vs "needs review"
- Reports summary via beads wisps, escalates on failure
- Runs hourly via cron gate, requires `GITHUB_TOKEN` env var

Split from #2615 for focused review.

## Test plan

- [ ] Verify `execute.sh` discovers rigs from `rigs.json`
- [ ] Verify PR categorization with valid `GITHUB_TOKEN`
- [ ] Verify graceful skip when `GITHUB_TOKEN` is not set

### DIFF
--- a/plugins/github-sheriff/execute.sh
+++ b/plugins/github-sheriff/execute.sh
@@ -10,8 +10,34 @@ set -euo pipefail
 
 readonly SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 readonly GITHUB_TOKEN="${GITHUB_TOKEN:-}"
-readonly RIGS=("beads" "debt_buying" "gleam" "laser" "payment_portal" "shuffle" "teleport" "gastown")
-readonly DOGS_DIR="/Users/seanbearden/gt/deacon/dogs/charlie"
+
+# Discover town root — walk up from script dir until we find .gt or rigs.json
+find_town_root() {
+    local dir="$SCRIPT_DIR"
+    while [ "$dir" != "/" ]; do
+        if [ -f "$dir/rigs.json" ] || [ -d "$dir/.gt" ]; then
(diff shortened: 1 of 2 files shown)


steveyegge:
Thanks for working on this @seanbearden — PR monitoring is useful and the rig discovery from `rigs.json` is a nice touch. Unfortunately there are two fatal bugs that prevent this from working:

**1. Subshell variable loss**

The `| while read` pipeline runs in a subshell, so all modifications to `easy_wins`, `needs_review`, and `processed_prs` are lost when it exits:
```bash
echo "$prs_response" | jq -r '.[] | @json' | while read -r pr_json; do
    easy_wins+=("...")      # lost
    needs_review+=("...")   # lost
    ((processed_prs++))     # lost
done
# Arrays are empty here, processed_prs is still 0
```
Fix: use process substitution (`while read ... < <(jq ...)`) or restructure to avoid piping into the loop.

**2. `additions`/`deletions` not in list endpoint**

The GitHub `/repos/.../pulls` list endpoint doesn't return `additions` or `deletions` — those fields only come from the individual PR detail endpoint (`/repos/.../pulls/{number}`). The code reads them from the list response, so `total_changes` is always 0, which breaks the categorization logic.

**Design note:** The existing `plugin.md` used `gh pr list` and `gh pr checks`, which handles auth, pagination, and rate limiting automatically. The raw `curl` approach loses those benefits and adds N+1 API call overhead (detail + status per PR). Consider building on the `gh`-based approach instead.

Happy to review a v2 that addresses these issues.

PR #2632, opened 2026-03-10
fix: resolve mayor dashboard runtime from agent config

jacopopiogargano:
## Problem
The dashboard Mayor banner always reported the runtime as `claude` once the session was attached, even when the Mayor was actually running on another agent such as `codex`.

That produced incorrect UI in real workspaces and was especially misleading after switching a town to a non-Claude default agent.

## What changed
- removed the hardcoded `claude` runtime from the dashboard Mayor fetch path
- resolve the runtime label from the Mayor session's `GT_AGENT` value when present
- fall back to Gas Town's canonical role/config resolution for Mayor when `GT_AGENT` is missing
- handle custom role aliases, `settings/agents.json` registry agents, provider-only aliases, and ephemeral cost-tier aliases
- keep an explicit fallback to the raw alias when a live session exposes an unknown agent name that current config cannot resolve

## Why this approach
The dashboard now uses the same config resolution paths Gas Town already uses for startup and role-agent selection instead of re-implementing ad hoc lookup logic in the web layer.

That keeps the dashboard aligned with:
- `role_agents`
- `default_agent`
- custom aliases in `settings/config.json`
- registry-defined agents in `settings/agents.json`
- ephemeral `GT_COST_TIER` role resolution

## Tests
Added coverage for:
- `resolveMayorRuntime()` with session `GT_AGENT`
- fallback to `default_agent`
- custom role aliases such as `claude-sonnet`
- registry-defined agents from `settings/agents.json`
- provider-only custom aliases
- ephemeral cost-tier aliases
- unknown alias fallback behavior
- direct runtime label formatting cases
- `FetchMayor()` end-to-end unit behavior

## Verification
- `go test ./internal/web -run 'TestResolveMayorRuntime|TestRuntimeLabelFromConfig|TestFetchMayor'`
- `go test ./internal/web`

## Manual check
Built the patched `gt` binary locally, launched the dashboard against an existing Gas Town workspace, and verified the Mayor banner showed `codex` instead of `claude`.

### DIFF
--- a/internal/web/fetcher.go
+++ b/internal/web/fetcher.go
@@ -18,6 +18,7 @@ import (
 	"github.com/steveyegge/gastown/internal/config"
 	"github.com/steveyegge/gastown/internal/constants"
 	"github.com/steveyegge/gastown/internal/session"
+	"github.com/steveyegge/gastown/internal/tmux"
 	"github.com/steveyegge/gastown/internal/workspace"
 )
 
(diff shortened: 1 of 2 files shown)


jacopopiogargano:
@michaellady FYI 

I am currently watching your video on YT - good stuff!

michaellady:
@jacopopiogargano thanks for watching! replying on stream now, lol https://www.youtube.com/watch?v=BAuCCFUFgI8

PR #2608, opened 2026-03-09
fix(convoy): count tracked externals in status and list

kirtangajjar:
## Problem

HQ convoy progress could show `0/0` even when the convoy was tracking real cross-rig work.

Typical failure mode:
- the convoy tracked cross-rig beads like `external:ghostty:ghostty-123`
- those tracked beads were visible under `bd show <convoy> --json`
- `gt convoy status` and `gt convoy list` only counted `bd dep list --type=tracks`
- progress under-counted or collapsed to `0/0`

```text
Before
  convoy tracks 21 cross-rig beads
    -> dep list sees 0
    -> status/list report 0/0

After
  convoy tracks 21 cross-rig beads
    -> merge dep list + show.dependencies
    -> normalize external:<rig>:<id>
    -> status/list count the real tracked set
```

## Related Work

While fixing this, I split unrelated shared compatibility and CI-harness cleanup into [#2648](https://github.com/steveyegge/gastown/pull/2648).

## Feature-only Review

Use this compare link to review only the external-tracking change:
- <https://github.com/kirtangajjar/gastown/compare/codex/convoy-shared-foundation-2646...codex/convoy-external-tracking-2605>

Line-level rationale is captured in the inline review comments on the diff.

## Validation

Targeted local validation:
- `go test ./internal/cmd -run 'TestGetTrackedIssues_FallsBackToShowTrackedDependencies|TestCollectConvoyBeads_ExternalTrackedIDs|TestTransitionConvoyToOpen_StagedReady'`

Closes #2605

PR #2648, opened 2026-03-10
fix(ci): stabilize shared bd compatibility and Windows command tests

### LINKED ISSUE #2646: Shared bd compatibility and Windows CI command-harness regressions
## Problem

While fixing [#2607](https://github.com/steveyegge/gastown/pull/2607) and [#2608](https://github.com/steveyegge/gastown/pull/2608), I found a separate set of cross-cutting defects in shared command helpers and the Windows CI harness.

These are not convoy-specific bugs. They affect shared `bd` reads, shared command execution helpers, and platform-specific test behavior.

## User-visible symptoms

- some shared `bd` read paths behave differently from direct CLI usage after the newer `--allow-stale` contract
- older `bd` installs can fail on read helpers that assume newer flags such as `--flat`
- Windows CI tests can fail even when the product logic is correct because mock commands do not behave like real `bd` or `codex` executables on Windows
- Git-for-Windows TLS tests can fail because the fixture pathing does not match the OpenSSL PEM path Git actually uses
- proxy TLS session resumption can bypass the custom revocation check unless session tickets are disabled

## What needs to change

- preserve caller environment during `bd` capability probing and use compatibility fallbacks for older `bd` behavior
- align shared read-only helpers with the current `--allow-stale` read contract
- make Windows CI install `bd` and make command-mocking tests use Windows-safe executables and path handling
- make the Git mTLS tests and proxy TLS behavior match actual runtime constraints

## Expected result

- shared read paths behave the same as the real CLI
- Linux and Windows CI exercise the same assumptions about `bd` and command execution
- cross-cutting compatibility fixes are reviewed once, separately from the convoy-specific behavior changes in [#2607](https://github.com/steveyegge/gastown/pull/2607) and [#2608](https://github.com/steveyegge/gastown/pull/2608)

kirtangajjar:
## Problem

This PR fixes a set of shared compatibility and CI defects that surfaced while working on [#2607](https://github.com/steveyegge/gastown/pull/2607) and [#2608](https://github.com/steveyegge/gastown/pull/2608).

The changes here are intentionally limited to cross-cutting code that is not specific to either convoy behavior fix.

## Review

Line-level rationale is captured in the inline review comments on the diff.

## Validation

Validated locally with the same command shapes used in CI:
- exact `Test` workflow command
- exact `Integration Tests` workflow command
- exact `golangci-lint --timeout=5m`

Closes #2646

### DIFF
--- a/internal/cmd/convoy.go
+++ b/internal/cmd/convoy.go
@@ -2033,13 +2033,26 @@ func applyFreshIssueDetails(dep *trackedDependency, details *issueDetails) {
 // getTrackedIssues uses bd dep list to get issues tracked by a convoy.
 // Returns issue details including status, type, and worker info.
 func getTrackedIssues(townBeads, convoyID string) ([]trackedIssueInfo, error) {
-	// Use bd dep list to get tracked dependencies
-	// Run from town root (parent of .beads) so bd routes correctly
 	townRoot := filepath.Dir(townBeads)
-	out, err := runBdJSON(townRoot, "dep", "list", convoyID, "--direction=down", "--type=tracks", "--json")
-	if err != nil {
-		return nil, fmt.Errorf("querying tracked issues for %s: %w", convoyID, err)
+	env := filterEnvKey(os.Environ(), "BEADS_DIR")
+	depArgs := beads.MaybePrependAllowStaleWithEnv(env, []string{
+		"dep", "list", convoyID, "--direction=down", "--type=tracks", "--json",
+	})
+	var stdout, stderr bytes.Buffer
+	cmd := exec.Command("bd", depArgs...)
+	cmd.Dir = townRoot
+	cmd.Env = env
+	cmd.Stdout = &stdout
+	cmd.Stderr = &stderr
+
+	if err := cmd.Run(); err != nil {
+		if errMsg := strings.TrimSpace(stderr.String()); errMsg != "" {
+			return nil, fmt.Errorf("querying tracked issues for %s: bd dep: %s", convoyID, errMsg)
+		}
+		return nil, fmt.Errorf("querying tracked issues for %s: bd dep: %w", convoyID, err)
 	}
+
+	out := stdout.Bytes()
 
 	// Parse the JSON output - bd dep list returns full issue details
 	var deps []trackedDependency


.github/workflows/windows-ci.yml
```diff
@@ -57,6 +57,9 @@ jobs:
           Copy-Item $bin.FullName "$gopathBin\dolt.exe"
           & "$gopathBin\dolt.exe" version
 
+      - name: Install bd
+        run: go install github.com/steveyegge/beads/cmd/bd@v0.57.0
```
kirtangajjar, line 61: Windows CI was missing bd entirely, which meant bead-dependent tests were running under a different dependency model from Linux and macOS. Installing bd here makes the Windows job exercise the same command paths as the rest of CI instead of failing for an environment gap.

internal/beads/beads.go
```diff
@@ -62,6 +68,9 @@ func BdSupportsAllowStale() bool {
 	}
 
 	cmd := exec.Command(bdPath, "--allow-stale", "version") //nolint:gosec // G204: bd is a trusted internal tool
+	if env != nil {
+		cmd.Env = env
```
kirtangajjar, line 72: The allow-stale capability probe has to inherit the caller environment. Without that, the probe could succeed or fail against a different bd context than the actual command invocation, which made compatibility decisions non-deterministic in routed towns and tests.

internal/beads/beads.go
```diff
@@ -597,6 +617,15 @@ func (b *Beads) List(opts ListOptions) ([]*Issue, error) {
 	}
 
 	out, err := b.run(args...)
+	if err != nil && strings.Contains(err.Error(), "unknown flag: --flat") {
```
kirtangajjar, line 620: Newer bd needs --flat for machine-readable list output, but older installs reject it. This fallback keeps the wrapper compatible across both versions instead of forcing every caller and every CI image to upgrade in lockstep.

internal/cmd/capacity_dispatch.go
```diff
@@ -304,9 +304,12 @@ func batchFetchBeadInfoByIDs(townRoot string, ids []string) map[string]beadStatu
 	// For simplicity, try all dirs - bd show will return results only for matching IDs
 	for _, dir := range beadsSearchDirs(townRoot) {
 		// bd show can accept multiple IDs
+		env := filterEnvKey(os.Environ(), "BEADS_DIR")
 		args := append([]string{"show", "--json"}, ids...)
-		showCmd := exec.Command("bd", args...)
+		showArgs := beads.MaybePrependAllowStaleWithEnv(env, args)
```
kirtangajjar, line 309: These batched show reads are still logically read-only, so they should follow the same allow-stale and stripped-BEADS_DIR contract as the direct CLI. That keeps capacity planning from accidentally querying the wrong database when the process inherits a rig-specific environment.

internal/cmd/capacity_dispatch.go
```diff
@@ -490,8 +493,11 @@ func listReadyWorkBeadIDsWithError(townRoot string) (map[string]bool, error) {
 	failCount := 0
 	var lastErr error
 	for _, dir := range dirs {
-		readyCmd := exec.Command("bd", "ready", "--json", "--limit=0")
+		env := filterEnvKey(os.Environ(), "BEADS_DIR")
+		readyArgs := beads.MaybePrependAllowStaleWithEnv(env, []string{"ready", "--json", "--limit=0"})
```
kirtangajjar, line 497: The ready scan has the same issue as the batched show path above: it is a routed read, not a write. Applying the same filtered environment here prevents inherited BEADS_DIR from biasing scheduler capacity decisions.

internal/cmd/convoy.go
```diff
@@ -2033,14 +2033,27 @@ func applyFreshIssueDetails(dep *trackedDependency, details *issueDetails) {
...
 // Returns issue details including status, type, and worker info.
 func getTrackedIssues(townBeads, convoyID string) ([]trackedIssueInfo, error) {
-	// Use bd dep list to get tracked dependencies
-	// Run from town root (parent of .beads) so bd routes correctly
 	townRoot := filepath.Dir(townBeads)
-	out, err := runBdJSON(townRoot, "dep", "list", convoyID, "--direction=down", "--type=tracks", "--json")
-	if err != nil {
-		return nil, fmt.Errorf("querying tracked issues for %s: %w", convoyID, err)
+	env := filterEnvKey(os.Environ(), "BEADS_DIR")
+	depArgs := beads.MaybePrependAllowStaleWithEnv(env, []string{
```
kirtangajjar, line 2038: This shared convoy read path is used by more than one follow-on fix, so it belongs in the common foundation PR. The important detail is that the allow-stale decision is made against the same filtered environment that the actual dep list command will run with.

internal/cmd/scheduler_epic.go
```diff
@@ -297,8 +297,9 @@ type epicChild struct {
 
 // getEpicChildren returns child issues of an epic via dependency lookup.
 func getEpicChildren(epicID string) ([]epicChild, error) {
-	depCmd := exec.Command("bd", "dep", "list", epicID,
-		"--direction=down", "--type=depends_on", "--json")
+	depArgs := beads.MaybePrependAllowStale([]string{"dep", "list", epicID,
```
kirtangajjar, line 300: Epic child discovery is a read-side dependency query. It now opts into the same allow-stale behavior as the rest of the shared bd readers so scheduler flows do not diverge on older or mixed bd setups.

internal/cmd/sling_convoy.go
```diff
@@ -66,7 +67,8 @@ func findConvoyByDescription(townRoot, beadID string) string {
 	townBeads := filepath.Join(townRoot, ".beads")
 
 	// Query all open convoys from HQ
-	listCmd := exec.Command("bd", "list", "--type=convoy", "--status=open", "--json")
+	listArgs := beads.MaybePrependAllowStale([]string{"list", "--type=convoy", "--status=open", "--json"})
```
kirtangajjar, line 70: This convoy lookup is part of scheduler and sling read behavior, not convoy mutation logic. Adding allow-stale here keeps the read contract consistent and avoids test stubs breaking just because the first argv token is now a global bd flag.

internal/cmd/sling_helpers.go
```diff
@@ -984,7 +984,6 @@ func hookBeadWithRetry(beadID, targetAgent, hookDir string) error {
 	for attempt := 1; attempt <= maxRetries; attempt++ {
 		err := BdCmd("update", beadID, "--status=hooked", "--assignee="+targetAgent).
 			Dir(hookDir).
-			WithAutoCommit().
```
kirtangajjar, line 987: This call is intentionally no longer forcing auto-commit. gt sling already controls transaction mode at the outer command level, and this helper was silently overriding that by turning auto-commit back on for one retry loop.

internal/cmd/wl_schema_evolution.go
```diff
@@ -99,13 +99,13 @@ func readDoltSchemaVersion(doltPath, forkDir, asOf string) (string, error) {
 // Returns (false, nil) when the fork lacks a _meta table or schema_version row
 // (pre-versioned fork) — the pull proceeds without interruption.
 func checkSchemaEvolution(doltPath, forkDir string, upgrade bool) error {
-	localVer, err := readDoltSchemaVersion(doltPath, forkDir, "HEAD")
-	if err != nil || localVer == "" {
+	localVer := readDoltSchemaVersion(doltPath, forkDir, "HEAD")
```
kirtangajjar, line 102: The old helper signature suggested that missing schema metadata was an error, but the real contract is that an unversioned fork simply returns no schema version. Collapsing that into a plain string result makes the upgrade decision logic match the underlying metadata semantics.

internal/beads/beads_agent_test.go
```diff
@@ -3,13 +3,42 @@ package beads
...
+	"runtime"
 	"testing"
 )
 
 func installMockBDFixedShowOutput(t *testing.T, showOutput string) {
 	t.Helper()
 
 	binDir := t.TempDir()
+	if runtime.GOOS == "windows" {
+		scriptPath := filepath.Join(binDir, "bd.cmd")
```
kirtangajjar, line 15: Windows does not execute a bare shell script named bd the same way Unix does. This test now installs a real bd.cmd wrapper so the mock behaves like an actual command discovered through PATH on Windows runners.

internal/config/loader_test.go
```diff
@@ -25,6 +25,23 @@ func skipIfAgentBinaryMissing(t *testing.T, agents ...string) {
 	}
 }
 
+func writeAgentStub(t *testing.T, binDir, name string) {
```
kirtangajjar, line 28: These startup-command tests stub agent binaries by name, so the Windows case needs executable .cmd shims instead of Unix shell scripts. Otherwise the test is only validating POSIX process lookup behavior, not the code under test.

internal/cmd/sling_rollback_cleanup_test.go
```diff
@@ -3,13 +3,34 @@ package cmd
...
 	"path/filepath"
+	"runtime"
 	"strings"
 	"testing"
 	"time"
 
 	"github.com/steveyegge/gastown/internal/config"
 )
 
+func writeRollbackCleanupBDStub(t *testing.T, binDir, unixScript, windowsScript string) {
```
kirtangajjar, line 14: The rollback-cleanup tests were relying on a Unix-only fake bd executable. This helper now builds a Windows-specific stub as well, so cleanup behavior can be validated without depending on shell semantics that do not exist on Windows.

internal/witness/handlers_test.go
```diff
@@ -328,16 +329,40 @@ func mockBd(execFn func(args []string) (string, error), runFn func(args []string
...
 }
 
+func stripMockBdFlags(args []string) []string {
+	for len(args) > 0 && strings.HasPrefix(args[0], "--") {
+		args = args[1:]
+	}
+	return args
+}
+
+func installFakeTmuxNoServer(t *testing.T) {
```
kirtangajjar, line 349: This fake tmux binary has to be executable through PATH on Windows too. Using a batch file here keeps the no-server simulation realistic across platforms instead of making the test fail before it even reaches the witness logic.

internal/beads/beads_types_test.go
```diff
@@ -3,12 +3,125 @@ package beads
...
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
 	"testing"
 
 	"github.com/steveyegge/gastown/internal/constants"
 )
 
+func installMockBDRecorder(t *testing.T) string {
```
kirtangajjar, line 13: These beads tests care about the commands that gt invokes, not about a real bd installation. Recording calls through a Windows-aware mock keeps the test focused on argument shape while still working on runners where Unix shell stubs are not executable.

internal/doctor/rig_check_test.go
```diff
@@ -3,10 +3,49 @@ package doctor
 import (
 	"os"
 	"path/filepath"
+	"runtime"
 	"strings"
 	"testing"
 )
 
+func installMockBdInitOnly(t *testing.T) {
```
kirtangajjar, line 11: The redirect check test only needs a deterministic bd init side effect. Using a Windows-aware mock here keeps the test pinned to the redirect behavior itself rather than to platform-specific command launch details.

internal/polecat/manager_test.go
```diff
@@ -357,6 +357,7 @@ func TestClearIssueWithoutAssignment(t *testing.T) {
 func TestAddWithOptions_HasAgentsMD(t *testing.T) {
 	// This test verifies that AGENTS.md exists in polecat worktrees after creation.
 	// AGENTS.md is critical for polecats to "land the plane" properly.
+	installMockBd(t)
```
kirtangajjar, line 360: Windows CI now has bd installed, so this test can no longer rely on tool absence. Installing the mock explicitly keeps the assertion focused on AGENTS.md propagation rather than whichever bd binary happens to be on PATH.

internal/doctor/beads_redirect_target_check_test.go
```diff
@@ -615,6 +616,10 @@ func TestBeadsRedirectTargetCheck_FixWithMissingConfigYaml(t *testing.T) {
 }
 
 func TestBeadsRedirectTargetCheck_FixMetadataRepairFails(t *testing.T) {
+	if runtime.GOOS == "windows" {
```
kirtangajjar, line 619: This failure path depends on POSIX permission semantics. Windows does not honor directory write bits the same way, so skipping here is not weakening the test; it is preventing a platform from pretending to validate behavior it cannot actually model.

internal/hooks/installer_test.go
```diff
@@ -122,6 +123,10 @@ func TestInstallForRole_EmptyProvider(t *testing.T) {
 }
 
 func TestInstallForRole_Permissions(t *testing.T) {
+	if runtime.GOOS == "windows" {
```
kirtangajjar, line 126: This assertion is specifically about POSIX mode bits on installed files. Skipping it on Windows keeps the test honest, because Windows does not preserve the same permission model through os.WriteFile.

internal/proxy/ca_test.go
```diff
@@ -49,6 +50,10 @@ func TestGenerateCA(t *testing.T) {
 	})
 
 	t.Run("ca.key file permissions are 0600", func(t *testing.T) {
+		if runtime.GOOS == "windows" {
```
kirtangajjar, line 53: These CA tests exercise POSIX permission behavior and chmod failure modes. Windows cannot reproduce those semantics reliably, so the skip makes the test reflect real platform constraints instead of producing false negatives.

internal/proxy/exec_test.go
```diff
@@ -229,12 +230,19 @@ func TestHandleExec(t *testing.T) {
 		// The script is placed in a temp dir added to PATH so AllowedCommands
 		// can reference it by plain name (no path separator — issue 12).
 		scriptDir := t.TempDir()
-		scriptPath := filepath.Join(scriptDir, "printenv.sh")
-		require.NoError(t, os.WriteFile(scriptPath, []byte("#!/bin/sh\nprintf '%s' \"$GT_PROXY_IDENTITY\"\n"), 0755))
-		t.Setenv("PATH", scriptDir+":"+os.Getenv("PATH"))
+		commandName := "printenv.sh"
+		if runtime.GOOS == "windows" {
```
kirtangajjar, line 234: AllowedCommands are resolved by executable name. On Windows, a .cmd shim is the closest equivalent to the tiny Unix shell helper used on other platforms, so this keeps the proxy execution test focused on environment injection rather than process-launch differences.

internal/proxy/git_test.go
```diff
@@ -692,12 +693,26 @@ func TestHandleReceivePackIntegration(t *testing.T) {
 
 	// pushRef runs git push with mTLS client cert/key/CA configured via -c flags.
 	pushRef := func(refspec string) ([]byte, error) {
-		cmd := exec.Command(gitPath,
+		args := []string{
+			"-c", "credential.helper=",
+		}
+		if runtime.GOOS == "windows" {
```
kirtangajjar, line 699: Git for Windows defaults to the schannel backend, but this test supplies PEM client cert and key files directly. Forcing the OpenSSL backend here makes the test exercise the same mTLS material we generate in the fixture instead of failing on a backend mismatch.

internal/cmd/install_test.go
```diff
@@ -13,6 +13,8 @@ import (
 
 func TestBuildBdInitArgs_AlwaysIncludesServerPort(t *testing.T) {
 	townDir := t.TempDir()
+	t.Setenv("GT_DOLT_PORT", "")
```
kirtangajjar, line 16: This test should not inherit a developer or CI Dolt port from the parent environment. Clearing both variables up front makes the bd init argument assertion deterministic and keeps it scoped to the helper under test.

internal/cmd/rig_integration_test.go
```diff
@@ -1329,6 +1329,7 @@ func checkWorktreeClean(t *testing.T, agent agentWorktree, hasTrackedBeads bool)
 		allowlist["?? .beads/interactions.jsonl"] = true   // Interactions log
 		allowlist["?? .beads/issues.jsonl"] = true         // Issues log
 		allowlist["?? .beads/metadata.json"] = true        // Beads metadata
+		allowlist["M .beads/metadata.json"] = true         // Tracked metadata is rewritten to the active Dolt server in tracked-beads mode
```
kirtangajjar, line 1332: Tracked-beads mode legitimately rewrites metadata.json to point at the active Dolt server, so this file can show up as modified even when the worktree is otherwise clean. Allowlisting that specific diff avoids flagging normal tracked-bead setup as a dirty checkout.

internal/cmd/scheduler_integration_test.go
```diff
@@ -60,6 +60,10 @@ func initBeadsDBForServer(t *testing.T, dir, prefix string) {
 	if err := os.WriteFile(issuesPath, []byte(""), 0644); err != nil {
 		t.Fatalf("create issues.jsonl in %s: %v", dir, err)
 	}
+
+	if err := beads.EnsureCustomTypes(filepath.Join(dir, ".beads")); err != nil {
```
kirtangajjar, line 64: Scheduler integration towns need custom bead types configured explicitly, otherwise later integration steps fail for test-fixture reasons instead of scheduler behavior. This keeps the fixture aligned with how a real initialized town looks.

internal/cmd/scheduler_integration_test.go
```diff
@@ -312,7 +316,8 @@ func TestSchedulerAutoConvoyCreation(t *testing.T) {
 
 	// Verify: convoy has a "tracks" dependency pointing to the rig bead.
 	// This is the core cross-rig link: convoy lives in HQ DB, bead in rig DB.
-	depCmd := exec.Command("bd", "dep", "list", fields.Convoy, "--direction=down", "--type=tracks", "--json")
+	depArgs := beads.MaybePrependAllowStale([]string{"dep", "list", fields.Convoy, "--direction=down", "--type=tracks", "--json"})
```
kirtangajjar, line 319: The integration assertion now prepends allow-stale exactly the way production code does. Without that, the test could fail only because the fixture command shape lagged behind the real bd invocation contract.

internal/cmd/sling_batch_test.go
```diff
@@ -41,6 +41,10 @@ func TestCreateBatchConvoy_CreatesOneConvoyTrackingAllBeads(t *testing.T) {
 	bdScript := `#!/bin/sh
 echo "CMD:$*" >> "` + logPath + `"
 cmd="$1"
+if [ "$cmd" = "--allow-stale" ]; then
```
kirtangajjar, line 44: These shell stubs record the first positional command, and newer bd calls may now begin with a global allow-stale flag. Shifting past that flag keeps the stub validating the real subcommand instead of failing on an implementation detail of command prefixing.

internal/refinery/batch_test.go
```diff
@@ -112,6 +112,12 @@ func makeMR(id, branch, target string) *MRInfo {
 	}
 }
 
+func failMarkerGateCmd() string {
```
kirtangajjar, line 115: The gate command already runs inside workDir, so baking an absolute Windows path into a POSIX shell test creates an MSYS path-parsing problem that has nothing to do with batch bisection. Using a relative marker check keeps the test portable and equivalent.

internal/mail/bd.go
```diff
@@ -95,11 +95,17 @@ func firstArg(args []string) string {
 }
 
 // bdReadCtx returns a context with the standard bd read timeout.
+//
+//nolint:gosec // The cancel function is returned to callers, who are responsible for invoking it.
```
kirtangajjar, line 99: These helpers intentionally return both the context and its cancel function to the caller. The linter warning here is about ownership, not an actual security problem, so the narrow suppression documents that the responsibility to cancel lives with the caller.

internal/polecat/manager.go
```diff
@@ -1225,8 +1225,10 @@ func forceRemoveDir(dir string) error {
 		}
 		// Make writable (0755 for dirs, 0644 for files)
 		if d.IsDir() {
+			//nolint:gosec // Controlled cleanup of a path inside the allocated polecat directory.
```
kirtangajjar, line 1228: This chmod is part of a controlled cleanup path inside an already allocated polecat directory immediately before deletion. The suppression is narrow because the code is not relaxing permissions for normal runtime use; it is only making stubborn files removable.

internal/proxy/server.go
```diff
@@ -211,6 +211,9 @@ func (s *Server) Start(ctx context.Context) error {
 		ClientAuth: tls.RequireAndVerifyClientCert,
 		ClientCAs:  pool,
 		MinVersion: tls.VersionTLS13,
+		// Disable session tickets so resumed sessions cannot bypass the custom
+		// revocation check implemented in VerifyPeerCertificate.
+		SessionTicketsDisabled: true,
```
kirtangajjar, line 216: The server does its own certificate revocation check in VerifyPeerCertificate. Session tickets have to be disabled here so a resumed TLS session cannot bypass that custom check after the original handshake.

PR #2650, opened 2026-03-10
feat(formula): implement extends and compose/expand support

r0man:
## Summary

- Add `Resolve()` to the formula package so workflow formulas can inherit steps from parent formulas via `extends` and replace target steps with expansion template steps via `compose.expand`
- Fix `validateWorkflow` to allow empty steps when `extends` is set (steps come from parent after resolution)
- Update integration test to use embedded formula FS instead of hardcoded paths; `shiny-enterprise` and `shiny-secure` now pass instead of being skipped

## Motivation

The `shiny-enterprise.formula.toml` uses `extends = ["shiny"]` and `[[compose.expand]]` but the formula parser had no support for these features, causing `gt prime --hook` to fail with _"workflow formula requires at least one step"_ for any formula that inherits steps from a parent.

Fixes ge-9hc.

## Test plan

- [x] `TestResolve_ShinyEnterprise` — resolves 9-step merged+expanded formula
- [x] `TestResolve_ShinySecure` — resolves 5-step inherited formula (aspects reserved)
- [x] `TestResolve_CycleDetection` — circular extends chain returns error
- [x] `TestResolve_NoExtends` — plain formula passes through unchanged
- [x] `TestParseRealFormulas/shiny-enterprise.formula.toml` — passes (was skipped)
- [x] `TestParseRealFormulas/shiny-secure.formula.toml` — passes (was skipped)
- [x] All other existing formula tests continue to pass

### DIFF
--- a/internal/formula/integration_test.go
+++ b/internal/formula/integration_test.go
@@ -1,67 +1,45 @@
 package formula
 
 import (
-	"os"
-	"path/filepath"
+	"io/fs"
 	"strings"
 	"testing"
 )
 
-// TestParseRealFormulas tests parsing actual formula files from the filesystem.
-// This is an integration test that validates our parser against real-world files.
+// TestParseRealFormulas tests parsing all embedded formula files.
+// Composition formulas (extends/compose) are now also resolved and validated.
 func TestParseRealFormulas(t *testing.T) {
-	// Find formula files - they're in various .beads/formulas directories
-	formulaDirs := []string{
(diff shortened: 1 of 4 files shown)


steveyegge:
Fix-merged to main as 7fcfe8e8. Applied with three fixes: deduplicated formula loading in applyExpandRule (reuses loadFormulaByName), fixed cycle detection test to actually verify circular extends errors (was passing for wrong reason), and added extends awareness to inferType. Thanks @r0man!

PR #2652, opened 2026-03-11
feat: github-sheriff v2 — fix subshell loss, use gh CLI for PR categorization

seanbearden:
## Summary

v2 addressing review feedback from #2617:

- **Fix subshell variable loss**: Replace `| while read` pipeline with process substitution (`< <(...)`) so array modifications to `easy_wins`, `needs_review`, and `processed_prs` persist after the loop
- **Fix missing `additions`/`deletions`**: Use `gh pr list --json additions,deletions,mergeable,statusCheckRollup` — single GraphQL call returns all fields, eliminating N+1 per-PR API overhead
- **Use `gh` CLI instead of raw `curl`**: Handles auth, pagination, and rate limiting automatically; builds on the existing `gh pr list`/`gh pr checks` approach from the original `plugin.md`

### What changed

**`plugin.md`**: Preserves original frontmatter exactly (cooldown gate, 5m duration, 2m timeout, severity low) — passes `TestParsePluginMD_GitHubSheriff`. Enhances the bash code blocks with PR categorization (easy wins vs needs review) using `gh --json` fields.

**`execute.sh`** (new): Standalone multi-rig script that discovers rigs from `rigs.json`, fetches PRs via `gh`, categorizes them, deduplicates CI failures against existing beads, and reports via wisps.

### Key design decisions

- `gh pr list --json statusCheckRollup` provides CI status inline — no separate `gh pr checks` call needed per PR
- `gh pr list --json additions,deletions,mergeable` provides size/conflict data that the REST list endpoint doesn't return
- Process substitution (`while read ... < <(jq ...)`) throughout — no subshell variable loss anywhere

## Test plan

- [x] `TestParsePluginMD_GitHubSheriff` passes (frontmatter preserved)
- [ ] `execute.sh` syntax validated (`bash -n`)
- [ ] Manual test with `GITHUB_TOKEN` set against a rig with open PRs
- [ ] Verify graceful skip when `gh auth status` fails


steveyegge (changes requested):
## Review: github-sheriff v2

Thanks for the follow-up, @seanbearden. The core improvements to `plugin.md` are solid — the subshell fix and N+1 elimination are real wins. A few things need addressing before this can land.

### plugin.md changes — looks good with minor notes

The three key fixes are all correct and welcome:
- **Subshell variable loss**: `< <(jq ...)` process substitution instead of `| while read` — correct fix.
- **N+1 elimination**: Single `gh pr list --json additions,deletions,mergeable,statusCheckRollup` call replaces per-PR `gh pr checks` — much better.
- **PR categorization** (easy wins vs needs review): Useful addition, well-scoped.

### execute.sh — should not be included

1. **Wrong filename**: The Deacon plugin system recognizes `run.sh`, not `execute.sh`. See `internal/plugin/scanner_test.go` — `HasRunScript` checks for `run.sh` specifically. This script would be silently ignored by the plugin runner.

2. **Scope creep**: The original `plugin.md` operates on a single rig (the current one, via `$GT_RIG_ROOT`). The new `execute.sh` iterates *all* rigs from `rigs.json`, which is a fundamentally different execution model. Multi-rig patrol is a separate concern that shouldn't be bundled here.

3. **Duplication**: `execute.sh` largely duplicates the logic already in `plugin.md`. The plugin.md *is* the executable specification — having both creates maintenance divergence.

4. **`mapfile -t`**: Requires bash 4+. macOS ships bash 3.2 by default. If this needs to work on stock macOS, this would fail silently.

**Recommendation**: Drop `execute.sh` entirely from this PR. The `plugin.md` improvements stand on their own. If multi-rig standalone execution is desired, that's a separate PR (and should use `run.sh` as the filename).

### Minor: `$ci_pass` used as bare command

```bash
if [ "$MERGEABLE" = "MERGEABLE" ] && $CI_PASS && [ "$TOTAL_CHANGES" -lt 200 ]; then
```

Using `$CI_PASS` as a bare command (`true`/`false`) works but is fragile — if the variable is ever empty or unset, it becomes a syntax error. Prefer `[ "$CI_PASS" = true ]` for robustness.

### Summary

- **plugin.md changes**: Approve (with the minor `$CI_PASS` style nit)
- **execute.sh**: Request removal — wrong filename for the plugin system, scope creep, and duplicates plugin.md logic

steveyegge:
Fix-merged to main as 59783678. Thanks @seanbearden!

Changes from v2 preserved: single GraphQL call, PR categorization, process substitution fix. Added: echo of categorized PR lists (easy wins / needs review were built but never printed), preserved check URLs in failure bead descriptions where available from statusCheckRollup, kept $SKIPPED count in summary.

PR #2658, opened 2026-03-11
fix(ci): restore cross-platform harness drift and deterministic convoy ID tests

### LINKED ISSUE #2657: fix(ci): restore shared main-red failures and harness drift
## Summary
Several failures showing up under the `#2607` branch were not introduced by that convoy work. They were already broken on `main` or were test-harness mismatches that only became visible while rerunning the exact GitHub workflows locally.

Found while working on:
- https://github.com/steveyegge/gastown/pull/2607

## Confirmed problems
1. Windows compile regression in the nudge poller
- `internal/nudge/poller.go` wrote `syscall.SysProcAttr{Setpgid: true}` directly.
- That field does not exist on Windows, so the shared Windows build job fails.

2. `FetchMayor` bypasses the test command seam
- `internal/web/fetcher.go` called `runCmd(...)` directly instead of the injectable `fetcherRunCmd(...)` seam.
- That makes the runtime-resolution test fail even though nearby fetcher code already uses the seam.

3. Parked-rig detection only read `mayor/rigs.json`
- `internal/cmd/rig_helpers.go` ignored rigs that only had a prefix in `rig/config.json`.
- Result: `TestIsRigParked_WhenOnlyBeadLabelPresent` fails and parked/docked detection can miss valid rig-local config.

4. Hook integration fixture was not a real town root
- `internal/cmd/hook_slot_integration_test.go` created nested rig data but no town marker.
- Workspace discovery could stop at the wrong place and build invalid paths such as `<tmp>/gastown/gastown`.

5. A few tests on `main` were stale after intentional behavior changes
- dog env now sets `BD_ACTOR=deacon/dogs/<name>` and `GT_DOG_NAME`
- tmux registry now uses the default socket for unset/default/auto
- witness startup command now uses `exec env ... run ...`

6. Prime CLI tests depended on whichever `gt` happened to be on `PATH`
- In long `internal/cmd` runs this could resolve to the wrong binary and produce empty or misleading failures.
- Those tests should build and run the current test binary directly.

7. tmux zombie-session test was flaky
- The test asserted directly on `IsAgentRunning(...)` during shell startup, which is timing-sensitive.
- The behavior we actually need is simpler: no agent is alive, and `EnsureSessionFresh(...)` must still recover the session.

8. Lint cleanup
- unchecked `resp.Body.Close()` calls
- ignored `os.WriteFile(...)` result in JSONL backup helper
- unused parameter in compact-resume helper

## Why this matters
These failures create noise in unrelated PRs and make `main` look healthier than it is.

## Validation
Run the same workflows locally:

```bash
go build -v ./cmd/gt
gotestsum --format testname --junitfile junit.xml -- -race -short -timeout=10m -coverprofile=coverage.out ./...
```

```bash
golangci-lint run --timeout=5m
```

```bash
GOOS=windows GOARCH=amd64 go build -v ./cmd/gt
```

```bash
docker pull dolthub/dolt-sql-server:1.83.0
make build
cp gt "$(go env GOPATH)/bin/"
gotestsum --format testname --junitfile junit-integration.xml -- -tags=integration -timeout=15m -v ./internal/cmd/...
```

kirtangajjar:
## Summary
This fixes a set of shared failures that were already broken outside the convoy branch and were making `#2607` look red for the wrong reasons.

Closes #2657.

Found while working on:
- #2607

## What changed
- use the cross-platform `util.SetProcessGroup(...)` helper in the nudge poller so Windows builds compile again
- route `FetchMayor()` through the existing `fetcherRunCmd(...)` seam so the runtime-resolution test exercises the right command path
- fall back to `rig/config.json` when parked/docked rig detection cannot resolve the beads prefix from `mayor/rigs.json`
- make the hook integration fixture a real town by writing `mayor/town.json`
- update stale tests to match intentional shipped behavior for dog env, tmux default sockets, and witness startup command ordering
- make the Windows config-agent assertions accept the platform-specific `claude.cmd` stub path instead of assuming a Unix-style `/claude` binary name
- make convoy short-ID tests deterministic by routing entropy through a reader seam instead of asserting that 1000 random draws never collide
- make prime CLI tests build and execute the current `gt` binary instead of depending on ambient `PATH`
- remove the redundant flaky `IsAgentRunning(...)` precondition from the zombie-session tmux test
- fix the lint findings around unchecked closes, ignored writes, and the unused compact-resume parameter

## Review notes
- the Windows config-test updates are test-only: they acknowledge the real platform behavior after the `.cmd` stub fix rather than weakening command-resolution checks
- the convoy ID change is also test-driven: production still uses crypto randomness, but tests no longer rely on a statistically unsafe “1000 random draws never collide” assumption

## Validation
```bash
go build -v ./cmd/gt
gotestsum --format testname --junitfile junit.xml -- -race -short -timeout=10m -coverprofile=coverage.out ./...
```

```bash
golangci-lint run --timeout=5m
```

```bash
GOOS=windows GOARCH=amd64 go build -v ./cmd/gt
```

```bash
docker pull dolthub/dolt-sql-server:1.83.0
make build
cp gt "$(go env GOPATH)/bin/"
gotestsum --format testname --junitfile junit-integration.xml -- -tags=integration -timeout=15m -v ./internal/cmd/...
```

### DIFF
--- a/internal/cmd/convoy.go
+++ b/internal/cmd/convoy.go
@@ -26,13 +26,19 @@ import (
 	"github.com/steveyegge/gastown/internal/workspace"
 )
 
+var convoyIDEntropy io.Reader = rand.Reader
+
 // generateShortID generates a collision-resistant convoy ID suffix using base36.
 // 5 chars of base36 gives ~60M possible values (36^5 = 60,466,176).
 // Birthday paradox: ~1% collision at ~1,100 IDs — safe for convoy volumes. (#2063)
 func generateShortID() string {
+	return generateShortIDFromReader(convoyIDEntropy)
+}
+
+func generateShortIDFromReader(r io.Reader) string {
 	const alphabet = "0123456789abcdefghijklmnopqrstuvwxyz"
 	b := make([]byte, 5)
-	_, _ = rand.Read(b)
+	_, _ = io.ReadFull(r, b)
 	for i := range b {
 		b[i] = alphabet[int(b[i])%len(alphabet)]
 	}
--- a/internal/cmd/convoy_id_test.go
+++ b/internal/cmd/convoy_id_test.go
@@ -4,6 +4,20 @@ import (
 	"testing"
 )
 
+type uniqueBase36Reader struct {
+	n uint64
+}
+
+func (r *uniqueBase36Reader) Read(p []byte) (int, error) {
+	v := r.n
+	r.n++
+	for i := len(p) - 1; i >= 0; i-- {
+		p[i] = byte(v % 36)
+		v /= 36
+	}
+	return len(p), nil
+}
+
 func TestGenerateShortID_Length(t *testing.T) {
 	id := generateShortID()
 	if len(id) != 5 {
@@ -31,8 +45,9 @@ func TestGenerateShortID_ValidChars(t *testing.T) {
 func TestGenerateShortID_Uniqueness(t *testing.T) {
 	seen := make(map[string]bool)
 	const n = 1000
+	reader := &uniqueBase36Reader{}
 	for i := 0; i < n; i++ {
-		id := generateShortID()
+		id := generateShortIDFromReader(reader)
 		if seen[id] {
 			t.Errorf("collision after %d IDs: %q", i, id)
 		}


internal/config/loader_test.go
```diff
@@ -1540,8 +1541,8 @@ func TestResolveRoleAgentConfig_FallsBackOnInvalidAgent(t *testing.T) {
 
 	// Should fall back to default (claude) when agent is invalid
 	rc := ResolveRoleAgentConfig(constants.RoleRefinery, townRoot, rigPath)
-	// Command can be "claude" or full path to claude
-	if rc.Command != "claude" && !strings.HasSuffix(rc.Command, "/claude") {
+	// Command can be "claude" or a resolved platform-specific claude binary path.
+	if !isClaudeCommand(rc.Command) {
```
kirtangajjar, line 1545: Windows resolves the test stub to `claude.cmd`, so the earlier `== "claude" || strings.HasSuffix(..., "/claude")` check was rejecting the correct command on this platform. Using `isClaudeCommand(...)` keeps the assertion strict about the executable identity while allowing the real Windows resolution path.

internal/cmd/convoy.go
```diff
@@ -26,13 +26,19 @@ import (
 	"github.com/steveyegge/gastown/internal/workspace"
 )
 
+var convoyIDEntropy io.Reader = rand.Reader
+
 // generateShortID generates a collision-resistant convoy ID suffix using base36.
 // 5 chars of base36 gives ~60M possible values (36^5 = 60,466,176).
 // Birthday paradox: ~1% collision at ~1,100 IDs — safe for convoy volumes. (#2063)
 func generateShortID() string {
+	return generateShortIDFromReader(convoyIDEntropy)
```
kirtangajjar, line 35: This reader seam keeps production entropy on `rand.Reader` but lets tests inject deterministic bytes. The bug here was not the ID generator itself; it was that the test suite had no way to validate uniqueness without gambling on random output.

internal/cmd/convoy_id_test.go
```diff
@@ -31,8 +45,9 @@ func TestGenerateShortID_ValidChars(t *testing.T) {
 func TestGenerateShortID_Uniqueness(t *testing.T) {
 	seen := make(map[string]bool)
 	const n = 1000
+	reader := &uniqueBase36Reader{}
```
kirtangajjar, line 48: This helper counts upward in base36 so the uniqueness test exercises the mapping logic deterministically. That removes the probabilistic collision failure without weakening the actual alphabet or width checks.

internal/cmd/escalate_impl.go
```diff
@@ -664,7 +664,7 @@ func sendEscalationSlack(cfg *config.EscalationConfig, beadID, severity, descrip
 	if err != nil {
 		return fmt.Errorf("posting to slack: %w", err)
 	}
-	defer resp.Body.Close()
+	defer func() { _ = resp.Body.Close() }()
```
kirtangajjar, line 667: The only reason for this wrapper is lint correctness: the important failure path is still the webhook status/body handling, while any `Close()` error on a fully-read response is intentionally ignored because there is no recovery action here.

internal/cmd/hook_slot_integration_test.go
```diff
@@ -34,6 +36,21 @@ func setupHookTestTown(t *testing.T) (townRoot, polecatDir, rigPrefix string) {
 
 	townRoot = t.TempDir()
 
+	// Create a real town marker so workspace discovery resolves the outer town
```
kirtangajjar, line 39: The hook fixture now writes a real town marker so workspace discovery stops at the outer temp town root. Without this, tests could accidentally resolve the nested rig as the workspace root and build invalid `gastown/gastown` paths.

internal/cmd/prime.go
```diff
@@ -205,7 +205,7 @@ func runPrime(cmd *cobra.Command, args []string) (retErr error) {
 // Unlike the full prime path, this outputs a brief recovery line instead of
 // the full AUTONOMOUS WORK MODE block. This prevents agents from re-announcing
 // and re-initializing after compaction. (GH#1965)
-func runPrimeCompactResume(ctx RoleContext, cwd string) {
+func runPrimeCompactResume(ctx RoleContext) {
```
kirtangajjar, line 208: Dropping the unused `cwd` parameter keeps the compact-resume path honest: this helper only needs the resolved role context, and the extra argument was dead plumbing that lint correctly flagged.

internal/cmd/prime_test.go
```diff
@@ -105,11 +105,7 @@ func TestGetAgentBeadID_UsesRigPrefix(t *testing.T) {
 }
 
 func TestPrimeFlagCombinations(t *testing.T) {
-	// Find the gt binary - we need to test CLI flag validation
-	gtBin, err := exec.LookPath("gt")
-	if err != nil {
-		t.Skip("gt binary not found in PATH")
-	}
+	gtBin := buildGT(t)
```
kirtangajjar, line 108: These CLI tests now build the current branch binary instead of trusting whatever `gt` happens to be on PATH. That matters in long CI or local sessions where an older installed binary can make the assertions fail for the wrong reason.

internal/cmd/rig_helpers.go
```diff
@@ -156,6 +144,22 @@ func IsRigParkedOrDocked(townRoot, rigName string) (bool, string) {
 	return false, ""
 }
 
+func rigBeadsPrefix(townRoot, rigPath, rigName string) string {
```
kirtangajjar, line 147: This fallback helper preserves the original `mayor/rigs.json` lookup but also honors rigs that only declare their beads prefix in `rig/config.json`. The intent is to fix parked/docked detection, not to change any naming rules.

internal/config/agents_test.go
```diff
@@ -8,10 +8,12 @@ import (
 	"testing"
 )
 
-// isClaudeCmd checks if a command is claude (either "claude" or a path ending in "/claude").
+// isClaudeCmd checks if a command resolves to the claude binary on any platform.
 // Note: Named differently from loader_test.go's isClaudeCommand to avoid redeclaration.
 func isClaudeCmd(cmd string) bool {
```
kirtangajjar, line 13: The earlier helper was Unix-biased because it only accepted `claude` or `/claude`. Normalizing through `filepath.Base` and `filepath.Ext` keeps the test strict about the executable identity while allowing Windows `claude.cmd` stubs.

internal/config/env_test.go
```diff
@@ -125,7 +125,8 @@ func TestAgentEnv_Dog(t *testing.T) {
 	})
 
 	assertEnv(t, env, "GT_ROLE", "dog")
-	assertEnv(t, env, "BD_ACTOR", "dog/alpha")
+	assertEnv(t, env, "GT_DOG_NAME", "alpha")
```
kirtangajjar, line 128: This test update documents the shipped dog environment contract: the dog name is exposed explicitly and `BD_ACTOR` uses the deacon-managed dog path, which is what the runtime now emits.

internal/config/loader_test.go
```diff
@@ -1049,7 +1049,8 @@ func TestBuildAgentStartupCommand(t *testing.T) {
 	if !strings.Contains(cmd, "BD_ACTOR=gastown/witness") {
 		t.Error("expected BD_ACTOR in command")
```
kirtangajjar, line 1050: This assertion now inspects the final argv shape instead of matching a Unix-specific substring. The behavior under test is still “startup command ends in the Claude executable plus the safety flag”, but it no longer rejects `claude.cmd`.

internal/config/loader_test.go
kirtangajjar, line 1545: This fallback case was another Windows false negative: command resolution was correct, but the old assertion only accepted Unix path forms. Reusing `isClaudeCommand(...)` keeps the check aligned with the cross-platform command helper.

internal/config/loader_test.go
```diff
@@ -3655,8 +3656,8 @@ func TestResolveRoleAgentConfig(t *testing.T) {
 
 	t.Run("town-level role (no rigPath) uses town RoleAgents", func(t *testing.T) {
 		rc := ResolveRoleAgentConfig("mayor", townRoot, "")
-		// mayor is in town's RoleAgents - command can be "claude" or full path to claude
-		if rc.Command != "claude" && !strings.HasSuffix(rc.Command, "/claude") {
+		// mayor is in town's RoleAgents and may resolve to a platform-specific claude binary path.
+		if !isClaudeCommand(rc.Command) {
```
kirtangajjar, line 3660: Town-level role resolution goes through the same executable lookup path, so this assertion needs the same platform-neutral check. Otherwise Windows would fail even though the config loader returned the right Claude command.

internal/config/test_main_test.go
```diff
@@ -26,7 +26,14 @@ func TestMain(m *testing.M) {
 	}
 	for _, name := range binaries {
 		path := filepath.Join(stubDir, name)
-		if err := os.WriteFile(path, stub, 0755); err != nil {
+		stub := []byte("#!/bin/sh\nexit 0\n")
+		mode := os.FileMode(0755)
+		if runtime.GOOS == "windows" {
```
kirtangajjar, line 31: The test harness now creates `.cmd` stubs on Windows because `exec.LookPath` and command resolution there do not treat a bare shell script as an executable agent binary. This is a harness realism fix, not a production behavior change.

internal/daemon/jsonl_git_backup.go
```diff
@@ -582,7 +582,10 @@ func ensureGitIgnore(gitRepo, entry string) {
 		content += "\n"
 	}
 	content += entry + "\n"
-	os.WriteFile(ignorePath, []byte(content), 0644)
+	if err := os.WriteFile(ignorePath, []byte(content), 0644); err != nil {
```
kirtangajjar, line 585: Ignoring a write failure here is deliberate and scoped: the backup workflow still works without the ignore entry, so the code now acknowledges the error explicitly instead of dropping it silently to satisfy lint.

internal/nudge/poller.go
```diff
@@ -68,9 +69,7 @@ func StartPoller(townRoot, session string) (int, error) {
 	cmd.Dir = townRoot
 	cmd.Stdout = nil // discard
 	cmd.Stderr = nil // discard
-	cmd.SysProcAttr = &syscall.SysProcAttr{
-		Setpgid: true, // detach from parent process group
-	}
+	util.SetProcessGroup(cmd)
```
kirtangajjar, line 72: Using `util.SetProcessGroup(cmd)` moves the process-group setup behind the existing cross-platform abstraction. The prior inline `Setpgid` field was the direct source of the Windows compile break.

internal/nudge/poller.go
```diff
@@ -109,15 +108,14 @@ func StopPoller(townRoot, session string) error {
 		return nil // corrupt PID file, clean up
 	}
 
-	proc, err := os.FindProcess(pid)
-	if err != nil {
+	if !pollerProcessAlive(pid) {
```
kirtangajjar, line 111: The liveness check is extracted so the poller logic can ask “is this PID alive?” without relying on Unix signal-0 semantics. That separation is what makes the same stop/alive code work on both Unix and Windows.

internal/nudge/poller_process_unix.go
```diff
@@ -0,0 +1,21 @@
+//go:build !windows
+
+package nudge
+
+import (
+	"os"
+	"syscall"
+)
+
+func pollerProcessAlive(pid int) bool {
```
kirtangajjar, line 10: On Unix we still use the lightweight signal-0 probe. Splitting it into a build-tagged file makes the platform behavior explicit instead of hiding it behind conditional branches in the main poller logic.

internal/nudge/poller_process_windows.go
```diff
@@ -0,0 +1,22 @@
...
+
+package nudge
+
+import (
+	"math"
+
+	"golang.org/x/sys/windows"
+)
+
+func pollerProcessAlive(pid int) bool {
```
kirtangajjar, line 11: Windows cannot use signal-0, so this version uses `OpenProcess(PROCESS_QUERY_LIMITED_INFORMATION)` as the existence probe. Treating `ACCESS_DENIED` as alive matches the practical question the caller is asking: whether the PID still refers to a live process.

internal/session/registry_socket_test.go
```diff
@@ -75,13 +75,11 @@ func TestInitRegistry_SocketFromTownName(t *testing.T) {
 			_ = InitRegistry(townRoot)
 
 			got := tmux.GetDefaultSocket()
-			// Socket should start with the sanitized basename
-			base := sanitizeTownName(tt.townDir)
-			if !strings.HasPrefix(got, base+"-") {
-				t.Errorf("after InitRegistry(%q) with GT_TMUX_SOCKET=%q:\n  socket = %q, want prefix %q-",
-					townRoot, tt.gtTmuxSocket, got, base)
+			if got != "" {
```
kirtangajjar, line 78: The socket expectation changed here because unset/default/auto now intentionally mean “use tmux default socket”, not “derive a town-specific socket”. This update aligns the test with the current registry behavior rather than preserving stale assumptions.

internal/tmux/tmux_test.go
```diff
@@ -248,10 +248,10 @@ func TestEnsureSessionFresh_ZombieSession(t *testing.T) {
 		t.Skip("session unexpectedly has agent running - can't test zombie case")
 	}
 
-	// Verify generic agent check also treats it as not running (shell session)
-	if tm.IsAgentRunning(sessionName) {
-		t.Fatalf("expected IsAgentRunning(%q) to be false for a fresh shell session", sessionName)
-	}
+	// Fresh tmux sessions can briefly report transient pane commands while the
```
kirtangajjar, line 251: This test stopped asserting on `IsAgentRunning(...)` during shell startup because that signal is transient and timing-sensitive. The actual invariant we care about is that there is no agent runtime yet, so `EnsureSessionFresh(...)` should rebuild the session.

internal/tmux/tmux_test.go
```diff
@@ -2173,19 +2173,29 @@ func TestCheckSessionHealth_ActivityCheck(t *testing.T) {
 
 func TestValidateCommandBinary(t *testing.T) {
 	t.Parallel()
+
+	absoluteShell := "/bin/sh"
```
kirtangajjar, line 2177: Using a runtime-discovered shell path avoids hard-coding `/bin/sh` into a Windows run. The test is still validating absolute-binary handling; it just now asks for an absolute shell path that exists on the current platform.

internal/web/fetcher.go
```diff
@@ -1462,7 +1462,7 @@ func (f *LiveConvoyFetcher) FetchMayor() (*MayorStatus, error) {
 	mayorSessionName := session.MayorSessionName()
 
 	// Check if mayor tmux session exists
-	stdout, err := runCmd(f.tmuxCmdTimeout, "tmux", "list-sessions", "-F", "#{session_name}:#{session_activity}")
+	stdout, err := fetcherRunCmd(f.tmuxCmdTimeout, "tmux", "list-sessions", "-F", "#{session_name}:#{session_activity}")
```
kirtangajjar, line 1465: Routing `FetchMayor()` through `fetcherRunCmd(...)` brings this call site back under the same injectable seam as the rest of the fetcher tests. That makes the runtime-resolution tests exercise the real fetcher path instead of bypassing the stubbed command runner.

internal/witness/manager_test.go
```diff
@@ -18,7 +18,7 @@ func TestBuildWitnessStartCommand_UsesRoleConfig(t *testing.T) {
 		t.Fatalf("buildWitnessStartCommand: %v", err)
 	}
 
-	want := "env -u CLAUDECODE NODE_OPTIONS='' exec run --town /town --rig gastown --role witness"
+	want := "exec env -u CLAUDECODE NODE_OPTIONS='' run --town /town --rig gastown --role witness"
```
kirtangajjar, line 21: This expectation changed because the witness start command now wraps execution as `exec env ... run ...`. The test update is documenting the current command ordering, not broadening what the command builder accepts.

PR #2667, opened 2026-03-11
fix: use prefix routing in runHookShow for rig-scoped agents

r0man:
## Summary

- `runHookShow` used `findLocalBeadsDir()` which naively walks up from CWD, finding the wrong `.beads` directory when called from mayor context for rig-scoped agents (e.g. `gastown_el/refinery`)
- Apply the same prefix routing used by `runMoleculeStatus`: build the agent bead ID via `buildAgentBeadID`, then resolve the correct beads directory via `beads.ResolveHookDir` before querying for hooked work

Fixes: ge-0rs

## Test plan

- [ ] `go test ./internal/cmd/... -run Hook` passes (all hook tests green)
- [ ] `gt hook show gastown_el/refinery` from mayor context returns the correct hooked issue
- [ ] `gt hook show` with town-level agents (mayor, deacon) still works correctly

### DIFF
--- a/internal/cmd/hook.go
+++ b/internal/cmd/hook.go
@@ -461,6 +461,16 @@ func runHookShow(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("not in a beads workspace: %w", err)
 	}
 
+	// For rig-scoped agents, resolve the correct beads directory via prefix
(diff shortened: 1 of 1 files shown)


r0man:
Closing: needs proper test coverage before merging. Will resubmit with tests.

PR #2671, opened 2026-03-11
fix: remove unused dbName parameter in batchDeleteRows (lint)

seanbearden:
## Summary

- Remove unused `dbName` parameter in `batchDeleteRows` (flagged by `unparam` linter)
- Only lint issue found by `golangci-lint run ./...` on current main

## Test plan

- [x] `golangci-lint run ./...` reports 0 issues
- [x] `go test ./internal/reaper/...` passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2672, opened 2026-03-11
fix: move --flat fallback into Beads.run to prevent re-injection

seanbearden:
## Summary

- Move the `--flat` retry logic from `List()` into `Beads.run()` to fix integration test failures
- `List()` had a fallback that stripped `--flat` and retried via `b.run()`, but `run()` calls `InjectFlatForListJSON()` which re-added `--flat` — making the retry always fail identically
- By handling the fallback inside `run()` (after `InjectFlatForListJSON` has already executed), the retry correctly strips `--flat` without re-injection

## Test plan

- [x] `TestBeadsListFromPolecatDirectory` — was failing, now fixed
- [x] `TestBeadsListFromCrewDirectory` — was failing, now fixed
- [x] `go test ./internal/beads/...` passes
- [x] `golangci-lint run ./internal/beads/...` — 0 issues
- [x] CI integration tests pass

### DIFF
--- a/internal/beads/beads.go
+++ b/internal/beads/beads.go
@@ -427,6 +427,28 @@ func (b *Beads) run(args ...string) (_ []byte, retErr error) {
 	cmd.Stderr = &stderr
 
 	err := cmd.Run()
+
+	// If bd doesn't support --flat, retry without it. The retry is done here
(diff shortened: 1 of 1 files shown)


seanbearden:
Full CI verified green in combination with #2671 (lint fix) via #2673.

Issue #2690, opened 2026-03-12
gt dolt restart has race window where rogues grab port


Random-Word:
`gt dolt restart` stops the running server and then starts a new one. Between the stop and start, there is a window where any active Gas Town process (deacon, witnesses, refineries, polecats) that triggers a `bd` operation can auto-spawn a rogue Dolt server that grabs the port before the canonical server starts.

## Reproduction

1. Have a Gas Town with active agents running
2. Run `gt dolt restart`

**Expected:** Canonical server restarts cleanly from `.dolt-data/`
**Actual:** A rogue from `.beads/dolt/` grabs the port in the gap. `gt dolt start` then adopts the rogue as canonical (see #TBD — status verification issue).

## Impact

In an active town, `gt dolt restart` is unreliable. We observed rogues spawning within milliseconds of the old server being terminated, from various `.beads/dolt` locations.

## Proposed fix options

1. **Atomic restart**: Bind the port with a placeholder listener before stopping the old server, then hand off to the new canonical server
2. **Spawn lock**: Write a lock file that `bd` checks before auto-spawning (e.g., `.dolt-data/.restarting`)
3. **Agent quiesce**: Temporarily pause agents that trigger `bd` during the restart window
4. **Port reservation**: Start the new server on a temp port, verify it's healthy, then atomically swap the port binding

## Workaround

Manually stop all agents before restarting, then restart agents after. This is what we do in `restart-clean.sh`.

## Related

- #2578 — `gt down` orphaned processes

DreadPirateRobertz:
The `Start()` function in `doltserver.go` already has mitigation for this:

1. **`StopIdleMonitors()`** (line 1291) — stops background processes that auto-spawn rogue servers before the restart
2. **Lock file** (line 1253) — exclusive lock prevents concurrent starts
3. **`VerifyServerDataDir()` + `KillImposters()`** (lines 1319-1338) — detects and kills imposters on the port

The remaining gap: between `Stop()` and the lock acquisition in `Start()`, a rogue could still grab the port. A tighter fix would be to hold the lock across both stop and start in the `restart` code path, rather than releasing and re-acquiring.

Is the issue still reproducing with the current code, or was this filed against an older version?

DreadPirateRobertz:
Triage: Classic TOCTOU race. The fix is to make `gt dolt restart` a single atomic operation that prevents rogues from grabbing the port during the restart window.

[...]

PR #2698, opened 2026-03-12
gt dolt start: close restart race window by stopping idle-monitors and waiting for port release

Random-Word:
## Problem

When restarting a Dolt server with `gt dolt restart` (or `stop` + `start`), there is a race window where `bd dolt idle-monitor` processes can spawn a rogue Dolt server on the configured port before the canonical server starts. The sequence:

1. `gt dolt stop` sends SIGTERM to the canonical server
2. The port becomes free
3. An idle-monitor detects the server is gone and spawns a replacement from a rig's `.beads/dolt/` directory
4. The spawned rogue binds to the configured port
5. `gt dolt start` fails because the port is already in use

This was observed in a production deployment where restarts would consistently fail, requiring manual process cleanup.

PR #2586 (port-file sync) and commit db23a4368 (stale port-file cleanup) mitigate the spawn trigger but don't close the race window — idle-monitors that are already running can still spawn between stop and start.

Related: #2690 (restart race window)

## Root Cause

There is no coordination between `gt dolt stop/start` and the background idle-monitor processes. The `Start()` function assumes the configured port will be available, but idle-monitors operate on their own polling interval and will respawn a Dolt server whenever they detect the canonical server is gone.

## Solution

Close the race window by: (1) stopping idle-monitors *before* stopping the server, and (2) adding a port-release verification step between stop and start.

### Changes

**`internal/doltserver/doltserver.go`:**

- **`StopIdleMonitors(townRoot string)`**: New exported function that discovers and terminates all `bd dolt idle-monitor` processes scoped to the given town. Uses path-boundary-aware matching to avoid false matches on sibling towns. Terminates with SIGTERM and waits up to 500ms for clean exit before forced termination.

- **`waitForPortRelease(port int, timeout time.Duration)`**: New function that polls the configured port until it's free or timeout expires. Used between stopping the old server and starting the new one.

- **`Start()` updated**: Calls `StopIdleMonitors` and `waitForPortRelease` before starting the new server, closing the race window.

- **`containsPathBoundary(line, path string)`**: Helper for path-boundary-aware matching. Ensures that a path match in process arguments is a complete path component, not a prefix of a longer path (e.g., `/tmp/gt` won't match `/tmp/gt-old`).

- **Structural `--port` argument parsing**: Port detection in process arguments parses `--port <value>` and `--port=<value>` as discrete arguments rather than substring matching. This prevents false matches on PIDs or other numeric values that happen to contain the port number.

### Key Design Decisions

- **Pre-start cleanup**: Idle-monitors are stopped *before* the canonical server is stopped. This prevents the race entirely rather than trying to win it.

- **Port release verification**: Even after stopping idle-monitors and the old server, `waitForPortRelease` adds a safety net for any process that might be holding the port. The 10-second timeout with 100ms polling is generous enough for clean shutdown.

- **Scoped process matching**: Both `StopIdleMonitors` and orphan detection use `containsPathBoundary` to scope matches to the specific town. This is critical in multi-town deployments.

### Alternatives Considered

- **File-lock approach** (lock file on the port during restart): Rejected as overly complex. Idle-monitors don't check for lock files, and adding that protocol would require changes across multiple components.

- **Disabling idle-monitors globally**: Rejected because idle-monitors serve a legitimate purpose — keeping Dolt available for rigs when the canonical server crashes. The goal is to coordinate them during intentional restarts, not remove them.

- **Using `SIGSTOP`/`SIGCONT` to pause idle-monitors**: Considered but rejected because stopped processes can become zombies if the parent doesn't wait, and there's no guarantee the parent is `gt`. Clean termination + restart is safer.

- **Socket-level port reservation** (bind the port between stop and start): Rejected because Go's net.Listen would need to be threaded through the Dolt server startup, which uses its own listener.

## Testing

- `go build ./...` passes
- `go test ./internal/doltserver/...` passes (15s)
- Manually verified in Gas Town deployment: restart no longer fails due to port conflicts

### DIFF
--- a/internal/doltserver/doltserver.go
+++ b/internal/doltserver/doltserver.go
@@ -957,6 +957,31 @@ func KillImposters(townRoot string) error {
 	return nil
 }
 
+// containsPathBoundary checks whether line contains path as a complete path
+// (not a prefix of a longer path). The character after the match must be a
+// path separator, whitespace, or end-of-string.
+func containsPathBoundary(line, path string) bool {
+	if path == "" {
+		return false
+	}
+	for start := 0; start < len(line); {
+		idx := strings.Index(line[start:], path)
+		if idx < 0 {
+			return false
+		}
+		end := start + idx + len(path)
+		if end >= len(line) {
+			return true
+		}
+		c := line[end]
+		if c == filepath.Separator || c == ' ' || c == '\t' {
+			return true
+		}
+		start = start + idx + 1
+	}
+	return false
+}
+
 // StopIdleMonitors finds and terminates "bd dolt idle-monitor" processes
 // associated with this town. These background processes auto-spawn rogue
 // Dolt servers from per-rig .beads/dolt/ directories when the canonical
@@ -985,8 +1010,9 @@ func StopIdleMonitors(townRoot string) int {
 			continue
 		}
 
-		// Scope to this town: match by path in args
-		matchesTown := strings.Contains(line, absRoot) || strings.Contains(line, townRoot)
+		// Scope to this town: match by path in args using path-boundary check
+		// to avoid false matches on sibling paths (e.g., /tmp/gt matching /tmp/gt-old)
+		matchesTown := containsPathBoundary(line, absRoot) || containsPathBoundary(line, townRoot)
 		if !matchesTown {
 			// Check for --port <portStr> as a discrete argument to avoid
 			// false matches on PIDs or other numeric substrings


steveyegge:
Landed via fix-merge in 496d69a8. Thanks @Random-Word for the solid race condition fix — the idle-monitor coordination and port-release polling are clean solutions. Cherry-picked your 3 commits directly to main with attribution.

PR #2711, opened 2026-03-13
Fix idle polecat reuse when tmux session is still live

Henry-E:
## Summary
- stop treating polecats with live tmux sessions as idle/reusable even when their agent bead says `idle`
- revalidate session state under the reuse lock and skip branch-reuse fallback when the existing session is still alive
- add an integration regression test covering a live tmux session plus `agent_state=idle`

## Testing
- `go test ./internal/polecat ./internal/cmd`
- `go test -tags=integration ./internal/polecat -run TestManagerDoesNotTreatLiveSessionAsIdle -count=1`
- `go test ./...` *(fails on pre-existing unrelated `internal/config` Dolt port tests: `TestAgentEnv_NoDoltPortWithoutTownRoot`, `TestResolveDoltPort_NoConfig`, `TestAgentEnv_NoDoltPortWithoutConfig`, `TestResolveDoltPort_FromDaemonJSON`)*

### DIFF
--- a/internal/cmd/polecat_spawn.go
+++ b/internal/cmd/polecat_spawn.go
@@ -2,6 +2,7 @@
 package cmd
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"os/exec"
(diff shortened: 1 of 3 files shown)


Henry-E:
Hopefully this helps to harden `gt sling`. Ran into this a number of times last night

PR #2713, opened 2026-03-13
fix(mail): retry bd list without --flat for bd < v0.59 compatibility

athosmartins:
## Problem

bd v0.57.0 does not support `--flat`. The mail router's `runBdCommand` was not retrying without `--flat` like `beads/beads.go` does, causing `gt mail send` to fail with 'no agent found' for all dog/crew recipients.

Also fixes dolt-archive plugin PROD_DBS: 'bd' database does not exist on this instance.

## Changes

- `internal/mail/bd.go`: retry `bd list` without `--flat` when command fails (matching existing pattern in `beads/beads.go`)
- `plugins/dolt-archive/plugin.md`: correct PROD_DBS list (hq, gt, property_scrapers, wa)

## Test

Verified `gt mail send` works correctly after this fix on bd v0.57.0.

### DIFF
--- a/internal/mail/bd.go
+++ b/internal/mail/bd.go
@@ -76,6 +76,27 @@ func runBdCommand(ctx context.Context, args []string, workDir, beadsDir string, 
 
 	runErr := cmd.Run()
 
+	// If bd doesn't support --flat (pre-v0.59), retry without it.
+	if runErr != nil && strings.Contains(stderr.String(), "unknown flag: --flat") {
+		retryArgs := make([]string, 0, len(args))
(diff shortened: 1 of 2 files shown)


steveyegge:
Closing — while the code is correct, adding per-callsite retry scaffolding for bd < v0.59 is compat debt heading in the wrong direction as both products approach v1.0. The existing `--flat` pattern in beads.go should be removed rather than replicated. A better approach would be a minimum bd version check at startup. The dolt-archive PROD_DBS doc fix is valid but that part already landed separately. Thanks for the contribution!

PR #2714, opened 2026-03-13
fix(compactor-dog): base32 hash validation + correct DEFAULT_DBS

athosmartins:
## Problem

Two bugs in the compactor-dog plugin:

1. `validate_hash` regex was `^[a-fA-F0-9]+$` (hex) but Dolt uses base32 encoding for commit hashes (lowercase a-z, 0-9). This caused all valid Dolt hashes to fail validation.
2. `DEFAULT_DBS` was `hq,bd,gt` — the `bd` database does not exist on this instance. Correct list is `hq,gt,property_scrapers,wa`.

## Changes

- `plugins/compactor-dog/run.sh`: change hash regex to `^[a-z0-9]+$` (base32)
- `plugins/compactor-dog/run.sh`: fix DEFAULT_DBS to `hq,gt,property_scrapers,wa`

## Test

Verified compactor-dog runs successfully against all 4 databases after this fix.

### DIFF
--- a/plugins/compactor-dog/run.sh
+++ b/plugins/compactor-dog/run.sh
@@ -24,7 +24,7 @@ DOLT_PORT="${DOLT_PORT:-3307}"
 DOLT_USER="${DOLT_USER:-root}"
 COMMIT_THRESHOLD="${COMMIT_THRESHOLD:-500}"
 # Default production databases (matches reaper.DefaultDatabases)
-DEFAULT_DBS="hq,bd,gt"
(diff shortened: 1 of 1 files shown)


steveyegge:
Closing — the hash validation fix landed via #2693 which uses the more precise base32 range (`[a-v0-9]` matching Dolt's noms encoding). The DEFAULT_DBS change adds user-specific databases that don't belong in the default config. Thanks for the contribution\!

PR #2730, opened 2026-03-13
feat(plugins): add submodule-gitignore plugin for rig repo hygiene

outdoorsea:
## Summary

- New town-level plugin (12h cooldown) that scans rig project repos for untracked Gas Town operational files
- Appends a standard guard block to submodule .gitignore when GT files appear untracked
- Idempotent via guard marker comment

## Why

Gas Town drops operational files into any directory it works in. External repos and submodules lack GT gitignore entries, causing persistent noise in git status.

## Test plan
- [ ] Runs without error on rig with no submodules
- [ ] Detects untracked GT files and appends guard block
- [ ] Re-running is idempotent

Generated with Claude Code

### DIFF
--- a/plugins/submodule-gitignore/plugin.md
+++ b/plugins/submodule-gitignore/plugin.md
@@ -1,0 +1,35 @@
++++
+name = "submodule-gitignore"
+description = "Inject Gas Town gitignore entries into rig project repos to prevent GT operational files from appearing as untracked"
+version = 1
+
+[gate]
+type = "cooldown"
+duration = "12h"
+
+[tracking]
+labels = ["plugin:submodule-gitignore", "category:git-hygiene"]
+digest = true
+
+[execution]
+type = "script"
+timeout = "5m"
+notify_on_failure = true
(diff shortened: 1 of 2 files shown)


steveyegge:
Thanks for this, Jeremy — the problem is real. GT operational files showing up as untracked in project repos is annoying.

However, we think this is better handled outside core plugins:

- A documented `.gitignore` recipe (or a one-time `gt rig init` step that seeds the entries) fits better than a recurring 12h plugin that auto-commits to project repos
- Auto-pushing gitignore changes to upstream repos is high blast radius for something that only needs to happen once per project
- The hardcoded path in plugin.md (`/Users/jeremy/gt/`) would need to be generalized

If you want to ship this as a user script or a `gt rig setup-gitignore` one-shot command, we'd be happy to look at that approach. The pattern list itself is solid — it's the delivery mechanism (recurring auto-commit plugin) that doesn't fit.

Also see #2731 which we're closing for similar reasons.

PR #2731, opened 2026-03-13
feat(plugins): add gitignore-reconcile and submodule-commit plugins

outdoorsea:
## Summary

Two new automated git hygiene plugins that address recurring problems Gas Town doesn't currently fix automatically:

### gitignore-reconcile (6h cooldown)
- Scans all rig repos for tracked files matching an active .gitignore rule
- Root cause: .gitignore only blocks NEW files; files committed before the rule was added stay tracked
- On clean main branches: runs git rm --cached + commits the untracking
- On dirty/polecat branches: creates a chore bead instead to avoid interference

### submodule-commit (2h cooldown, opt-in per rig)
- Reads .gitmodules in opt-in rigs, checks each submodule for uncommitted changes
- Commits on the current branch with [skip ci], pushes with || true (local commit is priority)
- Updates parent repo submodule pointer when submodules were committed
- Opt-in via rig plugin config: plugins[submodule-commit].enabled = true

## Why

Polecats only operate on parent repo worktrees. Submodules accumulate real code changes with no automated commit workflow. Similarly, .gitignore rules added after first commit leave files permanently tracked without intervention.

## Test plan
- [ ] gitignore-reconcile: runs clean on repos with no tracked+ignored files
- [ ] gitignore-reconcile: untracks and commits files on clean main
- [ ] gitignore-reconcile: creates chore bead on dirty/polecat branches
- [ ] submodule-commit: skips rigs without opt-in config
- [ ] submodule-commit: commits submodule changes and updates parent pointer

Generated with Claude Code

### DIFF
--- a/plugins/gitignore-reconcile/plugin.md
+++ b/plugins/gitignore-reconcile/plugin.md
@@ -1,0 +1,138 @@
++++
+name = "gitignore-reconcile"
+description = "Auto-untrack files that are tracked but match an active .gitignore rule"
+version = 1
+
+[gate]
+type = "cooldown"
+duration = "6h"
+
+[tracking]
+labels = ["plugin:gitignore-reconcile", "category:git-hygiene"]
+digest = true
+
+[execution]
+timeout = "10m"
+notify_on_failure = true
+severity = "low"
++++
+
+# Gitignore Reconcile
+
+Scans all rig repos for files that are tracked in git but now match an active
+`.gitignore` rule. On clean `main` branches, runs `git rm --cached` to untrack
+them and commits. On dirty branches or active polecat worktrees, creates a
+chore bead instead to avoid interference.
+
+Root cause: `.gitignore` rules only block NEW files. Files committed before the
+rule was added continue to be tracked until manually untracked.
+
+## Step 1: Enumerate rig repos
+
+```bash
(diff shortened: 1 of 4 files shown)


steveyegge:
Thanks Jeremy — both problems here are legit, but we think the solutions work better as external tooling rather than core plugins.

**gitignore-reconcile**: This is the most useful of the batch. `git rm --cached` for retroactively-ignored files is a real need. But as a recurring auto-commit plugin it's risky — other collaborators who don't have the gitignore rule would see files disappear on pull. Better as a one-shot command (`gt rig cleanup-gitignore`) that a human runs deliberately.

**submodule-commit**: Auto-committing with `git add -A` and pushing submodule changes is too high-blast-radius for unattended operation. The `[skip ci]` commits could mask broken state, and the fake author (`Gas Town <gastown@local>`) muddies git history. This also references a specific rig (`lilypad_chat`) making it quite user-specific.

**General note**: Both products are heading toward v1.0 and we're being selective about new plugin surface area. These would be great as:
- User scripts in `~/bin/` or a personal plugins repo
- A `gt rig cleanup` subcommand (for gitignore-reconcile specifically)
- Recipes in the docs that people can run when needed

Happy to revisit if you want to reshape any of these into one of those forms.

PR #2737, opened 2026-03-13
fix: auto-supersede old MRs when new MR created for same issue

seanbearden:
## Summary

- When `gt mq submit` creates a new MR bead, it checks for other open MRs with the same `source_issue` and auto-closes them with `CloseReasonSuperseded`
- **Also deletes the superseded polecat remote branch**, which triggers GitHub's `head_ref_delete` to auto-close the orphaned PR
- Only polecat branches (`polecat/*`) are deleted — non-polecat branches are left alone to avoid closing contributor fork PRs (GH#2669)
- Adds `FindOpenMRsForIssue()` and `MatchesMRSourceIssue()` to the beads package with 9 unit tests

**Problem**: On laser-platform, issue `la-cagb2` produced 6 abandoned PRs before one finally merged. Each new polecat attempt created a fresh branch/MR, but nobody closed the old ones or their GitHub PRs.

## Test plan

- [ ] Submit MR for an issue that already has an open MR — old MR should be auto-closed, old polecat branch deleted, GitHub PR auto-closed
- [ ] Submit MR for an issue with no prior MRs — no supersession (normal path)
- [ ] Re-submit same branch (idempotent) — no supersession (existing MR reused)
- [ ] Non-polecat branch MR superseded — MR bead closed but branch NOT deleted (fork safety)
- [x] Unit tests for `MatchesMRSourceIssue`: exact match, partial ID prefix/suffix non-match, boundary cases
- [ ] Build clean

Fixes: gt-tt37

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2739, opened 2026-03-13
feat: inject prior attempt context when re-dispatching issue to polecat

seanbearden:
## Summary

- When sling dispatches an issue that already has an open MR from a prior polecat attempt, the new polecat now receives `prior_branch` and `prior_failure` as formula variables
- The `mol-polecat-work` formula template instructs the polecat to cherry-pick prior commits onto its fresh branch instead of reimplementing from scratch
- Adds `FindOpenMRsForIssue` to beads for querying open MRs by source issue

## Design alignment

This feature works **within** the existing fresh-branch isolation design, not against it:

- **Fresh branches are preserved**: Each polecat still gets its own unique `@timestamp` branch from main. The isolation guarantee (collision-free, no TOCTOU races) is untouched.
- **No branch reuse**: The new polecat does NOT check out the old branch. It creates its own fresh branch and cherry-picks commits from the prior attempt as a starting point.
- **Prior branch availability**: The old branch exists on the remote because `gt done` pushed it, and the refinery only deletes remote branches after successful merge. Failed MR branches persist.
- **Timing**: The lookup runs at sling time (before the new polecat starts work), while MR supersession (#2737) runs at `gt done` time (after the new polecat finishes). So the old MR is still open and discoverable when we look for it.

The enhancement adds **context** to the dispatch, not structural changes. The polecat gets told "there was a prior attempt on branch X that failed for reason Y" and can choose to cherry-pick that work rather than starting from zero.

## Problem

On laser-platform, issue `la-cagb2` required 6 polecat attempts — each started from scratch and lost all prior work. The prior polecat's commits were available on the remote but the new polecat had no way to know about them.

## How it works

1. Sling dispatches issue `la-cagb2` to a new polecat
2. `lookupPriorAttempt` queries `FindOpenMRsForIssue("la-cagb2")`
3. Finds open MR with `source_issue: la-cagb2`, extracts `branch: polecat/furiosa/la-cagb2@mm4heq3e`
4. Injects `prior_branch` and `prior_failure` as formula variables
5. Polecat creates its own fresh branch, then cherry-picks from `origin/<prior_branch>`

## Test plan

- [ ] Sling an issue that has a prior open MR — polecat should see `prior_branch` in formula vars and "Prior attempt found" in sling output
- [ ] Sling an issue with no prior MR — no prior attempt context injected (normal path)
- [ ] Polecat cherry-picks prior commits during branch-setup step
- [ ] Build clean, existing tests pass

**Note**: `FindOpenMRsForIssue` also appears in #2737 (auto-supersede). Whichever merges first, the other will need a rebase.

Fixes: gt-zqvj

### DIFF
--- a/internal/beads/beads_mr.go
+++ b/internal/beads/beads_mr.go
@@ -45,3 +45,26 @@ func (b *Beads) findMRForBranch(branch string, skipClosed bool) (*Issue, error) 
 
 	return nil, nil
 }
+
+// FindOpenMRsForIssue returns all open merge-request beads whose source_issue
+// matches the given issue ID. Used to find prior attempts when re-dispatching
+// an issue and to supersede old MRs when a new one is created.
+func (b *Beads) FindOpenMRsForIssue(issueID string) ([]*Issue, error) {
+	needle := "source_issue: " + issueID + "\n"
+
+	issues, err := b.ListMergeRequests(ListOptions{
+		Status: "open",
+		Label:  "gt:merge-request",
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	var matches []*Issue
+	for _, issue := range issues {
+		if strings.Contains(issue.Description, needle) {
+			matches = append(matches, issue)
+		}
+	}
+	return matches, nil
+}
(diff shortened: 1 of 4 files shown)


steveyegge:
Landed via fix-merge in f613ef14. Thanks @seanbearden — clean solution for preserving polecat work across re-dispatches. Cherry-picked to main with attribution.

PR #2747, opened 2026-03-14
fix(compactor-dog): auto-discover databases via SHOW DATABASES instead of hardcoded names

DreadPirateRobertz:
## Problem

`compactor-dog/run.sh` had `DEFAULT_DBS="hq,bd,gt"` hardcoded. Actual database names on the server are `beads_hq`, `bd_001`, `gt` — so the script failed immediately when run by agent dogs, unable to find any databases.

Reported by `deacon/dogs/charlie` (escalation hq-mry1, 2026-03-13).

## Fix

1. **Default to `auto`** — the script already had a `SHOW DATABASES` discovery path when `--databases auto` was passed. This PR makes that the default instead of the broken hardcoded list.

2. **Fix test-DB exclusion regex** — the previous pattern used `^...$` anchors which only excluded exact names like `"testdb_"` (unlikely to exist), not prefixed names like `testdb_foo123`. Split into two greps: exact matches for system DBs, prefix-match for ephemeral test DBs.

## Changes

`plugins/compactor-dog/run.sh`:
- `DEFAULT_DBS="hq,bd,gt"` → `DEFAULT_DBS="auto"`  
- Fix `grep -v` to use prefix match for test DB patterns

## Testing

Compactor-dog can now be run without `--databases` override and will correctly discover `beads_hq`, `bd_001`, `gt` (and any future databases) via `SHOW DATABASES`.

### DIFF
--- a/plugins/compactor-dog/run.sh
+++ b/plugins/compactor-dog/run.sh
@@ -23,8 +23,9 @@ DOLT_HOST="${DOLT_HOST:-127.0.0.1}"
 DOLT_PORT="${DOLT_PORT:-3307}"
 DOLT_USER="${DOLT_USER:-root}"
 COMMIT_THRESHOLD="${COMMIT_THRESHOLD:-500}"
-# Default production databases (matches reaper.DefaultDatabases)
-DEFAULT_DBS="hq,bd,gt"
+# Default: auto-discover production databases via SHOW DATABASES.
+# Override with --databases db1,db2,... for an explicit list.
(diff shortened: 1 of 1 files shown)


steveyegge:
Landed via fix-merge in c44432fd (stripped beads contamination from the .beads/backup/ files). Thanks @DreadPirateRobertz — the auto-discovery default and regex fix are clean. For future PRs, run `bd preflight --check` to catch beads data leaking into the diff.

PR #2751, opened 2026-03-14
fix: shell-quote agent args and respect role_agents model flags

aphexcx:
## Summary

- **BuildCommand() now shell-quotes args** containing special characters (e.g., `sonnet[1m]`). Without quoting, zsh in tmux `respawn-pane` glob-expands `[1m]` as a character class, killing refinery/deacon sessions on startup.
- **Witness/daemon startup respects role_agents model flags.** The TOML `start_command` shortcut for Claude agents hardcoded `exec claude --dangerously-skip-permissions` without the `--model` flag from `role_agents`. Witnesses configured for `claude-sonnet` silently ran with the default model instead of `sonnet[1m]`. Now falls through to `BuildStartupCommandFromConfig` for proper model resolution.
- Custom (non-Claude) `start_command` patterns still use the TOML template expansion path.

## Test plan

- [x] Existing `TestBuildWitnessStartCommand_UsesRoleConfig` passes (custom non-Claude commands still use TOML pattern)
- [x] `TestBuildWitnessStartCommand_DefaultsToRuntime` passes
- [x] `TestRuntimeConfigBuildCommand` passes (args with special chars are quoted)
- [x] `TestShellQuote` passes
- [x] Manually verified all 9 sonnet-role agents (4 witnesses, 4 refineries, deacon) start with `--model 'sonnet[1m]'`

### DIFF
--- a/internal/config/types.go
+++ b/internal/config/types.go
@@ -782,16 +782,23 @@ func DefaultRuntimeConfig() *RuntimeConfig {
 }
 
 // BuildCommand returns the full command line string.
-// For use with tmux SendKeys.
+// For use with tmux SendKeys and respawn-pane, where the string is
+// interpreted by the user's shell. Args containing shell-special
+// characters (e.g., brackets in "sonnet[1m]") are quoted to prevent
+// glob expansion.
 func (rc *RuntimeConfig) BuildCommand() string {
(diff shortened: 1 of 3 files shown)


steveyegge:
Merged to main. Shell-quoting fix and model flag resolution landed.

Issue #2707, opened 2026-03-13
session-hygiene plugin uses rig name instead of session prefix, kills all legitimate sessions


kingb:
## Summary

The `session-hygiene` plugin (`plugins/session-hygiene/plugin.md`) incorrectly identifies legitimate rig sessions as zombies and kills them. On 2026-03-13, it killed 22 sessions across all 9 active rigs, taking down the entire town. The deacon caught the incident, restarted all rigs, and disabled the plugin (gate set to `manual`).

## Root Cause

The plugin builds its allowlist of valid session prefixes from **rig names** (`.[].name`) but tmux sessions use the **beads prefix** (`.[].beads_prefix`). For any rig where these differ, every session is classified as a zombie and killed.

### The mismatch

The plugin does this (line 37):
```bash
VALID_PREFIXES=$(echo "$RIG_JSON" | jq -r '.[].name // empty' 2>/dev/null)
```

This produces: `acorn`, `agent_forge`, `claude_code`, `flora_data`, `gastown`, `mc_and_pip`, `seldon`, `shepherd`, `snoozegoose`, `tercuman`

But actual tmux sessions use the `beads_prefix` as their prefix:

| Rig name | beads_prefix (session prefix) | Match? |
|----------|-------------------------------|--------|
| acorn | acorn | ✅ |
| agent_forge | af | ❌ |
| claude_code | cc | ❌ |
| flora_data | fd | ❌ |
| gastown | gs | ❌ |
| mc_and_pip | map | ❌ |
| seldon | se | ❌ |
| shepherd | shep | ❌ |
| snoozegoose | sg | ❌ |
| tercuman | te | ❌ |

Only 1 of 10 rigs has a matching name/prefix. The plugin sees sessions like `af-witness`, `cc-refinery`, `se-crew-hari` and treats them all as zombies because `af`, `cc`, `se` aren't in its allowlist of rig names.

### Why this setup is vulnerable

Gas Town allows custom `beads_prefix` values (short abbreviations) for ergonomic bead IDs. Most rigs use an abbreviated prefix rather than the full rig name. The plugin was written assuming `name == session prefix`, which is only true for the default case where no custom prefix is set.

## Impact

- 22 legitimate tmux sessions killed (all witnesses, refineries, and active crew across 9 rigs)
- Complete town outage until the deacon detected mass death and restarted everything
- The plugin had a 30-minute cooldown gate, meaning it would re-kill everything every 30 minutes if not disabled

## The Fix

[...]

Issue #2687, opened 2026-03-12
gt dolt start should clean stale /tmp/mysql.sock


daniel-jasinski:
## Summary

After a Dolt crash, the stale `/tmp/mysql.sock` file prevents clean restarts:

```
level=warning msg="unix socket set up failed: file already in use: /tmp/mysql.sock"
```

The Dolt server continues with TCP-only, but subsequent restart attempts also hit this issue. Manual `rm -f /tmp/mysql.sock` is required before `gt dolt start` works cleanly.

## Suggested fix

Add `rm -f /tmp/mysql.sock` (or the configured socket path) to the `gt dolt start` sequence before launching `dolt sql-server`.

## Environment

- gt v0.12.0
- WSL2 (Linux 5.15)

PR #2755, opened 2026-03-14
fix: session prefix bugs and stale dolt socket cleanup

DreadPirateRobertz:
## Summary

- **stuck-agent-dog** (gh-2707): Fixed session name construction to use beads prefix from `rigs.json` instead of rig name from `gt rig list --json`. When prefix != name (e.g. rig "cfutons" with prefix "CF"), the plugin was looking for sessions like `cfutons-polecat-foo` instead of `CF-polecat-foo`, making it blind to all stuck/crashed polecats.
- **session-hygiene** (gh-2707): Added kill threshold safeguard — if >50% of sessions would be classified as zombies, abort and escalate instead of killing. Prevents a data source bug from causing a town-wide outage.
- **doltserver** (gh-2687): Clean stale `/tmp/mysql.sock` (or port-specific variant) before starting Dolt server. After a crash, the orphaned socket file causes "unix socket set up failed" warnings. Safe because `IsRunning()` already returned false at this point.

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./internal/doltserver/` passes
- [ ] Verify stuck-agent-dog correctly detects stuck polecats on rigs where beads prefix != rig name
- [ ] Verify session-hygiene aborts when >50% kill threshold exceeded
- [ ] Verify `gt dolt start` cleans stale socket after simulated crash

### DIFF
--- a/internal/doltserver/doltserver.go
+++ b/internal/doltserver/doltserver.go
@@ -1355,6 +1355,20 @@ func Start(townRoot string) error {
 		return err
 	}
 
+	// Clean stale Unix socket from prior crash. Dolt creates /tmp/mysql.sock by
+	// default (or a port-specific variant). If the server crashed, the socket file
+	// persists and Dolt warns "unix socket set up failed: file already in use".
+	// Safe to remove: if a Dolt server were actually running, IsRunning() above
+	// would have detected it and we'd have returned already. (gh-2687)
+	socketPath := "/tmp/mysql.sock"
+	if config.Port != 3306 {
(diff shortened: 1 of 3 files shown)


DreadPirateRobertz:
Note: The stale socket cleanup (gh-2687) in this PR overlaps with PR #2753 which has a more robust implementation using `lsof`. If #2753 lands first, the socket portion of this PR should be dropped — the session prefix fixes (stuck-agent-dog + session-hygiene threshold) remain independent.

Issue #2596, opened 2026-03-09
gt stale-binary warning is triggered by bd backup-only commits even with BD_BACKUP_ENABLED=false


Henry-E:
`gt` can immediately report itself as stale right after a clean rebuild in `gastown/mayor/rig`, even when no source files changed.

What’s happening:
- `make build` or `make install` builds `gt` from commit `A`
- afterward, Beads creates an automatic `bd: backup ...` commit on `main`
- that commit only updates `.beads/backup/*`
- `gt` stale checking compares the embedded build commit against repo `HEAD`, sees `A != B`, and warns that the binary is stale

Impact:
- repeated false-positive stale warnings
- confusing rebuild loops
- `origin/main` drifts because backup-only commits land on the build branch

Important detail:
- Gas Town already sets `BD_BACKUP_ENABLED=false`
- that setting is present in agent env setup specifically to prevent Beads from force-adding `.beads/backup/*` and committing/pushing them
- despite that, backup commits are still being created in some execution paths
- so the disable is not being honored universally, or some `bd` invocation path is bypassing the protected env

Nuance:
- this was observed on a Gastown fork
- in this repo, `.beads/backup/*` is already tracked, which amplifies the impact because backup writes become real commits immediately

Expected:
- `BD_BACKUP_ENABLED=false` should suppress these backup commits across all `gt`/`bd` execution paths
- and backup-only Beads commits should not make `gt` look stale even if they do occur

PR #2758, opened 2026-03-14
fix: skip stale-binary warning for beads-only commits

DreadPirateRobertz:
## Summary

- Added `onlyBeadsChanges()` helper that checks whether all commits between the binary's build commit and HEAD exclusively modify `.beads/` files
- When HEAD advanced only via beads-only commits (e.g., `bd backup`), `CheckStaleBinary` now returns `IsStale=false`
- This eliminates false-positive stale warnings and confusing rebuild loops when no source code actually changed

Closes gh-2596

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./internal/version/` passes
- [x] `go vet ./internal/version/` passes
- [ ] Create a beads-only commit on main, verify `gt doctor` does not report binary as stale
- [ ] Create a real source commit on main, verify stale warning still triggers correctly

### DIFF
--- a/internal/version/stale.go
+++ b/internal/version/stale.go
@@ -101,6 +101,14 @@ func CheckStaleBinary(repoDir string) *StaleBinaryInfo {
 	// Compare commits using prefix matching (handles short vs full hash)
 	// Use the shorter of the two commit lengths for comparison
 	if !commitsMatch(info.BinaryCommit, info.RepoCommit) {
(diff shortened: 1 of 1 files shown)


Henry-E:
lol, it would have been nicer to see a fix to prevent beads from autocommiting but i guess this doesn't matter anymore anyway because beads are no longer located inside of the repos / rigs?

PR #2776, opened 2026-03-14
fix: harden Dolt server ownership detection

Henry-E:
## Summary
- make Dolt server ownership checks prefer live process evidence over stale state
- use one consistent ownership predicate across `IsRunning`, `CheckPortConflict`, `VerifyServerDataDir`, and `KillImposters`
- add regression tests for rogue `--config` / stale-state scenarios

## Problem
Gas Town can adopt a rogue rig-local `dolt sql-server` just because it owns the configured port. When that happens, `gt dolt status` can look healthy while the active server is actually serving a rig-local `.beads/dolt` directory instead of the town `.dolt-data`, which makes databases like `hq` or `ba` appear missing.

## Fix
This change makes ownership detection prefer live process evidence in this order:
1. `--data-dir`
2. `--config`
3. process cwd
4. state file as a last-resort fallback

That same rule is now used consistently by startup, verification, conflict detection, and imposter killing, so stale state can no longer mask a foreign Dolt server.

## Testing
- `go test ./internal/doltserver`

### DIFF
--- a/internal/doltserver/doltserver.go
+++ b/internal/doltserver/doltserver.go
@@ -485,20 +485,7 @@ func IsRunning(townRoot string) (bool, int, error) {
 					// Verify it's actually serving on the expected port.
 					// More reliable than ps string matching (ZFC fix: gt-utuk).
 					if isDoltServerOnPort(config.Port) {
-						// Verify data-dir to avoid claiming another town's Dolt.
-						serverDataDir := getServerDataDir(townRoot, pid)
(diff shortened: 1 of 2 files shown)


steveyegge:
Landed via fix-merge in c44432fd. Thanks @Henry-E — excellent refactor. The single doltProcessMatchesTown() predicate with evidence hierarchy is exactly the right approach. Great test coverage too.

PR #2783, opened 2026-03-14
fix: force-push to DoltHub after compaction in Go daemon

seanbearden:
## Summary

- The Go daemon (`compactor_dog.go`) flattens commit history but never pushes to DoltHub
- After flatten, the commit graph is rewritten — standard push always fails with non-fast-forward
- The shell script (`run.sh:397`) already has `DOLT_PUSH('--force')`, but the daemon (the normal runtime path) was missing it
- Adds `compactorForcePush()` after successful compaction + GC: discovers remote from `dolt_remotes`, runs `DOLT_PUSH('--force')`
- Non-fatal: silently skips if no remote configured, logs warning on push failure

## Test plan

- [ ] Run compactor-dog daemon cycle on a database with DoltHub remote — should force-push after compaction
- [ ] Run on a database with no remote — should skip silently (no error)
- [ ] Verify DoltHub shows the flattened history after push
- [ ] Build clean

Fixes: gt-pszx

### DIFF
--- a/internal/daemon/compactor_dog.go
+++ b/internal/daemon/compactor_dog.go
@@ -22,6 +22,8 @@ const (
 	compactorQueryTimeout = 30 * time.Second
 	// compactorGCTimeout is the timeout for CALL dolt_gc() after compaction.
 	compactorGCTimeout = 5 * time.Minute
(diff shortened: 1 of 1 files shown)


steveyegge:
Content already landed on main. Force-push compaction cycle merged.

PR #2796, opened 2026-03-14
fix: suppress clone progress output when --json flag is set

### LINKED ISSUE #2793: bug: gt wl browse --json does not produce valid JSON because of "Cloning..." header
## Bug Description
`gt wl browse --json | jq` does not work because the output contains
```
Cloning hop/wl-commons...
✓ Cloned successfully
```

## Steps to Reproduce
1. gt wl browse --json | jq
2. `jq: parse error: Invalid numeric literal at line 1, column 8`

## Expected Behavior
only the json is, at least piped, to the next command. Not the meta information of pulling the DB.

## Actual Behavior
the download headers and the JSON are output, breaking the assumption that --json outputs only valid JSON

## Environment
- **OS**: macOS + Docker
- **Go version**: 
- **Gas Town version**: (`gt version`) gt @ da0d15ead9e895ffb8e7c21b4f76a26ca489f993

mmlac:
## Summary

Fixes #2793 — `gt wl browse --json | jq` fails because "Cloning..." and "Cloned successfully" messages are printed to stdout before the JSON output, producing invalid JSON.

## Changes

- Wrap progress messages in `wl_browse.go` with `if !wlBrowseJSON` guards
- Suppress clone stderr in JSON mode to prevent dolt clone progress from contaminating output

## Test plan

- [ ] `gt wl browse --json | jq` produces valid JSON without parse errors
- [ ] `gt wl browse` (without --json) still shows clone progress as before

### DIFF
--- a/internal/cmd/wl_browse.go
+++ b/internal/cmd/wl_browse.go
@@ -74,14 +74,20 @@ func runWLBrowse(cmd *cobra.Command, args []string) error {
 	cloneDir := filepath.Join(tmpDir, commonsDB)
 
 	remote := fmt.Sprintf("%s/%s", commonsOrg, commonsDB)
-	fmt.Printf("Cloning %s...\n", style.Bold.Render(remote))
+	if !wlBrowseJSON {
+		fmt.Printf("Cloning %s...\n", style.Bold.Render(remote))
+	}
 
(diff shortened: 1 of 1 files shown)


mmlac:
wrong base branch - closing

PR #2809, opened 2026-03-15
fix: compactor-dog run.sh false positive on concurrent writes

athosmartins:
## Problem

The integrity check in `plugins/compactor-dog/run.sh` used strict inequality (`!=`) to detect data integrity failures. This incorrectly flags the case where `postCount > preCount`, which is actually safe — concurrent writes during the flatten operation can add rows, since the merge base shifts and transactions are preserved.

Only data loss (`postCount < preCount`) should fail.

## Fix

Change the check from:
```bash
if [[ "$POST_COUNT" != "$PRE" ]]; then
```
to:
```bash
if [[ -n "$POST_COUNT" && "$POST_COUNT" -lt "$PRE" ]]; then
```

This matches the logic in `compactor_dog.go` and the same fix already applied in the town-level `plugins/` directory.

## Testing

The fix is consistent with the Go implementation in `compactor_dog.go` which uses `<` for the row count comparison.

### DIFF
--- a/plugins/compactor-dog/run.sh
+++ b/plugins/compactor-dog/run.sh
@@ -350,8 +350,11 @@ for entry in "${CANDIDATES[@]}"; do
       log "  WARNING: Table $TABLE appeared after compaction (new table?)"
       continue
     fi
-    if [[ "$POST_COUNT" != "$PRE" ]]; then
-      log "  INTEGRITY FAILURE: $DB.$TABLE — pre=$PRE post=$POST_COUNT"
+    # Only fail on data loss (postCount < preCount).
+    # postCount > preCount is safe: concurrent writes during flatten add rows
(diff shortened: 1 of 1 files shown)


seanbearden:
Heads up — our PR #2811 overlaps with this. Same file, same problem, different approach.

**#2809**: Change `!=` to `-lt` (only flag row loss, ignore concurrent inserts).

**#2811**: Redesigned verification — table presence as primary check, row counts as secondary with >100 threshold. Also adds `--set-upstream` to the force-push.

The `-lt` approach in #2809 is simpler but has a gap: it wouldn't catch a scenario where flatten loses rows in table A while concurrent writes add rows to table B — the net count could mask real loss. #2811's table-level check + high threshold catches this.

Happy to defer to whichever approach the maintainer prefers, or we can combine the best of both.

PR #2813, opened 2026-03-15
fix: wait-idle uses queue for Claude Code (tmux send-keys delivery lost)

seanbearden:
## Summary

- For Claude Code sessions (`HasTurnBoundaryDrain=true`), wait-idle now queues first, then runs `watchAndDeliver` as a fallback for idle agents
- Claude's `UserPromptSubmit` hook drains the queue on the next turn boundary (covers busy agents)
- `watchAndDeliver` polls for idle and delivers via `NudgeSession` if the hook doesn't drain first (covers idle agents with no incoming turns)
- Refactored agent preset lookup to happen once at the top of the wait-idle block

**Problem**: Direct `NudgeSession` (tmux send-keys) to idle Claude Code sessions was unreliable — text could go into the void. Queue-only delivery fixed busy agents but left idle agents stuck (no turns = no hook fire = queue never drains).

**Solution**: Queue + watcher. Two delivery paths race:
1. **Hook drain** — if Claude starts a new turn, `UserPromptSubmit` fires and drains the queue
2. **Watcher drain** — if Claude stays idle, `watchAndDeliver` detects idle and delivers via `NudgeSession`

Whichever fires first wins. The queue's atomic drain (rename-based claiming) prevents double delivery.

## Test plan

- [x] Nudge busy Claude session — queue drains on next turn boundary
- [x] Nudge idle Claude session — watcher detects idle and delivers
- [ ] Non-Claude agents (Gemini, Codex) — unchanged behavior
- [ ] Queue failure fallback — should still attempt direct NudgeSession
- [ ] Build clean

### DIFF
--- a/internal/cmd/nudge.go
+++ b/internal/cmd/nudge.go
@@ -186,45 +186,65 @@ func deliverNudge(t *tmux.Tmux, sessionName, message, sender string) error {
 			// rather than silently degrading to immediate (destructive) delivery.
 			return fmt.Errorf("--mode=wait-idle requires a Gas Town workspace")
 		}
-		// Check if the target agent supports prompt-based idle detection.
-		// WaitForIdle uses Claude Code's prompt pattern (❯) and status bar (⏵⏵).
-		// Non-Claude agents (Gemini, Codex, etc.) have no ReadyPromptPrefix,
-		// so WaitForIdle produces false positives — it sees no busy indicator
-		// and matches stale prompt characters in the pane buffer. (GH#gt-5ey3)
-		// Degrade to queue mode for agents without prompt-based detection.
+		// Resolve agent preset for delivery strategy decisions.
+		var preset *config.AgentPresetInfo
 		if agentName, err := t.GetEnvironment(sessionName, "GT_AGENT"); err == nil && agentName != "" {
-			preset := config.GetAgentPresetByName(agentName)
-			if preset != nil && preset.ReadyPromptPrefix == "" {
-				fmt.Fprintf(os.Stderr, "wait-idle: %s agent %q has no prompt detection, using queue mode\n", sessionName, agentName)
-				if qErr := nudge.Enqueue(townRoot, sessionName, nudge.QueuedNudge{
+			preset = config.GetAgentPresetByName(agentName)
+		}
+
+		// Non-Claude agents (Gemini, Codex, etc.) have no ReadyPromptPrefix,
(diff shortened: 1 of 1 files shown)


DreadPirateRobertz (commented):
Good approach — queue-first for Claude Code sessions is the right call. The `UserPromptSubmit` hook drain mechanism is elegant since it naturally fires at turn boundaries.

One question: is there a timeout on the `watchAndDeliver` fallback? If the agent is busy for an extended period (e.g., long-running tool call), does the watcher give up and leave the message in the queue, or does it keep polling indefinitely?

DreadPirateRobertz (approved):
Code review: **A-** (strong approval)

The three-path refactor (no-prompt → queue-only, turn-boundary-drain → queue+watcher, other → WaitForIdle) is clean and well-reasoned. Key observations:

1. **Queue-first is correct**: For Claude Code, queuing before watching ensures the message survives even if the watcher times out. The `UserPromptSubmit` hook drain + `watchAndDeliver` backup covers both busy and idle agents.

2. **`watchAndDeliver` design is sound**: The exit conditions (queue empty, session gone, timeout) prevent leaks. Using `WaitForIdle` with a short timeout inside the watcher gets the consecutive-poll guard for free.

3. **Preset lookup refactor**: Moving `GetAgentPresetByName` to the top of the block and reusing it across all three paths eliminates redundant lookups.

4. **Minor nit**: The comment block at line 291-295 uses `/ Exit conditions:` (single-slash) instead of `// Exit conditions:`. Cosmetic only — doesn't affect functionality.

LGTM. This fixes a real delivery reliability issue for Claude Code sessions. Thanks @seanbearden!

DreadPirateRobertz (commented):
Reviewed by batty. The refactoring is clean — three delivery strategies (no-prompt → queue+poller, turn-boundary-drain → queue+watcher, default → WaitForIdle) are now clearly separated instead of nested.

Key observations:
1. Preset lookup lifted to top — good, avoids duplicate resolution
2. `watchAndDeliver` is the right fallback for idle Claude Code sessions where the hook won't fire
3. Queue failure fallback to NudgeSession is consistent across all paths
4. The `HasTurnBoundaryDrain` gate is correct — only Claude Code has this today

No issues found. Recommend merge.

seanbearden:
Closing — upstream already addresses this via WaitForIdle + queue + watchAndDeliver fallback. The poller.go change is also contradicted by upstream commit e6e0e87b which restored detached process groups.

PR #2819, opened 2026-03-15
feat(dolt): add automatic failover/failback for external Dolt servers

### LINKED ISSUE #2817: Bug: Dolt failover never reverts to external host after recovery
## Summary

When an external Dolt server becomes unreachable, the daemon correctly detects the outage (`DOLT_UNHEALTHY` signal, alerts to mayor/witnesses), but there is no mechanism to fail over to an alternate host or fail back when the primary recovers.

On 2026-03-04, when mini2 (100.111.197.110) went down, an agent improvised by mutating `~/.zshenv` to set `GT_DOLT_HOST=127.0.0.1`. When mini2 recovered, the system never failed back — the stale env var persisted, causing `bd`/mail to connect to localhost (which only had system DBs), triggering dozens of false escalations over days.

## Gap Analysis

`daemon.json` already declares `fallback_hosts`:
```json
"dolt_server": {
  "host": "0.0.0.0",
  "external": true,
  "fallback_hosts": ["100.111.197.110:3307", "100.86.9.58:3307", "127.0.0.1:3307"]
}
```

But nothing in the codebase reads `fallback_hosts`:

- **`DoltServerConfig` struct** (`internal/daemon/dolt.go`): No `FallbackHosts` field — the JSON key is silently ignored during unmarshalling.
- **`IsDoltUnhealthy()`**: Exported but never called outside its own package.
- **Recovery callback** (`daemon.go:358`): Only triggers a convoy sweep — no host failover.
- **Witness patrol formula**: No dolt health or failover steps.
- **Deacon `dolt-health` step**: Runs `gt health --json` and escalates — no failover.
- **Doctor Dog**: Read-only probing, no corrective action.
- **No agent documentation** (CLAUDE.md, design docs, formulas) mentions failover, `GT_DOLT_HOST` mutation, or `~/.zshenv` as a failover mechanism.

## Three Root Problems

1. **No automatic failover**: `fallback_hosts` config exists but is never parsed or used. When the primary host is unreachable, the daemon just alerts and hopes someone fixes it.
2. **No automatic failback**: Even if failover happened (via agent improvisation), there's no mechanism to detect primary recovery and revert.
3. **`~/.zshenv` is not a failover mechanism**: An agent wrote `GT_DOLT_HOST=127.0.0.1` to `~/.zshenv` as an improvised fix. This persists across reboots and overrides all other config, with no expiry or health check. The env var approach is fundamentally wrong for transient failover state.

## Proposed Solution

Build failover into the daemon's `DoltServerManager`:

1. Parse `fallback_hosts` from `DoltServerConfig`
2. On health check failure for external server, try the next host in the fallback list
3. Track the active host as runtime state in `daemon/dolt-state.json` (not `~/.zshenv`)
4. On each health tick, probe the primary host — if it recovers, fail back automatically
5. Add agent documentation prohibiting `~/.zshenv` mutation for failover

## Impact

Without this fix, any future external Dolt outage will either:
- Go unresolved until a human intervenes, or
- Get "fixed" by an agent improvising env var mutations that cause cascading failures when the primary recovers

trillium:
# Overview

 I have a setup where I dolt on an `always on` system and access at over tail scale. I'm running into problems where  agents are manually sitting my dolt db in  my bash profile to a local instance, causing me to have to go back and investigate over and over again. This Is hopefully a solved that problem.

> ⚠️ **Note:** This PR fixes the gastown (gt) side only. The `bd` binary (steveyegge/beads) hardcodes `DefaultDoltServerHost = "127.0.0.1"` in `internal/configfile/configfile.go:210` and `internal/storage/dolt/store.go:383`. Every `.beads/config.yaml` created without explicit `dolt.host` silently falls back to localhost. `bd` does not read `GT_DOLT_HOST` or the failover state file — a companion beads PR is needed to close the loop fully. Without it, every new rig/worktree/polecat workspace will still default to localhost and require manual patching.

## Summary

- Adds automatic failover/failback for external Dolt servers using the existing `fallback_hosts` config in daemon.json (which was previously declared but never parsed)
- When the primary host fails health checks, the daemon walks `fallback_hosts` in order and switches to the first reachable one
- On every subsequent health tick, the daemon probes the primary and automatically fails back when it recovers
- Active host state is persisted to `daemon/dolt-failover-state.json` — child processes (`bd`, `gt` subcommands) read this via `doltserver.DefaultConfig()`, no `~/.zshenv` mutation needed
- Agent documentation updated to explain the failover system and explicitly prohibit `~/.zshenv` mutation as a failover workaround

Closes #2817

## Files changed

| File | What |
|------|------|
| `internal/daemon/dolt.go` | `FallbackHosts` field, failover state, `tryFailover()`, `tryFailback()`, state persistence |
| `internal/daemon/dolt_failover_test.go` | 12 new tests |
| `internal/doltserver/doltserver.go` | `DefaultConfig` reads `dolt-failover-state.json` |
| `docs/design/dolt-storage.md` | New "External Server Failover" section |
| `internal/templates/townroot/claude.md` | Agent-facing failover docs + prohibition on `~/.zshenv` mutation |

## Test plan

- [x] `TestFailover_SwitchesToFallbackOnPrimaryFailure` — switches to fallback when primary unreachable
- [x] `TestFailover_NoFallbackHosts` — errors gracefully with no fallbacks configured
- [x] `TestFailover_AllFallbacksUnreachable` — errors when all fallbacks down
- [x] `TestFailback_RevertsWhenPrimaryRecovers` — reverts when primary comes back
- [x] `TestFailback_NoRevertWhenPrimaryStillDown` — stays on fallback when primary still down
- [x] `TestFailoverState_WrittenAndReadable` — state file written and parseable
- [x] `TestFailoverState_ClearedOnFailback` — state file removed on failback
- [x] `TestReadFailoverState_NoFile` — nil when no state file
- [x] `TestReadFailoverState_NotInFailover` — nil when InFailover=false
- [x] `TestFailover_SkipsPrimaryAndCurrentInFallbackList` — skips already-tried hosts
- [x] `TestEnsureRunning_ExternalFailoverIntegration` — end-to-end failover via EnsureRunning
- [x] `TestEnsureRunning_ExternalFailbackIntegration` — end-to-end failback via EnsureRunning
- [x] All existing daemon tests pass (no regressions)

### DIFF
--- a/internal/daemon/lifecycle_defaults_test.go
+++ b/internal/daemon/lifecycle_defaults_test.go
@@ -143,6 +143,7 @@ func TestEnsureLifecycleDefaults_FullyConfigured(t *testing.T) {
 			JsonlGitBackup:       &JsonlGitBackupConfig{Enabled: false},
 			DoltBackup:           &DoltBackupConfig{Enabled: false},
 			ScheduledMaintenance: &ScheduledMaintenanceConfig{Enabled: false, Threshold: &threshold},
+			Handler:              &PatrolConfig{Enabled: false},
 		},
 	}
 


trillium:
Closing in favor of #2827.

The root cause wasn't a failover problem — it was a configuration propagation gap. gt was propagating `BEADS_DOLT_PORT` to bd subprocesses but never `BEADS_DOLT_SERVER_HOST`. bd already supports this env var (has since at least v0.56), but gt never set it, so every new rig/worktree silently fell back to `127.0.0.1`.

#2827 fixes this by translating `GT_DOLT_HOST` → `BEADS_DOLT_SERVER_HOST` at all 6 env injection points, matching the existing port propagation pattern. No failover machinery needed — the agents just needed to be told where Dolt is.

PR #2827, opened 2026-03-15
fix: propagate BEADS_DOLT_SERVER_HOST so bd doesn't default to localhost

trillium:
## Summary

- Adds `BEADS_DOLT_SERVER_HOST` propagation everywhere gt already propagates `BEADS_DOLT_PORT`
- Translates `GT_DOLT_HOST` → `BEADS_DOLT_SERVER_HOST`, matching the existing port pattern
- Prevents bd from falling back to `127.0.0.1` when Dolt runs on a remote machine

## How the Dolt host flows through the system

```
~/.zshenv
  export GT_DOLT_HOST=100.111.197.110
       │
       ▼
  gt up / gt daemon start
  └─ doltserver.DefaultConfig() reads GT_DOLT_HOST
  └─ os.Setenv("BEADS_DOLT_SERVER_HOST", host)   ← THIS PR ADDS THIS
       │
       ▼
  AgentEnv builder (config/env.go)
  └─ Reads BEADS_DOLT_SERVER_HOST from process env
  └─ Falls back to GT_DOLT_HOST if BEADS_DOLT_SERVER_HOST not set
  └─ Injects into spawned agent tmux sessions
       │
       ▼
  Agent session (mayor, polecat, witness, etc.)
  └─ Inherits BEADS_DOLT_SERVER_HOST in env
       │
       ▼
  bd subprocess (via beads.go wrapper)
  └─ translateDoltPort() translates GT_DOLT_HOST → BEADS_DOLT_SERVER_HOST
  └─ filterBeadsEnv() preserves BEADS_DOLT_SERVER_HOST (not stripped)
       │
       ▼
  bd internal (store.go:540)
  └─ if BEADS_DOLT_SERVER_HOST set → use it
  └─ else → 127.0.0.1  ← the broken default
```

**Before this PR**: Steps 2-4 only propagated port, never host. Every bd subprocess fell back to `127.0.0.1`.

**Fragility note**: The entire chain starts from `~/.zshenv`. There is no town-level config file for the Dolt host. If `.zshenv` isn't sourced (new machine, non-zsh subprocess), the host is lost. A future improvement would be persisting this in `daemon.json` or a `gt config` file.

## Problem

bd (beads) hardcodes `DefaultDoltServerHost = "127.0.0.1"`. It already supports `BEADS_DOLT_SERVER_HOST` as an env var override (since `store.go` and `configfile.go` both check it), but gt never set it. Every new rig, worktree, or polecat workspace silently defaults to localhost, requiring manual `dolt.host` patching in each `.beads/config.yaml`.

This was causing recurring configuration drift — 23 config files had to be manually patched in a single session.

## Files changed

| File | What |
|------|------|
| `internal/cmd/up.go` | Set `BEADS_DOLT_SERVER_HOST` from `doltserver.DefaultConfig()` at startup |
| `internal/cmd/dashboard.go` | `ensureDoltPortEnv` now also sets host |
| `internal/daemon/daemon.go` | Daemon startup propagates host to process env |
| `internal/config/env.go` | `AgentEnv` builder includes host for spawned agents |
| `internal/beads/beads.go` | `filterBeadsEnv` preserves host; `translateDoltPort` translates `GT_DOLT_HOST` → `BEADS_DOLT_SERVER_HOST` |
| `internal/rig/manager.go` | Rig init env includes host translation |
| `internal/testutil/cmd.go` | `CleanGTEnv` preserves `GT_DOLT_HOST` |

## Test plan

- [x] `go test ./internal/beads/` — pass
- [x] `go test ./internal/config/ -run Env` — all AgentEnv tests pass
- [x] `go test ./internal/testutil/` — CleanGTEnv tests pass
- [x] `go test ./internal/cmd/ -run "Dashboard|EnsureDoltPort"` — all pass

### DIFF
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -177,10 +177,14 @@ func New(config *Config) (*Daemon, error) {
 			logger.Printf("Dolt server management enabled (port %d)", patrolConfig.Patrols.DoltServer.Port)
 			// Propagate Dolt connection info to process env so AgentEnv() passes it to
 			// all spawned agent sessions. Without this, bd in agent sessions
-			// auto-starts rogue Dolt instances. (GH#2412)
+			// auto-starts rogue Dolt instances or connects to localhost. (GH#2412)
 			portStr := strconv.Itoa(patrolConfig.Patrols.DoltServer.Port)
 			os.Setenv("GT_DOLT_PORT", portStr)
 			os.Setenv("BEADS_DOLT_PORT", portStr)
+			if patrolConfig.Patrols.DoltServer.Host != "" {
+				os.Setenv("GT_DOLT_HOST", patrolConfig.Patrols.DoltServer.Host)
+				os.Setenv("BEADS_DOLT_SERVER_HOST", patrolConfig.Patrols.DoltServer.Host)
+			}
 		}
 	}
 


trillium:
## Open question: should the Dolt host be persisted in a config file?

Right now the entire propagation chain starts from `export GT_DOLT_HOST=...` in `~/.zshenv`. This works but is fragile:

- New machine setup requires knowing to add it to `.zshenv`
- Non-zsh subprocesses or environments that don't source `.zshenv` lose it
- There's no single source of truth that's part of the town itself

I noticed that `DoltServerConfig` in `daemon/dolt.go` already has a `Host` field (`json:"host,omitempty"`), which maps to `daemon.json` → `patrols.dolt_server.host`. So there's infrastructure for a config-file-based approach that isn't being used — `daemon.json` currently has an empty `dolt_server: {}`.

Possible approaches:
1. **`daemon.json`** — `patrols.dolt_server.host` already exists in the struct, just needs to be populated and read as a source of truth
2. **A town-level config** (e.g., `~/.config/gastown/config.yaml` or `gt config set dolt.host ...`)
3. **Keep `.zshenv`** — simplest, already works, just needs documentation

I don't know what the convention is here or whether there's already a preferred pattern for town-level persistent config that I'm not seeing. Deferring to maintainer on the right approach.

PR #2841, opened 2026-03-15
fix: GetRigLED checks operational state before session state

### LINKED ISSUE #2555: fix: GetRigLED shows green for parked/docked rigs when sessions still running

DreadPirateRobertz:
## Summary
- `GetRigLED()` now checks parked/docked state BEFORE checking session state
- Parked rigs always show parking indicator, docked rigs always show stop indicator
- Previously, sessions still running during the teardown race window would show green/yellow instead of the correct park/dock indicator

Fixes #2555

## Test plan
- [x] `go build ./...` passes
- [x] `TestGetRigLED` updated and passing — now verifies parked/docked overrides session state

### DIFF
--- a/internal/cmd/rig.go
+++ b/internal/cmd/rig.go
@@ -687,20 +687,23 @@ func runRigAdd(cmd *cobra.Command, args []string) error {
 //   - 🅿️ = parked (intentionally paused)
 //   - 🛑 = docked (global shutdown)
 func GetRigLED(hasWitness, hasRefinery bool, opState string) string {
+	// Check operational state FIRST — parked/docked overrides session state.
+	// Sessions may still be running during the race window after park/dock
+	// but before sessions are killed (GH#2555).
+	switch opState {
+	case "PARKED":
+		return "🅿️"
+	case "DOCKED":
+		return "🛑"
+	}
+
 	if hasWitness && hasRefinery {
 		return "🟢"
 	}
 	if hasWitness || hasRefinery {
 		return "🟡"
 	}
-	switch opState {
-	case "PARKED":
-		return "🅿️"
-	case "DOCKED":
-		return "🛑"
-	default:
-		return "⚫"
-	}
+	return "⚫"
 }
 
 // rigStatePriority returns a sort priority for a rig's state.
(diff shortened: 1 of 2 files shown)


steveyegge (approved):
DPR review: clean behavioral fix. Reorders GetRigLED to check operational state (PARKED/DOCKED) before session state, fixing race window where sessions still running after park/dock showed green/yellow instead of park/dock indicator (GH#2555). Tests updated consistently. CI failures (TestEnsureLifecycleDefaults_FullyConfigured, TestSchedulerDirectConvoyDispatch) are pre-existing and unrelated.

PR #2872, opened 2026-03-16
fix: send MERGE_READY mail to refinery when polecat MR is pending (gs-9px)

alexsiri7:
## Summary
- Witnesses detect polecat MR submissions but never sent MERGE_READY mail to refinery
- Added protocol.NewMergeReadyMessage() calls to notifyRefineryMergeReady() and processDiscoveredCompletion()
- Passes mail.Router through the call chain

## Context
Bug: gs-9px - Witnesses not sending MERGE_READY mail to refineries


DreadPirateRobertz (approved):
Clean PR — properly threads mail.Router through the call chain so MERGE_READY mail actually gets sent. Tests updated correctly (nil router for unit tests). The nil guard on router is good defensive practice. LGTM.

PR #2884, opened 2026-03-16
fix: cross-prefix deps queryable via raw SQL fallback

### LINKED ISSUE #2786: bd dep add succeeds for cross-prefix deps but dep is not persisted/queryable
When adding a `tracks` dependency between beads in different databases (e.g., `hq-cv-*` convoy in HQ → `gt-*` issue in a rig), `bd dep add` reports success but the dependency is silently lost.

## Reproduction

```bash
# Add cross-prefix tracking dep
bd dep add hq-cv-fts6o gt-pye --type=tracks
# → {"depends_on_id": "gt-pye", "issue_id": "hq-cv-fts6o", "status": "added", "type": "tracks"}

# Query it back — empty
bd dep list hq-cv-fts6o --direction=down --type=tracks --json
# → []

# Show also has no deps
bd show hq-cv-fts6o --json | jq '.[0].dependencies'
# → null
```

## Impact
- All convoy progress tracking shows `0/0 completed` for single-bead slings where the convoy (`hq-cv-*`) tracks a rig bead (`gt-*`, `bd-*`, etc.)
- `gt convoy status`, `gt convoy list`, `gt convoy check` (auto-close), and `gt convoy land` all rely on `getTrackedIssues()` which calls `bd dep list` then falls back to `bd show` deps — both return empty
- Convoys never auto-close because they can never see their tracked issues as complete
- `convoy.go` line 2141 already has a comment: *"cross-database deps where the JOIN fails — see GH #2624"*

## Root cause hypothesis
The dependency is being written to one Dolt database (likely the HQ database where the convoy lives) but the cross-database JOIN needed to resolve the `gt-*` prefix target fails silently. The `bd dep add` command doesn't verify the dep is round-trippable after writing.

## Expected behavior
Either (a) cross-prefix deps should work end-to-end, or (b) `bd dep add` should fail loudly when it detects the target is in a different database and the dep won't be queryable.

## Workaround
None for convoy progress display. Workflows that manually track bead status (e.g., `bd list --status=closed`) are unaffected.

## Related
- #2624

DreadPirateRobertz:
## Summary
- `getTrackedIssues()` now uses `bdDepListRawIDs` as a fallback when `bdDepListTracked` returns empty
- This resolves cross-database dependency queries where the JOIN-based `bd dep list` silently fails
- Convoys tracking beads in other databases (e.g., `hq-cv-*` → `gt-*`) now correctly show progress
- Added test for the cross-prefix raw SQL resolution path

## Test plan
- [x] `go build ./cmd/gt` — compiles
- [x] `TestGetTrackedIssues_FallsBackToShowTrackedDependencies` — passes (full fallback chain)
- [x] `TestGetTrackedIssues_RawSQLResolvesCrossPrefixDeps` — passes (new test, raw SQL succeeds)
- [x] `TestRunConvoyList_UsesTownRootAndStripsBeadsDir` — passes
- [x] `TestRunConvoyStatus_UsesTownRootAndStripsBeadsDir` — passes

Fixes #2786

### DIFF
--- a/internal/cmd/convoy.go
+++ b/internal/cmd/convoy.go
@@ -2236,8 +2236,16 @@ func getTrackedIssues(townBeads, convoyID string) ([]trackedIssueInfo, error) {
 		return nil, fmt.Errorf("querying tracked issues for %s: %w", convoyID, err)
 	}
 
-	// Fallback: when dep list returns empty (common for cross-database deps),
-	// parse tracked dependencies from bd show output.
+	// Fallback 1: raw SQL query on dependencies table — works for cross-database
+	// deps where bd dep list's JOIN fails silently. (GH#2786, GH#2624)
+	if len(trackedIDs) == 0 {
+		trackedIDs, err = bdDepListRawIDs(townBeads, convoyID, "down", "tracks")
+		if err != nil {
+			return nil, fmt.Errorf("raw sql dep query for %s: %w", convoyID, err)
+		}
+	}
+
+	// Fallback 2: parse tracked dependencies from bd show output.
 	if len(trackedIDs) == 0 {
 		trackedIDs, err = bdShowTrackedDeps(townBeads, convoyID)
 		if err != nil {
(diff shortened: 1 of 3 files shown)


DreadPirateRobertz:
Closing in favor of #3085 — clean rebase with conflict resolved. Same fix, no accumulated drift.

Issue #2656, opened 2026-03-11
gt down race condition: daemon respawns immediately after shutdown


Henry-E:
## Bug

`gt down` stops the daemon (Phase 4) before killing all agent sessions (Phases 1-3 handle refineries/witnesses, but the daemon restart happens during teardown). A still-alive agent or the boot dog can restart the daemon within 1 second of it being killed.

### Evidence from daemon.log

```
2026/03/10 23:52:30 Received signal terminated, shutting down
2026/03/10 23:52:30 Daemon shutting down
2026/03/10 23:52:30 Daemon stopped
2026/03/10 23:52:31 Daemon starting (PID 61389)
2026/03/10 23:52:31 Daemon running, recovery heartbeat interval 3m0s
```

The daemon came back 1 second after `gt down` killed it. All tmux sessions remained running overnight.

### Root cause

Looking at `down.go`, the shutdown order is:

1. Polecats (only with `--polecats`)
2. Refineries
3. Witnesses
4. **Daemon** (Phase 4)
5. Dolt (Phase 4b)

The daemon is killed while agents may still be running commands that trigger `ensureDaemon` or `daemon start`. The daemon should be the last thing stopped, not the first infrastructure component.

## Expected

`gt down` should stop the daemon last (after all agent sessions are dead), or prevent the daemon from being restarted during shutdown (e.g., by writing a shutdown sentinel file that `daemon start` checks).

DreadPirateRobertz:
**Analysis:**

The shutdown order in \`down.go\` (Phases 1-4) stops refineries → witnesses → town-level sessions → daemon → Dolt. The problem is crew sessions are never stopped by \`gt down\` (only polecats with \`--polecats\`), so a running crew session can trigger \`ensureDaemon\` after the daemon is killed.

**Proposed fix — shutdown sentinel approach:**

1. Before Phase 4 (daemon stop), write a sentinel file: \`$TOWN_ROOT/daemon/.shutting-down\`
2. \`daemon.Start()\` / \`ensureDaemon()\` checks for this sentinel and refuses to start
3. Phase 4 stops the daemon
4. Phase 4b stops Dolt
5. After all phases, remove the sentinel

This prevents respawn without requiring crew sessions to be killed. The sentinel is cleaned up at the end of shutdown or on next \`gt up\`.

The sentinel file would also help with the related \`--all\` cleanup path (Phase 5), which currently waits 500ms and checks for respawned processes — a sentinel would prevent respawn in the first place.

[...]

PR #2885, opened 2026-03-16
fix: prevent daemon respawn during gt down shutdown (#2656)

DreadPirateRobertz:
## Summary
- `ensureDaemon()` now checks `IsShutdownInProgress()` before spawning a new daemon process
- `Daemon.Run()` also refuses to start if the shutdown lock is held
- Fixes the race where a dying session (e.g., mayor) triggers `ensureDaemon` after `gt down` kills the daemon in Phase 4, spawning a replacement that outlives the teardown

## Root cause
`gt down` kills the daemon in Phase 4, but town sessions (mayor, boot, deacon) are killed in Phase 3. If a session's exit hook or respawn logic calls `ensureDaemon` before the process fully dies, a new daemon starts within ~1 second. The new daemon has no awareness of the in-progress shutdown and keeps all tmux sessions alive overnight.

## Fix
Both `ensureDaemon` and `Daemon.Run` now check the existing `shutdown.lock` (held by `gt down` for the entire teardown). If the lock is held, they silently return without starting.

## Test plan
- [x] `go build ./internal/cmd/... ./internal/daemon/...` — compiles clean
- [x] `go test ./internal/daemon/ -run TestShutdown` — passes
- [ ] Manual: `gt up`, then `gt down` — verify daemon does not respawn

### DIFF
--- a/internal/cmd/up.go
+++ b/internal/cmd/up.go
@@ -476,7 +476,18 @@ func disableCurrentAgentDND(townRoot string) (bool, error) {
 }
 
 // ensureDaemon starts the daemon if not running.
+// Refuses to start during an active shutdown to prevent the race condition
+// where a dying session restarts the daemon after gt down kills it (#2656).
 func ensureDaemon(townRoot string) error {
+	// Check shutdown lock before starting a new daemon.
+	// During gt down, sessions are killed after the daemon. A dying session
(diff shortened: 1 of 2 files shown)


DreadPirateRobertz:
Closing as duplicate of #2907 (already merged).

PR #2886, opened 2026-03-16
fix: implement Dolt failover/failback for multi-host setups (#2817)

### LINKED ISSUE #2817: Bug: Dolt failover never reverts to external host after recovery

DreadPirateRobertz:
## Summary

- Adds `FallbackHosts` field to `DoltServerConfig` so `fallback_hosts` in `daemon.json` is no longer silently ignored
- On health check failure, the daemon cycles through fallback hosts and activates the first reachable one
- Periodically probes the primary host while on a fallback, and automatically fails back when it recovers
- Propagates active host to process env vars (`GT_DOLT_HOST`, `BEADS_DOLT_SERVER_HOST`) so spawned agents connect to the correct server

Closes #2817

## Test plan

- [ ] Unit tests for `initFailoverHosts` (dedup, bare host normalization, empty case)
- [ ] Unit tests for `tryFailover` with live TCP listener as fallback
- [ ] Unit tests for `tryFailback` with rate limiting and recovery detection
- [ ] Verify `go build ./...` passes
- [ ] Verify `go vet ./internal/daemon/` passes

### DIFF
--- a/internal/daemon/dolt.go
+++ b/internal/daemon/dolt.go
@@ -78,6 +78,16 @@ type DoltServerConfig struct {
 	// detection of Dolt server crashes without changing the overall
 	// heartbeat frequency. Default 30s.
 	HealthCheckInterval time.Duration `json:"health_check_interval,omitempty"`
+
+	// FallbackHosts is an ordered list of host:port pairs to try when the
+	// primary host becomes unreachable. The daemon cycles through them on
+	// health check failure and fails back to the primary when it recovers.
+	// Example: ["100.111.197.110:3307", "100.86.9.58:3307", "127.0.0.1:3307"]
+	FallbackHosts []string `json:"fallback_hosts,omitempty"`
+
+	// FailbackProbeInterval is how often to probe the primary host when
+	// running on a fallback, to detect recovery and fail back. Default 60s.
+	FailbackProbeInterval time.Duration `json:"failback_probe_interval,omitempty"`
 }
 
 // DefaultDoltServerConfig returns sensible defaults for Dolt server config.
@@ -142,6 +152,13 @@ type DoltServerManager struct {
 	// Protected by mu.
 	onRecoveryFn func()
 
+	// Failover state: tracks which host we're actively connected to.
+	// allHosts is [primary, fallback1, fallback2, ...] built at init.
+	// activeHostIdx indexes into allHosts; 0 = primary.
+	allHosts       []string // primary + fallback hosts as "host:port"
+	activeHostIdx  int      // index into allHosts; 0 = primary
+	lastFailback   time.Time // last time we probed primary for failback
+
 	// Test hooks (nil = use real implementations; set only in tests)
 	healthCheckFn      func() error
 	writeProbeCheckFn  func() error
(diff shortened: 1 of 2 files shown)


DreadPirateRobertz:
CI failures (Test: `TestEnsureLifecycleDefaults_FullyConfigured` and Windows Build) are pre-existing failures in main, not caused by this PR. These were fixed by commits `3106337a` and `252afed5` which landed after this branch was created. A rebase against main should clear them.

Issue #2767, opened 2026-03-14
agent-beads-exist overcounts missing crew identities when cross-rig worktrees exist in crew directories


Henry-E:
## Summary

`agent-beads-exist` appears to treat every non-hidden directory under `<rig>/crew` as a real crew worker. That inflates the missing-agent count when a crew directory contains extra worktrees, such as fix branches or cross-rig worktrees.

This is probably not the main source of the recurring failures, but it does make the reported count noisier and less trustworthy.

## Symptoms

In a town with a cross-rig or extra worktree under a crew directory, `gt doctor` reports extra missing crew beads for directories that are not canonical crew identities.

Example shape:
- `gastown/crew/scout`
- `gastown/crew/scout-fix-polecat-reuse`

The second directory is a worktree, not a real crew identity, but it is still counted as an expected crew agent bead.

## Expected behavior

Only canonical registered crew identities should be counted when computing expected crew agent beads.

## Actual behavior

The check enumerates directories on disk and assumes each is a worker identity.

## Evidence

`AgentBeadsCheck` builds expected crew identities from `listCrewWorkers()`:

- `internal/doctor/agent_beads_check.go`

`listCrewWorkers()` currently:
- reads `<rig>/crew`
- returns every non-hidden directory name

There is no filtering for:
- cross-rig worktrees
- temporary/fix worktrees
- directories that are not registered crew identities

In a live town this can inflate the missing count by at least one for a single extra worktree.

## Suspected root cause

The check uses raw filesystem enumeration as a proxy for identity membership, but crew directories can contain non-identity worktrees.

## Proposed fix

Possible approaches:
- derive expected crew identities from registered crew state rather than raw directory listing
- or at least filter out known cross-rig/temporary worktrees
- or reuse the logic already behind the `crew-worktrees` doctor check to exclude non-canonical worktrees

## Acceptance criteria

- extra worktrees in crew directories do not create fake missing-agent failures
- `agent-beads-exist` only counts canonical crew identities
- regression coverage exists for a rig with one real crew worker and one extra worktree directory

DreadPirateRobertz:
Fixed in PR #2836: `listCrewWorkers()` now checks if `.git` is a directory (real clone) vs a file (worktree symlink). Worktrees created by `gt worktree` in crew directories are excluded from identity counting.

DreadPirateRobertz:
## Triage Analysis

[...]

Issue #2766, opened 2026-03-14
gt upgrade does not repair missing identity beads, and rig add/adopt treat identity creation as warning-only


Henry-E:
## Summary

Missing rig and agent identity beads are not reliably repaired by normal upgrade/setup flows.

There appear to be two contributing gaps:

1. `gt upgrade` does not run the doctor checks that recreate missing rig and agent identity beads
2. `gt rig add` and `gt rig adopt` treat rig/witness/refinery identity bead creation failures as warnings and continue

This means a transient creation failure can leave a town permanently missing identity beads, and the normal upgrade path does not converge the workspace back to healthy state.

## Symptoms

In a real town, `gt doctor` reports failures such as:

- `rig-beads-exist: 23 rig identity bead(s) missing`
- `agent-beads-exist: ...` for witness/refinery identities
- `rig-config-sync: 23 missing identity bead(s)`

Running `gt upgrade` does not repair these missing identity beads.

## Expected behavior

- `gt upgrade` should backfill missing identity beads as part of converging the town to healthy state
- `gt rig add` / `gt rig adopt` should not silently succeed if core identity beads were not created

## Actual behavior

Full `gt doctor` includes the relevant checks:
- `NewAgentBeadsCheck()`
- `NewRigBeadsCheck()`

But `upgradeDoctor()` does not register them.

Also, rig setup paths print warnings and continue if identity bead creation fails.

## Evidence

Full doctor registers identity repair checks:

- `internal/cmd/doctor.go`
  - includes `NewAgentBeadsCheck()`
  - includes `NewRigBeadsCheck()`

Upgrade omits them:

- `internal/cmd/upgrade.go`
  - `upgradeDoctor()` does not register those checks

Rig creation/adoption is warning-only:

- `internal/cmd/rig.go`
  - `gt rig add` warns and continues if rig identity bead creation fails
  - `gt rig add` warns and continues if witness/refinery bead creation fails
  - `gt rig adopt` does the same

Town install also downgrades missing town-level agent bead creation to a warning:

- `internal/cmd/install.go`

## Suspected root cause

Identity beads are currently treated as semi-optional during setup, while upgrade only runs a subset of doctor checks and misses the ones that would repair this class of damage.

## Proposed fix

- Add `NewAgentBeadsCheck()` to `upgradeDoctor()`
- Add `NewRigBeadsCheck()` to `upgradeDoctor()`
[...]

PR #2887, opened 2026-03-16
fix: stop creating agent identity beads as ephemeral wisps

### LINKED ISSUE #2768: agent identity beads are created as wisps and get deleted by routine wisp GC
## Summary

Agent identity beads appear to be modeled as ephemeral wisps, but routine wisp GC treats any old open wisp as abandoned and deletes it. In practice this makes `gt doctor` repeatedly report large numbers of missing agent beads even in otherwise normal towns.

This looks like the main root cause of the recurring `agent-beads-exist` failures.

## Symptoms

In a real town, `gt doctor` repeatedly reports failures like:

- `agent-beads-exist: 85 agent bead(s) missing`

The count closely matches the expected identity population:
- town-level agents: mayor, deacon
- per-rig agents: witness, refinery
- crew identities
- polecat identities

When inspecting sampled databases, the expected `gt:agent` rows are actually absent rather than merely misdetected.

## Expected behavior

Agent identity beads should be durable operational identities. They should survive normal patrol cycles, `gt doctor --fix`, and routine wisp cleanup.

## Actual behavior

Agent identity beads are created as ephemeral wisps and are eligible for generic wisp GC once they age past the stale threshold.

## Evidence

Agent creation uses `--ephemeral`:

- `internal/beads/beads_agent.go`
  - `CreateAgentBead()` always includes `--ephemeral`

Gas Town's own GC logic treats old open wisps as abandoned:

- `internal/doctor/wisp_check.go`
  - counts open wisps older than threshold as abandoned
  - `Fix()` runs `bd mol wisp gc`

Patrol formulas also invoke raw wisp GC:

- `internal/formula/formulas/mol-witness-patrol.formula.toml`
- `internal/formula/formulas/mol-deacon-patrol.formula.toml`
- `internal/formula/formulas/mol-refinery-patrol.formula.toml`

And `bd mol wisp gc --help` explicitly says:
- any wisp not updated within the age threshold and not closed is considered abandoned

So long-lived open agent identities look exactly like stale wisps.

## Suspected root cause

Agent identity state was moved onto the wisp substrate, but no exemption or alternate storage model was added for long-lived identities.

## Proposed fix

Preferred:
- stop creating agent identity beads as ephemeral wisps

Alternative:
- add a hard GC exemption for agent identity wisps, e.g. `type=agent` and/or `label=gt:agent`

I would strongly prefer making agent identities non-GC-able by construction rather than relying on GC exclusions.

## Acceptance criteria

- Agent identity beads survive repeated patrol cycles
- `gt doctor --fix` does not delete healthy agent identities
- `agent-beads-exist` stays clean after 24h+ of normal operation
- regression coverage exists for interaction between agent identities and wisp GC

DreadPirateRobertz:
## Summary

- Remove `--ephemeral` flag from `CreateAgentBead()` — agent beads now go to the issues table instead of wisps
- Remove `--ephemeral` update from `CreateOrReopenAgentBead()` — reopened agent beads also stay persistent
- Existing agent beads in the wisps table remain discoverable via `ListAgentBeads()` which queries both tables

**Problem**: Agent identity beads were created with `--ephemeral`, making them wisps subject to wisp GC. The reaper deleted them on every patrol cycle, causing `gt doctor` to report 85+ missing agent beads. This is the root cause of the identity beads cluster (#2766, #2767).

**Solution**: Stop creating agent beads as ephemeral. The `--ephemeral` flag was originally added to avoid Dolt commit noise from agent state changes, but the GC side-effect made agent beads unreliable. When `bd` gains `--no-history` support (beads PR #2622), agent beads should switch to that for the best of both worlds.

Fixes #2768

## Test plan

- [x] `go build ./internal/beads/` — clean
- [x] `go vet ./internal/beads/` — clean
- [x] `go test ./internal/beads/ -run Agent` — passes
- [ ] Verify new agent beads appear in issues table (not wisps)
- [ ] Verify wisp GC no longer deletes agent beads
- [ ] Verify `gt doctor` agent-beads-exist check passes after patrol cycle

### DIFF
--- a/internal/beads/beads_agent.go
+++ b/internal/beads/beads_agent.go
@@ -221,7 +221,6 @@ func (b *Beads) CreateAgentBead(id, title string, fields *AgentFields) (*Issue, 
 			"--description=" + description,
 			"--type=agent",
 			"--labels=gt:agent",
-			"--ephemeral",
 		}
 		if NeedsForceForID(id) {
 			a = append(a, "--force")
@@ -234,8 +233,9 @@ func (b *Beads) CreateAgentBead(id, title string, fields *AgentFields) (*Issue, 
 		return a
 	}
 
-	// Create ephemeral agent bead (wisps table). Agent operational state has
-	// zero git history consumers (gt-bewatn.9).
+	// Create persistent agent bead (issues table). Agent beads were previously
+	// ephemeral (wisps) but wisp GC deleted them (GH#2768). When bd gains
+	// --no-history support, switch to that to avoid Dolt commit noise.
 	out, err := b.run(buildArgs()...)
 	if err != nil {
 		out, err = b.run(buildArgs()...)
@@ -344,12 +344,8 @@ func (b *Beads) CreateOrReopenAgentBead(id, title string, fields *AgentFields) (
 	if _, err := target.run("update", id, "--type=agent"); err != nil {
 		return nil, fmt.Errorf("fixing agent bead type: %w", err)
 	}
-	// Ensure agent bead is ephemeral (wisp) — agent operational state has
-	// zero git history consumers (gt-bewatn.9)
-	if _, err := target.run("update", id, "--ephemeral"); err != nil {
-		// Non-fatal: the bead is functional without ephemeral flag
-		_ = err
-	}
+	// Agent beads are now persistent (GH#2768) — wisp GC was deleting them.
+	// Will switch to --no-history when bd supports it.
 
 	// Note: role slot no longer set - role definitions are config-based
 


mmlac:
Dolt 0.61.0 has --no-history now, I suggest switching to it right away instead of removing --ephemeral and adding a comment.

DreadPirateRobertz:
CI Test/Windows failures are the pre-existing failures fixed by #2892. Once that merges, rebase to clear. The fix itself is clean — 5 additions, 9 deletions, stops agent beads from being GC'd by wisp reaper.

mmlac:
@DreadPirateRobertz --no-history fix has already been merged into main. This PR should be closed.

steveyegge:
Closing — this is superseded by PR #2965 (merged 2026-03-17), which addressed the same GH#2768 issue with a `--no-history` flag instead of simply removing `--ephemeral`. The code on main no longer uses `--ephemeral` for agent beads.

PR #2892, opened 2026-03-16
fix(test): resolve 3 pre-existing CI failures blocking all PRs

DreadPirateRobertz:
## Summary
- Adds missing `Handler` field to lifecycle defaults test fixture (fixes `TestEnsureLifecycleDefaults_FullyConfigured`)
- Adds `t.Cleanup` DB drop to multi-rig scheduler setup (fixes `TestSchedulerMultiRigConvoyAutoResolve` flakiness)
- Adds Windows `bd.cmd`+PowerShell mock for verify persistence test (fixes `TestEnsureCustomTypes_VerifyPersistence` on Windows)

These 3 pre-existing test failures block **all** gastown PR merges. This combined fix unblocks the entire CI pipeline.

Fixes #2832

## Test plan
- [x] `go vet` passes on all 3 packages
- [x] `TestEnsureCustomTypes_VerifyPersistence` passes locally (Unix)
- [ ] CI Integration Tests pass (scheduler cleanup)
- [ ] CI Windows Build passes (PowerShell mock)
- [ ] CI Unit Tests pass (lifecycle defaults)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2914, opened 2026-03-16
fix: add --flat to bd list --json calls to guarantee JSON output (#2499)

### LINKED ISSUE #2499: gt doctor fails agent-beads-exist check due to bd list --json output format and ephemeral bead label skipping

DreadPirateRobertz:
## Summary
- Add `--flat` flag to `bd list --json` calls that parse JSON output
- Prevents tree-mode from overriding `--json` flag and returning non-JSON text
- Fixed in 4 locations: agent bead listing, mail router, daemon lifecycle, deacon stale hooks

## Context
When tree-view is the default `bd` display mode, `bd list --json` can return tree-formatted text instead of JSON. This causes `json.Unmarshal` failures, leading `gt doctor` to report missing agent beads and mail routing to fail silently.

Fixes #2499

## Test plan
- [ ] `gt doctor` agent-beads-exist check passes with tree-view as default bd mode
- [ ] Mail routing correctly discovers agent beads across rigs
- [ ] Daemon lifecycle queries return valid JSON

### DIFF
--- a/internal/beads/beads_agent.go
+++ b/internal/beads/beads_agent.go
@@ -221,7 +221,6 @@ func (b *Beads) CreateAgentBead(id, title string, fields *AgentFields) (*Issue, 
 			"--description=" + description,
 			"--type=agent",
 			"--labels=gt:agent",
-			"--ephemeral",
 		}
 		if NeedsForceForID(id) {
 			a = append(a, "--force")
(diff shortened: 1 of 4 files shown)


DreadPirateRobertz:
CI failures are the pre-existing failures being fixed in #2892. Needs rebase after that merges. The `--flat` flag addition is the right fix — prevents tree-mode from overriding `--json` and producing non-parseable output in downstream JSON consumers.

steveyegge:
Fix-merged to main as 96b72f71. Applied only the --flat changes (4 locations). Note: PR diff also included ephemeral→persistent agent bead changes, but those were already on main (GH#2768).

PR #2936, opened 2026-03-16
fix(sling): kill session unconditionally when reusing idle polecat

quad341:
## Summary
- When `gt sling` reuses an idle polecat, any existing tmux session is now killed unconditionally (not just when stale)
- Previously, a non-stale session caused `ReuseIdlePolecat` to return `ErrSessionRunning`, forcing allocation of a new polecat instead of reuse
- Heartbeat freshness can race with session lifecycle (compact/resume hooks refresh the heartbeat while Claude sits idle at a dead `❯` prompt), leaving the old session alive and new work undiscovered
- Also cleans up the heartbeat file after killing the session to prevent stale data from confusing `SessionManager.Start`

## Test plan
- [x] `TestReuseIdlePolecat_KillsLiveSession` — verifies a live (non-stale) session is killed instead of returning `ErrSessionRunning`
- [x] `TestReuseIdlePolecat_KillsStaleSession` — regression: stale session path still works
- [x] `TestReuseIdlePolecat_NoSessionNoop` — no-session case proceeds normally
- [x] Existing heartbeat tests pass
- [x] Full build passes

### DIFF
--- a/internal/cmd/polecat_spawn.go
+++ b/internal/cmd/polecat_spawn.go
@@ -2,7 +2,6 @@
 package cmd
 
 import (
-	"errors"
 	"fmt"
 	"os"
 	"os/exec"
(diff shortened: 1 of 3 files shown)


quad341:
i'll get to the test failures in a minute

quad341:
CI failures (`TestBuildDoltSQLCmd_Remote` and `TestEnsureLifecycleDefaults_FullyConfigured`) are pre-existing — they reproduce on main locally. Fix is in #2941.

quad341:
Flaky integration test (`TestSchedulerDirectConvoyDispatch`) fixed in #2942. Unit test failures fixed in #2941.

quad341:
All CI-blocking test fixes consolidated in #2942 (supersedes #2941). Once #2942 merges, this PR's CI should go green. The remaining `TestAutoRespawnHook_RespawnWorks` failure is a flaky tmux timing test that passes on main.

PR #2939, opened 2026-03-16
fix(prime): replace nonexistent /ralph-loop with inline instructions

quad341:
## Summary
- Ralph mode polecats died immediately (0 commits, 5 attempts) because `gt prime --hook` told Claude to run `/ralph-loop` — a slash command that was never provisioned
- Replaced `outputRalphLoopDirective` to emit inline iterative workflow instructions: shows formula steps (same as normal mode), adds iteration guidance, ends with `gt done`
- Removed dead `buildRalphPromptFromMolecule` helper that built the prompt for the nonexistent command

## Test plan
- [x] `TestOutputRalphLoopDirective_NoSlashCommand` — verifies no `/ralph-loop` reference, has inline instructions + `gt done`
- [x] `TestOutputRalphLoopDirective_WithFormula` — verifies formula steps shown inline in ralph mode
- [x] Full build passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2941, opened 2026-03-16
fix(tests): update stale assertions in doltserver and daemon tests

quad341:
## Summary
- `TestBuildDoltSQLCmd_Remote` expected `cmd.Dir=""` for remote connections, but GH#2537 intentionally changed `buildDoltSQLCmd` to always set `Dir` to `DataDir`. Updated assertion to match.
- `TestEnsureLifecycleDefaults_FullyConfigured` was missing the `Handler` patrol config added in PR #2775. Added it to the fixture so the test is truly "fully configured".

These failures are pre-existing on main (reproducible locally) and are blocking CI for PRs #2936 and #2939.

## Test plan
- [x] Both tests pass locally
- [x] Full build passes

### DIFF
--- a/internal/daemon/lifecycle_defaults_test.go
+++ b/internal/daemon/lifecycle_defaults_test.go
@@ -137,12 +137,13 @@ func TestEnsureLifecycleDefaults_FullyConfigured(t *testing.T) {
 		Type:    "daemon-patrol-config",
 		Version: 1,
 		Patrols: &PatrolsConfig{
-			WispReaper:   &WispReaperConfig{Enabled: false},
-			CompactorDog: &CompactorDogConfig{Enabled: false},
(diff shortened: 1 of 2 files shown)


quad341:
Integration test failure is `TestSchedulerDirectConvoyDispatch` — a flaky timing race (`issues.jsonl is newer than last import`, 1-second clock skew). Unrelated to this PR's changes (we only touch `doltserver_test.go` and `lifecycle_defaults_test.go`). Needs a re-run.

quad341:
Flaky integration test (`TestSchedulerDirectConvoyDispatch`) root-caused and fixed in #2942 — `runBdJSON` was missing `--allow-stale` on bd calls.

quad341:
Superseded by #2942 which now includes all three test fixes (scheduler flaky test + doltserver stale assertion + lifecycle defaults stale assertion).

PR #2946, opened 2026-03-16
fix(config): wire up Provider and preset defaults for custom agents

### LINKED ISSUE #2944: gt sling needs --model flag for per-polecat model override
## Summary

There is no way to override the Claude model for a single `gt sling` invocation. The `ANTHROPIC_MODEL` env var is propagated to the tmux session (confirmed via `tmux show-environment`) but Claude Code ignores it and uses its default model.

## Steps to Reproduce

1. Run: `ANTHROPIC_MODEL=claude-opus-4-6 gt sling <bead> <rig>`
2. Check tmux env: `tmux show-environment -t <session> | grep MODEL` → shows `ANTHROPIC_MODEL=claude-opus-4-6`
3. Observe Claude Code banner: shows "Sonnet 4.6" instead of "Opus 4.6"

## Expected Behavior

Either:
- Claude Code should respect `ANTHROPIC_MODEL` env var for model selection, OR
- `gt sling` should accept a `--model` flag that injects `--model <model>` into the Claude startup args

## Root Cause

Claude Code's model selection doesn't check `ANTHROPIC_MODEL` env var. The model must be passed via `--model` on the command line. Gas Town's `gt sling` has no `--model` flag to inject this.

The `--agent` flag was tried as a workaround (custom agent with `--model` in args), but custom agents break session lifecycle (#2943).

## Suggested Fix

Add `--model <model-id>` flag to `gt sling` that appends `--model <model-id>` to the Claude startup command args. This is simpler than fixing Claude Code's env var handling and gives Gas Town explicit control.

## Workaround

None currently. Role-level model config via `role_agents` works for persistent per-role overrides, but there's no per-invocation override.

## Environment

- Gas Town version: current main branch
- Claude Code version: v2.1.76
- OS: Linux (Docker container)

mmlac:
## Summary
- `gt config agent set` now infers Provider from the command binary name when it matches a known preset (e.g., "claude", "gemini", "codex"). Users can also supply `--provider` explicitly.
- `fillRuntimeDefaults` auto-fills Session, Tmux, PromptMode, and Instructions fields from preset defaults when not explicitly set, so custom agents inherit lifecycle metadata (ProcessNames, SessionIDEnv, etc.).
- Removed dead pi-specific special-case block (superseded by generic preset-based auto-fill).
- Added PromptMode: "arg" to the pi preset so it comes from the registry.

Fixes https://github.com/steveyegge/gastown/issues/2943
Fixes #2944 

## Test plan
- [x] New tests for `gt config agent set` provider inference
- [x] New tests for `fillRuntimeDefaults` preset merging
- [X] Manual: `gt config agent set myagent claude -- claude` then verify `gt config agent get myagent` shows Session/Tmux fields

### DIFF
--- a/internal/cmd/config.go
+++ b/internal/cmd/config.go
@@ -79,10 +79,16 @@ to all rigs in the town.
 The command can include arguments. Use quotes if the command or
 arguments contain spaces.
 
+The provider preset is inferred from the command binary name when it
+matches a known preset (e.g., "gemini", "claude"). Use --provider to
+set it explicitly for custom binary names. The provider controls
+session handling, tmux detection, hooks, and other runtime defaults.
+
 Examples:
   gt config agent set claude-glm \"claude-glm --model glm-4\"
   gt config agent set gemini-custom gemini --approval-mode yolo
-  gt config agent set claude \"claude-glm\"  # Override built-in claude`,
+  gt config agent set claude \"claude-glm\"  # Override built-in claude
+  gt config agent set my-bot my-bot-cli --provider claude  # Use Claude defaults`,
 	Args: cobra.ExactArgs(2),
 	RunE: runConfigAgentSet,
 }
@@ -225,7 +231,8 @@ Examples:
 
 // Flags
 var (
-	configAgentListJSON bool
+	configAgentListJSON  bool
+	configAgentSetProvider string
 )
 
 // AgentListItem represents an agent in list output.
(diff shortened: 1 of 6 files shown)


steveyegge:
Fix-merged to main (5b8b0a64). Resolved conflicts with recent preset changes and applied cleanly. Thank you for the contribution — custom agents now properly inherit lifecycle defaults from their provider preset.

PR #2954, opened 2026-03-16
fix(beads): fall back to bd show --json when bd list returns text

outdoorsea:
## Summary

- `bd list --json` has a regression in bd v0.59.0+ where it returns human-readable text instead of a JSON array under some conditions (e.g. when the result set is non-empty but the `--json` flag is ignored).
- Previously, non-JSON output from `bd list` was silently treated as an empty result (`return nil, nil`), causing issues to go missing from Gas Town tooling.
- This fix detects non-JSON output, parses bead IDs from the text using a regex, and fetches full issue data via `bd show --json` as a fallback.

## Changes

- `internal/beads/beads.go`: Replace silent nil return with ID-parsing fallback in `List()`. Add `bdListIDRe` regex and `parseIDsFromBDListText` helper. Add `regexp` import.
- `internal/beads/beads_list_test.go`: Unit tests for `parseIDsFromBDListText` covering empty input, "No issues found." text, single ID, and deduplication.

## Test plan

- [ ] `go build ./...` compiles cleanly
- [ ] `go test ./internal/beads/... -run TestParseIDs` passes
- [ ] Manually verify that `gt list` / issue listing works correctly when `bd list --json` returns text output

### DIFF
--- a/internal/beads/beads.go
+++ b/internal/beads/beads.go
@@ -10,6 +10,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"regexp"
 	"strings"
 	"sync"
 	"time"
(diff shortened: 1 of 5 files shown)


steveyegge:
The underlying bd list --json regression was fixed in beads v0.61 (now on main). The JSON output works correctly — no fallback needed. Thanks for flagging the issue.

PR #2957, opened 2026-03-17
feat: configurable window tint (window-style) theming

mrhbriggs:
## Summary

- **Adds configurable window background (tmux `window-style`) theming** that mirrors the existing status bar theme config system
- Default behavior is **OFF** — no window background coloring unless explicitly enabled via `theme.window_tint.enabled: true` in config
- When enabled with no specific theme, falls back to matching each session's status bar colors
- Resolution chain mirrors status bar: per-rig role → global role → per-rig → global → disabled (nil)

## Background

Commit `91e592d4` introduced `ApplyWindowStyle` which unconditionally applied status bar colors to the entire terminal window background. This was disruptive for users who prefer their terminal's default background.

This PR makes window tinting opt-in and fully configurable, using the same layered config system as status bar theming:

```json
{
  "theme": {
    "window_tint": {
      "enabled": true,
      "name": "ocean",
      "role_tints": {
        "witness": "rust",
        "refinery": "plum"
      }
    }
  }
}
```

## Changes

- **`tmux/theme.go`**: Add `WindowStyle` type, `Theme.Window` field
- **`tmux/tmux.go`**: Consolidate old `ApplyWindowStyle(Theme)` + `ApplyPaneStyle(*PaneStyle)` into single `ApplyWindowStyle(*WindowStyle)` — nil resets to terminal defaults, non-nil applies colors
- **`config/types.go`**: Add `WindowTint` config struct with `enabled`, `name`, `custom`, `role_tints` fields (JSON key: `window_tint`)
- **`session/window_tint.go`**: `ResolveWindowTint()` — 5-level resolution chain matching status bar theming. `IsWindowTintEnabled()` — checks if any config level enables window tinting
- **`cmd/theme.go`**: `gt theme apply` reads window tint config and applies/resets per session

## Design decisions

- **Zero caller changes**: `ConfigureGasTownSession` (10+ callers) passes `theme.Window` through — nil by default, so existing behavior is unchanged
- **Only `gt theme apply` reads config**: Session creation doesn't read window tint config; users opt in by running `gt theme apply` after configuring
- **Consolidated apply/reset**: Single `ApplyWindowStyle(*WindowStyle)` handles both set and reset (nil = `bg=default,fg=default`), avoiding branching logic at call sites

## Test plan

- [x] `gt theme apply --all` with `window_tint.enabled: true` — window backgrounds match status bar colors
- [x] `gt theme apply --all` with `window_tint.enabled: false` — window backgrounds reset to terminal defaults
- [x] `go test ./internal/tmux/ ./internal/session/ ./internal/config/ ./internal/cmd/` — all pass
- [ ] Verify per-rig role tint overrides work
- [ ] Verify custom color overrides work

### DIFF
--- a/internal/cmd/theme.go
+++ b/internal/cmd/theme.go
@@ -80,6 +80,7 @@ func init() {
 	themeCmd.AddCommand(themeCLICmd)
 	themeCmd.Flags().BoolVarP(&themeListFlag, "list", "l", false, "List available themes")
 	themeApplyCmd.Flags().BoolVarP(&themeApplyAllFlag, "all", "a", false, "Apply to all rigs, not just current")
+
 }
 
 func runTheme(cmd *cobra.Command, args []string) error {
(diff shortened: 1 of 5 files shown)


mrhbriggs:
Note: status bar theming at boot also doesn't read config — all boot-time callers use `tmux.AssignTheme(rig)` (hash-based) or hardcoded `tmux.MayorTheme()`/`tmux.DeaconTheme()`. Config-based resolution (`getThemeForRole`, `loadRigTheme`) only runs via `gt theme apply`.

Window tint should match this pattern: hash/hardcoded at boot, config-based on `gt theme apply`.

steveyegge:

PR #2960, opened 2026-03-17
fix: convoy create/add route cross-rig beads via town root

dannomayernotabot:
## Summary
- **Root cause**: `getTownBeadsDir()` returns the town root, but `filepath.Dir(townBeads)` was used to derive the working directory for `bd dep add` — going one level *above* the workspace. This caused `bd` to fail with "no beads database found" when tracking beads from other rigs (e.g., `mo-2sh.1` from the mayor directory).
- **Fix**: Use `townBeads` directly as the `Dir` for dep add commands (it already IS the town root), removing the incorrect `filepath.Dir()` calls in `runConvoyCreate`, `runConvoyAdd`, `findStrandedConvoys`, `notifyMayorSession`, and `getExternalIssueDetails`.
- **Bonus**: `getIssueDetails()` and `getIssueDetailsBatch()` now set `Dir` to the town root and strip `BEADS_DIR`, so cross-rig bead lookups during auto-naming also work correctly.

## Test plan
- [x] New test `TestConvoyCreate_DepAddUsesTownRoot` verifies dep add runs from town root with BEADS_DIR stripped
- [x] Existing `TestRunConvoyList_UsesTownRootAndStripsBeadsDir` passes
- [x] Existing `TestRunConvoyStatus_UsesTownRootAndStripsBeadsDir` passes
- [x] All convoy tests pass (`go test ./internal/cmd/ -run "Convoy|convoy"`)
- [ ] Manual: `gt convoy create "test" mo-2sh.1` from mayor directory should succeed

### DIFF
--- a/internal/cmd/convoy.go
+++ b/internal/cmd/convoy.go
@@ -707,9 +707,9 @@ func runConvoyCreate(cmd *cobra.Command, args []string) error {
 
 	// Notify address is stored in description (line 166-168) and read from there
 
-	// Run dep add from town root (parent of .beads) so bd routes correctly
-	// across rigs via routes.jsonl. Running from .beads breaks cross-rig routing.
-	townRoot := filepath.Dir(townBeads)
+	// Run dep add from town root so bd routes correctly across rigs via
+	// routes.jsonl. getTownBeadsDir() already returns the town root.
+	// StripBeadsDir prevents inherited BEADS_DIR from overriding routing.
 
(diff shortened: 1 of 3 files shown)


steveyegge:
Fix-merged to main. Verified all convoy routing tests pass. Pre-existing macOS symlink flake in TestConvoyResolveBeadsDir unrelated.

PR #2969, opened 2026-03-17
fix: prevent polecats from exiting without committing work

dannomayernotabot:
## Summary
- Auto-commits uncommitted polecat work in `gt done` instead of blocking with an error (the #1 polecat failure mode)
- Moves the commit step earlier in `mol-polecat-work` formula (right after implement, before review/build)
- Adds stronger "commit as you go" language to prevent work loss from session death

## Source
Polecat branch: `polecat/furiosa/gt-ou68@mmuyw1as`
MR bead: `gt-wisp-hfb`

## Test plan
- [x] Build passes (`go build ./...`)
- [x] Done-related tests pass

### DIFF
--- a/plugins/resource-monitor/plugin.md
+++ b/plugins/resource-monitor/plugin.md
@@ -1,0 +1,36 @@
++++
+name = "resource-monitor"
+description = "Monitor disk, memory, and working polecat count — escalate on threshold breach"
+version = 1
+
+[gate]
+type = "cooldown"
+duration = "15m"
+
+[tracking]
+labels = ["plugin:resource-monitor", "category:health"]
+digest = true
+
+[execution]
+timeout = "2m"
+notify_on_failure = true
+severity = "high"
++++
+
+# Resource Monitor
+
+Checks disk usage, memory availability, and working polecat count.
+Escalates via `gt escalate` if thresholds are breached.
+
+## Thresholds
+
+- Disk usage > 80% → HIGH escalation
+- Memory available < 10Gi → HIGH escalation
+- Working polecats > 22 → MEDIUM escalation
+
+## Usage
+
+```bash
+./run.sh                    # Normal execution
+./run.sh --rig <rig>        # Monitor polecats in a specific rig
+```


steveyegge:
Thanks for tackling this — losing uncommitted polecat work is a real pain point and we appreciate you digging into it.

However, we're going to close this because the fix is at the wrong layer. Polecats don't always write code — they might be doing reviews, investigations, triage, or other non-coding work. Hardcoding "must commit before exit" into `gt done` would break those workflows.

The right way to enforce a particular polecat workflow is through **formulas** (molecules). In fact, the existing `mol-polecat-work` formula already has this built in — `commit-changes` is step 7, before `submit-and-exit` at step 10. When you sling work using this formula, the polecat follows those steps in order.

If you want a stricter version that commits even more aggressively (e.g., right after implement, before self-review), you'd create a custom formula. Here's a sketch:

```toml
# mol-polecat-aggressive-commit.formula.toml
#
# Like mol-polecat-work, but commits immediately after implementation
# to minimize the window for work loss from session death.

[formula]
name = "mol-polecat-aggressive-commit"
type = "workflow"
description = "Polecat work lifecycle with early commit to prevent work loss."

[formula.variables.issue]
description = "The issue ID assigned to this polecat"
required = true

# Steps: same as mol-polecat-work but commit moved right after implement
[[steps]]
name = "load-context"
description = "Load context and verify assignment"

[[steps]]
name = "branch-setup"
description = "Set up working branch"
needs = ["load-context"]

[[steps]]
name = "implement"
description = "Implement the solution"
needs = ["branch-setup"]

[[steps]]
name = "checkpoint-commit"
description = "Commit work-in-progress immediately to prevent loss"
needs = ["implement"]
# Key insight: commit early, even before tests pass.
# A WIP commit on a branch is better than lost work.

[[steps]]
name = "self-review"
description = "Self-review changes"
needs = ["checkpoint-commit"]

[[steps]]
name = "run-tests"
description = "Run tests and verify coverage"
needs = ["self-review"]

[[steps]]
name = "final-commit"
description = "Amend or squash with final tested version"
needs = ["run-tests"]

[[steps]]
name = "submit-and-exit"
description = "Submit work and self-clean via gt done"
needs = ["final-commit"]
```

Then dispatch with:
```bash
gt sling <bead> <rig> --formula mol-polecat-aggressive-commit
```

This way you get the "don't lose work" safety net without breaking polecats that aren't writing code.

This PR also bundles the beads v0.61 upgrade and env var docs — the beads upgrade is being handled separately in #3019.

PR #2970, opened 2026-03-17
fix(doltserver): fix timing race and lsof dependency in dolt startup

### LINKED ISSUE #2968: gt dolt start fails on Linux when dolt takes >500ms to start (PID file wrongly removed + lsof missing)
## Environment
- gastown 0.12.1
- Linux (Kubernetes pod, CSI block storage)
- `lsof` not installed

## Symptom

`gt dolt start` consistently fails with:
```
Error: Dolt server failed to start (check logs with 'gt dolt logs')
```

## Root Cause

Two bugs in `internal/doltserver/doltserver.go` compound to cause this failure:

### Bug 1: `IsRunning` removes valid PID file (timing race)

In `Start`'s retry loop, `IsRunning` is called at 500ms after dolt is launched. When dolt is alive (`signal(0)` succeeds) but not yet listening on its port, `isDoltServerOnPort` returns false. The code then incorrectly removes the PID file as "stale" and returns `false`. The retry loop sees `!running` and exits immediately.

On systems with slow storage (CSI/NFS block storage in Kubernetes), dolt can take **1.5–2 seconds** to start listening. The first retry at 500ms fires well before dolt is ready.

The problematic code path in `IsRunning`:
```go
if err := process.Signal(syscall.Signal(0)); err == nil {
    if isDoltServerOnPort(config.Port) {
        // return true if process matches town
    }
    // Falls through if port not yet listening
}
// BUG: reached for alive-but-not-yet-listening processes too
_ = os.Remove(config.PidFile)  // ← removes valid PID file!
```

**Proposed fix** for `Start`'s retry loop — use `cmd.Process.Signal(0)` directly instead of `IsRunning`, avoiding the PID-removal side effect:
```go
    // Check if process is still alive (WITHOUT calling IsRunning which removes PID file)
    if err := cmd.Process.Signal(syscall.Signal(0)); err != nil {
        return fmt.Errorf("Dolt server process died during startup: %w", err)
    }
    // Check if accepting connections yet
    if err := CheckServerReachable(townRoot); err == nil {
        return nil  // success
    }
    lastErr = err
}
return fmt.Errorf("Dolt server started (PID %d) but not accepting connections after 5s: %w", cmd.Process.Pid, lastErr)
```

### Bug 2: `findDoltServerOnPort` requires `lsof` (not universally available)

`findDoltServerOnPort` calls `lsof -i :PORT -sTCP:LISTEN -t`. On many Linux systems (containers, minimal installs, Kubernetes pods), `lsof` is not installed. The function silently returns 0, making it impossible to detect a running dolt server by port — breaking the fallback path entirely.

```go
cmd := exec.Command("lsof", "-i", fmt.Sprintf(":%d", port), "-sTCP:LISTEN", "-t")
output, err := cmd.Output()
if err != nil { return 0 }  // lsof not found → always returns 0
```

**Proposed fix**: Use `ss -tlnp` as primary (widely available on modern Linux without extra packages), with `/proc/net/tcp` as a zero-dependency fallback:
```go
func findDoltServerOnPort(port int) int {
    // Try ss first (iproute2, available on most modern Linux)
    if pid := findPortWithSS(port); pid > 0 { return pid }
    // Fallback: /proc/net/tcp (no external tools required)
    if pid := findPortViaProcFS(port); pid > 0 { return pid }
    // Last resort: lsof (macOS, or Linux systems that have it)
    if pid := findPortWithLsof(port); pid > 0 { return pid }
    return 0
}
```

## Impact

`gt dolt start` and `gt up` are completely broken on this class of systems. The CHANGELOG for 0.12.1 mentions "Fixed stale socket cleanup, server ownership detection" — but these bugs still exist in 0.12.1.

## Steps to Reproduce

1. Use a Linux system where `lsof` is not installed (e.g., minimal Kubernetes pod)
2. Use a system where dolt takes >500ms to start (e.g., slow CSI block storage)
3. Run `gt dolt start`

## Diagnostics

Verified with `strace -e trace=openat,connect,socket,fcntl,flock,unlink,execve,kill`:

1. At t+500ms: TCP connect to `127.0.0.1:3307` → EPOLLERR (connection refused)
2. Immediately after: `unlink("dolt.pid")` — removes the valid PID file
3. `execve("/usr/sbin/lsof", ...)` → `ENOENT` — lsof not found
4. `IsRunning` returns false
5. `Start` returns "Dolt server failed to start"

Dolt startup time measured at **~1691ms** on this system (CSI block storage overhead).

The `WaitForReady` function already exists in the codebase with a similar retry-with-polling pattern — the `Start` function should use the same approach for the process-alive check.

baumgold:
## Summary

Fixes two compounding bugs that cause `gt dolt start` (and `gt up`) to always fail on Linux systems with slow storage and/or no `lsof` installed.

- **Bug 1 (timing race)**: `Start()`'s retry loop called `IsRunning()` to check if the dolt process was alive. When dolt is alive but not yet listening on its port, `IsRunning()` treats the PID file as stale and removes it, then returns `false`. The loop sees `!running` and exits with "Dolt server failed to start" — even though dolt is starting up normally. On CSI/NFS-backed storage (common in Kubernetes), dolt takes 1.5–2s to bind its port; the first check at 500ms consistently fires too early.

- **Bug 2 (lsof dependency)**: `findDoltServerOnPort()` used `lsof` as its only method. On many Linux systems (containers, minimal installs, Kubernetes pods), `lsof` is not installed. The function silently returned 0, breaking the fallback port-based detection in `IsRunning()`.

## Regression

Bug 1 is a regression introduced by commit 1cd600fc (*"fix: replace ps string matching with structural identity checks (gt-utuk)"*, 2026-03-01). That commit replaced `isDoltProcess(pid)` (process name check) with `isDoltServerOnPort(port)` (TCP dial) inside `IsRunning()`, conflating two distinct questions: process liveness vs. port readiness.

## Changes

**Fix for Bug 1** — in `Start()`'s retry loop, replace `IsRunning()` with a direct `cmd.Process.Signal(0)` check:
- A starting-up process is alive but not yet listening — this is normal, not stale
- `cmd.Process` is the exact process we launched, so no PID file lookup is needed
- Actual process death (signal returns error) still exits the loop immediately

**Fix for Bug 2** — add `ss` (iproute2) as a fallback in `findDoltServerOnPort()`:
1. `lsof` — kept as primary (cross-platform, reliable when available)
2. `ss` (iproute2) — standard on modern Linux; no extra packages required

## Test plan

- [x] `go build ./internal/doltserver/...` passes
- [x] `gt dolt start` succeeds on the affected system (Linux Kubernetes pod, CSI storage, no `lsof`) after this fix

Closes #2968

---

### DIFF
--- a/internal/doltserver/doltserver.go
+++ b/internal/doltserver/doltserver.go
@@ -665,30 +665,42 @@ func CheckPortConflict(townRoot string) (int, string) {
 }
 
 // findDoltServerOnPort finds a process listening on the given port.
-// Returns the PID or 0 if not found. Uses lsof to identify the listener PID.
+// Returns the PID or 0 if not found.
 // Does not verify process identity via ps string matching (ZFC fix: gt-utuk).
+//
+// Tries lsof first (macOS and most Linux), then ss (iproute2) as a fallback
+// for Linux systems where lsof is not installed.
 func findDoltServerOnPort(port int) int {
-	// Use lsof to find the LISTENING process on port (not clients connected to it).
+	// Try lsof — preferred when available (cross-platform).
 	// Without -sTCP:LISTEN, lsof returns client PIDs (e.g., gt daemon) first,
 	// which aren't dolt processes — causing false negatives.
 	cmd := exec.Command("lsof", "-i", fmt.Sprintf(":%d", port), "-sTCP:LISTEN", "-t")
-	output, err := cmd.Output()
-	if err != nil {
-		return 0
+	if output, err := cmd.Output(); err == nil {
+		lines := strings.Split(strings.TrimSpace(string(output)), "\n")
+		if len(lines) > 0 && lines[0] != "" {
+			if pid, err := strconv.Atoi(lines[0]); err == nil {
+				return pid
+			}
+		}
 	}
 
-	// Parse first PID from output
-	lines := strings.Split(strings.TrimSpace(string(output)), "\n")
-	if len(lines) == 0 || lines[0] == "" {
-		return 0
+	// Fall back to ss (iproute2) — standard on modern Linux, no extra packages needed.
+	// Example output line: LISTEN 0 128 *:3307 *:* users:(("dolt",pid=12345,fd=7))
+	cmd = exec.Command("ss", "-tlnp", fmt.Sprintf("sport = :%d", port))
+	if output, err := cmd.Output(); err == nil {
+		for _, line := range strings.Split(string(output), "\n") {
+			if idx := strings.Index(line, "pid="); idx >= 0 {
+				rest := line[idx+4:]
+				if end := strings.IndexAny(rest, ",)"); end > 0 {
+					if pid, err := strconv.Atoi(rest[:end]); err == nil && pid > 0 {
+						return pid
+					}
+				}
+			}
+		}
 	}
 
-	pid, err := strconv.Atoi(lines[0])
-	if err != nil {
-		return 0
-	}
-
-	return pid
+	return 0
 }
 
 // DoltListener represents a Dolt process listening on a TCP port.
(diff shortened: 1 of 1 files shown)


baumgold:
## Regression introduced in commit 1cd600fc

This fixes a regression from commit 1cd600fc8d21b1ed8ea1bfe8f79a1a7337b7f6bf (*"fix: replace ps string matching with structural identity checks (gt-utuk)"*, merged 2026-03-01).

That commit replaced `isDoltProcess(pid)` (a `ps` command-line check) with `isDoltServerOnPort(port)` (a TCP dial) inside `IsRunning()`. The original check correctly answered "is this process our dolt?" without caring about port readiness. The replacement conflates process liveness with port readiness, causing the PID file to be wrongly removed for a process that is alive but not yet bound to its port.

See issue #2968 for full diagnosis.

baumgold:
@steveyegge - do you have time to take a look at this one?  Thanks!

DreadPirateRobertz (approved):
Clean fix for both bugs. The key insight in the PR summary is correct: `IsRunning()` conflates two questions — process liveness and port readiness — and calling it during startup causes it to clean up its own PID file on slow storage.

**Bug 1 (timing race):** Using `cmd.Process.Signal(0)` is the right call here. We have a direct reference to the process we launched, so there's no need to go through PID file lookup. The PID file removal side effect in `IsRunning` is the actual hazard, and this sidesteps it cleanly.

**Bug 2 (lsof dependency):** The `ss` fallback is correct. The pid= extraction from ss output is simple and the sport filter means we are not grepping across unrelated lines. The pid= prefix is stable across all modern iproute2 versions.

Build check passes. This is a regression fix worth merging promptly given the impact on Kubernetes/containerized deployments.

PR #2972, opened 2026-03-17
fix(beads): route cross-rig agent bead creation from town root

EchBTI:
## Summary
Fix cross-rig agent bead creation when a routed bead ID is created from an imported rig or nested Gas Town checkout.

This change:
- makes `FindTownRoot` prefer the outermost town root
- runs routed `CreateAgentBead` calls from the town root so `bd` resolves the target route only once
- keeps hook slot writes on the same target-scoped execution path

## Why
On latest `main`, routed agent bead creation can still double-stack rig paths in imported-rig layouts, which breaks polecat/crew identity bead creation in cross-rig scenarios.

## Tests
- `go test ./internal/beads`
- added unit coverage for outermost town-root detection
- added a mock-`bd` regression test that asserts routed `CreateAgentBead` executes from town root with town-level `BEADS_DIR`

### DIFF
--- a/internal/beads/beads_agent.go
+++ b/internal/beads/beads_agent.go
@@ -212,6 +212,15 @@ func (b *Beads) CreateAgentBead(id, title string, fields *AgentFields) (*Issue, 
 	// Don't bail out — try the bd create calls anyway (GH#1769).
 	_ = EnsureCustomTypes(targetDir)
 
+	// For routed cross-rig bead IDs, run bd from the town root so bd's own
+	// prefix router resolves the target once. Running from a rig worktree with
+	// a routed BEADS_DIR can double-stack the path for imported rigs.
+	target := b
+	townRoot := b.getTownRoot()
+	if townRoot != "" && ExtractPrefix(id) != "" {
+		target = NewWithBeadsDir(townRoot, filepath.Join(townRoot, ".beads"))
+	}
+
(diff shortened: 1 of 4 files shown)


steveyegge:
Thanks for the detailed analysis and test coverage. However, we're going to close this because we think it's solving a problem that shouldn't exist rather than fixing the root cause.

A rig with its own `mayor/town.json` (creating a town-inside-a-town) is a misconfiguration, not a routing scenario that the code should accommodate. The right fix is to catch this with `gt doctor` rather than changing `FindTownRoot` traversal semantics.

The `FindTownRoot` change from "return first" to "return outermost" has a broad blast radius — it affects every caller, not just agent bead creation. In legitimate nested-town scenarios (dev/testing), you'd want the innermost root, not the outermost. Changing this globally to work around a layout that shouldn't exist risks breaking cases that do.

If you're hitting double-stacked paths in practice, we'd love to hear more about the specific layout that causes it — there may be a narrower fix at the rig-add or worktree layer.

EchBTI:
I’ll check, I had done a migration from local device to a server (mainly for a much needed reinstall) but also sometimes the Mayor under Opus goes rogue and just does whatever fixes it sees fit to get something working. I’ll assess and try send over specifics on this chain.

________________________________

PR #2973, opened 2026-03-17
fix(mail): fall back to explicit agent workspaces

EchBTI:
## Summary
Teach mail recipient validation to accept explicit workspace-style agent addresses when bead lookup fails.

This change:
- accepts `rig/crew/name` and `rig/polecats/name` in filesystem fallback
- preserves route-query errors without letting them mask a valid on-disk workspace
- adds regression coverage for route-error plus workspace-fallback behavior

## Why
In practice, agents sometimes need durable mail sent to explicit reviewer or polecat paths even when agent-bead queries are temporarily failing. Latest `main` still rejects some of those valid addresses.

## Tests
- `go test ./internal/mail -run 'TestValidateRecipientFilesystemFallback|TestValidateRecipientFilesystemFallbackWithRouteErrors'`
- `go test ./internal/mail`

### DIFF
--- a/internal/mail/router.go
+++ b/internal/mail/router.go
@@ -945,6 +945,7 @@ func (r *Router) validateRecipient(identity string) error {
 	}
 
 	// Query agents from rig-level beads via routes.jsonl
+	var routeQueryErr error
 	if r.townRoot != "" {
 		townBeadsDir := filepath.Join(r.townRoot, ".beads")
 		routes, err := beads.LoadRoutes(townBeadsDir)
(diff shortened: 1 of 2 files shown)


DreadPirateRobertz (commented):
The fix is correct — the original code returned the route-query error immediately instead of falling through to the filesystem check. Saving the error and only returning it after the filesystem fallback also fails is the right ordering.

The case-3 / paths in  fill a gap — these explicit workspace addresses (e.g., `gastown/crew/batty`) should work even if the agent bead lookup fails. The existing code would only accept them if the filesystem happened to have the right structure for the generic case. LGTM.

PR #2974, opened 2026-03-17
fix(codex): wait for idle before draining queued nudges

EchBTI:
## Summary
Reduce Codex session interruption by using Codex prompt/busy markers and deferring queued nudge delivery until the session is actually idle.

This change:
- gives the Codex preset a prompt prefix (`› `) for tmux idle detection
- makes tmux idle checks use the target session's agent prompt marker
- treats `esc to interrupt` as a generic busy indicator, not a Claude-only one
- updates the nudge poller to skip drain cycles while prompt-aware agents are still busy

## Why
Latest `main` already starts the Codex nudge poller correctly, but queued notifications can still inject mid-turn because the poller drains even after idle wait fails. That shows up as avoidable Codex conversation interruptions.

## Tests
- `go test ./internal/config ./internal/tmux ./internal/cmd -run 'TestCodexRuntimeConfigHasPromptDetection|TestHasBusyIndicator|TestShouldSkipDrainUntilIdle'`
- `go test ./internal/cmd`
- `go test ./internal/tmux`


steveyegge:

PR #2976, opened 2026-03-17
feat: gt up detects and recovers orphaned hooked beads after crash (gas-udp)

quad341:
## Summary
- After a crash, `gt up` now scans for beads in hooked/in_progress status assigned to dead polecats
- Checks if polecats actually exist (tmux session, worktree)
- For dead polecats: unhooks bead (reset to open), notifies mayor via mail

Issue: gas-udp
Polecat: jasper
Branch: polecat/jasper/gas-udp@mmv7yf01

---
*Merged by Gas Town Refinery*

### DIFF
--- a/internal/cmd/up.go
+++ b/internal/cmd/up.go
@@ -997,6 +997,7 @@ func recoverOrphanedBeads(townRoot string, rigs []string, prefetchedRigs map[str
 
 	for _, rigName := range rigs {
 		if _, ok := prefetchedRigs[rigName]; !ok {
+			fmt.Fprintf(os.Stderr, "[orphan-recovery] skipping rig %s (failed to load)\n", rigName)
 			continue // Rig failed to load — skip
 		}
 
--- a/internal/cmd/up_test.go
+++ b/internal/cmd/up_test.go
@@ -236,6 +236,10 @@ func TestRecoverOrphanedBeads_SkipsUnloadedRigs(t *testing.T) {
 
 func TestRecoverOrphanedBeads_NoOrphansCleanRig(t *testing.T) {
 	// Set up a rig directory with no beads — should produce no services.
+	// Note: Full recovery-path tests (hooked bead + dead polecat → reset to open)
+	// require a live Dolt server and are covered by DetectOrphanedBeads tests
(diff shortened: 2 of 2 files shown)


DreadPirateRobertz:
**Review: Correct approach, test coverage is thin**

The integration point is right: run orphan recovery in `runUp` after Dolt is ready, before witnesses start their own patrol. This catches the crash-recovery case cleanly.

**Concerns:**

1. **Test coverage**: The diff shows `TestRecoverOrphanedBeads_NoRigs` (empty rig list). Is there a test that covers the actual recovery path — a bead in `hooked` status with a dead polecat getting reset to `open`? That's the core behavior this PR adds, and it should have explicit coverage. If `witness.DetectOrphanedBeads` has its own tests, that's fine, but the integration with `runUp` should have at least one scenario with real orphans.

2. **`router.WaitPendingNotifications()` placement**: This is called after the rig loop. Is `DetectOrphanedBeads` sending notifications asynchronously? If the router queues notifications internally and `Wait` flushes them, the placement is correct. Just want to confirm this isn't a case where notifications could be dropped if the function returns early (e.g., `continue` on rig load failure).

3. **Silent skip on prefetch failure**: `if _, ok := prefetchedRigs[rigName]; !ok { continue }` silently skips rigs that failed to load. That's correct behavior (rig is broken anyway), but it might be worth logging a debug message so orphan recovery can be attributed in the `gt up` output.

**Everything else**: The `ServiceStatus` reporting is clean. Calling `DetectOrphanedBeads` from witness package is the right separation of concerns.

DreadPirateRobertz (commented):
Approach is correct — detecting orphaned hooked beads on `gt up` is exactly the right recovery hook. Test coverage is thin though: the current tests don't exercise the crash-recovery path (orphan present + bead still hooked + no active session). Can you add a test that simulates that state and verifies the recovery fires? Happy to approve once that's in.

DreadPirateRobertz (commented):
test

steveyegge:
Cherry-pick merged to main (up.go + up_test.go only). The .gitignore and compactor_dog changes were omitted — please submit those separately if still needed.

PR #2963, opened 2026-03-17
fix(polecat): provision CLAUDE.md in worktree so polecats call gt done

quad341:
## Summary
- Embed `templates/polecat-CLAUDE.md` and write it to polecat worktrees during spawn and repair
- The template has 3 separate `gt done` warning sections ("IDLE POLECAT HERESY", "Completion Protocol", "FINAL REMINDER") — but was **never provisioned** into worktrees
- The old approach relied on ephemeral `gt prime` output which scrolls past and gets lost after compaction
- CLAUDE.md is always loaded by Claude Code, making `gt done` instructions persistent across sessions
- Add `CLAUDE.md` to `gasTownIgnorePatterns()` so it doesn't pollute `git status` in worktrees

Fixes: polecats finish work then sit idle instead of calling `gt done` (BUG-polecats-no-gt-done.md)

## Test plan
- [x] `TestCreatePolecatCLAUDEmd` — verifies placeholders replaced, `gt done` present, heresy warning present
- [x] `TestCreatePolecatCLAUDEmd_NoOverwrite` — verifies existing CLAUDE.md preserved
- [x] Overlay/gitignore tests updated for new `CLAUDE.md` pattern
- [x] `go build ./...` clean
- [ ] CI passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2978, opened 2026-03-17
fix: gt polecat list shows idle for actively working polecats (gas-8e3)

quad341:
## Summary
- effectivePolecatState: promote StateIdle→StateWorking when session running
- loadFromBeads: relaxed agent bead HookBead fallback for cross-rig beads
- ReuseIdlePolecat: remove stale heartbeat on reuse

Fixes gas-8e3.

### DIFF
--- a/internal/daemon/compactor_dog.go
+++ b/internal/daemon/compactor_dog.go
@@ -450,19 +450,19 @@ func (d *Daemon) surgicalRebaseOnce(dbName string, keepRecent int) error {
 	// []uint8 (e.g. "1.00") which cannot be scanned directly into int.
 	var minOrderStr, maxOrderStr string
 	if err := db.QueryRowContext(ctx, "SELECT MIN(rebase_order), MAX(rebase_order) FROM dolt_rebase").Scan(&minOrderStr, &maxOrderStr); err != nil {
-		d.surgicalAbortAndCleanup(db, workBranch)
+		d.surgicalAbortAndCleanup(db)
 		return fmt.Errorf("read rebase bounds: %w", err)
 	}
 	minOrder, maxOrder, err := parseRebaseOrder2(minOrderStr, maxOrderStr)
 	if err != nil {
-		d.surgicalAbortAndCleanup(db, workBranch)
+		d.surgicalAbortAndCleanup(db)
 		return fmt.Errorf("parse rebase bounds: %w", err)
 	}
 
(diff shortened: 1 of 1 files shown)


DreadPirateRobertz (commented):
The `effectivePolecatState` logic for promoting Idle→Working when a session is running is the right fix. The current state was misleading — a running session should read as Working, not Idle.

The heartbeat clearing on reuse (`ReuseIdlePolecat`) prevents stale state carrying over, which is an important correctness fix. LGTM.

DreadPirateRobertz:
**Review: Mostly good — one edge case concern in loadFromBeads fallback**

**`effectivePolecatState` change**: Correct. A running session is ground truth for working state, regardless of stale beads state. The new test case covers it.

**`ReuseIdlePolecat` heartbeat removal**: Correct. Without this, the new session inherits a stale heartbeat and the `polecatSessionState` staleness check fires prematurely, causing `gt polecat list` to show idle. Good find.

**`loadFromBeads` relaxed fallback**: This is the one I'd like a second look at. The new path:
```go
if running, _ := m.polecatSessionState(name); running {
    return &Polecat{..., Issue: fields.HookBead}, nil
}
```
...trusts the `hook_bead` value even when `isCurrentHookedIssueForAssignee` failed. The failure cases for that check include:
- bead exists but wrong assignee (cross-rig sling timing) ← safe to trust session
- bead exists but wrong status (status changed) ← probably safe
- bead was deleted ← **`Show(fields.HookBead)` returns non-nil err**, so this path never fires (the `err == nil` guard catches it). Actually reading again: the outer `if hookIssue, err := m.beads.Show(fields.HookBead); err == nil` means the bead exists, so the fallback is only reached when the bead exists but `isCurrentHookedIssueForAssignee` returns false. In that case, trusting `hook_bead` when the session is running seems fine.

Actually on re-reading, the concern is resolved — `Show` success is the outer guard. The fallback only fires when the bead exists + session running + assignee/status mismatch. That's the cross-rig timing case the comment describes.

**Verdict**: LGTM. The `.beads/.gitignore` conflict note from #2963 applies here too.

steveyegge:
Cherry-pick merged to main (polecat.go + test). The manager.go changes were skipped (main already handles session-reuse differently). The compactor_dog lint fix was also skipped (main uses nolint:unparam instead).

PR #2980, opened 2026-03-18
fix(rig): scaffold default_agent config dir on gt rig add

baumgold:
## Summary

- When `default_agent=opencode` is set in town settings, `gt rig add` now scaffolds `.opencode/` in witness, refinery, and polecat directories instead of hardcoding `.claude/`
- Previously, the presence of `.claude/` caused `gt` to detect claude-code as the agent, overriding the town default
- Fix reads `default_agent` from town settings in `AddRig` and passes it to `ProvisionFor` dynamically (falls back to `claude` if unset)

## Test

Added `TestRigAddRespectsDefaultAgent` in `internal/cmd/rig_integration_test.go` — creates a town with `default_agent=opencode`, runs `AddRig`, and asserts `polecats/.opencode/` exists while `polecats/.claude/` does not.

Fixes: gt-vdx

### DIFF
--- a/internal/cmd/rig_integration_test.go
+++ b/internal/cmd/rig_integration_test.go
@@ -499,6 +499,59 @@ func TestRigAddCreatesCorrectStructure(t *testing.T) {
 	}
 }
 
+// TestRigAddRespectsDefaultAgent verifies that gt rig add scaffolds the polecat
+// config directory matching the town's default_agent setting (gt-vdx).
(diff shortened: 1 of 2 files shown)


DreadPirateRobertz (commented):
Clean fix. The approach of reading `TownSettings.DefaultAgent` in `AddRig` and passing it to `ProvisionFor` is the right level of abstraction. 

The test is complete — checks both that the correct dir exists and the wrong one doesn't. LGTM.

PR #2981, opened 2026-03-18
fix: repair PROJECT IDENTITY MISMATCH after crash or rig addition (gas-tc4)

quad341:
## Summary

- Fixes bd/Dolt PROJECT IDENTITY MISMATCH after crash or rig addition — server serves wrong database
- Adds `gt repair` command for identity mismatch detection and repair
- Updates doltserver and polecat/rig managers for robustness

**Issue**: gas-tc4
**Polecat**: obsidian

---

### DIFF
--- a/internal/rig/manager.go
+++ b/internal/rig/manager.go
@@ -860,19 +860,15 @@ func (m *Manager) verifyRigIdentity(rigPath, rigName string) error {
 
 	// Verify the database name matches what we expect.
 	// The database should be named after the rig (e.g., "gastown") not after
-	// a bd init artifact (e.g., "beads_gt").
+	// a bd init artifact (e.g., "beads_gt") or a stale value from another rig.
 	if metadata.DoltDatabase != "" && metadata.DoltDatabase != rigName {
-		// Check if this is a known prefix-based alias (short prefixes like "be" for "beads_el")
-		// In that case, the mismatch is intentional.
-		if strings.HasPrefix(metadata.DoltDatabase, "beads_") {
-			// This is a bd init artifact — EnsureMetadata should have fixed it.
-			// Force-repair it now.
-			if repairErr := doltserver.EnsureMetadata(m.townRoot, rigName); repairErr != nil {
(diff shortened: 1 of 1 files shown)


DreadPirateRobertz:
**Review: Good fix with two items to clarify**

**`EnsureMetadata` database mismatch repair**: Correct and important. The wrong `dolt_database` field causes silent connection to the wrong database, which is hard to diagnose. The stderr warning on auto-repair is appropriate.

**`gt repair` command**: Useful focused recovery command. The choice to run only `RigConfigSyncCheck` and `StaleDoltPortCheck` is reasonable for the identity-mismatch scenario described. Clear that `gt doctor` exists for broader diagnostics.

**Two concerns:**

1. **`verifyRigIdentity` only repairs `beads_*` prefixed databases.** The logic is:
```go
if strings.HasPrefix(metadata.DoltDatabase, "beads_") {
    // repair
}
```
But the issue title says "any wrong database name after crash or rig addition." What if the wrong name doesn't have a `beads_` prefix? The check silently passes in that case without warning or repair. Should the outer condition be `metadata.DoltDatabase != rigName` → warn (not just the `beads_` case → repair)?

2. **`gt repair` vs `gt doctor --fix` separation**: `gt repair` auto-fixes every issue it finds (no user confirmation), while `gt doctor --fix` also auto-fixes. If `gt repair` is a subset of `gt doctor --fix`, consider whether users need both or if `gt doctor --fix --checks=identity` would be clearer. Not a blocker, but worth discussing.

**Everything else**: The test for `EnsureMetadata_RepairsWrongDoltDatabase` directly covers the failing scenario and is well-structured. The `verifyRigIdentity` call placement (post-`success = true` in rig add flow) is correct — the rig is created, identity just might be wrong.

DreadPirateRobertz (commented):
test

steveyegge:
Cherry-pick merged to main (EnsureMetadata fix, verifyRigIdentity, improved error messages). The gt repair command was omitted — consider making it a thin wrapper around gt doctor --fix with targeted checks instead of a separate implementation.

PR #2982, opened 2026-03-18
feat: periodic checkpoint watchdog for polecat uncommitted work (gas-7d4)

quad341:
## Summary

- Adds checkpoint squash and watchdog for polecat sessions to periodically save uncommitted work
- New `gt checkpoint-watchdog` command with platform-specific signal handling
- Integrates with polecat session manager and `gt done` flow

**Issue**: gas-7d4
**Polecat**: quartz

---

### DIFF
--- a/internal/checkpoint/squash.go
+++ b/internal/checkpoint/squash.go
@@ -99,7 +99,9 @@ func listCommitMessages(workDir, baseRef string) ([]string, error) {
 		return nil, err
 	}
 
-	cmd := exec.Command("git", "log", "--format=%B---COMMIT-SEP---", mergeBase+"..HEAD")
+	// Use null-byte separator (%x00) instead of a text marker to avoid
+	// collision with commit messages that might contain the separator string.
+	cmd := exec.Command("git", "log", "--format=%B%x00", mergeBase+"..HEAD")
 	cmd.Dir = workDir
 	out, err := cmd.Output()
 	if err != nil {
@@ -107,7 +109,7 @@ func listCommitMessages(workDir, baseRef string) ([]string, error) {
 	}
 
 	var messages []string
-	for _, msg := range strings.Split(string(out), "---COMMIT-SEP---") {
+	for _, msg := range strings.Split(string(out), "\x00") {
 		msg = strings.TrimSpace(msg)
 		if msg != "" {
 			messages = append(messages, msg)


DreadPirateRobertz:
**Review: Valuable feature, a few things to verify**

The checkpoint watchdog solves a real problem: polecats losing work between sessions. The implementation is well-structured with platform-specific files and proper setsid-based daemonization.

**Items to verify:**

1. **`SquashWIPCommits` squashes ALL commits when rebasing to merge-base.** The operation:
   - Soft-reset to merge-base (all commits gone, changes staged)
   - Create one combined commit
   
   This means `feat: add auth` + `WIP checkpoint` + `fix: typo` becomes a single commit. Non-WIP commit subjects are preserved in the body, but the commit structure is flattened. Is this the intended behavior when WIP commits are mixed with real commits? Some users might want the real commits preserved separately. The PR description says "squash", so this may be intentional — just confirm it's the expected semantics.

2. **`---COMMIT-SEP---` separator** in `listCommitMessages`: This string could appear in a commit message body (`%B` format includes full body). Rare but possible. `git log --format=%x00` (null-byte separator) would be more robust.

3. **PID file at `/tmp/gt-checkpoint-<session>.pid`**: On shared systems, any user could create a file with this name to prevent the watchdog from starting or kill the wrong process. Consider `os.TempDir()` or a user-specific temp dir. Minor but worth noting.

4. **`DeactivateCheckpointWatchdog` called before `KillSessionWithProcesses`**: This ordering is correct (kill watchdog first, then the session). Confirmed in the diff tail. Good.

5. **Windows no-op stub**: Appropriate — checkpoint watchdog requires fork/setsid which isn't available on Windows. The stub allows the code to compile cross-platform.

**`checkpoint/squash_test.go`**: Tests cover the basic cases (no WIP, mixed WIP, all WIP). The test uses `setupGitRepo` which I assume initializes a git repo with an initial commit on `main`. Confirm the base ref (`"main"`) in tests matches what `SquashWIPCommits` would use in production (the actual branch the polecat branched from).

**Overall**: LGTM pending the squash semantics confirmation. The setsid-based daemonization and PID-file cleanup are solid.

steveyegge:
Closing: good idea, wrong architecture. This should be a daemon dog patrol (like compactor_dog, doctor_dog, etc.) rather than a standalone watchdog process with its own CLI command and session monitoring. The daemon already has the lifecycle management, interval scheduling, and rig-awareness needed. We'll implement this as a checkpoint_dog patrol. Thanks for the concept work — the squash-on-done integration is a nice touch we'll preserve.

PR #2986, opened 2026-03-18
feat: unique polecat names across rigs via global namepool (gas-21k)

quad341:
## Summary

- Ensures polecat names are unique across all rigs, not just within a single rig
- Adds global namepool with cross-rig deduplication
- Updates rig manager to check global pool during polecat creation

**Issue**: gas-21k
**Polecat**: jasper

---

### DIFF
--- a/internal/cmd/rig.go
+++ b/internal/cmd/rig.go
@@ -636,6 +636,9 @@ func runRigAdd(cmd *cobra.Command, args []string) error {
 		}
 	}
 
+	// Auto-assign a namepool theme that doesn't collide with other rigs (gas-21k).
+	autoAssignNamepoolTheme(townRoot, name, mgr)
+
 	// Sync hooks for the new rig's targets
 	if err := syncRigHooks(townRoot, name); err != nil {
 		fmt.Fprintf(os.Stderr, "Warning: failed to sync hooks for new rig: %v\n", err)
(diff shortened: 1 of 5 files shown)


DreadPirateRobertz:
**Review: Correct for new rigs — note existing rigs aren't retroactively fixed**

`ThemeForRigAvoiding` correctly picks an unused theme at `rig add`/`rig adopt` time, and the test coverage handles the key cases (avoids used themes, falls back to hash when all taken, empty used list).

**One item to note:**

The deduplication is prospective only — it applies to newly added rigs. Existing rigs that already share the same hash-based theme (because they were added before this feature) will continue to collide. Is there a migration plan, or is this intentional (don't retroactively rename names that polecats are currently using)?

If retroactive repair is desired, `gt doctor --fix` or `gt repair` could check for theme collisions and reassign. Not a blocker for this PR — the prospective fix is valuable on its own — but worth documenting the limitation.

**`autoAssignNamepoolTheme` integration**:
- Called in both `runRigAdd` and `runRigAdopt` — correct, covers both add paths
- Only sets theme if not already configured (`rigSettings.Namepool.Style != ""`) — correct, won't override intentional config
- Non-fatal warning on failure — appropriate

**`UsedNamepoolThemes(polecat.ThemeForRig)` call**: Passes `ThemeForRig` as a function to get the default theme for rigs without explicit config. Verified that this returns the deterministic hash-based theme, which is what we want for the "already used" set.

**`.beads/.gitignore` conflict note**: Applies here as in the rest of the batch.

LGTM.

PR #2987, opened 2026-03-18
fix(doltserver): fix gt dolt start on Linux with slow storage or no lsof (#2968)

### LINKED ISSUE #2968: gt dolt start fails on Linux when dolt takes >500ms to start (PID file wrongly removed + lsof missing)

DreadPirateRobertz:
## Summary

Fixes two bugs in `gt dolt start` that caused it to fail on Linux systems with slow storage (Kubernetes CSI block storage) or without `lsof` installed.

### Bug 1 (critical): startup loop removes valid PID file too early

The retry loop called `IsRunning()` to check liveness, but `IsRunning` has a side effect: when the process is alive but not yet listening on the port, it removes the PID file as "stale" and returns `false`. The loop then exited immediately on the first retry (500ms) — well before dolt was ready (~1.5s on slow storage).

**Fix**: replace `IsRunning()` in the retry loop with a direct `cmd.Process.Signal(0)` check. This tests liveness without PID file side effects.

### Bug 2: `findDoltServerOnPort` requires `lsof`

`findDoltServerOnPort` only called `lsof`, which is not installed on many Linux systems (containers, minimal pods, Kubernetes). When `lsof` was unavailable, it silently returned 0.

**Fix**: try `ss` (iproute2, available on most modern Linux) first on Linux, fall back to `lsof`, then `/proc/net/tcp` as a zero-dependency last resort.

## Test plan

- [x] `go test ./internal/doltserver/...` — all existing tests pass
- [x] New tests: `TestFindPortViaProcNet_NotFound`, `TestFindPortWithSS_NotFound`

Closes #2968

### DIFF
--- a/internal/doltserver/doltserver.go
+++ b/internal/doltserver/doltserver.go
@@ -665,30 +665,136 @@ func CheckPortConflict(townRoot string) (int, string) {
 }
 
 // findDoltServerOnPort finds a process listening on the given port.
-// Returns the PID or 0 if not found. Uses lsof to identify the listener PID.
-// Does not verify process identity via ps string matching (ZFC fix: gt-utuk).
+// Returns the PID or 0 if not found.
+// Uses ss (Linux), lsof (macOS/Linux), or /proc/net/tcp (Linux fallback)
+// to identify the listener PID. (gh-2968: lsof not universally available)
 func findDoltServerOnPort(port int) int {
-	// Use lsof to find the LISTENING process on port (not clients connected to it).
+	// Try ss first on Linux (iproute2, available on most modern Linux).
+	if runtime.GOOS == "linux" {
+		if pid := findPortWithSS(port); pid > 0 {
+			return pid
+		}
+	}
+
+	// Try lsof (available on macOS, and many Linux systems).
 	// Without -sTCP:LISTEN, lsof returns client PIDs (e.g., gt daemon) first,
 	// which aren't dolt processes — causing false negatives.
 	cmd := exec.Command("lsof", "-i", fmt.Sprintf(":%d", port), "-sTCP:LISTEN", "-t")
 	output, err := cmd.Output()
+	if err == nil {
+		lines := strings.Split(strings.TrimSpace(string(output)), "\n")
+		if len(lines) > 0 && lines[0] != "" {
+			if pid, err := strconv.Atoi(lines[0]); err == nil {
+				return pid
+			}
+		}
+	}
+
+	// Last resort on Linux: parse /proc/net/tcp (no external tools required).
+	if runtime.GOOS == "linux" {
+		if pid := findPortViaProcNet(port); pid > 0 {
+			return pid
+		}
+	}
+
+	return 0
+}
+
+// findPortWithSS finds the PID listening on the given port using ss(8).
+// ss is part of iproute2 and available on most modern Linux systems.
+// Returns 0 if ss is not available or the port is not in use.
+func findPortWithSS(port int) int {
+	cmd := exec.Command("ss", "-tlnpH", fmt.Sprintf("sport = :%d", port))
+	output, err := cmd.Output()
+	if err != nil {
+		return 0
+	}
+	// ss output with -p includes: users:(("dolt",pid=12345,fd=10))
+	// Extract pid=<N> from the output.
+	for _, line := range strings.Split(string(output), "\n") {
+		if idx := strings.Index(line, "pid="); idx >= 0 {
+			rest := line[idx+4:]
+			end := strings.IndexAny(rest, ",)")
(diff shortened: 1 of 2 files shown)


DreadPirateRobertz:
Closing in favor of PR #2970 by @baumgold (the issue reporter) which has the same core fixes with a cleaner implementation. 

My PR also added:
- `/proc/net/tcp` as a zero-dependency Linux fallback in `findDoltServerOnPort`  
- Unit tests for the new helper functions

Happy to add those as follow-up commits to #2970 if wanted.

PR #2992, opened 2026-03-18
feat: add 'gt rig repair' to fix partial beads initialization (#2922)

### LINKED ISSUE #2922: Beads initialization is fragile: partial init leaves rigs broken with no clean recovery
## Summary

When `gt rig add` can't fully initialize beads (e.g. Dolt server is down), the rig is left in a partially initialized state that's confusing to diagnose and easy to make worse. The current initialization flow has several compounding issues that make this harder than it needs to be.

## The partial initialization problem

`gt rig add` writes the route to `routes.jsonl` and prefix to `rigs.json` early in the process. Beads initialization happens later via `InitBeads()`, which shells out to `bd init`. If Dolt is down at that point:

- The route exists (prefix `tr-` for a rig called `test-rig`)
- `config.yaml` gets written (fallback)
- **`metadata.json` is never created** — only `bd init` writes this, and it failed
- The database doesn't exist on the server

The rig now exists in Gas Town's registry but can't do anything with beads. `gt sling`, `gt mail`, witness patrols — anything touching beads on this rig will fail. The error messages don't point you toward the root cause (missing init), and there's no obvious command to complete the initialization.

## bd init is easy to misuse

If someone (or something) tries to fix the broken rig by running `bd init` from the rig directory:

- **Without `--prefix`**: bd auto-generates a prefix from the directory name, creating a database with a different name than what `routes.jsonl` expects. This is silent — nothing warns you that a route already exists with a different prefix.
- **With the wrong `--prefix`**: creates a database that doesn't match the route, same problem.
- **On a rig that's already initialized**: bd does correctly refuse (guard exists). But if init partially failed (no `metadata.json`), bd thinks it's a fresh init and will happily create a conflicting setup.

There's no way for `bd init` to know what Gas Town intended, because the intended prefix lives in `routes.jsonl` which bd doesn't read.

## InitBeads works against bd init

Even when `bd init` succeeds, `InitBeads()` immediately cleans up after it:

1. `bd init --prefix xyz` creates a database named `beads_xyz` on the server
2. Gas Town uses `xyz` as the database name (via `InitRig` / `EnsureMetadata`)
3. `InitBeads` deletes the `beads_xyz` orphan (line 965, comment: *"Drop the orphaned beads\_\<prefix\> database created by bd init"*)
4. `InitBeads` overwrites `metadata.json` to point at the correct database
5. `InitBeads` re-runs `bd config set issue_prefix` because `bd init --prefix` didn't persist it

This suggests `bd init` and Gas Town have different ideas about database naming, and Gas Town is working around it.

## Recovery is non-obvious

`gt doctor --fix` can detect and repair missing beads initialization. It looks up the correct prefix from `rigs.json` and runs `bd init --prefix <correct> --server`. But:

- You have to know to run `gt doctor --fix` (the error messages from failed `gt sling` / `gt mail` don't suggest this)
- It only runs when explicitly invoked — there's no automatic recovery when Dolt comes back up
- If someone already ran `bd init` incorrectly before `gt doctor --fix`, the rig has a `.beads/` with wrong metadata and the doctor check may think it's already initialized

## What would be better

1. **Partial init should be safe**: If Dolt is down, `gt rig add` should still write `metadata.json` and `config.yaml` with the correct values. These are just config files — they don't need a running database. The database itself can be created later when Dolt is available.

2. **Clear error messages**: When `gt sling` or `gt mail` hits a rig with no database, the error should say something like *"Beads database not initialized for rig X. Run `gt doctor --fix --rig X` to complete setup."*

3. **Automatic recovery**: When the daemon detects Dolt is back up, it could check for rigs with `metadata.json` but no database and create them automatically.

4. **bd init should check for existing routes**: If `bd init` is run in a directory that already has a route in `routes.jsonl`, it should either use that prefix or warn that a route already exists. Running `bd init` without `--prefix` in a Gas Town rig should not silently create a mismatched database.

## Real-world impact

We had several rigs added while Dolt was temporarily down. All ended up with mismatched database names — `routes.jsonl` said one thing, `metadata.json` said another. Witnesses, polecats, and mail all failed silently. Diagnosis required manually comparing `routes.jsonl`, `metadata.json`, and the Dolt server's database list across each rig. Fix required deleting `.beads/` and re-running `bd init --prefix <correct>` for each one.

DreadPirateRobertz:
## Summary

When \`gt rig add\` completes but Dolt goes down mid-initialization, the rig is left in a broken state: \`EnsureMetadata\` and \`bd config set issue_prefix\` are non-fatal, so the rig appears to succeed but has \`metadata.json\` pointing at \`beads_<prefix>\` instead of \`<rigName>\`, and \`issue_prefix\` is unset. All subsequent \`gt sling\`, \`gt mail\`, and witness patrols on that rig fail.

Until now the only recovery was \`gt doctor --fix --rig <name>\`, which re-runs \`bd init\` but skips the \`EnsureMetadata\` + \`issue_prefix\` repair steps — the exact steps that were broken.

New command: **\`gt rig repair <name>\`**

1. Ensures the Dolt database exists (`InitRig`, idempotent)
2. Runs `bd init` if `.beads/metadata.json` is missing entirely
3. Fixes `metadata.json` to use the correct `dolt_database` and server mode (`EnsureMetadata` — the step that was non-fatal and often left broken)
4. Re-sets `issue_prefix` config on the server-side database

All steps are idempotent — safe to run on a healthy rig.

## Test plan
- [x] Unit tests for all error paths: rig not found, invalid config, missing prefix, Dolt not running
- [x] `go build ./internal/cmd/` and `go vet ./internal/cmd/` clean
- [x] `gt rig repair --help` shows correct usage
- [x] Full repair flow requires Dolt (integration test candidate for follow-up)

Fixes #2922

### DIFF
--- a/internal/cmd/rig.go
+++ b/internal/cmd/rig.go
@@ -339,6 +339,25 @@ var (
 	promptYesNoUnsafeProceed = promptYesNo
 )
 
+var rigRepairCmd = &cobra.Command{
+	Use:   "repair <rig>",
+	Short: "Re-initialize beads for a rig with a partial or broken beads setup",
+	Long: `Re-run beads initialization for a rig that was left in a broken state.
+
+Common cause: 'gt rig add' succeeded but Dolt went down mid-init, leaving
+metadata.json pointing at the wrong database or missing issue_prefix config.
+
+repair:
+  1. Ensures the rig's Dolt database exists (idempotent)
+  2. Fixes metadata.json to use the correct database name and server mode
+  3. Re-runs 'bd init' if .beads/ is missing entirely
+  4. Re-sets issue_prefix config on the server-side database
+
+Safe to run on a healthy rig — all steps are idempotent.`,
+	Args: cobra.ExactArgs(1),
+	RunE: runRigRepair,
+}
+
 func init() {
 	rootCmd.AddCommand(rigCmd)
 	rigCmd.AddCommand(rigAddCmd)
@@ -346,6 +365,7 @@ func init() {
 	rigCmd.AddCommand(rigListCmd)
 	rigCmd.AddCommand(rigRebootCmd)
 	rigCmd.AddCommand(rigRemoveCmd)
+	rigCmd.AddCommand(rigRepairCmd)
 	rigCmd.AddCommand(rigResetCmd)
 	rigCmd.AddCommand(rigRestartCmd)
 	rigCmd.AddCommand(rigShutdownCmd)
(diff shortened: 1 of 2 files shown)


steveyegge:
Closing — the repair logic should live in `gt doctor --fix` rather than a new CLI entrypoint. The right fix is adding EnsureMetadata + issue_prefix steps to the existing doctor checks. More CLI surface area = more to maintain.

PR #2993, opened 2026-03-18
fix(config): custom agents inherit Session/Tmux from matched preset

DreadPirateRobertz:
## Summary

- `fillRuntimeDefaults` now auto-fills `Session` and `Tmux` from the matched preset when a custom agent's `Command` matches a known preset (e.g. `claude`), using the same nil-guard pattern already used for `Hooks`, `Env`, and `ACP`
- The pi-specific Tmux block runs first, so pi keeps its existing hardcoded values; general preset filling only applies when `Tmux` is still nil
- Fixes the root cause: `WaitForRuntimeReady` returns immediately when `rc.Tmux == nil`, so the startup beacon was sent before Claude finished loading — agent idled at welcome screen

## Root Cause

When `gt config agent set claude-opus 'claude --model claude-opus-4-6'` creates a custom agent, only `Command` and `Args` are stored. `fillRuntimeDefaults` filled `Hooks` from the claude preset but left `Session` and `Tmux` nil.

`tmux.WaitForRuntimeReady` has an early-return guard: `if rc == nil || rc.Tmux == nil { return nil }`. So for custom agents, readiness detection was skipped entirely — the startup beacon was injected before Claude displayed its prompt and got lost.

`handoff.go` also checks `runtimeConfig.Session != nil` before exporting `GT_SESSION_ID_ENV`, so session ID propagation was broken for custom agents.

## Test Plan

- [x] `TestFillRuntimeDefaults/custom_claude_agent_inherits_Session_and_Tmux_from_preset` — new test verifying Session.SessionIDEnv and Tmux.ReadyPromptPrefix are filled
- [x] `TestFillRuntimeDefaults/explicit_Session_config_is_not_overridden_by_preset` — user-set Session is preserved
- [x] `TestFillRuntimeDefaults/explicit_Tmux_config_is_not_overridden_by_preset` — user-set Tmux is preserved
- [x] Updated `TestFillRuntimeDefaults/nil_nested_structs_are_auto-filled_from_preset_for_known_agents`
- [x] All config tests pass (`go test ./internal/config/... -count=1`)
- [x] All internal tests pass (pre-existing failures in `acp` and `cmd` packages unrelated)

Fixes gt-7z76.

### DIFF
--- a/internal/config/loader.go
+++ b/internal/config/loader.go
@@ -1852,6 +1852,31 @@ func fillRuntimeDefaults(rc *RuntimeConfig) *RuntimeConfig {
 		}
 	}
 
+	// Auto-fill Session defaults from preset when not explicitly set.
+	// Custom agents (e.g., "claude-opus" with Command:"claude") inherit
+	// SessionIDEnv/ConfigDirEnv from the matched preset, enabling session
+	// resume and GT_SESSION_ID_ENV propagation in handoffs.
+	if result.Session == nil && preset != nil && (preset.SessionIDEnv != "" || preset.ConfigDirEnv != "") {
+		result.Session = &RuntimeSessionConfig{
+			SessionIDEnv: preset.SessionIDEnv,
+			ConfigDirEnv: preset.ConfigDirEnv,
+		}
+	}
+
+	// Auto-fill Tmux defaults from preset for process detection and readiness.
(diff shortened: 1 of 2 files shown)


steveyegge:
Fix-merged to main in 05ca00da. Thanks for the contribution — the Session/Tmux preset inheritance fix is solid.

PR #2999, opened 2026-03-18
Fix vitals zombie detection flagging sibling Gas Town instances

### LINKED ISSUE #2998: gt vitals zombie detection flags Dolt from other Gas Town workspaces
## Bug

`findVitalsZombies()` in `internal/cmd/vitals.go` lists ALL Dolt listeners on non-production ports as "test zombies" without checking workspace ownership. It simply finds all Dolt processes via `FindAllDoltListeners()` and labels anything not on the current workspace's production port as a zombie.

## Impact

When multiple Gas Town instances run on the same machine (e.g., different workspaces for different projects), the deacon acts on the vitals zombie report and kills legitimate Dolt servers belonging to sibling Gas Town instances. This causes data corruption risks and disrupts work in other workspaces.

## Root Cause

`findVitalsZombies()` has no workspace ownership check — it assumes any Dolt process on a non-production port is a leftover test server.

Meanwhile, `KillImposters()` in `internal/doltserver/doltserver.go` already correctly checks data-dir ownership by inspecting the process's `--data-dir` flag and comparing it against the expected workspace data directory. The vitals code should use the same approach.

## Fix

Check workspace ownership before labeling a Dolt process as a zombie:
1. Read the process's `--data-dir` flag via `GetDoltDataDirFromProcess()`
2. If the data-dir is a `.dolt-data` directory under a valid Gas Town workspace (checked via `workspace.IsWorkspace()`), label it as "foreign workspace" instead of "test zombie"
3. Only processes that don't belong to any valid workspace are true zombies

Cdfghglz:
## Summary

- `findVitalsZombies()` now checks workspace ownership before labeling Dolt processes as zombies, using the same `--data-dir` inspection approach that `KillImposters()` already uses
- Processes belonging to sibling Gas Town workspaces are labeled "foreign workspace" (dimmed) instead of "test zombie" (warning), preventing the deacon from killing them
- Exports `GetDoltDataDirFromProcess()` so the vitals command can use it

## Test plan

- [ ] Run `gt vitals` on a machine with multiple Gas Town workspaces — sibling Dolt servers should show as "foreign workspace" not "test zombie"
- [ ] Run `gt vitals` on a machine with actual test zombie Dolt processes — they should still be flagged as "test zombie"
- [ ] Verify existing vitals tests pass (`go test ./internal/cmd/ -run Vitals`)

Fixes #2998

### DIFF
--- a/internal/cmd/vitals.go
+++ b/internal/cmd/vitals.go
@@ -58,21 +58,51 @@ func printVitalsDoltServers(townRoot string) {
 	}
 
 	// Zombie dolt processes (test servers not cleaned up)
-	for _, z := range findVitalsZombies(config.Port) {
-		fmt.Printf("  %s :%s test zombie PID %s\n", style.Warning.Render("○"), z.port, z.pid)
+	for _, z := range findVitalsZombies(townRoot, config.Port) {
+		if z.foreign {
+			fmt.Printf("  %s :%s foreign workspace PID %s\n", style.Dim.Render("○"), z.port, z.pid)
+		} else {
+			fmt.Printf("  %s :%s test zombie PID %s\n", style.Warning.Render("○"), z.port, z.pid)
+		}
 	}
 }
 
-type vitalsZombie struct{ pid, port string }
+type vitalsZombie struct {
+	pid, port string
+	foreign   bool // true if process belongs to another Gas Town workspace
+}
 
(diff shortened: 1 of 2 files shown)


steveyegge:
Fix-merged to main (16cc705e). Just needed the function rename resolved — the rest applied cleanly. Good catch on the multi-workspace zombie detection. Thank you!

PR #3009, opened 2026-03-18
feat: add gt followup to prevent deferred work from stalling

Rome-1:
## Summary

- Adds `gt followup` command—persistent reminders that survive session restarts, stored under `.runtime/followups/`
- Adds a guard in `gt done --status DEFERRED` that warns agents exiting without a follow-up set
- Updates polecat and crew role templates with explicit "Waiting for Feedback?" instructions
- Adds `FollowupThresholds` config (default delay: 30m, configurable via `operational.followup.default_delay`)

## Problem

Crew members say "I'll wait for their feedback and iterate" then end the session. Nothing re-engages them, so deferred work stalls until a human manually prompts them again.

## Solution

Three-layer fix:
1. **`gt followup` command** — `gt followup "topic" --in 30m` creates a persistent JSON record. `gt followup list` and `gt followup resolve <id>` for lifecycle management.
2. **`gt done` guard** — When a polecat defers without any pending followups, a warning is printed with the exact command to schedule one.
3. **Template instructions** — Both polecat and crew templates now explicitly say "NEVER say 'I'll wait for feedback' without scheduling a followup first."

Deacon patrol integration (scanning overdue followups and re-engaging agents) is designed for but not included in this PR—follow-up work.

## Design note: mail reliability

Followup storage deliberately uses plain JSON files in `.runtime/` rather than beads/mail. This avoids known Dolt contention issues that affect polecat mail under load (async notification fire-and-forget, ack failures swallowed silently, send timeouts under lock contention). The future deacon patrol integration will need to re-engage agents via nudges or mail and should account for these reliability gaps—but the storage layer itself is not affected.

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./internal/followup/` — 10 tests covering create, list, resolve, overdue detection, path traversal rejection, empty state
- [x] `go test ./internal/config/` — existing config tests still pass
- [ ] Manual: run `gt followup "test" --in 5m` in a polecat session, verify file appears in `.runtime/followups/`
- [ ] Manual: run `gt done --status DEFERRED` without a followup set, verify warning appears

## CI note

Lint and integration test failures are pre-existing on main (unparam in `show.go`/`compactor_dog.go`, scheduler epic tests). Not introduced by this PR.


Rome-1:
## Alternative: Make `--followup-in` mandatory on DEFERRED (Option A)

After thinking more about this, I think the current approach has a fundamental weakness: **it's opt-in for agents who are already failing to opt-in to follow-up behavior.** An agent that forgets to follow up is the same agent that needs to remember to run `gt followup`. The soft warning in `gt done` helps but doesn't enforce.

### Proposed change

Make `gt done --status DEFERRED` **require** either `--followup-in <duration>` or `--no-followup`:

```bash
gt done --status DEFERRED --followup-in 30m    # creates followup inline, exits
gt done --status DEFERRED --no-followup        # explicit opt-out ("I know what I'm doing")
gt done --status DEFERRED                      # ERROR: must specify one of the above
```

### What stays the same

- `internal/followup` package (storage, list, resolve, deacon patrol API)—unchanged
- `gt followup list` / `gt followup resolve`—still useful for inspection
- Template instructions, config thresholds—unchanged
- All tests—unchanged

### What changes

- `done.go`: Replace the soft `warnNoFollowupOnDefer()` warning with a hard error. Add `--followup-in` and `--no-followup` flags. When `--followup-in` is provided, `done.go` calls `followup.Create()` directly before proceeding.
- `gt followup "<topic>" --in 30m` (standalone create) can stay for cases where an agent wants to schedule a followup without exiting, but it's no longer the primary path.

### Why this is better

| | Current (soft warning) | Option A (hard gate) |
|---|---|---|
| Agent forgets | Warning printed, work stalls anyway | Command fails, agent must decide |
| New concepts | Must learn `gt followup` command | Just new flags on existing `gt done` |
| Explicit opt-out | Not required | `--no-followup` forces conscious choice |
| Enforcement point | Before `gt done` (separate step) | Inside `gt done` (atomic) |

The core insight: **enforcement at the exact moment of failure beats a tool the agent has to remember to use.** The `--no-followup` escape hatch prevents false blocks while making "I chose not to follow up" a deliberate act rather than a default.

### Scope

This would be a small follow-up commit on this PR—~20 lines changed in `done.go`, everything else stays.

Rome-1:
### Addendum to Option A: mail reliability consideration

Investigation into polecat mail reliability found several systemic issues worth noting for follow-up work:

- **Async notification fire-and-forget**: `notifyRecipient()` spawns a goroutine—if the CLI exits first, the nudge never fires. Errors go to stderr only, caller never knows.
- **Dolt lock contention**: Every mail send is a Dolt write. Under concurrent polecat load, sends compete for locks. Timeout was bumped 30s→60s as a band-aid; messages can still time out with undefined persistence state.
- **Ack failures silently swallowed**: `AcknowledgeDeliveries()` errors are logged to stderr and ignored. Receiver thinks ack succeeded.

**Impact on this PR**: None directly. Followup storage uses plain JSON in `.runtime/`, bypassing Dolt entirely. That's a deliberate choice.

**Impact on future deacon patrol integration**: The patrol will need to re-engage agents via nudges or mail when followups go overdue. Both paths inherit these reliability issues. The patrol should prefer nudges (filesystem-based, no Dolt) over mail where possible, and handle delivery failures gracefully rather than assuming fire-and-forget is sufficient.

This doesn't change the Option A proposal—enforcement at the \`gt done\` gate is still the right move regardless of how the deacon eventually re-engages agents.

steveyegge:
Thanks for identifying the problem — deferred work stalling is real and frustrating. However, we don't think a dedicated followup system belongs in gt core.

The concern is that this adds a new storage layer (`.runtime/followups/`), a new lifecycle concept (followup records), and template changes to shared infrastructure — all for something that should be solvable with existing primitives:

- **Beads** already track deferred work (`--status deferred`) and are queryable
- **Formulas** can encode "check back on deferred work" as a patrol step
- **Plugins** can run periodic scans for stale deferred beads and nudge the responsible agent
- **The deacon** already patrols for operational health — a "stale deferred bead" check fits naturally in a deacon plugin

A plugin like `stale-deferred-check` that runs `bd list --status=deferred` and nudges agents with overdue work would achieve the same goal without adding a parallel storage system or new core commands.

We appreciate the thorough design and test coverage — hope you'll consider the plugin approach.

PR #3020, opened 2026-03-18
fix(doctor): detect and repair relocated worktree gitdir paths

bobvan:
## Summary
- After rsync/move between machines, worktree `.git` files contain absolute paths to the old location, breaking all git operations
- The `worktree-gitdir-valid` doctor check now infers the correct `.repo.git` path from the current town root instead of trusting the stale path
- Scans `deacon/dogs/` for cross-rig worktrees (previously missed entirely)
- Falls back to manual worktree registration when `git worktree add` fails on existing directories with content

## Test plan
- [x] All 12 unit tests pass (8 existing + 4 new: relocated worktree, deacon dogs, multiple dogs, no deacon)
- [x] Tested on live system: 13 broken worktrees detected and repaired after rsync from macOS to Linux
- [x] `gt doctor` shows clean after fix
- [x] `git status` / `git log` work in repaired worktrees

### DIFF
--- a/internal/doctor/worktree_gitdir_check.go
+++ b/internal/doctor/worktree_gitdir_check.go
@@ -12,19 +12,26 @@ import (
 // gitdir paths. When a worktree's .git file contains "gitdir: /path/to/.repo.git/worktrees/X"
 // but the referenced path doesn't exist, all git operations in that worktree fail.
 //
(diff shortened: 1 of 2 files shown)


steveyegge:

PR #3036, opened 2026-03-19
fix: auto-close stale wisps from escalations and plugin runs

EchBTI:
## Summary

- **Escalation close** (`gt escalate close`) now closes open mail wisps tagged with `escalation:<id>`. Extracted into testable `closeEscalationMailWisps()` helper with `escalationWispCloser` interface.
- **Plugin run recording** (`RecordRun`) now auto-closes plugin-run beads immediately after creation. `GetLastRun`/`GetRunsSince` already use `--all` so queries are unaffected.
- These two sources accounted for ~80% of wisp accumulation that triggered recurring reaper alerts (threshold 500).

## Test plan

- [x] `TestCloseEscalationMailWisps`: 4 cases — no wisps, multiple wisps, list error, close error
- [x] `TestRecordRunBuildsCorrectLabels`: verifies label assembly for plugin runs
- [x] `TestRecordRunLabelsNoRig`: verifies rig label omitted when empty
- [x] `TestQueryRunsIncludesClosedBeads`: documents `--all` contract
- [x] Full test suite passes (`go test ./...`) — only pre-existing config/Dolt port test failures
- [x] Arch review: approved at aabdd99a

Fixes: hq-im27o

### DIFF
--- a/internal/cmd/escalate_impl.go
+++ b/internal/cmd/escalate_impl.go
@@ -330,16 +330,8 @@ func runEscalateClose(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("closing escalation: %w", err)
 	}
 
-	// Close any mail beads tagged with this escalation to prevent wisp accumulation.
-	// When an escalation is sent, mail beads are annotated with "escalation:<id>".
-	// These mail beads stay open even after the escalation itself is closed.
-	mailBeads, _ := bd.List(beads.ListOptions{
-		Label:  fmt.Sprintf("escalation:%s", escalationID),
-		Status: "open",
-	})
-	for _, mb := range mailBeads {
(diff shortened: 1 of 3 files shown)


steveyegge:
Thanks for tracking down the wisp accumulation sources. However, closing wisps at creation time across 15 files is treating the symptom rather than the cause. Wisp lifecycle is the reaper's job — if escalation and plugin-run wisps are accumulating past the threshold, the right fix is tuning the reaper's TTL or adding a dog patrol step that targets these wisp types, not scattering close-on-create logic throughout the codebase. A compactor-dog or reaper-dog plugin that runs `bd list --wisp-type escalation,plugin-run --status open` and closes stale ones would handle this cleanly from one place.

EchBTI:
Ah okay, makes sense. Cheers

________________________________

PR #3038, opened 2026-03-19
fix: session-hygiene plugin should skip non-GT sessions

Rome-1:
## Summary
- Session-hygiene plugin was killing user-created tmux sessions (e.g., `scratch`, `debug`, `notes`) every 30 minutes because their names don't match any known rig prefix
- GT sessions always contain a dash (`hq-mayor`, `gt-furiosa`, `cr-crew-charlie`); user sessions typically don't
- Skip sessions without a dash before prefix extraction — zero config, correct by convention

## Test plan
- [ ] Create a dashless tmux session (`tmux new -s scratch`) and verify it survives a `./run.sh` cycle
- [ ] Verify GT sessions with invalid prefixes are still killed (e.g., `fake-session`)
- [ ] Run `./run.sh --dry-run` and confirm dashless sessions don't appear in output

### DIFF
--- a/plugins/session-hygiene/run.sh
+++ b/plugins/session-hygiene/run.sh
@@ -84,6 +84,12 @@ ZOMBIES=()
 while IFS= read -r SESSION; do
   [ -z "$SESSION" ] && continue
 
+  # Sessions without a dash are user-created, not GT-managed — skip them.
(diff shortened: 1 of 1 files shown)


steveyegge:
The session-hygiene plugin was already removed from main (a26928d1) because it kept killing crew sessions. This PR fixes a plugin that no longer exists. Thanks though — the dash-check heuristic was a good idea, it just came after the plugin was already pulled.

PR #3052, opened 2026-03-19
fix(dolt): allow hyphens in validDBName regex (GAS-zb8)

daronstinnett:
## Summary
- Allow hyphens in `ValidateDBName` regex so `gt-dolt` and other hyphenated database names are accepted
- Add dirty-check before `DOLT_COMMIT` and persist `GT_DOLT_LOGLEVEL`

## Changes
- `internal/reaper/reaper.go`: `^[a-zA-Z0-9_]+$` → `^[a-zA-Z0-9_-]+$`
- `internal/daemon/jsonl_git_backup.go`: same regex fix
- Dirty-check guard before committing to avoid empty-commit errors

## Why
`gt-dolt` was being rejected by the validator, making the gt-dolt database inaccessible to the reaper and backup daemon.

Closes GAS-zb8

### DIFF
--- a/internal/daemon/jsonl_git_backup.go
+++ b/internal/daemon/jsonl_git_backup.go
@@ -42,8 +42,8 @@ var testPollutionPatterns = []*regexp.Regexp{
 	regexp.MustCompile(`-wisp-`),                                       // id: wisp-pattern IDs leaked into issues table
 }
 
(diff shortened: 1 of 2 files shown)


steveyegge:
Duplicate of #3053.

PR #3057, opened 2026-03-19
Two Dolt stability fixes: startup timeout scaling + rig-config-sync DB naming

### LINKED ISSUE #3056: gt dolt start: fixed 5s startup timeout ignores database count
## Problem

`gt dolt start` reports *"Dolt server process started (PID …) but not accepting connections after 5s"* and exits with error — even though the Dolt process is healthy and eventually comes up successfully ~10–15 seconds later.

### Root cause

The health-check loop in `internal/doltserver/doltserver.go` is hardcoded to **10 attempts × 500 ms = 5 s**:

```go
    ...
}
return fmt.Errorf("... not accepting connections after 5s ...")
```

Each database Dolt loads adds ~1 s of startup overhead (LevelDB compaction, stats loading, catalog visibility probing). A workspace with 8 rig databases therefore needs ~8–12 s to start, well beyond the 5 s limit.

### Symptoms
- `gt dolt start` exits 1 with the 5 s error message
- Dolt PID is still alive and eventually comes up
- Agents and `gt doctor` report Dolt as down, creating cascading failures
- The workaround is `sleep 15 && gt dolt status` after the failed `gt dolt start`

## Fix

Scale the timeout by database count: **5 seconds × number of databases** (minimum 5 s for 0 or 1 database):

```go
dbCount := len(databases)
if dbCount < 1 {
    dbCount = 1
}
maxAttempts := dbCount * 10 // 10 × 500ms = 5s per database
```

This gives a workspace with 8 databases up to 40 s to start, which comfortably covers observed startup times, while keeping the default at 5 s for small workspaces.

The error message is also updated to show the actual timeout used and the database count.

## PR

https://github.com/baumgold/gastown/tree/fix/dolt-startup-timeout-per-db

### LINKED ISSUE #3058: gt doctor --fix rig-config-sync renames Dolt DB to prefix instead of rig name, breaking beads
## Problem

`gt doctor --fix` destroys beads access whenever a rig's beads prefix differs from its directory name (e.g. prefix `ts` in rig directory `trading_scripts`).

### Root cause

`internal/doctor/rig_config_sync_check.go` line 172:

```go
// Database name should match the prefix
expectedDBName := configPrefix  // BUG
```

The check uses the beads **prefix** (`ts`) as the expected Dolt database name. But the authoritative convention — established by `doltserver.EnsureMetadata` — is that the database name must equal the **rig directory name** (`trading_scripts`):

```go
// From doltserver.go EnsureMetadata:
// Default: rigName (correct when db-name == rig-dir-name)
effectiveDB := rigName
```

### Failure loop

1. `gt doctor` detects: `metadata has 'trading_scripts', should be 'ts' (prefix)` — incorrect
2. `--fix` renames `.dolt-data/trading_scripts` → `.dolt-data/ts` and writes `ts` to `metadata.json`
3. `doltserver.EnsureMetadata` later reverts `metadata.json` to `trading_scripts`
4. `bd` commands fail: `database 'trading_scripts' not found on server` (only `ts` exists on disk)
5. Loop repeats on next `gt doctor --fix`

### Impact

Every `gt doctor --fix` run silently breaks beads for any rig whose prefix ≠ directory name, causing cascading failures (witnesses can't patrol, refineries can't process queue, polecats can't write work tracking).

## Fix

Change line 172 from `configPrefix` to `rigName`:

```go
// Database name should match the rig directory name, not the beads prefix.
expectedDBName := rigName
```

Also update the diagnostic message and comments to say "rig name" instead of "prefix".

## PR

https://github.com/baumgold/gastown/pull/3057 (same branch — second commit adds this fix)

baumgold:
Fixes #3056 — `gt dolt start` times out with fixed 5s regardless of database count
Fixes #3058 — `gt doctor --fix` rig-config-sync renames Dolt DB to prefix instead of rig name

---

## Fix 1: Scale Dolt startup timeout by database count (5s per DB)

**File:** `internal/doltserver/doltserver.go`

The health-check loop used a fixed 10 attempts × 500 ms = **5 s** timeout, insufficient for workspaces with many rig databases. Each database adds ~1 s of startup overhead.

```go
// Before

// After
maxAttempts := dbCount * 10 // 5s per database
for attempt := 0; attempt < maxAttempts; attempt++ {
```

`databases` is already computed earlier in the function for LOCK cleanup — no extra I/O.

| Databases | Before | After |
|-----------|--------|-------|
| 1         | 5 s    | 5 s   |
| 4         | 5 s    | 20 s  |
| 8         | 5 s    | 40 s  |

---

## Fix 2: rig-config-sync uses rig name (not prefix) as expected DB name

**File:** `internal/doctor/rig_config_sync_check.go`

The check computed the expected Dolt database name as the beads prefix, but `doltserver.EnsureMetadata` establishes that the database name must equal the **rig directory name**.

```go
// Before (wrong)
expectedDBName := configPrefix  // e.g. "ts" for trading_scripts rig

// After (correct)
expectedDBName := rigName  // e.g. "trading_scripts"
```

This caused a destructive loop: `--fix` renamed `.dolt-data/trading_scripts → ts`, then `EnsureMetadata` reverted `metadata.json` to `trading_scripts`, then `bd` failed because the directory was now named `ts` again.

---

## Testing

`go build ./...` passes. Both changes are targeted and minimal — no logic changes beyond the bug fixes.

### DIFF
--- a/internal/doctor/rig_config_sync_check.go
+++ b/internal/doctor/rig_config_sync_check.go
@@ -168,14 +168,15 @@ func (c *RigConfigSyncCheck) Run(ctx *CheckContext) *CheckResult {
 
 		// Check if Dolt database exists (only for server mode)
 		if metadata.DoltMode == "server" {
-			// Database name should match the prefix
-			expectedDBName := configPrefix
-			if expectedDBName == "" {
-				expectedDBName = metadata.DoltDatabase // fallback to what's in metadata
-			}
+			// Database name should match the rig directory name (rigName), not the beads
+			// prefix. This is the convention established by doltserver.EnsureMetadata:
+			// the Dolt database identifier is the rig's directory name so that rigs
+			// with short prefixes (e.g. "ts" for trading_scripts) don't collide and
+			// bd can always locate the right database without extra config.
+			expectedDBName := rigName
 
 			if expectedDBName != "" {
-				// Check if database name matches prefix
+				// Check if database name matches the rig directory name
 				if metadata.DoltDatabase != expectedDBName {
 					c.dbNameMismatches = append(c.dbNameMismatches, dbMismatch{
 						rigName:    rigName,
@@ -184,7 +185,7 @@ func (c *RigConfigSyncCheck) Run(ctx *CheckContext) *CheckResult {
 						expectedDB: expectedDBName,
 					})
 					details = append(details, fmt.Sprintf(
-						"Rig %s database name mismatch: metadata has '%s', should be '%s' (prefix)",
+						"Rig %s database name mismatch: metadata has '%s', should be '%s' (rig name)",
 						rigName, metadata.DoltDatabase, expectedDBName))
 				}
 
@@ -343,7 +344,7 @@ func (c *RigConfigSyncCheck) Fix(ctx *CheckContext) error {
 		}
 	}
 
-	// Fix database name mismatches - rename database to match prefix
+	// Fix database name mismatches - rename database to match rig directory name
 	renamedDBs := false
 	for _, mismatch := range c.dbNameMismatches {
 		rigPath := filepath.Join(ctx.TownRoot, mismatch.rigName)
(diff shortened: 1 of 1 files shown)


baumgold:
Closing in favour of two smaller PRs: #3059 (doltserver timeout scaling) and #3060 (rig-config-sync DB name fix).

PR #3019, opened 2026-03-18
fix: beads v0.61 upgrade + plugin GT_ROOT fixes + remove pane background

ktoulgaridis:
## Summary
- **beads v0.59 → v0.61**: Needed for `--no-history` flag that `gt sling` uses
- **convoy_manager.go**: Adapt to beads v0.61 API changes (`GetAllEventsSince` takes `time.Time`; `Event.ID` is `string`)
- **plugins/dolt-backup/run.sh**: Use `GT_ROOT` for paths; auto-discover databases
- **plugins/dolt-archive/run.sh**: Use `GT_ROOT` for paths
- **internal/tmux/tmux.go**: Remove `ApplyWindowStyle` — eliminates orange pane background tint

Cherry-picked from @dannomayernotabot's #3004, rebased onto current main.

## Test plan
- [x] `golangci-lint run` — 0 issues
- [x] `go build ./...` — clean
- [x] Integration tests pass (`TestSchedulerMultiRigEpicAutoResolve`, `TestSchedulerEpicDetection`, `TestSchedulerDirectEpicDispatch`) — #3004 failures were from missing #3007 fix, resolved by rebasing onto main

PR #3062, opened 2026-03-19
fix: upgrade beads dependency to v0.61.0, fix convoy event polling

macneale4:
## Summary

Beads v0.61.0 changed event IDs from sequential BIGINT to UUID CHAR(36) and GetAllEventsSince now takes time.Time instead of int64. Update the convoy manager's high-water mark from int64 to time.Time and use CreatedAt for cursor advancement instead of ID comparison.

## Related PR:
https://github.com/steveyegge/beads/pull/2691

## Changes
Upgrade all go dependencies, but we care about moving to beads 0.61.0

## Testing

- [ - ] Unit tests pass (`go test ./...`)
Tests don't pass. Fail the same way before and after this change.

- [ + ] Manual testing performed
Running GT no longer gives errors about comparing a UUID to an int.

## Checklist
- [ + ] Code follows project style
- [ + ] Documentation updated (if applicable)
- [ + ] No breaking changes (or documented in summary)


steveyegge:
Superseded — beads v0.61 upgrade was fix-merged from PR #3019. Convoy event polling fix is included.

PR #3067, opened 2026-03-20
fix(reaper): handle 'nothing to commit' in purge/auto-close

OS-joaocastilho:
## Summary

- Add `isNothingToCommit(err)` guard to `purgeClosedWisps`, `purgeOldMail`, and `AutoClose` DOLT_COMMIT error handlers
- Aligns with existing pattern in `Reap` (line 404) and `closePluginReceipts` (line 812)
- Eliminates recurring false `dolt_commit_failed` anomalies that trigger escalations from the Reaper Dog

## Context

When Dolt auto-commit processes DELETEs before the explicit `COMMIT` + `DOLT_COMMIT` sequence, the working set already matches HEAD, producing a benign "nothing to commit" error. The `Reap` function handled this correctly; the purge and auto-close paths did not.

This caused recurring MEDIUM/HIGH escalations from the Reaper Dog to overseer and mayor on every 30-minute cycle.

## Test plan

- [x] `go test ./internal/reaper/` passes
- [x] Manual: `gt reaper purge --db=hq --json` no longer reports false anomalies
- [ ] Integration: run full reaper cycle and verify no escalations


steveyegge:
Superseded by #3072 (coryshepherdson) which covers the same reaper nothing-to-commit fix with tests.

PR #3074, opened 2026-03-20
fix(doctor): extend hooks-sync to detect stale gemini settings

seanbearden:
## Summary

- Extends the `hooks-sync` doctor check to detect and auto-fix stale `.gemini/settings.json` files across all rigs
- Adds `Provider` field to `hooks.Target` to distinguish claude vs gemini targets
- `DiscoverTargets` now scans crew member, witness, refinery, and polecat work directories for `.gemini/settings.json`
- Gemini targets are compared against resolved templates (with `{{GT_BIN}}` substituted) rather than the Claude base+override merge system

## Context

Gemini crew sessions (e.g., laneassist/gisele) had stale hook files missing `GT_HOOK_SOURCE=compact` on PreCompress and other env var prefixes. Without these, `isCompactResume()` returns false, the full prime path runs, exceeds gemini's hook timeout, and the hook fails. The root cause was that `gt doctor hooks-sync` only checked Claude settings files — gemini settings were invisible to the check.

## Test plan

- [x] New test `TestDiscoverTargets_GeminiTargets` — verifies gemini targets are discovered with correct keys, roles, and provider
- [x] New test `TestComputeExpectedTemplate_Gemini` — verifies template resolution for autonomous vs interactive roles
- [x] New test `TestTemplateContentEqual` — verifies JSON structural comparison with whitespace normalization
- [x] All existing `hooks` and `doctor` package tests pass
- [ ] Manual: `gt doctor -v` shows gemini targets as out-of-sync for stale installs (laser/crew/jen, laser/witness, laser/refinery)
- [ ] Manual: `gt doctor --fix` re-installs gemini settings from current templates

Closes gt-0j3q

### DIFF
--- a/internal/doctor/hooks_sync_check.go
+++ b/internal/doctor/hooks_sync_check.go
@@ -44,6 +44,14 @@ func (c *HooksSyncCheck) Run(ctx *CheckContext) *CheckResult {
 
 	var details []string
 	for _, target := range targets {
+		if target.Provider == "gemini" {
+			if detail := c.checkGeminiTarget(target); detail != "" {
+				details = append(details, detail)
+			}
+			continue
+		}
+
+		// Claude targets: use base+override merge system
 		expected, err := hooks.ComputeExpected(target.Key)
 		if err != nil {
 			details = append(details, fmt.Sprintf("%s: error computing expected: %v", target.DisplayKey(), err))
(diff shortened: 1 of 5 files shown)


DreadPirateRobertz:
Good fix for a real pain point — Gemini hook staleness causing the full prime path on compact is a session-killer. The template comparison approach (structural JSON compare with `{{GT_BIN}}` substitution) is cleaner than the Claude merge strategy since Gemini settings are fully template-driven.

One question: `DiscoverTargets` now scans crew member, witness, refinery, and polecat work directories — does it also cover city-level or town-level Gemini sessions? Town-level sessions that run long-lived Gemini instances may have settings in a different path than rig-scoped sessions.

Also worth cross-referencing with PR #3080 (hooks sync for non-Claude agents) — both touch `hooks.Target` struct. PR #3080 adds `DiscoverRoleLocations`/`DiscoverWorktrees`, and this PR adds a `Provider` field. If both merge independently they'll conflict. Might be worth coordinating with @jrwoolley to decide which lands first or whether they should be combined.

PR #3079, opened 2026-03-20
Fix convoy UUID event polling and shell PATH precedence

### LINKED ISSUE #3076: Convoy event polling breaks on UUID event IDs (beads schema mismatch)
## Summary
`gt daemon run` / `gt daemon start` can leave the Convoy event poller broken when the workspace is backed by a beads schema where `events.id` is UUID / `char(36)` instead of an auto-increment integer.

In that state, the daemon starts, but Convoy polling logs repeated scan errors and never processes close events correctly.

## Environment
- Host: macOS
- Workspace: town at `~/gt`
- Repro date: March 20, 2026
- `bd version`: 0.61.0
- Installed `gt` before local fix: Homebrew `gastown 0.12.1`

## Observed Behavior
`~/gt/daemon/daemon.log` repeatedly shows errors like:

```text
Convoy: event poll error (hq): failed to scan event: sql: Scan error on column index 0, name "id": converting driver.Value type []uint8 ("6af709a5-1c6d-4bda-ba60-224ce2e1ebaa") to a int64: invalid syntax
```

The daemon otherwise continues to run, but Convoy event polling is effectively broken.

## Why this looks like a compatibility bug
The local source tree was pinned to `github.com/steveyegge/beads v0.59.0`, which still models:
- `Event.ID` as `int64`
- `GetAllEventsSince(ctx, sinceID int64)`

But the actual workspace schema / newer beads API uses:
- `Event.ID` as UUID string
- time-based event polling cursor

So an older `gastown` binary against a newer beads schema fails at runtime inside the event poller.

## Repro
1. Start from a working town using a newer beads/Dolt schema.
2. Run `gt daemon run` or `gt daemon start` with a binary built against the old beads event API.
3. Watch `~/gt/daemon/daemon.log`.
4. Within the event poll interval, scan errors start repeating.

## Expected Behavior
Convoy event polling should work regardless of whether `events.id` is integer or UUID, or `gastown` should fail fast with a clear compatibility error instead of silently degrading into repeated poll errors.

## Local confirmation / fix
I confirmed locally that updating `gastown` to use the newer beads event API and switching the Convoy poll high-water mark from integer event IDs to `created_at` / `time.Time` eliminates the repeated `event poll error` lines on the same workspace.

## Suggested fix
- Upgrade `gastown` to the beads API that treats event IDs as strings / UUIDs.
- Change ConvoyManager event polling to use a time-based cursor instead of integer event IDs.
- Consider adding a startup compatibility check so older binaries fail explicitly on incompatible workspace schemas.

### LINKED ISSUE #3077: Source install can stay shadowed by Homebrew gt because ~/.local/bin is not prioritized
## Summary
Gas Town's canonical install path and shell integration can leave users running a stale `gt` binary even after successfully building and installing a newer one.

This happens when an older Homebrew install exists at `/usr/local/bin/gt` and the newer source-built install lands at `~/.local/bin/gt`, but the user's shell PATH does not prioritize `~/.local/bin`.

## Environment
- Host: macOS / zsh
- Existing install: `/usr/local/bin/gt` -> Homebrew `gastown 0.12.1`
- Source install target: `~/.local/bin/gt`
- Repro date: March 20, 2026

## Observed Behavior
After building and installing from source, the new binary is present at `~/.local/bin/gt`, but a fresh shell still resolves:

```text
which gt
/usr/local/bin/gt
```

So the user thinks they are testing the newly built binary, but they are still running the old Homebrew binary.

## Why this looks like a product / install UX bug
`make install` intentionally installs to `~/.local/bin/gt`, but:
- it only removes stale binaries from `~/go/bin` and `~/bin`
- it does not warn when `/usr/local/bin/gt` still shadows the canonical install
- the shell integration file did not ensure `~/.local/bin` was placed before `/usr/local/bin`

This makes upgrades and source testing easy to get wrong, especially on macOS systems that already have a Homebrew install.

## Repro
1. Install `gt` via Homebrew so `/usr/local/bin/gt` exists.
2. Clone the repo and run `make install`.
3. Open a new zsh shell.
4. Run `which gt`.
5. Observe that the shell may still resolve to `/usr/local/bin/gt` instead of `~/.local/bin/gt`.

## Expected Behavior
One of these should happen:
- shell integration ensures `~/.local/bin` takes precedence
- `make install` warns loudly when another `gt` earlier in PATH will still be used
- installer / doctor explicitly detects and reports that the canonical install is shadowed by Homebrew

## Local workaround
I fixed this locally by prepending:

```sh
export PATH="$HOME/.local/bin:$HOME/go/bin:$PATH"
```

to `~/.zshrc`.

## Suggested fix
- Update shell install / enable flow so `~/.local/bin` is guaranteed to be early in PATH.
- Add a `gt doctor` or post-install check for binary shadowing.
- Warn specifically about `/usr/local/bin/gt` when Homebrew and source installs coexist.

fusioncid:
## Summary
- update convoy event polling to use beads v0.61 time-based cursors instead of assuming integer event IDs
- add shell-hook PATH normalization so source installs in ~/.local/bin and ~/go/bin are not shadowed by older gt binaries earlier in PATH
- cover both changes with focused daemon and shell integration tests

## Testing
- go test ./internal/shell ./internal/daemon -count=1
- go test -run ^ ./...
- manually rebuilt gt, ran daemon against ~/gt, and verified new daemon log output no longer emitted the UUID -> int64 convoy poll error

Closes #3076
Closes #3077

### DIFF
--- a/internal/shell/integration.go
+++ b/internal/shell/integration.go
@@ -156,6 +156,22 @@ var shellHookScript = `#!/bin/bash
 # Installed by: gt install --shell
 # Location: ~/.config/gastown/shell-hook.sh
 
+_gastown_prepend_path() {
+    local dir="$1"
+    [[ -z "$dir" ]] && return 0
+    [[ -d "$dir" ]] || return 0
+    case ":$PATH:" in
+        *":$dir:"*) ;;
+        *) PATH="$dir${PATH:+:$PATH}" ;;
+    esac
+}
+
+_gastown_ensure_path() {
+    _gastown_prepend_path "$HOME/.local/bin"
+    _gastown_prepend_path "$HOME/go/bin"
+    export PATH
+}
+
 _gastown_enabled() {
     [[ -n "$GASTOWN_DISABLED" ]] && return 1
     [[ -n "$GASTOWN_ENABLED" ]] && return 0
@@ -231,6 +247,8 @@ _gastown_offer_add() {
             ;;
     esac
 }
+
+_gastown_ensure_path
 
 _gastown_hook() {
     local previous_exit_status=$?
--- a/internal/shell/integration_test.go
+++ b/internal/shell/integration_test.go
@@ -130,3 +130,21 @@ func TestUpdateRCFile(t *testing.T) {
 		t.Errorf("RC file has %d start markers, want 1", startCount)
 	}
 }
+
+func TestShellHookScript_EnsuresCanonicalBinPaths(t *testing.T) {
+	if !strings.Contains(shellHookScript, `_gastown_prepend_path "$HOME/.local/bin"`) {
+		t.Error("shell hook should prepend ~/.local/bin to PATH")
+	}
+	if !strings.Contains(shellHookScript, `_gastown_prepend_path "$HOME/go/bin"`) {
+		t.Error("shell hook should prepend ~/go/bin to PATH")
+	}
+	if !strings.Contains(shellHookScript, "_gastown_ensure_path") {
+		t.Error("shell hook should call _gastown_ensure_path")
+	}
+}
+
+func TestShellHookScript_AvoidsDuplicatePathEntries(t *testing.T) {
+	if !strings.Contains(shellHookScript, `case ":$PATH:" in`) {
+		t.Error("shell hook should guard against duplicate PATH entries")
+	}
+}


xzy818:
The underlying fix is a significant improvement, and it's worth taking a moment to detail precisely why the old approach was fragile and how the new timestamp-based polling mechanism improves robustness.

  Here is a detailed analysis for the record:
  **1. The Core Flaw of the Original Integer ID Polling**
  The bug originated from a fundamental mismatch between the application's assumption and the database's reality.
     1. Faulty Assumption: The gastown application was originally built against an older version of the beads library. In that
      library's schema, the events.id primary key was an auto-incrementing int64. The application code was therefore written with the explicit expectation of scanning an integer from this column.
     2. Changing Reality: The beads library evolved. In version v0.61.0 and later, to better support distributed systems and data federation, the events.id column was changed to a UUID (a 36-character string).
     3. The Bug: When the gastown daemon (linked against the old beads logic) ran against a workspace using the new schema, the event poller would fail. The Go SQL driver would attempt to scan a UUID string (e.g., "a7a2886c-...") into an int64 variable. This type conversion is impossible, triggering a sql: Scan error ... invalid syntax exception, which crashed the entire event polling loop.

  **2. Full Bug Scenario**
  A typical scenario to reproduce this failure is as follows:
   1. Setup: A user has an older build of gastown but a newer version of the beads tool. They create a new workspace (bd new) ormigrate an old one, resulting in a database where events.id is a UUID.
   2. Trigger: The user starts the gastown daemon (gt daemon run).
   3. Failure Cascade:
       * The ConvoyManager's event poller queries the beads database for new events.
       * As it tries to read the first row of the result, the sql.Scan error described above occurs.
       * This error crashes the polling goroutine or throws it into a perpetual retry-fail loop, flooding the daemon's log file
         with errors.
   4. End Result: The gastown automation engine is crippled. It can no longer perceive new events from the beads database (such as issue closures), and all convoy automations grind to a halt.

xzy818:
**Risk Analysis of the New Timestamp-Based Solution**

  Switching the polling cursor from id to created_at is a standard and much more robust architectural pattern. It decouples gastown from the implementation details of the beads primary key. However, it does introduce a dependency on clock synchronization, which carries its own set of risks.

   **1. Risk 1: Clock Rollbacks (The Primary Risk)**
       * Scenario: The greatest risk is a system clock moving backward, typically due to NTP adjustments or manual changes. If the database server's clock is set to an earlier time T0 after the gastown poller has already registered a high-water mark of T1 (where T1 > T0), a critical failure occurs.
       * Consequence: Any new events written to the database between T0 and T1 will have timestamps before the poller's high-water mark. When the poller next queries for events with WHERE created_at > T1, these events will be permanently missed.
       * Impact: This would result in lost events, silently breaking the integrity of the automation workflow.

   **2. Risk 2: Clock Skew & Timestamp Precision**
       * Scenario: In a distributed environment with multiple nodes writing to the database, minor clock drift between nodes is possible. It's also theoretically possible for two events to have the exact same high-precision timestamp.
       * Consequence: The query WHERE created_at > 'high_water_mark' is strict. If an event's timestamp is identical to the high-water mark, it could be missed.
       * Impact: This PR's implementation correctly mitigates this by fetching a batch of events and then setting the new high-water mark to the maximum CreatedAt value within that batch. This ensures that even if multiple events share a timestamp, they are all processed before the high-water mark is advanced past them.

  **Conclusion on Robustness:**
  This fix is the correct architectural choice. It moves the system from a "guaranteed to fail" state (due to the type mismatch) to a "works correctly under normal clock conditions" state.
  While the risks associated with timekeeping are real, they are a well-understood class of infrastructure problems. The solution is not for the application to build complex logic to handle time travel, but for the underlying infrastructure to ensure reliable clock synchronization via NTP. This PR replaces a fragile, application-level design flaw with a dependency on standard, stable infrastructure—a significant improvement in robustness by any measure.

steveyegge:
The convoy UUID→time polling fix is already on main (from the beads v0.61 upgrade). The shell PATH manipulation is too invasive — hardcoding ~/.local/bin and ~/go/bin into every user's shell hook makes assumptions about their setup. If users have stale gt binaries shadowing newer ones, that's a `gt doctor` check or install-time concern, not something the shell hook should silently fix. Thanks for flagging both issues though.

PR #3080, opened 2026-03-21
fix: make hooks sync support non-Claude agents

jrwoolley:
## Summary

- `gt hooks sync` only discovered `.claude/settings.json` targets, silently skipping all non-Claude agents (OpenCode, Cursor, Gemini, Copilot, Pi, OMP)
- Add `SyncForRole` for content-aware template sync (separate from `InstallForRole` which remains create-only to protect Claude's merged settings)
- Add `DiscoverRoleLocations` and `DiscoverWorktrees` for agent-agnostic target discovery
- Update `hooks sync` to resolve the configured agent per role and sync template-based agents to their individual worktrees

## Test plan

- [x] Unit tests for `SyncForRole` (create, update, unchanged, error paths)
- [x] Unit tests for `DiscoverRoleLocations` and `DiscoverWorktrees`
- [x] Integration test for `runHooksSync` with non-Claude agent (OpenCode)
- [x] Integration test for dry-run mode
- [x] Manual test: OpenCode plugin synced to worktree, updated when stale, unchanged on re-sync
- [x] Manual test: Cursor hooks.json synced to worktree, updated when stale
- [x] Manual test: Claude settings.json untouched by non-Claude sync path
- [x] Manual test: Full end-to-end with OpenCode as mayor — meta-prompt injection, bead creation, polecat dispatch
- [x] All existing tests pass (`go test ./internal/hooks/... ./internal/runtime/...`)
- [x] Pre-existing `TestConvoyResolveBeadsDir_RegressionEmptyConvoy` failure confirmed unrelated (macOS `/tmp` → `/private/tmp` symlink)

### DIFF
--- a/internal/doctor/hooks_sync_check.go
+++ b/internal/doctor/hooks_sync_check.go
@@ -1,18 +1,32 @@
 package doctor
 
 import (
+	"bytes"
 	"fmt"
 	"os"
 	"path/filepath"
 	"strings"
 
+	"github.com/steveyegge/gastown/internal/config"
 	"github.com/steveyegge/gastown/internal/hooks"
 )
 
-// HooksSyncCheck verifies all settings.json files match what gt hooks sync would generate.
+// templateTarget tracks a non-Claude template-based agent file that is out of sync.
+type templateTarget struct {
+	path           string
+	dir            string
+	provider       string
+	role           string
+	hooksDir       string
+	settingsFile   string
+	useSettingsDir bool
+}
+
+// HooksSyncCheck verifies all hook/settings files match what gt hooks sync would generate.
 type HooksSyncCheck struct {
 	FixableCheck
-	outOfSync []hooks.Target
+	outOfSync         []hooks.Target   // Claude targets
+	templateOutOfSync []templateTarget // Non-Claude template-based targets
 }
 
 // NewHooksSyncCheck creates a new hooks sync validation check.
@@ -28,10 +42,15 @@ func NewHooksSyncCheck() *HooksSyncCheck {
 	}
 }
 
-// Run checks all managed settings.json files for sync status.
+// Run checks all managed hook/settings files for sync status.
 func (c *HooksSyncCheck) Run(ctx *CheckContext) *CheckResult {
 	c.outOfSync = nil
+	c.templateOutOfSync = nil
 
+	var details []string
+	totalTargets := 0
+
+	// Loop 1: Claude targets — use base+override merge system via DiscoverTargets.
 	targets, err := hooks.DiscoverTargets(ctx.TownRoot)
 	if err != nil {
 		return &CheckResult{
@@ -42,16 +61,9 @@ func (c *HooksSyncCheck) Run(ctx *CheckContext) *CheckResult {
 		}
 	}
 
-	var details []string
 	for _, target := range targets {
-		if target.Provider == "gemini" {
-			if detail := c.checkGeminiTarget(target); detail != "" {
-				details = append(details, detail)
-			}
-			continue
-		}
+		totalTargets++
 
-		// Claude targets: use base+override merge system
 		expected, err := hooks.ComputeExpected(target.Key)
 		if err != nil {
 			details = append(details, fmt.Sprintf("%s: error computing expected: %v", target.DisplayKey(), err))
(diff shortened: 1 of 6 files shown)


DreadPirateRobertz:
Solid PR — the separation between `SyncForRole` (update-safe, content-aware) and `InstallForRole` (create-only, protects Claude's merged settings) is the right design. A few things worth checking before merge:

1. **Non-Claude agents lose settings on sync** — `SyncForRole` appears to do a full template overwrite for OpenCode/Cursor/etc. Claude uses a merge strategy to preserve manual settings additions. Is a full overwrite correct for template-based agents, or could users have local customizations in e.g. `.opencode/hooks.json` that would get clobbered?

2. **`DiscoverWorktrees` scope** — does this discover worktrees for all roles (witness, refinery, crew, polecat), or only crew? Polecat worktrees in particular can be ephemeral and recreated frequently — sync running against a stale worktree path is low-risk but worth noting.

3. **+677/-41 line count** — significantly larger than the doctor Gemini PR (#3074) which covers similar ground. Some overlap risk. Worth confirming these two don't conflict on the `hooks.Target` struct (PR #3074 adds a `Provider` field that may need to be present here too).

jrwoolley:
Thanks for the review @DreadPirateRobertz — good questions.

**1. Template overwrite and user customizations**

`SyncForRole` only overwrites the specific gastown-managed file (e.g., `.opencode/plugins/gastown.js`), not the whole directory. Checked the OpenCode source — it loads all `*.js`/`*.ts` files from `.opencode/plugins/`, so user plugins alongside `gastown.js` are unaffected. For agents like Cursor (`.cursor/hooks.json`), the file is fully template-driven — same as how Claude's hooks section is fully managed by `syncTarget`.

**2. DiscoverWorktrees scope**

It runs for all non-settingsDir rig roles: crew, polecat, witness, and refinery. Syncing ephemeral polecat worktrees is harmless — worst case the file lands in a directory that gets cleaned up later. Witness and refinery typically have a single subdirectory.

**3. Overlap with #3074**

We rebased onto the latest main which includes #3074's `Provider` field on `Target` and `discoverGeminiTargets` — no struct conflict.

But looking more closely, the two PRs solve the same underlying problem from different angles:
- **#3074** adds Gemini-specific discovery and check/fix to `gt doctor`
- **#3080** adds generic infrastructure (`SyncForRole`, `DiscoverRoleLocations`, `DiscoverWorktrees`) to `gt hooks sync` for all non-Claude agents

What's missing is connecting them: `gt doctor hooks-sync` could use our generic functions instead of the Gemini-specific `checkGeminiTarget`/`fixGeminiTarget`/`discoverGeminiTargets` code, making the doctor agent-agnostic too. Happy to extend this PR to update `hooks_sync_check.go` to use `SyncForRole` if that's the preferred direction — it would subsume #3074's Gemini-specific code and cover all template-based agents in one shot.

DreadPirateRobertz:
Good answers on all three points. The `SyncForRole`-based approach making `gt doctor` agent-agnostic is the right direction — it eliminates the Gemini-specific proliferation in `hooks_sync_check.go` that would need to be repeated for every new agent type.

Yes, extend #3080 to update `hooks_sync_check.go` to use `SyncForRole`. Subsuming #3074's Gemini-specific code into the generic infrastructure is cleaner than having both live side by side. One PR that covers all template-based agents is the right call.

steveyegge:
Note: PR #3074 (Gemini hooks-sync doctor check) just merged. When you address mel's review feedback, please also subsume #3074's Gemini-specific code paths (`checkGeminiTarget`/`fixGeminiTarget`/`discoverGeminiTargets`) into the generic `SyncForRole` infrastructure — that was the agreed direction and will eliminate the per-provider branching in `hooks_sync_check.go`.

jrwoolley:
Addressed all review feedback:

1. **ComputeExpectedTemplate dedup** — now delegates to `resolveAndSubstitute` (identical logic removed)
2. **hooks_sync_check.go made agent-agnostic** — replaced Gemini-specific `checkGeminiTarget`/`fixGeminiTarget` with generic two-loop architecture using `DiscoverRoleLocations` + `SyncForRole`. `gt doctor hooks-sync` now covers all template-based agents (OpenCode, Cursor, Gemini, Copilot, etc.)
3. **SyncForRole JSON comparison** — uses `TemplateContentEqual` for `.json` files (whitespace-tolerant), `bytes.Equal` for non-JSON
4. **Removed `discoverGeminiTargets`** — subsumed by the generic `DiscoverRoleLocations` path

Manually tested with a 4-rig workspace (Claude, OpenCode, Gemini, Cursor) across all code paths: initial sync, idempotent re-sync, tampered file detection/update, `gt doctor` check + fix, and dry-run.

PR #3082, opened 2026-03-21
fix(doltserver): quarantine moves corrupted DBs instead of deleting them

DreadPirateRobertz:
## Problem

The startup quarantine logic used `os.RemoveAll` to delete database directories with a missing `noms/manifest`. This caused permanent data loss when a manifest was temporarily absent — e.g., after a crash or incomplete write during Dolt startup.

This is what caused `beads_hq` to disappear from `.dolt-data/` after a Dolt restart (gt-0hqx), blocking `gt mail inbox` for `gastown/refinery`.

## Fix

Move corrupted directories to `.dolt-data/.quarantine/<name>` instead of deleting them. The data is preserved for manual recovery. Falls back to `RemoveAll` only if the `os.Rename` itself fails (e.g., cross-device move).

```
Before: Quarantine: removing corrupted database dir "beads_hq" (missing noms/manifest)
After:  Quarantine: moved corrupted database dir "beads_hq" to .quarantine/ (missing noms/manifest)
```

## Test plan

- [ ] `go test ./internal/doltserver/...` passes
- [ ] Startup with a phantom `.dolt` dir (no `noms/manifest`) moves it to `.quarantine/` rather than deleting it
- [ ] Server starts successfully after quarantine
- [ ] Legitimate databases are unaffected

Fixes gt-0hqx.


DreadPirateRobertz:
**PR quality audit — bundled changes need splitting**

The doltserver quarantine-move fix is solid (doltserver.go +19/-4). The approach is right — move over remove, with RemoveAll fallback on cross-device rename failure.

However this PR bundles unrelated wl_stamp changes:
- Removes `endorsement` stamp type
- Removes `pilot_cohort` column/flag entirely
- SQL INSERT columns changed

This is causing the lint failures and ~10 extra test failures beyond the 5 pre-existing ones. These wl changes need to land in a separate PR.

**Recommendation**: Extract the wl_stamp changes to their own PR with context on why endorsement/pilot_cohort are being removed. Keep this PR to the doltserver quarantine fix only. The core fix here is correct and reviewable — don't let the bundled noise block it.

DreadPirateRobertz:
Rebased onto current main (ref: 9d67d3de).

The apparent "wl_stamp bundling" was a false positive — the branch was based on 853f9e93, which predated 331e708a (pilot_cohort/endorsement addition). The diff vs main showed those as "removals" because the branch simply predated them, not because it was intentionally stripping them.

Post-rebase, the diff is now exactly doltserver.go only (+19/-4). No wl_stamp changes involved. CI failures should reduce to just the 5 pre-existing ones fixed by #3096 — this PR can merge once #3096 lands.

steveyegge:

PR #2857, opened 2026-03-15
fix: deduplicate CRASHED_POLECAT alerts in daemon (#2795)

DreadPirateRobertz:
## Summary
- Add `crashNotified` map to daemon to track which polecats have already had CRASHED_POLECAT alerts sent
- Skip re-notification on subsequent heartbeat cycles until the polecat's session comes back alive
- Prevents the witness inbox from being flooded with duplicate crash alerts (86+ messages observed in production)

Fixes #2795

## How it works
- First crash detection for a polecat: send CRASHED_POLECAT mail to witness, record in `crashNotified`
- Subsequent heartbeat cycles: skip notification (log "Skipping duplicate")
- When session comes back alive: clear the entry from `crashNotified`, enabling future crash detection

## Test plan
- [ ] `go build ./...` passes
- [ ] New test `TestCheckPolecatHealth_DeduplicatesCrashNotification` verifies dedup behavior
- [ ] Existing tests continue to pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #2795, opened 2026-03-14
bug: witness floods mayor inbox with duplicate RECOVERY_NEEDED alerts for intentionally stopped polecats


drdanmaggs:
## Bug Description

When polecats complete their work and shut down cleanly (or are nuked with \`gt polecat nuke --force\`), the overseer continues to fire \`RECOVERY_NEEDED\` / \`CRASHED_POLECAT\` alerts for those sessions on every patrol cycle (~30s). The mayor inbox grows unboundedly until the rig is parked.

## Steps to Reproduce

1. Sling a large batch of beads to polecats (7+ in parallel): \`gt sling <bead> <rig>\`
2. Wait for polecats to complete and shut down (state: \`done\`, \`session_running: false\`)
3. Observe witness flooding mayor inbox with repeated \`RECOVERY_NEEDED\` alerts for the completed polecats
4. Inbox reaches 86+ messages within ~30 minutes, growing ~6 messages per cycle with no end state

## Expected Behavior

Once a polecat has shut down cleanly (state: \`done\`) or been explicitly nuked (\`gt polecat nuke --force\`), the overseer should not fire further \`RECOVERY_NEEDED\` alerts for that session. Alerts should be idempotent — if the mayor has acknowledged a recovery alert for a given polecat, no further alerts should be sent.

## Actual Behavior

Overseer re-alerts on every patrol cycle regardless of whether:
- The polecat completed normally (\`state: done\`)
- The polecat was intentionally nuked (\`gt polecat nuke --force\`)
- The mayor already replied to a prior alert for the same polecat

The only way to stop the flood is \`gt rig park <rig>\`, which also stops the refinery mid-queue.

## Impact

- Mayor inbox unusable (86+ messages, growing)
- Witness spends every cycle replying to duplicate alerts instead of monitoring live work
- Rig must be parked to stop flooding, which halts the refinery merge queue

## Environment

- **OS**: macOS 15 (Darwin 24.5.0)
- **Gas Town version**: \`gt version 0.12.0 (dev)\`
- **Triggered by**: Running 7 parallel polecats, nuking 4 mid-flight, remainder completing cleanly

drdanmaggs:
[...]

PR #3084, opened 2026-03-21
fix: deduplicate CRASHED_POLECAT notifications in daemon heartbeat (#2795)

DreadPirateRobertz:
## Summary

Consecutive daemon heartbeat cycles were sending duplicate CRASHED_POLECAT notifications to the witness for the same crashed polecat, flooding the witness inbox (GH#2795).

**Root cause**: `checkPolecatHealth` notifies the witness unconditionally whenever a polecat session is dead with a hook_bead. On each 3-minute heartbeat, the same crash triggers another notification.

**Fix**: Add `crashNotified map[string]time.Time` to the `Daemon` struct. Before notifying the witness, check if we've already notified for this polecat. Skip if so. Clear the entry when the session comes back alive, so real recoveries + new crashes still trigger alerts.

## Changes

- `internal/daemon/daemon.go`: Add `crashNotified` field + dedup guard before `notifyWitnessOfCrashedPolecat`
- `internal/daemon/polecat_health_test.go`: Add `TestCheckPolecatHealth_DeduplicatesCrashNotification`

## Test

```
go test ./internal/daemon/ -run TestCheckPolecatHealth -v
```

All polecat health tests pass. This replaces PR #2857 (which had accumulated unrelated diffs and conflicts).

Closes #2857

### DIFF
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -99,6 +99,12 @@ type Daemon struct {
 	// lastMaintenanceRun tracks when scheduled maintenance last ran.
 	// Only accessed from heartbeat loop goroutine - no sync needed.
 	lastMaintenanceRun time.Time
+
+	// crashNotified tracks polecats that have already been notified as crashed.
+	// Prevents flooding the witness with duplicate CRASHED_POLECAT alerts on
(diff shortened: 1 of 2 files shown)


steveyegge:

PR #3089, opened 2026-03-21
fix(witness): cap await-signal backoff-max at 2m to catch remote crash notifications

### LINKED ISSUE #3075: Witness await-signal misses crash notifications from overseer
## Bug

The witness uses `bd activity --follow` in `await-signal` to wake between patrol cycles. When a polecat crashes, the overseer detects it and sends `CRASHED_POLECAT` mail to the witness. However, mail arrival does not trigger bd activity, so the witness stays asleep and never reads the crash notification.

## Observed Impact

`stt/chrome` crash-looped 11 times over ~35 minutes while the witness slept. The witness had 15 unread messages (crash reports + merge notifications) and only processed them when manually nudged by the mayor.

## Expected Behavior

The `await-signal` subscription should also wake on incoming mail, or the witness should poll mail on a bounded timer (e.g. max 2 minutes) regardless of bd activity. Without this, any polecat crash during a quiet period goes undetected until the 5-minute backoff cap expires -- or longer if the bd activity subscription is stale.

## Affected Components

- `gt mol step await-signal` -- only subscribes to `bd activity`, not mail arrival
- Witness formula (`mol-witness-patrol`) -- context-check step assumes await-signal covers all wake conditions

## Reproduction

1. Start a rig with a witness patrolling
2. Let the witness enter `await-signal` (idle, no bd activity)
3. Crash a polecat (kill its tmux session)
4. Observe: overseer sends `CRASHED_POLECAT` mail to witness
5. Witness does not wake up until backoff timer expires or manual nudge

DreadPirateRobertz:
## Problem

The witness patrol used `--backoff-max 5m` in `await-signal`, meaning it could sleep up to 5 minutes between patrol cycles when idle. Crash notifications from remote rigs (e.g. `stt/chrome`) do not write to the **local** `.events.jsonl`, so `await-signal` would not wake early — the witness missed crash reports for up to the full 5-minute cap.

Observed: `stt/chrome` crash-looped 11 times over ~35 minutes while the witness slept through 15 unread messages.

## Fix

Cap `--backoff-max` at 2 minutes. The witness now wakes at least every 2 minutes regardless of event source, ensuring crash notifications from remote rigs are never delayed more than 2 minutes.

## Test plan

- [ ] `go test ./internal/formula/...` passes
- [ ] Witness patrol formula contains `--backoff-max 2m`
- [ ] Witness wakes within 2 minutes even with no local events activity

Fixes #3075.


steveyegge:

PR #3093, opened 2026-03-21
fix(rig): check for empty remote before cloning in gt rig add (#3054)

### LINKED ISSUE #3054: rig add fails with confusing error on empty repositories
## Summary

`gt rig add` fails with a confusing error when the target repository is empty (no commits/branches):

```
$ gt rig add bot https://github.com/chandrameenamohan/taskflow-bot
Creating rig bot...
  Repository: https://github.com/chandrameenamohan/taskflow-bot
  Cloning repository (this may take a moment)...
Error: adding rig: creating bare repo: fetching origin main: fatal: couldn't find remote ref main
```

## Root Cause

The clone flow in `AddRig()` does:
1. `git clone --bare --single-branch --depth 1` — succeeds even for an empty repo
2. `configureRefspec()` tries to fetch the default branch (`main`) — **fails** because there are no refs

The code already has an `IsEmpty()` check in `manager.go` (~line 423-430) with a good error message:
```
"repository %s is empty (no commits). Push at least one commit before adding it as a rig"
```

But this check runs **after** `configureRefspec()` is called during the clone, so the fetch error surfaces first with the cryptic "couldn't find remote ref main" message.

## Expected Behavior

A clear error message like:
> "Repository https://github.com/chandrameenamohan/taskflow-bot is empty (no commits). Push at least one commit before adding it as a rig."

## Suggested Fix

Either:
1. **Pre-validate the remote** with `git ls-remote` before cloning — if no refs are returned, fail early with the friendly message.
2. **Reorder the checks** so `IsEmpty()` runs before `configureRefspec()`, or catch the fetch error in `configureRefspec()` and check for empty repo before returning.

## Environment

- gt version: 0.12.1 (Homebrew)
- macOS Darwin 22.4.0

DreadPirateRobertz:
## Summary

- Adds `IsRemoteEmpty(url string)` to `git.go` — uses `ls-remote --heads` to check for branches without cloning
- Calls it in `AddRig` before the clone attempt
- Produces the existing clear error message instead of git's opaque "couldn't find remote ref main"

## Root Cause

`git clone --branch main --depth 1` fails with "couldn't find remote ref main" for repos with no commits. The existing `IsEmpty()` check in `manager.go` (post-clone) never gets a chance to run. `wrapCloneError` doesn't recognize the empty-repo case.

## Fix

Check the remote before cloning via `git ls-remote --heads <url>`. If no heads exist, return the clear message already written:
> "repository X is empty (no commits). Push at least one commit before adding it as a rig"

## Test plan

- [x] `go test ./internal/git/...` passes
- [ ] Manual: `gt rig add emptyrepo <url-of-empty-repo>` shows clear error (not git clone error)

Fixes #3054.

### DIFF
--- a/internal/git/git.go
+++ b/internal/git/git.go
@@ -991,6 +991,17 @@ func (g *Git) IsEmpty() (bool, error) {
 	return strings.TrimSpace(out) == "", nil
 }
 
+// IsRemoteEmpty returns true if the remote URL has no refs (no commits).
+// Uses ls-remote --heads to check for branches without cloning.
+// Returns (true, nil) for empty repos and (false, nil) for non-empty repos.
+func (g *Git) IsRemoteEmpty(url string) (bool, error) {
+	out, err := g.run("ls-remote", "--heads", url)
+	if err != nil {
+		return false, fmt.Errorf("checking remote refs: %w", err)
+	}
+	return strings.TrimSpace(out) == "", nil
+}
+
 // RemoteBranchExists checks if a branch exists on the remote.
 func (g *Git) RemoteBranchExists(remote, branch string) (bool, error) {
 	out, err := g.run("ls-remote", "--heads", remote, branch)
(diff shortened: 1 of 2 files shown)


steveyegge:
Duplicate of #3116 (both address GH#3054). Closing this one — #3116 is the cleaner version.

PR #3096, opened 2026-03-21
fix(ci): repair 5 pre-existing test failures blocking all PRs

DreadPirateRobertz:
## Summary

Five pre-existing test failures were blocking CI on all open PRs. This fixes them all.

**Root causes:**

1. **`TestStrandedScanQueryShape`** — assumed `lines[0]` was the `bd list` call, but `BdSupportsAllowStale()` now emits a `bd --allow-stale version` probe first. Fixed: search all lines for "list" instead of hardcoding index 0.

2. **`TestGetTrackedIssues_FallsBackToShowTrackedDependencies`** — stub had a catch-all `exit 1` that rejected the version probe, causing `BdSupportsAllowStale()` to cache false → `--allow-stale` stripped → expected args didn't match. Fixed: add `"--allow-stale version") exit 0 ;;` to stub.

3. **`TestRunConvoyList/Status_UsesTownRootAndStripsBeadsDir`** — same probe issue but worse: the test sets `BEADS_DIR=/wrong/.beads` to verify it gets stripped, and the stub exits 1 when `BEADS_DIR` is set. The probe runs before `runBdJSON`'s BEADS_DIR stripping, so the probe fails. Fixed: exempt the probe from the BEADS_DIR guard (before the `if BEADS_DIR` check).

4. **`TestConvoyResolveBeadsDir_RegressionEmptyConvoy`** — macOS-specific symlink mismatch: `t.TempDir()` returns `/var/folders/...` but `filepath.EvalSymlinks` inside `ResolveBeadsDir` returns `/private/var/folders/...`. Fixed: normalize both paths with `EvalSymlinks` before comparing.

5. **`TestDoctorDogDatabases`** — test expected 6 default databases but `doctorDogDatabases()` was updated to return 3 (`hq`, `gt`, `mo`) without updating the test. Fixed: update expected count to 3.

**Bonus:** Remove unused `timeout` param from `(*Tmux).sendMessageToTarget` (all callers always passed `constants.NudgeReadyTimeout`), fixing the `unparam` lint warning.

## Test plan

- [x] `go test ./internal/cmd/ -run "TestStrandedScan|TestGetTrackedIssues|TestRunConvoyList|TestRunConvoyStatus|TestConvoyResolveBeadsDir"` — all pass
- [x] `go test ./internal/daemon/ -run "TestDoctorDog"` — pass
- [x] `go build ./...` — clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3101, opened 2026-03-21
fix(reaper): DiscoverDatabases returns error instead of hardcoded fallback (#3013)

DreadPirateRobertz:
## Summary

Fixes GH#3013 — `DiscoverDatabases` silently returned `DefaultDatabases` on any failure, causing the reaper to operate on a stale hardcoded list when `SHOW DATABASES` failed.

**Before:** any error (connection failure, timeout, empty result) → silently returns `["hq"]`

**After:** returns `([]string, error)` — callers fail the reaper cycle explicitly and log the discovery error

## Changes

- `internal/reaper/reaper.go`: `DiscoverDatabases` now returns `([]string, error)`. Removed the three silent fallback return paths.
- `internal/cmd/reaper.go`: both `databases` and `run` subcommands return error on discovery failure.
- `internal/daemon/wisp_reaper.go`: logs discovery failure and calls `mol.failStep("scan", ...)`.
- `DefaultDatabases` kept for reference but removed from the fallback path.

## Test plan

- [x] `go test ./internal/reaper/` passes
- [x] `go build ./internal/reaper/ ./internal/cmd/ ./internal/daemon/` clean
- [x] `TestDoctorDogDatabases` failure is pre-existing (same failure on main — tracked in PR #3096)

### DIFF
--- a/internal/cmd/reaper.go
+++ b/internal/cmd/reaper.go
@@ -46,7 +46,10 @@ var reaperDatabasesCmd = &cobra.Command{
 	Use:   "databases",
 	Short: "List databases available for reaping",
 	RunE: func(cmd *cobra.Command, args []string) error {
-		dbs := reaper.DiscoverDatabases(reaperHost, reaperPort)
+		dbs, err := reaper.DiscoverDatabases(reaperHost, reaperPort)
+		if err != nil {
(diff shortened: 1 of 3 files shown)


DreadPirateRobertz:
Note: this PR does **not** depend on #3096 and can be reviewed/merged independently.

Issue #3006, opened 2026-03-18
Daemon inherits agent identity from launch environment, causing misattributed system messages


Henry-E:
**Summary**
When `gt daemon run` is started from an agent session, the daemon inherits that session's identity environment, including `GT_ROLE` and `BD_ACTOR`. Any daemon-spawned subprocess that derives sender/actor identity from ambient environment can then be misattributed to that agent instead of to a system/service identity.

**Observed behavior**
A daemon process was found running with an inherited agent environment such as:
- `GT_ROLE=<rig>/crew/<agent>`
- `GT_RIG=<rig>`
- `GT_CREW=<agent>`
- `BD_ACTOR=<rig>/crew/<agent>`

The daemon then emitted `CRASHED_POLECAT` notifications that were recorded with:
- `Owner: <rig>/crew/<agent>`
- `from:<rig>/crew/<agent>`

...even though they were generated by daemon logic, not by that crew agent.

**Why this happens**
The daemon shells out to `gt mail send` for crash notifications. The daemon code attempts to set `BD_ACTOR=daemon`, but `gt mail send` derives sender identity from `GT_ROLE`-based detection, so the inherited agent role wins and the message is attributed to the launching agent.

**Impact**
This affects provenance and auditability:
- system-generated mail can appear to come from a human/agent session
- attribution for daemon-generated actions becomes unreliable
- any daemon subprocess that consults ambient agent identity may be contaminated by launch context

**Expected behavior**
The daemon should not inherit crew/polecat identity in a way that causes service-generated actions to be attributed to the agent session that happened to launch it.

**Confirmed scope**
Confirmed by:
- inspecting the live daemon process environment
- observing `CRASHED_POLECAT` beads/messages attributed to a crew identity
- tracing daemon notification code into `gt mail send`
- tracing mail sender detection to `GT_ROLE`-based identity resolution

**Relevant code paths**
- `internal/cmd/daemon.go`
- `internal/daemon/daemon.go`
- `internal/cmd/mail_send.go`
- `internal/cmd/mail_identity.go`

Henry-E:
In general it's pretty dumb that mail from the daemon look like they're coming from whatever place the daemon was started and if it can break based on being init from a crew's rig.

DreadPirateRobertz:
[...]

PR #3102, opened 2026-03-21
fix(daemon): clear agent identity env vars at startup (#3006)

DreadPirateRobertz:
Fixes GH#3006 — daemon inherits agent identity from launch environment.

When `gt daemon run` is launched from an agent session, `GT_ROLE`, `GT_CREW`, and `BD_ACTOR` are inherited from that session. Any subprocess that derives sender identity from these vars (e.g. `gt mail send`) is then misattributed to the launching agent.

**Fix**: add three lines to `runDaemonRun` before daemon creation:
```go
os.Unsetenv("GT_ROLE")
os.Unsetenv("GT_CREW")
os.Setenv("BD_ACTOR", "daemon")
```

The `BD_ACTOR=daemon` override already existed on some `exec.Command` calls (`daemon.go:2154`) using `os.Environ() + append`, but that pattern loses to GT_ROLE-based sender detection. Clearing at process startup ensures no residual agent identity leaks through any subprocess.

## Test plan
- [x] `go build ./internal/cmd/` clean

### DIFF
--- a/internal/cmd/daemon.go
+++ b/internal/cmd/daemon.go
@@ -343,6 +343,15 @@ func runDaemonRun(cmd *cobra.Command, args []string) error {
 		return fmt.Errorf("not in a Gas Town workspace: %w", err)
 	}
 
+	// Clear agent identity env vars inherited from the launch environment.
+	// When the daemon is started from an agent session, GT_ROLE/GT_CREW are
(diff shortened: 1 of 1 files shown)


DreadPirateRobertz:
Note: this PR does **not** depend on #3096 and can be reviewed/merged independently.

PR #3105, opened 2026-03-21
fix(test): add 1ms sleep in TestGenerateStampID_Unique to fix Windows flakiness

### LINKED ISSUE #3104: bug: TestGenerateStampID_Unique flaky on Windows (nanosecond timer resolution)
## Summary

`TestGenerateStampID_Unique` in `internal/cmd/wl_stamp_test.go:144` is flaky on Windows CI. It calls `generateStampID()` twice and expects different values because the function uses a nanosecond timestamp. On Windows, the timer resolution is typically 100ns–15ms (depending on hardware), causing consecutive calls to sometimes return the same timestamp and produce identical IDs.

## Observed

The test appears as `=== PAUSE TestGenerateStampID_Unique` in Windows CI runs — it either hangs waiting for the clock to advance, or times out.

## Fix

Two options:
1. Add a small sleep between the two calls: `time.Sleep(time.Microsecond)` is sufficient to force clock advance on all platforms
2. Change `generateStampID` to use a monotonic counter as the uniqueness component instead of (or in addition to) nanoseconds — this is more robust

The function already uses nanoseconds for uniqueness, but on Windows the granularity is insufficient for back-to-back calls in the same test.

## Environment

- Observed on Windows CI (GitHub Actions, Windows Server runner)
- Does NOT affect Linux/macOS (nanosecond resolution is reliable)

DreadPirateRobertz:
## Summary

Fixes Windows CI flakiness in `TestGenerateStampID_Unique` (issue #3104).

**Root cause:** Windows timer resolution is ~15ms (100ns–15ms depending on hardware). Back-to-back calls to `generateStampID()` can return identical nanosecond timestamps, making the two IDs collide and the uniqueness assertion fail.

**Fix:** Add a 1ms `time.Sleep` between the two calls. This is longer than Windows' worst-case timer resolution and shorter than any CI timeout. The test remains fast (~1ms overhead) and reliable on all platforms.

## Test plan

- [x] `go test ./internal/cmd/ -run "TestGenerateStampID"` — passes locally
- [x] No functional changes — test-only fix

Closes #3104

### DIFF
--- a/internal/cmd/wl_stamp_test.go
+++ b/internal/cmd/wl_stamp_test.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"strings"
 	"testing"
+	"time"
 
 	"github.com/steveyegge/gastown/internal/doltserver"
 )
@@ -144,6 +145,7 @@ func TestGenerateStampID_Format(t *testing.T) {
 func TestGenerateStampID_Unique(t *testing.T) {
 	t.Parallel()
 	id1 := generateStampID("alice", "bob", `{"quality":4}`, "c-abc")
+	time.Sleep(time.Millisecond) // ensure clock advances on Windows (coarse timer resolution)
 	id2 := generateStampID("alice", "bob", `{"quality":4}`, "c-abc")
 	if id1 == id2 {
 		t.Error("sequential stamp IDs should be unique (includes nanosecond timestamp)")


DreadPirateRobertz:
Closing in favor of #3106 (batty's atomic counter approach) — atomic counter is strictly better than a sleep: no timer dependency, works on all platforms regardless of clock resolution. Both fix #3104.

PR #3106, opened 2026-03-21
fix(wl): atomic counter in generateStampID fixes Windows flakiness (GH#3104)

### LINKED ISSUE #3104: bug: TestGenerateStampID_Unique flaky on Windows (nanosecond timer resolution)

DreadPirateRobertz:
## Problem

`TestGenerateStampID_Unique` is flaky on Windows CI. `generateStampID` uses `time.Now().UTC().Format(time.RFC3339Nano)` for uniqueness, but Windows' timer resolution (~100ns–15ms) means two consecutive calls can return the same timestamp → same SHA-256 input → identical IDs.

## Fix

Add a package-level `atomic.Uint64` counter that increments on every call. The counter value is included alongside the timestamp in the SHA-256 input:

```go
var stampCounter atomic.Uint64

func generateStampID(author, subject, valence, contextID string) string {
    now := time.Now().UTC().Format(time.RFC3339Nano)
    seq := stampCounter.Add(1)
    input := fmt.Sprintf("%s|%s|%s|%s|%s|%d", author, subject, valence, contextID, now, seq)
    hash := sha256.Sum256([]byte(input))
    return fmt.Sprintf("s-%s", hex.EncodeToString(hash[:])[:12])
}
```

This guarantees uniqueness across back-to-back calls on all platforms — no sleep, no timing dependency.

## Test plan

- [ ] `TestGenerateStampID_Format` — still passes (format unchanged)
- [ ] `TestGenerateStampID_Unique` — deterministically unique now

Fixes #3104

### DIFF
--- a/internal/cmd/wl_scorekeeper.go
+++ b/internal/cmd/wl_scorekeeper.go
@@ -54,10 +54,10 @@ func runWlScorekeeper(cmd *cobra.Command, args []string) error {
 	}
 
 	store := doltserver.NewWLCommons(townRoot)
-	return runScorekeeperWithStore(store, townRoot)
+	return runScorekeeperWithStore(store)
 }
 
-func runScorekeeperWithStore(store doltserver.WLCommonsStore, townRoot string) error {
+func runScorekeeperWithStore(store doltserver.WLCommonsStore) error {
 	if !wlScorekeeperJSON {
 		fmt.Printf("%s Running scorekeeper...\n", style.Bold.Render("⚡"))
 	}


DreadPirateRobertz:
Rebased onto current main (c74602ba) — was 11 commits behind which included the 5 pre-existing test failures now fixed in #3096.

The unparam fix for `sendMessageToTarget` (unused timeout param) dropped cleanly as already upstream. Only the atomic counter fix for `generateStampID` remains.

CI should now show only legitimate failures from this PR's code, if any. Lint was already passing.

PR #3109, opened 2026-03-21
fix(refinery): supersede MR on same-branch re-submission with new commits (GH#3032)

DreadPirateRobertz:
## Summary

- **Root cause**: Polecat branch `@timestamp` suffix is a base-36 millisecond timestamp, NOT a commit SHA. Multiple fix attempts share the same branch name, so `FindMRForBranch` always found the existing open MR and returned "idempotent" — even when the polecat had pushed new commits after a failure.
- **Fix**: Record the branch HEAD SHA as `head_sha` in the MR description. On re-submission, compare stored `head_sha` against current HEAD. If they differ (new commits pushed), close the stale MR and create a fresh one so the Refinery processes the new commits.
- **Backward-compatible**: Old MRs without `head_sha` continue to behave as before (idempotent by branch name).

## Changes

- `internal/beads/fields.go`: Add `HeadSHA` field to `MRFields`; parse `head_sha` key in `ParseMRFields`
- `internal/cmd/done.go`: Capture HEAD SHA at MR creation time; supersede old MR if SHA differs on re-submission
- `internal/cmd/mq_submit.go`: Same — HEAD SHA capture + supersede logic

## Test plan

- [ ] `go build ./...` passes
- [ ] `go test ./internal/beads/...` passes
- [ ] `go test ./internal/refinery/...` passes
- [ ] Simulate scenario: create MR, push new commits to same branch, re-run `gt done` → old MR is superseded, new MR created with updated `head_sha`

### DIFF
--- a/internal/cmd/done.go
+++ b/internal/cmd/done.go
@@ -861,9 +861,14 @@ func runDone(cmd *cobra.Command, args []string) (retErr error) {
 				style.PrintWarning("MR %s has stale HEAD (%s → %s) — superseding with fresh submission (GH#3032)",
 					existingMR.ID, short(existingFields.HeadSHA), short(headSHA))
 				if closeErr := bd.CloseWithReason(fmt.Sprintf("superseded: new commits on %s", branch), existingMR.ID); closeErr != nil {
-					style.PrintWarning("could not supersede stale MR %s: %v", existingMR.ID, closeErr)
+					// Close failed — keep existing MR to avoid two open MRs for the same branch.
+					style.PrintWarning("could not supersede stale MR %s (reusing it): %v", existingMR.ID, closeErr)
+					mrID = existingMR.ID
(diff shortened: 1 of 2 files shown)


DreadPirateRobertz:
Good work on this. Root cause identification (base-36 millisecond timestamp, not SHA) was non-obvious and the fix is clean.

The backward-compat approach (old MRs without `head_sha` keep original behavior) is correct — avoids breaking existing open MRs on upgrade.

One thing to verify: the supersede logic closes the stale MR before creating the new one. Make sure the close + create is either atomic or the close failure is handled gracefully (i.e., if close fails, don't create the new MR — avoid having two open MRs for the same branch). If close fails, should probably return an error rather than proceeding.

Otherwise looks good to merge after #3096. Tests pass, diff is tight.

steveyegge:
Fix-merged to main as 30839a1b. Reviewed: solid supersede logic with good safety fallbacks. Thanks @DreadPirateRobertz!

PR #3108, opened 2026-03-21
feat(sling): add --model flag to override Claude model at spawn time

DreadPirateRobertz:
## Summary

- Adds \`--model\` flag to \`gt sling\` to override the Claude model when spawning a polecat (e.g., \`gt sling gt-abc myrig --model claude-opus-4-6\`)
- Threads the model value through the full dispatch pipeline: \`ScheduleOptions\` → \`SlingContextFields\` (persisted for deferred dispatch) → \`DispatchParams\` → \`SlingParams\` → \`SlingSpawnOptions\` → polecat startup command (\`--model value\` appended)
- Works for both direct dispatch and deferred/scheduled dispatch via the capacity scheduler

## Files changed

- \`internal/cmd/sling.go\` — \`slingModel\` var + \`--model\` flag wired into all \`ScheduleOptions\` and \`ResolveTargetOptions\` sites
- \`internal/cmd/sling_batch.go\` — wired into batch \`SlingParams\`
- \`internal/cmd/sling_schedule.go\` — \`Model\` in \`ScheduleOptions\` + wired into \`SlingContextFields\`
- \`internal/cmd/sling_dispatch.go\` — \`Model\` in \`SlingParams\` + passed to \`SlingSpawnOptions\`
- \`internal/cmd/sling_target.go\` — \`Model\` in \`ResolveTargetOptions\` + both \`SlingSpawnOptions\` sites
- \`internal/cmd/polecat_spawn.go\` — \`Model\` in \`SlingSpawnOptions\` + \`SpawnedPolecatInfo\` + injected into startup command
- \`internal/cmd/capacity_dispatch.go\` — \`dp.Model\` passed to \`SlingParams\`
- \`internal/scheduler/capacity/pipeline.go\` — \`Model\` in \`SlingContextFields\` + \`DispatchParams\` + \`ReconstructFromContext\`

## Test plan

- [x] \`go build ./...\` passes clean
- [x] \`go test ./internal/cmd/ -run "TestSling|TestDispatch|TestCapacity|TestSchedule|TestBatch|TestStamp"\` passes
- [x] \`go test ./internal/scheduler/capacity/...\` passes

Closes #2944

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3111, opened 2026-03-21
fix(security): enforce full suffix validation in looksLikeBeadID (GH#3110)

### LINKED ISSUE #3110: security: looksLikeBeadID only validates first char of bead suffix, rest is unchecked
## Problem

`looksLikeBeadID` in `internal/cmd/handoff.go:1319` validates the prefix portion of a bead ID correctly (lowercase letters only) and checks that `rest[0]` is alphanumeric — but does **not** validate the remaining characters in `rest`.

A string like `gt-a$(cmd)` or `gt-a;evil` passes `looksLikeBeadID` today.

## Current code

```go
// Check rest starts with alphanumeric and contains only alphanumeric, dots, hyphens
first := rest[0]
if !((first >= 'a' && first <= 'z') || (first >= '0' && first <= '9')) {
    return false
}
return true  // rest[1:] is NOT validated
```

## Why risk is currently low but should be fixed

All downstream bead ID uses go through Go's `exec.Command` with argument slices (not shell strings), so shell injection via a malformed bead ID is not currently exploitable. However:

1. The comment says "only alphanumeric, dots, hyphens" but doesn't enforce it
2. If any future code path passes a bead ID through `fmt.Sprintf` into a shell string, the gap becomes exploitable
3. Defense-in-depth: the validation should match its stated intent

## Fix

Validate all characters in `rest` against `[a-z0-9.-]` (matching the comment's stated contract):

```go
for _, c := range rest[1:] {
    if !((c >= 'a' && c <= 'z') || (c >= '0' && c <= '9') || c == '.' || c == '-') {
        return false
    }
}
```

## Related

PR #3108 / cfdd9c82 established the input validation pattern for flag values that reach spawn paths. This applies the same pattern to bead ID validation.

DreadPirateRobertz:
## Problem

`looksLikeBeadID` only validated `rest[0]`, leaving the rest of the bead ID suffix unchecked. A string like `gt-a$(cmd)` passed the check despite containing shell metacharacters.

## Risk

Currently low: all downstream uses of bead IDs go through Go's `exec.Command` with argument slices, not shell strings. However the function's stated contract ("only alphanumeric, dots, hyphens") was not enforced, and a future code path passing a bead ID into `fmt.Sprintf` for a shell string would be exploitable.

Found during a dispatch pipeline security audit triggered by the `--model` injection fix in PR #3108.

## Fix

Validate every character in the suffix against `[a-z0-9.-]` with the constraint that the first character must be alphanumeric:

```go
for i, c := range rest {
    if (c >= 'a' && c <= 'z') || (c >= '0' && c <= '9') {
        continue
    }
    if i > 0 && (c == '.' || c == '-') {
        continue
    }
    return false
}
```

## Tests

Added 8 shell-metacharacter test cases to the existing `TestLooksLikeBeadID` table in `sling_test.go`. All pass.

## Checklist

- [x] `go test ./internal/cmd/... -run TestLooksLikeBeadID` passes
- [x] `go build ./internal/cmd/...` passes
- [x] Shell metacharacters (`$(`)`, `;`, `|`, `&`, `>`, backtick) now correctly rejected
- [x] Valid IDs (`gt-abc`, `hq-cv-abc`, `ap-qtsup.16`) still accepted

Closes #3110.

### DIFF
--- a/internal/cmd/handoff.go
+++ b/internal/cmd/handoff.go
@@ -1338,9 +1338,15 @@ func looksLikeBeadID(s string) bool {
 		return false
 	}
 
-	// Check rest starts with alphanumeric and contains only alphanumeric, dots, hyphens
-	first := rest[0]
-	if !((first >= 'a' && first <= 'z') || (first >= '0' && first <= '9')) {
+	// Check rest contains only alphanumeric, dots, hyphens — enforcing the stated contract.
+	// The first character must be alphanumeric (no leading dot/hyphen).
+	for i, c := range rest {
+		if (c >= 'a' && c <= 'z') || (c >= '0' && c <= '9') {
+			continue
+		}
+		if i > 0 && (c == '.' || c == '-') {
+			continue
+		}
 		return false
 	}
 
--- a/internal/cmd/sling_test.go
+++ b/internal/cmd/sling_test.go
@@ -1190,6 +1190,16 @@ func TestLooksLikeBeadID(t *testing.T) {
 		{"123-abc", false},      // numeric prefix
 		{"a-", false},           // nothing after hyphen
 		{"aaaaaa-b", false},     // prefix too long (6 chars)
+
+		// Shell metacharacters must be rejected (defense-in-depth, GH#3110)
+		{"gt-a$(cmd)", false},
+		{"gt-a;evil", false},
+		{"gt-a|pipe", false},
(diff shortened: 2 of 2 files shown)


steveyegge:
Fix-merged to main. Full suffix validation in looksLikeBeadID — shell metacharacters now rejected. All tests pass.

PR #3113, opened 2026-03-21
fix(formula): remove go test default from mol-refinery-patrol test_command

work2fly:
## Summary
- Removes the incorrect default `go test ./...` test_command from mol-refinery-patrol formula
- Source issue: gt-g7p

## Merge Details
- Branch: polecat/furiosa/gt-g7p@mn08jh0a
- Worker: furiosa
- Clean rebase on main, ff-only merge

Merged by gastown/refinery

### DIFF
--- a/docs/concepts/integration-branches.md
+++ b/docs/concepts/integration-branches.md
@@ -516,7 +516,7 @@ Commands run in this order (any can be empty = skip):
 
 ### Example Configurations
 
-**Go project** (default — only test_command is set by default):
+**Go project** (all commands empty by default — configure per-rig):
 ```json
 {
   "merge_queue": {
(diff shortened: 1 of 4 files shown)


DreadPirateRobertz:
LGTM. Removing `go test ./...` as the hardcoded default is the right call — test commands vary by rig configuration and having a blanket default silently runs potentially unintended test suites.

The doc updates (`reference.md`, `integration-branches.md`, and the example JSON) are consistent and clearly document the new "empty = skip" behavior. Formula table updated too.

One minor observation: it might be worth adding a note somewhere that projects can set `test_command: "go test ./..."` explicitly if they want Go test behavior — just so the intentional opt-in path is discoverable. Not a blocker.

✓ Approve

steveyegge:
Fix-merged to main as 86a67876. Reviewed and approved — change is correct, consistent with existing Go code defaults, and well-scoped.

PR #3116, opened 2026-03-21
fix(rig): pre-check remote for empty repo before cloning (GH#3054)

DreadPirateRobertz:
## Summary

- `gt rig add` was emitting a cryptic `fatal: couldn't find remote ref main` when the target repo had no commits. A user-friendly error message already existed but fired too late — `configureRefspec`'s fetch call would fail first.
- Fix: add `RemoteIsEmpty(url)` to `git.go` (uses `git ls-remote --refs`) and call it in `AddRig()` before the clone. If no refs exist, fail immediately with the existing clear message.

## Test plan

- [ ] `go build ./...` passes
- [ ] `go test ./internal/rig/... ./internal/git/...` passes
- [ ] `gt rig add <name> <empty-repo-url>` now shows: `repository <url> is empty (no commits). Push at least one commit before adding it as a rig`

### DIFF
--- a/internal/git/git.go
+++ b/internal/git/git.go
@@ -991,6 +991,17 @@ func (g *Git) IsEmpty() (bool, error) {
 	return strings.TrimSpace(out) == "", nil
 }
 
+// RemoteIsEmpty checks if a remote repository URL has no refs (no commits).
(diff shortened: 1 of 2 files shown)


steveyegge:

Issue #3042, opened 2026-03-19
7ea8586a reverts per-town tmux socket isolation, re-opens #761


scbrown:
## Summary

Commit `7ea8586a` ("fix: revert InitRegistry socket to empty default, remove unused dbName param") reverted the per-town tmux socket namespacing introduced in `86d3c77e` (cherry-pick from PR #2473, which closed #761).

The revert changed `InitRegistry` so that `GT_TMUX_SOCKET` unset/default/auto maps to `socket = ""` (tmux default server) instead of calling `townSocketName(townRoot)`.

## Impact

- **Re-opens #761**: Multiple towns sharing the same host will collide on the default tmux socket again.
- **Orphans existing sessions**: Any crew/agent sessions created on the per-town socket before the revert become invisible to `gt` after upgrading. `gt crew at` creates new sessions on the default socket while the old sessions (and their Claude processes) continue running on the previous socket. From the user's perspective, the crew "disappears."
- **Silent failure**: No warning is emitted about orphaned sessions on the old socket.

## Reproduction

1. Start crew sessions with a `gt` version that includes `86d3c77e` (per-town sockets)
2. Upgrade to a version that includes `7ea8586a` (revert)
3. Run `gt crew at <name>` — creates a new session on the default tmux socket
4. Old sessions remain running on the previous socket, unreachable by `gt`

## Root Cause

The commit message states the revert was done to match "main's test expectations." It appears tests assumed an empty socket rather than being updated to account for the new per-town socket naming.

## Suggested Fix

Restore `townSocketName()` in `InitRegistry` and update the affected tests to expect per-town socket names. The original PR #2473 (by @DreadPirateRobertz) was intentional and closed a real issue.

Additionally, `gt` should detect and warn about orphaned sessions on stale sockets during startup or `gt crew at`.

htristan:
Also hitting this with **mayor sessions**, not just crew. Running two towns on the same host (e.g. `~/town-a` and `~/town-b`), `gt mayor attach` from the second town attaches to the first town's `hq-mayor` session since they share the same default tmux socket.

Repro:
1. `cd ~/town-a && gt mayor start`
2. `cd ~/town-b && gt mayor start`
3. `gt mayor attach` from `town-b` — connects to the `town-a` mayor

Worth noting the issue affects all session types (mayor, deacon, crew, etc.), not just crew — the current description and repro steps are crew-focused which might cause this to be triaged too narrowly.

[...]

PR #3118, opened 2026-03-21
fix(tmux): restore per-town socket isolation reverted by 7ea8586a

### LINKED ISSUE #761: gt down kills services from other towns due to non-namespaced tmux sessions

htristan:
## Summary

- Restores `townSocketName(townRoot)` in `InitRegistry`, undoing the revert in 7ea8586a that broke per-town tmux socket isolation (re-opens #761)
- Fixes test assertions in `registry_socket_test.go` to expect per-town socket names instead of empty string
- Adds `TestInitRegistry_SocketFormat` validating the `basename-hash6` socket name format

## Context

Commit `86d3c77e` (cherry-pick from PR #2473) introduced per-town tmux socket namespacing to fix #761 (multi-town collisions on shared hosts). The next day, `7ea8586a` reverted the one critical line — changing `townSocketName(townRoot)` back to `""` — with the justification "matches main's test expectations."

The revert left the codebase in an internally inconsistent state:
- `townSocketName()` exists but is dead code (never called at runtime)
- Test names describe per-town behavior but assertions validate the reverted empty-socket behavior
- The doctor split-brain check (`tmux_socket_check.go`) short-circuits as no-op when socket is empty
- Comments in `down.go` describe per-town sockets as active, which was false after the revert

This PR restores the intended behavior. The existing `townSocketName`, `LegacySocketName`, legacy cleanup in `down.go`, and the doctor split-brain check were all written to support per-town sockets — they just needed the one-line revert undone.

Re-closes #761, addresses #3042.

## Test plan

- [x] `go test ./internal/session/` — all socket tests pass with corrected assertions
- [x] `go test ./internal/tmux/ ./internal/lock/ ./internal/doctor/` — no regressions
- [x] `go test ./internal/cmd/ -run "Socket|Legacy|SplitBrain"` — no regressions
- [x] `go build ./...` — compiles cleanly

### DIFF
--- a/internal/session/registry.go
+++ b/internal/session/registry.go
@@ -123,12 +123,12 @@ func InitRegistry(townRoot string) error {
 	var errs []error
 
 	// Determine the tmux socket name from GT_TMUX_SOCKET env var:
-	//   unset / "default" / "auto" → use the default tmux socket (empty name)
+	//   unset / "default" / "auto" → per-town socket derived from town directory path
 	//   any other value            → use that name as-is
 	socket := os.Getenv("GT_TMUX_SOCKET")
 	switch socket {
 	case "", "default", "auto":
-		socket = ""
+		socket = townSocketName(townRoot)
 	}
 	tmux.SetDefaultSocket(socket)
 
--- a/internal/session/registry_socket_test.go
+++ b/internal/session/registry_socket_test.go
@@ -10,7 +10,7 @@ import (
 )
 
 // TestInitRegistry_SocketFromTownName verifies GT_TMUX_SOCKET socket selection:
-//   - unset / "default" / "auto" → use the default tmux socket (empty name)
+//   - unset / "default" / "auto" → per-town socket derived from town directory path
 //   - explicit value              → that value verbatim
 func TestInitRegistry_SocketFromTownName(t *testing.T) {
 	origTMUX := os.Getenv("TMUX")
@@ -75,9 +75,10 @@ func TestInitRegistry_SocketFromTownName(t *testing.T) {
 			_ = InitRegistry(townRoot)
 
 			got := tmux.GetDefaultSocket()
-			if got != "" {
-				t.Errorf("after InitRegistry(%q) with GT_TMUX_SOCKET=%q:\n  socket = %q, want default socket \"\"",
-					townRoot, tt.gtTmuxSocket, got)
+			want := townSocketName(townRoot)
+			if got != want {
+				t.Errorf("after InitRegistry(%q) with GT_TMUX_SOCKET=%q:\n  socket = %q, want %q",
+					townRoot, tt.gtTmuxSocket, got, want)
 			}
 
 			tmux.SetDefaultSocket("")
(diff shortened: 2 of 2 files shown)


DreadPirateRobertz:
Reviewed — this is correct and the analysis is accurate.

**Verified**: The current `InitRegistry` (after `7ea8586a`) uses `GT_TMUX_SOCKET` env var but falls through to `socket = ""` for unset/default/auto — `townSocketName()` is dead code at runtime. The function exists at `registry.go:160` but is never called from `InitRegistry` after the revert.

**Impact**: Without per-town socket names, two gt instances on the same host (different town roots) share the default tmux socket and can clobber each other's sessions — exactly what #761 described. The `townSocketName`, `LegacySocketName`, `down.go` legacy cleanup, and doctor split-brain check are all wired for per-town sockets and work correctly once the one-line revert is undone.

**No behavioral regression risk**: The `GT_TMUX_SOCKET` override path is preserved. Single-town setups on isolated hosts are unaffected. The change only activates the already-written per-town socket logic for the default case.

Recommend merge after CI passes. This is a safe, well-scoped fix.

Henry-E:
This is an evil PR and should not have been merged

htristan:
Can you elaborate on what the issue is here? Muti-town was completely broken and messed up without this fix, and you aren't really sharing what is now not working or what this breaks. Do we not have test coverage for something else that is broken now? More context or detail would be very helpful.

Henry-E:
I'm going to make a PR that should be clearer. But essentially the first gas town should default to the empty / unnamed socket. Otherwise just like myself, everyone is going to run into a lot of issues when they upgrade with the whole split socket brain. Additionally for the majority of people who are not running multiple towns, the first gas town should default to empty / unnamed for ease of use, not to some obscure `-L gt-34jkfdsf` named tag.

I just updated to tip of main and just like the last time this was introduced it completely destroyed my whole town by messing with where agents were put etc. It's not a very nice PR, in the sense that, I'm really glad this solves your niche problems, but I'm fairly sure this will mess things up for 90% of people when they update. And this was already previously reverted within a day or two for just that reason and this just reintroduces the same issues with no consideration to why it was reverted in the first place.

htristan:
I don't think this PR specifically is "evil" — it restores functionality that was already in the codebase (`townSocketName`, `LegacySocketName`, the legacy cleanup in `down.go`, the doctor split-brain check) and was intentionally merged via PR #2473 to fix #761. The revert in 7ea8586a left all of that code as dead weight with no runtime path. This PR just reconnected the one line that made it work.

That said, I do appreciate the overall point you're making. The migration path for existing single-town users upgrading is a real concern — if you go from the default socket to a hashed socket name, your existing sessions become orphaned with no warning. That's a bad experience regardless of whether the end state is correct.

The `gt-a1b2c3` socket names aren't something I introduced — that's the format `townSocketName()` has always produced (it was written by @DreadPirateRobertz in #2473). But I agree that for single-town users who don't need isolation, it's an unnecessary change from what they're used to.

I'll take a look at #3352 and your proposed approach for handling this.

PR #3121, opened 2026-03-21
docs(security): formula variable security principle (gt-sec-001)

DreadPirateRobertz:
Documents the gt-sec-001 injection class: attack vector, why sanitization is insufficient, structural fix (bead ID as formula var), and general principle (system-generated IDs only in formula vars). Includes decision table. Dutch/mac STRIDE framing pending as follow-up.

### DIFF
--- a/docs/design/formula-variable-security.md
+++ b/docs/design/formula-variable-security.md
@@ -64,6 +64,25 @@ boundary. The gastown polecat then executes with gastown capabilities
 
 ---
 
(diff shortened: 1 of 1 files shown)


DreadPirateRobertz:
Mac STRIDE framing added (7f787917). STRIDE table now complete: EoP (High) as primary threat — cross-rig trust boundary via context injection. Includes "why EoP not Spoofing" clarification. Doc is ready for review.

PR #3126, opened 2026-03-21
fix(sling): serialize concurrent hook writes with per-assignee flock (#3114)

### LINKED ISSUE #3114: Race condition: hook persistence fails under concurrent batch sling
## Summary

When slinging multiple beads concurrently to the same rig (e.g., 8 beads in rapid succession), all sling operations report success but hook assignments don't persist on any polecat. `gt hook` shows nothing afterward, requiring manual re-attachment.

## Root Cause Analysis

Investigation of the Dolt server logs from the incident (2026-03-14) revealed dozens of "nothing to commit" warnings during the concurrent sling window:

```
[x2] time="2026-03-14T09:33:59-04:00" level=warning msg="error running query" ... error="nothing to commit"
[x3] time="2026-03-14T09:35:30-04:00" level=warning msg="error running query" ... error="nothing to commit"
```

This is a classic TOCTOU (Time-of-Check-Time-of-Use) race condition:

1. Multiple concurrent slings read the same `hook_bead` value from the `issues` table
2. All attempt to update `hook_bead` simultaneously
3. Dolt rolls back conflicting transactions silently — hooks don't persist
4. Sling reports success (polecat spawned, branch created) but the hook write was lost

The existing file locks at `.runtime/locks/sling/*.flock` are per-bead, not per-assignee, so they don't prevent concurrent writes to the same database row.

## Suggested Fix

Add row-level pessimistic locking when updating `hook_bead`:

```sql
START TRANSACTION;
SELECT * FROM issues WHERE id = ? FOR UPDATE;  -- lock the row
UPDATE issues SET hook_bead = ? WHERE id = ?;
COMMIT;
```

Alternatively, serialize sling operations per-polecat or per-assignee rather than per-bead.

## Reproduction

```bash
# Sling multiple beads rapidly to the same rig
for id in bead1 bead2 bead3 bead4 bead5 bead6 bead7 bead8; do
  gt sling $id <rig> &
done
wait
gt polecat list <rig>  # Polecats exist with correct branches
gt hook                 # But hooks are empty
```

## Workaround

Manually re-attach hooks after batch sling:
```bash
gt hook attach <bead-id> <polecat>
```

## Environment

- gastown v0.12.1 (Homebrew)
- Dolt backend on port 3307
- macOS (arm64)

DreadPirateRobertz:
## Summary

Fixes the TOCTOU race condition causing silent hook loss under concurrent batch sling (#3114).

**Root cause**: Multiple concurrent `gt sling` calls targeting the same polecat all raced to write `hook_bead` to Dolt simultaneously. Dolt's optimistic concurrency silently rolled back conflicting writes — polecats ended up with empty hooks even though each sling reported success.

**What was missing**: The per-bead flock (`tryAcquireSlingBeadLock`) serializes slings of the *same bead* but does not prevent multiple distinct beads from racing to write the *same assignee's* hook_bead field.

**Fix**: Add `tryAcquireSlingAssigneeLock()` — a per-assignee blocking flock (`assignee_<safe-agent>.flock`) in `.runtime/locks/sling/`. Acquired immediately before `hookBeadWithRetry`, released immediately after. This serializes hook writes per polecat without blocking unrelated sling operations.

## Test plan

- [x] `go vet ./internal/cmd/...` — clean  
- [x] Sling dispatch/lock tests — PASS
- [x] No new lock contention for different-assignee slings (distinct lock files per assignee)

Closes #3114.

### DIFF
--- a/internal/cmd/sling.go
+++ b/internal/cmd/sling.go
@@ -899,9 +899,19 @@ func runSling(cmd *cobra.Command, args []string) (retErr error) {
 
 	// Hook the bead with retry and verification.
 	// See: https://github.com/steveyegge/gastown/issues/148
+	//
+	// Acquire a per-assignee lock before writing hook_bead to serialize concurrent slings
+	// targeting the same polecat. Without this, multiple concurrent slings race on the
+	// same assignee's row in Dolt, causing silent rollbacks (issue #3114).
+	assigneeUnlock, assigneeLockErr := tryAcquireSlingAssigneeLock(townRoot, targetAgent)
+	if assigneeLockErr != nil {
+		return fmt.Errorf("serializing hook write for %s: %w", targetAgent, assigneeLockErr)
+	}
 	hookDir := beads.ResolveHookDir(townRoot, beadID, hookWorkDir)
-	if err := hookBeadWithRetry(beadID, targetAgent, hookDir); err != nil {
-		return err
+	hookErr := hookBeadWithRetry(beadID, targetAgent, hookDir)
+	assigneeUnlock()
+	if hookErr != nil {
+		return hookErr
 	}
 
 	// Emit a propulsion signal if the target is the mayor.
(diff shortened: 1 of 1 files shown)


DreadPirateRobertz:
Reviewed `tryAcquireSlingAssigneeLock`.

**Correct**:
- Blocking `LOCK_EX` flock is the right choice for serialization (not try-acquire)
- Acquired before `hookBeadWithRetry`, explicitly released immediately after — minimal lock scope is intentional
- Error path: if lock acquisition fails, `releaseSlingLock` (bead lock) still fires via `defer` — no leak
- `safeAgent` replacement of `/` and `:` → `_` is correct for filesystem safety
- Windows no-op handled in platform files

**One minor nit**: `beads.ResolveHookDir` is called between lock acquisition and `assigneeUnlock()`. If it panics, the flock file is held until process exit. In practice this is OS-handled (flock is FD-based), but adding a `defer assigneeUnlock()` immediately after acquisition — and then the explicit `assigneeUnlock()` before the hook call acts as early release for the non-panic path — isn't possible in Go since defer fires at function return. So the explicit release is the right pattern here; just worth a brief comment noting the panic edge case is OS-handled.

Overall: LGTM. The fix correctly serializes per-assignee hook writes without blocking unrelated slings (distinct lock file per assignee).

DreadPirateRobertz:
**Re: 6 failures vs 5 on other PRs** — `TestStrandedScanQueryShape` is pre-existing on main (verified locally). The 5→6 delta is because other PRs were cut before this test regressed on main, not a new failure introduced here. No action needed on this PR.

steveyegge:
Fix-merged to main as 35005ba7. Reviewed: clean per-assignee flock following existing lock pattern. Thanks @DreadPirateRobertz!

PR #3134, opened 2026-03-21
fix: ungate startup idle verification for hook-based agents

### LINKED ISSUE #3133: bug: hook-based polecats have zero startup fallback — sit idle indefinitely when hooks fail
## Summary

When Claude Code hooks fail to fire during polecat startup (settings not found, hook crashed, Claude updated), the agent sits at the welcome screen indefinitely. The system has **zero fallback mechanisms** for hook-based agents — no nudge, no beacon, no idle verification.

Non-hook agents (Codex, AMP) get `SendStartupNudge=true` which triggers `verifyStartupNudgeDelivery` (3 retries × 5s). Hook-based Claude agents get nothing.

This cascades into three user-visible failures:
1. **Startup hang** — polecat sits at Claude welcome screen for hours
2. **No commit / no gt done** — agent never started work, so it never finishes
3. **Anti-rework failures** — next polecat on the same task starts from scratch (no prior branch)

## Root Cause

Two gaps in the startup flow:

**1. `verifyStartupNudgeDelivery` is gated on `SendStartupNudge`**

In `session_manager.go`, the idle-verification safety net only runs when `fallbackInfo.SendStartupNudge` is true. `GetStartupFallbackInfo()` returns `SendStartupNudge=false` for hooks+prompt agents (Claude's default config), so the entire verification path is skipped. If hooks fail silently, nobody checks whether the agent actually started working.

**2. `containsPromptIndicator` false positives**

`containsPromptIndicator()` in `tmux.go` matches any line ending with `>`, causing false positives on HTML tags (`</div>`), progress arrows (`====>`), and error messages (`expected <string>`). This can cause `AcceptWorkspaceTrustDialog` to exit early, missing a trust dialog that hasn't rendered yet.

## Reproduction

The bug manifests in a real gastown environment when hooks fail silently during polecat startup. To reproduce:

1. **Set up a rig with a polecat worktree** (normal `gt sling` flow)
2. **Break the hooks**: In the polecat's worktree, rename or delete `.claude/settings.json` so the SessionStart hook won't fire. This simulates the real failure modes: settings not synced to worktree, hook definition changed by a Claude Code update, or hook command crashing.
3. **Sling work to the polecat**: `gt sling <issue-id>` — the polecat gets assigned work via its hook bead
4. **Observe the startup flow**: The polecat's tmux session will show Claude at the `❯` prompt, idle. The startup code in `session_manager.go` completes without error — `AcceptStartupDialogs` succeeds, `WaitForRuntimeReady` sees the prompt, but `GetStartupFallbackInfo` returns no nudge and `verifyStartupNudgeDelivery` never runs.
5. **Wait**: The polecat sits idle indefinitely. The Witness's `DetectStalledPolecats` will eventually fire blind dialog dismissal (Enter keys), but this doesn't help — the agent is past the dialogs, it just has no work instructions.

The key is that the startup code **reports success** — no errors, no warnings. The gap is silent: the system assumes hooks fired because it never checks.

### What correct behavior looks like

After startup completes, `verifyStartupNudgeDelivery` should run regardless of `SendStartupNudge`, detect the agent is idle at `❯`, and send a recovery nudge ("Check your hook with `gt hook`"). For non-hook agents this already works; hook agents are the gap.

## Affected Code Paths

- `internal/polecat/session_manager.go` → `Start()`: `verifyStartupNudgeDelivery` gated on `fallbackInfo.SendStartupNudge`
- `internal/runtime/runtime.go` → `GetStartupFallbackInfo()`: hooks+prompt branch returns zero fallback
- `internal/tmux/tmux.go` → `containsPromptIndicator()`: false positives on non-prompt `>` content
- `internal/witness/handlers.go` → `DetectStalledPolecats()`: only does dialog dismissal, doesn't expose pane/agent context for the Witness agent to reason about

## Environment

- gastown @ main
- Claude Code v2.1.81
- macOS (darwin 25.2.0)

htristan:
Closes #3133

## Summary

- **Ungate `verifyStartupNudgeDelivery`** from `SendStartupNudge` in `session_manager.go` so it runs for all agents with prompt detection, not just nudge-based agents. When hooks succeed, the check sees the agent is working and returns instantly (zero side effects). When hooks fail silently, it detects the idle prompt and sends a recovery nudge.
- **Tighten `containsPromptIndicator`** to reject non-prompt `>` content (HTML tags, progress arrows, version strings, error messages) that caused `AcceptWorkspaceTrustDialog` to exit early.
- **Enrich Witness `StalledResult`** with `PaneSnapshot`, `AgentState`, `HasHookedWork`, and classified `StallType` (`dialog-stall` vs `idle-stall` vs `startup-stall`) so the Witness agent can reason about recovery actions instead of only doing blind dialog dismissal.

## Test plan

- [x] `go test ./internal/tmux/ -run TestContainsPromptIndicator` — prompt indicator false positive fix + guardrails
- [x] `go test ./internal/polecat/ -run TestHookAgentStartup_VerificationRunsWithoutNudge` — integration test: real tmux session, hook agent config, verification runs without nudge
- [x] `go test ./internal/runtime/ -run TestGetStartupFallbackInfo` — fallback matrix unchanged for all agent configs
- [x] `go test ./internal/witness/ -run TestStalledResult` — enriched struct + stall type classification
- [x] `go test ./...` — full suite, no regressions

### DIFF
--- a/internal/polecat/session_manager.go
+++ b/internal/polecat/session_manager.go
@@ -455,12 +455,12 @@ func (m *SessionManager) Start(polecat string, opts SessionStartOptions) error {
 		}
 	}
 
-	// Verify startup nudge was delivered: poll for idle prompt and retry if lost.
-	// This fixes the Mode B race where the nudge arrives before Claude Code is ready,
-	// causing the polecat to sit idle at an empty prompt. See GH#1379.
-	if fallbackInfo.SendStartupNudge {
-		m.verifyStartupNudgeDelivery(sessionID, runtimeConfig)
-	}
+	// Verify the agent isn't stuck at the idle prompt after startup.
+	// For nudge-based agents, this catches lost nudges (GH#1379).
+	// For hook-based agents, this catches silent hook failures (GH#3133).
+	// verifyStartupNudgeDelivery self-guards: returns immediately if
+	// the runtime has no ReadyPromptPrefix (can't detect idle state).
+	m.verifyStartupNudgeDelivery(sessionID, runtimeConfig)
 
 	// Legacy fallback for other startup paths (non-fatal)
 	_ = runtime.RunStartupFallback(m.tmux, sessionID, "polecat", runtimeConfig)
--- a/internal/polecat/session_manager_test.go
+++ b/internal/polecat/session_manager_test.go
@@ -13,6 +13,7 @@ import (
 
 	"github.com/steveyegge/gastown/internal/config"
 	"github.com/steveyegge/gastown/internal/rig"
+	gtruntime "github.com/steveyegge/gastown/internal/runtime"
 	"github.com/steveyegge/gastown/internal/session"
 	"github.com/steveyegge/gastown/internal/tmux"
 )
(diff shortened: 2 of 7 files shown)


htristan:
**CI failures here are unrelated to this PR** — they're the `captureStdout` data race in `TestOutputRoleDirectives` tracked in #3140. The race is in `internal/cmd/prime_output_test.go` (pre-existing on `main`), not in any file this PR touches.

All 8 failures break down to:
- 5 direct: `TestOutputRoleDirectives` + 4 subtests — `captureStdout()` races on `os.Stdout` under `t.Parallel()`
- 2 collateral: `TestClassifySchemaChange`, `TestValidateConvoyStatusTransition` — race detector poisons the test binary
- 1 parent: `TestOutputRoleDirectives` itself

Fix is in #3141 — refactors `outputRoleDirectives` to accept `io.Writer` instead of writing to `os.Stdout`. Will rebase this PR once that merges.

steveyegge:
Hi @htristan — thanks for this. The core logic change (ungating verification from `SendStartupNudge`) is correct, and the `AgentState`/`HasHookedWork` fields on `StalledResult` are good structural additions.

However, we have ZFC concerns with two parts of this PR:

**1. Prompt detection is still string heuristics on pane content.**

The tightened `>` matching reduces false positives, but it's still screen-scraping to determine agent state. Making a heuristic better doesn't make it structural — it just shifts the failure cases. `"Type your question >"` or any agent using `>` differently will still false-match.

**2. `PaneSnapshot` elevates screen-scraping to first-class witness input.**

Feeding 20 lines of pane content to the witness for "agent-level reasoning" couples the witness to every agent's TUI layout. If Copilot or Gemini changes their welcome screen, the witness's reasoning breaks silently.

**What we'd prefer instead:**

The underlying problem — "hooks might fail silently" — should be solved structurally. Hook-based agents should report their own health: a ready marker file written by the session-start hook, a heartbeat, or structured hook output. One reliable health-check mechanism is better than two parallel detection systems (hooks + screen-scraping) that can conflict.

We'd welcome a resubmission that:
- Keeps the ungating of `verifyStartupNudgeDelivery` (the logic fix is correct)
- Keeps `AgentState`/`HasHookedWork` on `StalledResult` (structural state)
- Replaces `PaneSnapshot` and prompt heuristic changes with a hook-based health check (e.g., hook writes `~/.gt/ready/<session-id>` on success, verification checks for the file with a timeout)

Thanks for tackling this — the problem is real, just needs a structural solution.

PR #3124, opened 2026-03-21
fix(test): eliminate data race in TestValidateStampInputs (GH#3122)

htristan:
## Summary

Fixes #3122 — data race in `TestValidateStampInputs_*` tests that poisons unrelated tests under `-race`.

The four `TestValidateStampInputs_*` tests called `t.Parallel()` but concurrently read/wrote package-level globals (`wlStampQuality`, `wlStampSeverity`, etc.) via a save-and-restore pattern. Under `go test -race`, this caused collateral failures in `TestClassifySchemaChange`, `TestParseSchemaVersion_Valid`, and `TestParseSchemaVersion_Invalid`.

**Fix**: Extract a `stampInputs` struct and change `validateStampInputs` to accept it as a parameter instead of reading globals. The single production caller (`runWlStamp`) constructs the struct from cobra flag globals. Tests construct local values directly — no globals touched, `t.Parallel()` is safe.

**Files changed**:
- `internal/cmd/wl_stamp.go` — add `stampInputs` struct, refactor `validateStampInputs(in stampInputs)`, update caller
- `internal/cmd/wl_stamp_test.go` — remove save-and-restore boilerplate, pass `stampInputs` directly

## Test plan

- [x] `go test -race ./internal/cmd/ -run "TestValidateStampInputs|TestClassifySchemaChange|TestParseSchemaVersion"` — all pass, no race warnings
- [x] `go build ./internal/cmd/...` — clean
- [x] Linter clean

Made with [Cursor](https://cursor.com)

PR #3141, opened 2026-03-21
fix(test): refactor outputRoleDirectives to accept io.Writer, eliminating data race (GH#3140)

### LINKED ISSUE #3140: bug: data race in captureStdout causes TestOutputRoleDirectives and other cmd tests to fail under -race
## Summary

Multiple tests in `internal/cmd` are failing on CI with data race errors. The root cause is `captureStdout()` in `prime_test.go`, which swaps the global `os.Stdout` without synchronization. When parallel subtests run concurrently, they race on `os.Stdout`.

## Failing Tests

All in `internal/cmd`, all under `-race`:

- `TestOutputRoleDirectives` (and all subtests: `no_directives_emits_nothing_visible`, `town-level_directive_emits_town_header`, `rig-level_directive_emits_rig_header`, `both_levels_emits_combined_header`, `explain_mode_shows_file_paths`)
- `TestClassifySchemaChange` (collateral — race detected during same test binary execution)
- `TestValidateConvoyStatusTransition` (collateral)

## Race Stack

```
Write at 0x000003c78be8 by goroutine 3682:
  github..../gastown/internal/cmd.captureStdout()
      .../internal/cmd/prime_test.go:29 +0x105   // os.Stdout = w

Previous read at 0x000003c78be8 by goroutine 3684:
  fmt.Println()
      .../src/fmt/print.go:315 +0x9fc
  github..../gastown/internal/cmd.outputRoleDirectives()
      .../internal/cmd/prime_output.go:146 +0x99f
```

## Root Cause

`captureStdout()` (`prime_test.go:22-45`) does `os.Stdout = w` and later `os.Stdout = oldStdout`. This is not safe when multiple parallel subtests call it concurrently — one goroutine writes to `os.Stdout` while another reads it via `fmt.Println`.

## Fix

Same pattern used in #3123 for `TestValidateStampInputs`: either remove `t.Parallel()` from the subtests that use `captureStdout`, or refactor the functions under test to accept an `io.Writer` instead of writing to `os.Stdout` directly.

## Impact

Failing the `Test` CI job on `main` and all PRs when the race detector triggers (nondeterministic but frequent).

htristan:
## Summary

Fixes #3140 — data race in `captureStdout` causes `TestOutputRoleDirectives` and other `internal/cmd` tests to fail under `-race`.

### Root cause (two races)

1. **`os.Stdout` race**: `captureStdout()` swaps `os.Stdout = w` / `os.Stdout = oldStdout`. With `t.Parallel()` subtests in `TestOutputRoleDirectives`, multiple goroutines race on the global — one writes the pipe while another reads via `fmt.Println`.
2. **`primeExplain` race**: The "explain mode shows file paths" subtest mutates the `primeExplain` global under `t.Parallel()`, racing with any concurrent `explain()` reader.

### Fix

Refactor `outputRoleDirectives` to accept an `io.Writer` and `explainEnabled bool` as parameters instead of writing to `os.Stdout` and reading the `primeExplain` global:

- **`prime_output.go`**: Change signature to `outputRoleDirectives(ctx RoleContext, w io.Writer, explainEnabled bool)`. Replace `fmt.Println` with `fmt.Fprintln(w, ...)`. Replace `explain()` calls with a local `explainf` closure that writes to `w`.
- **`prime.go`**: Update the single production caller to pass `os.Stdout, primeExplain`.
- **`prime_output_test.go`**: All 6 subtests now pass a `bytes.Buffer` directly — no `captureStdout`, no `primeExplain` mutation. `t.Parallel()` is retained and safe.

### Context

The prior quick fix for a similar race (PR #3123: remove `t.Parallel()` from `TestValidateStampInputs`) was merged then bulk-reverted in `927c935`. I had also opened PR #3124 with the structural refactor approach (parameter injection via a `stampInputs` struct) but closed it in favor of #3123.

This PR applies that same structural pattern — dependency injection over global mutation — to the `outputRoleDirectives` race. It's the more robust fix: tests stay parallel, the function is more testable, and no globals are touched.

## Test plan

- [x] `go build ./internal/cmd/...` — compiles clean
- [x] `go test -race -count=1 ./internal/cmd/ -run "TestOutputRoleDirectives|TestClassifySchemaChange|TestValidateConvoyStatusTransition"` — all pass, zero race warnings
- [x] Linter clean

### DIFF
--- a/internal/cmd/prime.go
+++ b/internal/cmd/prime.go
@@ -408,7 +408,7 @@ func outputRoleContext(ctx RoleContext) (string, error) {
 		return "", err
 	}
 
-	outputRoleDirectives(ctx)
+	outputRoleDirectives(ctx, os.Stdout, primeExplain)
 	outputContextFile(ctx)
 	outputHandoffContent(ctx)
 	outputAttachmentStatus(ctx)
--- a/internal/cmd/prime_output.go
+++ b/internal/cmd/prime_output.go
@@ -4,6 +4,7 @@ import (
 	"github.com/steveyegge/gastown/internal/cli"
 	"encoding/json"
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
 	"strings"
@@ -96,7 +97,11 @@ func outputPrimeContext(ctx RoleContext) (string, error) {
 // outputRoleDirectives loads and emits operator-provided role directives.
 // These come from the directive file layout (town-level and/or rig-level)
 // and override formula defaults where they conflict.
-func outputRoleDirectives(ctx RoleContext) {
+//
+// w and explainEnabled are injected so tests can capture output without
+// mutating os.Stdout or the primeExplain global (avoiding data races
+// under t.Parallel).
+func outputRoleDirectives(ctx RoleContext, w io.Writer, explainEnabled bool) {
 	role := string(ctx.Role)
 	townRoot := ctx.TownRoot
 	rigName := ctx.Rig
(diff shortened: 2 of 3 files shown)


quad341:
This fix would unblock 3 PRs currently failing CI with the same `captureStdout` data race (#3151, #3152, #3155). CI is all green on this PR. Would be great to get this merged.

htristan:
Agreed - @steveyegge  - just flagging this here for you as it will unblock all the CI issues that we are seeing on other PRs currently.

Issue #2814, opened 2026-03-15
Federated completions are vulnerable to silent tool-call confabulation


smledbetter:
### Problem

This is a Wasteland-layer concern, but filing here since issues are disabled on the Wasteland repo.

When LLMs perform tool-calling tasks (e.g., checking dependencies against a vulnerability database), a subset of models silently skip tool calls and substitute results from training knowledge. In a controlled study across 9 models and 686 trials, 3 of 8 models confabulated at rates between 25% and 80% on the primary test condition.

This is a tier 2/3 model problem — the 5 frontier and mid-tier models tested (Sonnet, Flash, Haiku, GPT-4o Mini, Llama 70B) showed 0–5% confabulation. The confabulators are DeepSeek R1 (47.5%), DeepSeek V3 (45.0%), and Mercury 2 (50.0%). In a single-rig setup you can just avoid those models.

In federation you can't. Wasteland doesn't control which model is behind the rig that submitted work to a wanted board. Cost pressure at scale selects for cheaper models — and R1's partial-completion confabulation mixes real and fabricated results in a way that passes casual inspection. Worse, R1's confabulation rate is unstable across runs (60% in one batch, 22% in the next under identical conditions), so even benchmarking doesn't reliably predict it.

In a two-rig simulation, 82% of trials with confabulated input propagated fabricated data to the downstream agent — and in 73% of those, the downstream agent elaborated on the fabrication (inventing CVE numbers, patch versions, and remediation priorities for vulnerabilities that were never checked). The stamp system catches bad completions after the fact but can't detect confabulation at the point of origin.

### Proposed solution

HMAC-signed receipt gates attached to tool call responses. A deterministic verifier checks claimed results against a cryptographic receipt log before completions cross rig boundaries. In the same study:

- Receipt-gated propagation: **0%** (vs 82% ungated, p=0.00022)
- Mercury 2 confabulation with receipt+gate: **6.7%** (vs 80% baseline, p<0.0000004)

This fits naturally alongside the existing hash-based verification in `internal/inference/` and could integrate with the stamps system to penalize confabulated completions.

### Design doc

I've drafted a design proposal for where the receipt gate components would fit in Wasteland (completion verification, receipt signing, gate, stamps integration) and can submit it as a PR. The full study with methodology, raw data (686 trials), and experiment code is here: [Receipt-Gated Pipelines](https://github.com/smledbetter/receipt-gated-pipelines)

Happy to discuss the approach before submitting code.

[...]

PR #3142, opened 2026-03-22
design: receipt-gated completions for tool-call verification

smledbetter:
**Requested labels:** `area/wasteland`, `type/security` (per @DreadPirateRobertz in #2814 — labels don't exist yet)

## Summary

- Design doc proposing HMAC-signed receipt gates on tool-call completions in the Wasteland federation layer
- Gate fires at wanted board acceptance (before propagation), opt-in per board via `requiresToolReceipts` flag
- Stamps integration feeds gate verdicts into rig reputation for economic pressure on cost-cutting model choices

Addresses #2814. Research data: [receipt-gated-pipelines](https://github.com/smledbetter/receipt-gated-pipelines) (686 trials, 9 models, $15.42).

## Scope

This is a design proposal for review, not implementation. Three components:

1. **ToolShim receipt signing** — HMAC-SHA256 on tool call responses (integration point TBD by maintainer)
2. **Board-level receipt gate** — verification at wanted board acceptance with `requiresToolReceipts` flag
3. **Stamps integration** — confabulation verdicts feed into trust tier demotion

## What migrates to Wasteland

ToolShim and receipt prompts stay in Gastown. Gate, schema changes, stamps integration, and CLI flags move to Wasteland once issues are re-enabled.

### DIFF
--- a/docs/design/receipt-gated-completions.md
+++ b/docs/design/receipt-gated-completions.md
@@ -1,0 +1,360 @@
+# Receipt-Gated Completions for Federated Tool-Call Verification
+
+> **Date:** 2026-03-21
+> **Author:** smledbetter
+> **Status:** Proposal
+> **Issue:** [steveyegge/gastown#2814](https://github.com/steveyegge/gastown/issues/2814)
+> **Labels:** `area/wasteland`, `type/security`
+> **Related:** [Receipt-Gated Pipelines](https://github.com/smledbetter/receipt-gated-pipelines) (study),
+> `internal/wasteland/trust.go` (tier escalation), `internal/wasteland/spider.go` (collusion detection),
+> model-aware-molecules (capability routing)
+
+---
+
+## 1. Problem Statement
+
+When a rig submits a completion to the wanted board, there is no verification that
+tool calls claimed in the completion actually happened. A rig running a cheap model
+can submit a completion where the model skipped tool calls and fabricated results
+from parametric knowledge — and the fabrication looks like success.
+
+This is an NDI failure. The Wasteland assumes nondeterministic processes produce
+useful outcomes through orchestration and oversight. But tool-call confabulation
+breaks the feedback loop: stamps and trust tiers can only correct what they can
+observe, and confabulated completions are indistinguishable from correct ones by
+inspection. The oversight layer needs a ground-truth signal.
+
+Three of eight models tested confabulate tool results at 25-70% rates. The
+fabricated entries are often real facts drawn from training data, just never
+verified through the tool — completeness checks pass, spot-checking passes. In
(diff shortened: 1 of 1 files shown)


steveyegge:
Closing this — appreciate the research and the thorough writeup, but we don't think the design is viable as proposed. Detailed feedback below so you can see our reasoning.

## The core issue: invocation ≠ fidelity

The receipt gate verifies that a tool was *called*, but confabulation is about whether the model *used the tool's output faithfully*. These are fundamentally different questions, and the gap between them is where the dangerous confabulation lives.

The pattern the gate catches: model skips the tool call entirely and fabricates results. This is the dumb version.

The pattern the gate misses: model calls the tool, gets a result it doesn't like or doesn't understand, substitutes parametric knowledge, and echoes the valid `_call_id`. The gate sees a legitimate receipt and stamps VERIFIED on confabulated output.

This second pattern gets *worse* as models get smarter. Dumb models skip tool calls (caught). Smart models call the tool and selectively ignore results (not caught). The gate's effectiveness is inversely correlated with model capability — the opposite of what you want from a security mechanism. Your own data shows receipt prompts still leave 13% confabulation with Mercury — and those 13% would pass the gate as VERIFIED if the model echoes the call_id.

## The trust model is circular

Three problems that are actually one problem:

1. The rig can't hold its own HMAC signing key (it could forge receipts)
2. So the ToolShim must be external to the rig
3. But the doc doesn't commit to where ("integration point TBD")

You can't evaluate the security properties without knowing where the shim sits. And if it must be external, you're proposing infrastructure that intercepts every tool call at a layer the agent can't reach — that's a major architectural commitment that the doc treats as a detail.

## PARTIAL_CONFAB defeats the stated purpose

The doc says receipt gates catch confabulation "before propagation." But PARTIAL_CONFAB — a mix of verified and confabulated results — is accepted with a warning tag. The confabulated data still enters the commons. That's "after propagation with a label," not "before."

## The honest escape hatch has no stable equilibrium

The `unverified` field creates a perverse incentive. Penalize SELF_REPORTED_UNVERIFIED and models stop self-reporting. Don't penalize it and models game it by marking some items unverified to appear honest while confabulating the rest.

## What would actually work

The real problem — verifying output *fidelity*, not just invocation — likely requires structured tool output schemas and claim-level provenance tracking. Each specific claim in the final report would need to be bound to a specific field in a specific tool response, with hash verification at the claim level, not the call level. That's a much harder problem and may not be practical at the orchestration layer.

We'd rather see this problem explored further in research before committing to a design direction. The confabulation data you gathered is genuinely useful — the 686-trial study is good work. But the receipt mechanism doesn't close the gap it's designed to close.

PR #3152, opened 2026-03-22
feat: add push_strategy config — fork push + GitHub PR workflow

quad341:
## Summary

Adds `push_strategy: fork` to `TownSettings` so polecats can push to a fork remote and create GitHub PRs instead of attempting to push to origin.

**Issue:** gas-ddg

**Changes:**
- `internal/config/types.go`: Add `PushStrategy` field + `PushStrategyFork = "fork"` constant
- `internal/cmd/done.go`: Load `push_strategy` from town settings; add `runForkPushStrategy()` that sets up fork remote via `gh repo fork --clone=false`, pushes branch, resolves fork owner for `owner:branch` head ref, creates PR via `gh pr create`
- `internal/polecat/manager.go`: Add `setupForkRemoteInWorktree()` helper; call at spawn time in both `addWithOptionsLocked` and `AddWithOptions` when `push_strategy: fork` is set
- `internal/cmd/tap_guard.go`: Allow `gh pr create` when `push_strategy: fork` configured; add `isPushStrategyFork()` helper

## Usage

Set in `settings/config.json`:
```json
{
  "push_strategy": "fork"
}
```

When set, `gt done` pushes to the `fork` remote (set up by `gt sling` or `gh repo fork`) and creates a GitHub PR instead of creating an internal MR bead.

## Test plan
- [ ] Build passes (`go build ./...`)
- [ ] Config tests pass (`go test ./internal/config/...`)
- [ ] Pre-existing test failures unaffected by this change

### DIFF
--- a/internal/cmd/done.go
+++ b/internal/cmd/done.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strconv"
 	"strings"
@@ -344,6 +345,12 @@ func runDone(cmd *cobra.Command, args []string) (retErr error) {
 		defaultBranch = rigCfg.DefaultBranch
 	}
 
+	// Load push strategy from town settings (empty = default MR workflow).
+	pushStrategy := ""
+	if townSettings, err := config.LoadOrCreateTownSettings(config.TownSettingsPath(townRoot)); err == nil {
+		pushStrategy = townSettings.PushStrategy
+	}
+
 	// For COMPLETED, we need an issue ID and branch must not be the default branch
 	var mrID string
 	var pushFailed bool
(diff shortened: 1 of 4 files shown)


steveyegge:
Thanks for this, Jim — the fork push workflow is a real need and the implementation works. But we're going to close this in favor of a more general approach.

The submission path in `gt done` is already a multi-strategy state machine (MR beads, convoy, etc.), and adding fork as another hardcoded branch doesn't scale well. If someone later needs a GitLab push strategy, or a "push to staging remote" strategy, they'd have to add another block next to the fork one.

What we'd like instead is a **pluggable push strategy framework** — something like:

- A `PushStrategy` interface with `Push()` and `Submit()` methods
- Per-rig configuration (the current town-wide setting is too coarse — you'd want some rigs forking and others pushing direct)
- Tap guard integration with the strategy, rather than the strategy bypassing the guard
- The done command dispatching to the strategy without fork-specific knowledge

We'd happily accept a contribution for that framework. The fork workflow would be the first concrete implementation, and the current MR bead path would be refactored into the default strategy. That way the fork logic you've already written largely survives — it just gets wrapped in an interface.

A few other things to address if you pick this up:
- Tests for the push/submit failure paths
- Dedup the `setupForkRemoteInWorktree` calls (it's identical in both `addWithOptionsLocked` and `AddWithOptions`)
- The `isPushStrategyFork()` tap guard bypass needs to check strategy-level permissions rather than short-circuiting entirely

quad341:
good feedback; i'll make it so. or more formally, claude will make it so

Issue #3153, opened 2026-03-22
Bug: --agent override skips --settings flag, Claude Code hooks never fire for polecats


htristan:
## Summary

When `gt sling` is called with `--agent` (e.g., `--agent claude-sonnet`), the polecat's Claude Code instance launches without the `--settings` flag. Claude Code can't find `.claude/settings.json` (which lives at the shared `polecats/` level, not inside the git worktree), so **SessionStart hooks never fire**.

This means `gt prime --hook` never runs, the agent doesn't get role context injected, and `GT_AGENT_READY` is never set. The session falls back to the startup beacon prompt.

## Root Cause

`polecat_spawn.go` pre-builds the startup command via `BuildPolecatStartupCommandWithAgentOverride` when `--agent` is specified. This calls `BuildStartupCommandWithAgentOverride` with a non-empty `agentOverride`, which resolves the agent config through `ResolveAgentConfigWithOverride` — a path that **does not call `withRoleSettingsFlag`**.

The non-override path goes through `ResolveRoleAgentConfig`, which does call `withRoleSettingsFlag` and correctly adds `--settings`.

Additionally, since `startOpts.Command` is pre-set by `polecat_spawn.go`, the session manager's own `BuildStartupCommandFromConfig` (which would add `--settings`) is skipped entirely.

## Impact

- Affects all polecats launched with an explicit `--agent` flag
- SessionStart/PreCompact/UserPromptSubmit hooks silently don't fire
- Polecats don't get `gt prime --hook` context injection at startup
- `GT_AGENT_READY` never set (affects WaitForCommand readiness detection)

## Fix

Apply `withRoleSettingsFlag` in `BuildStartupCommandWithAgentOverride` after all agent resolution paths, with an idempotency guard to avoid double-adding when the non-override path already includes it.

## Affected Files

- `internal/config/loader.go` — `BuildStartupCommandWithAgentOverride`, `withRoleSettingsFlag`
- `internal/cmd/polecat_spawn.go` — builds command with override, sets `startOpts.Command`

PR #3154, opened 2026-03-22
fix: replace screen-scraping with hook-based ready marker for polecat startup (GH#3133)

### LINKED ISSUE #3133: bug: hook-based polecats have zero startup fallback — sit idle indefinitely when hooks fail

htristan:
Closes #3133

## Context

This is a resubmission of #3134, reworked based on @steveyegge's [review feedback](https://github.com/steveyegge/gastown/pull/3134#issuecomment-2743971625). The original PR correctly identified the problem — hook-based polecats have zero startup fallback and sit idle indefinitely when hooks fail silently — but solved it with screen-scraping heuristics that violate ZFC principles.

This second attempt replaces all heuristic-based detection with a structural, hook-based health check.

## What changed from #3134

**Kept** (per Steve's review):
- Ungating of startup verification so it runs for hook-based agents, not just nudge-based
- `AgentState` and `HasHookedWork` fields on `StalledResult` (structural state from beads)

**Removed** (per Steve's review):
- `containsPromptIndicator` heuristic tightening — still screen-scraping regardless of accuracy
- `PaneSnapshot` on `StalledResult` — couples Witness to agent TUI layout
- `classifyStallType` — heuristic stall classification based on pane content

**Added** (the structural replacement):
- **Ready marker files**: `gt prime --hook` writes `.runtime/ready/<session>.ready` on success, following the existing heartbeat file pattern. `verifyHookHealth` polls for this marker with a timeout and sends a recovery nudge if it's missing.
- **`ResolveCurrentSession`**: New `Tmux` method that walks the process ancestry PID chain and matches against pane PIDs on the named tmux socket. This is needed because Claude Code strips `$TMUX` and `$TMUX_PANE` from hook subprocesses — standard tmux env detection doesn't work.
- **Marker lifecycle**: Cleanup in session stop, `gt done`, and daemon shutdown paths alongside existing heartbeat cleanup.

## How it works

```
SessionStart hook fires
  → gt prime --hook runs
  → ResolveCurrentSession() walks PID ancestry to find our tmux session
  → WriteReadyMarker() writes .runtime/ready/<session>.ready
  → signalAgentReady() sets GT_AGENT_READY=1 in tmux env

Meanwhile, session_manager:
  → verifyHookHealth() polls for marker file with configurable timeout
  → Marker found → hooks succeeded, return immediately
  → Marker missing after timeout → send recovery nudge
```

One structural signal (file exists = hooks ran), no screen-scraping, no TUI coupling.

## Test plan

- `go test ./internal/polecat/ -run TestWriteReadyMarker` — marker write/read/exists/remove/staleness
- `go test ./internal/polecat/ -run TestVerifyHookHealth` — integration: marker present, missing, mid-poll, nudge error
- `go test ./internal/witness/ -run TestStalledResult` — structural fields + zero values
- `go test ./internal/polecat/ ./internal/witness/` — full package suite, no regressions
- Live tested: slung polecats with `gt sling`, confirmed ready markers written and hook health verified

## Related

- #3153 — separate bug discovered during testing: `--agent` override missing `--settings` flag (filed separately per ZFC separation of concerns)

### DIFF
--- a/internal/cmd/prime.go
+++ b/internal/cmd/prime.go
@@ -17,6 +17,7 @@ import (
 	"github.com/steveyegge/gastown/internal/beads"
 	"github.com/steveyegge/gastown/internal/cli"
 	"github.com/steveyegge/gastown/internal/lock"
+	"github.com/steveyegge/gastown/internal/polecat"
 	"github.com/steveyegge/gastown/internal/state"
 	"github.com/steveyegge/gastown/internal/style"
 	"github.com/steveyegge/gastown/internal/telemetry"
@@ -306,6 +307,15 @@ func handlePrimeHookMode(townRoot, cwd string) {
 	// remains "bash" even though the agent is running as a descendant.
 	signalAgentReady()
 
+	// Write ready marker so the SessionManager can verify hooks succeeded
+	// without screen-scraping. See GH#3133.
+	// Prefer tmux session name from GT_SESSION env var (set by SessionManager
+	// via tmux set-environment) since $TMUX may not be inherited by hook
+	// subprocesses in all agent runtimes.
+	if markerSession := resolveMarkerSessionName(); markerSession != "" {
+		_ = polecat.WriteReadyMarker(townRoot, markerSession)
+	}
+
 	// Store source for compact/resume detection in runPrime
 	primeHookSource = source
 
@@ -330,23 +340,31 @@ func hookSessionBeaconLines(sessionID, source string) []string {
 	return lines
 }
 
+// resolveMarkerSessionName returns the tmux session name for writing the ready
+// marker. Uses ResolveCurrentSession which matches our TTY against tmux panes
+// on the town socket — works even when Claude Code strips $TMUX from hooks.
+func resolveMarkerSessionName() string {
+	t := tmux.NewTmux()
+	name, err := t.ResolveCurrentSession()
+	if err != nil {
+		return ""
+	}
+	return name
+}
+
 // signalAgentReady sets GT_AGENT_READY=1 in the current tmux session environment.
 // Called from the agent's SessionStart hook to signal that the agent has started.
 // WaitForCommand polls for this variable as a ZFC-compliant alternative to
-// probing the process tree via IsAgentAlive. No-op when not in a tmux session.
+// probing the process tree via IsAgentAlive.
+// Uses ResolveCurrentSession to find our session on the town socket — raw
+// exec.Command("tmux", ...) would use the default socket and miss the gastown server.
 func signalAgentReady() {
-	if os.Getenv("TMUX") == "" {
+	t := tmux.NewTmux()
+	name, err := t.ResolveCurrentSession()
+	if err != nil || name == "" {
 		return
 	}
-	out, err := exec.Command("tmux", "display-message", "-p", "#{session_name}").Output()
-	if err != nil {
-		return
-	}
-	session := strings.TrimSpace(string(out))
-	if session == "" {
-		return
-	}
-	_ = exec.Command("tmux", "set-environment", "-t", session, tmux.EnvAgentReady, "1").Run()
+	_ = t.SetEnvironment(name, tmux.EnvAgentReady, "1")
 }
 
 // isCompactResume returns true if the current prime is running after compaction or resume.
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -2299,8 +2299,9 @@ func (d *Daemon) killIdlePolecat(rigName, polecatName, sessionName string, idleD
 		return
 	}
 
-	// Clean up heartbeat file
+	// Clean up heartbeat and ready marker files
 	polecat.RemoveSessionHeartbeat(d.config.TownRoot, sessionName)
+	polecat.RemoveReadyMarker(d.config.TownRoot, sessionName)
 
 	d.logger.Printf("Reaped idle polecat %s/%s — session killed, API slot freed", rigName, polecatName)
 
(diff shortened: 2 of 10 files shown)


htristan:
**CI note:** The Windows Build failure is unrelated to this PR — it's the pre-existing `TestOutputRoleDirectives` bug tracked in #3140. None of the files this PR touches are involved. All other checks (Test, Integration Tests, Lint) pass.

steveyegge:
Thanks for the rework — the structural approach is exactly what we asked for in the #3134 review, and the code quality is solid.

We cherry-picked the parts that carry their weight:

- **`ResolveCurrentSession()`** — genuinely valuable, fixes the bug where Claude Code strips `$TMUX` from hook subprocesses
- **Improved `signalAgentReady()`** — now uses the tmux abstraction and ResolveCurrentSession instead of raw exec.Command
- **`StalledResult` structural fields** — useful for witness diagnostics

What we didn't take: the `.runtime/ready/` marker file system. The existing `GT_AGENT_READY` tmux env var (which your improved `signalAgentReady` now sets correctly) provides the same "did hooks run?" signal without the file lifecycle management across 5 teardown paths. The only remediation for a missing marker was sending a nudge — and checking the tmux env var is sufficient to gate that.

Cherry-picked in 4bdfd637. Thanks for the contribution.

htristan:
@steveyegge Thanks for the cherry-pick and the credit — appreciate it.

One thing I wanted to flag for your take: the ungating of `verifyStartupNudgeDelivery` didn't make it into the cherry-pick. On main it's still behind `if fallbackInfo.SendStartupNudge`, which is false for hook+prompt agents (Claude Code).

With `signalAgentReady()` now working correctly, `GT_AGENT_READY` gets set when hooks succeed — that's the signal fix. But if hooks fail silently (the original #3133 scenario), nothing actively detects that during startup for hook-based agents. The verification loop is still gated off for them, and the only backstop is the Witness zombie patrol on a longer cycle.

Is the intent that the Witness handles this case, or would it make sense to also ungate the verification (or wire up a `GT_AGENT_READY` check) so there's an active startup recovery path for hook-based agents? Happy to put up a follow-up PR if useful.

PR #3156, opened 2026-03-22
fix: apply --settings flag for Claude agents launched with --agent override

### LINKED ISSUE #3153: Bug: --agent override skips --settings flag, Claude Code hooks never fire for polecats

htristan:
## Summary

Fixes #3153

- When `gt sling` uses `--agent` (e.g., `--agent claude-sonnet`), the polecat's Claude Code instance launched without `--settings`, so `SessionStart` hooks never fired and `gt prime --hook` context injection was silently skipped.
- Root cause: `BuildStartupCommandWithAgentOverride` resolves through `ResolveAgentConfigWithOverride`, which does not call `withRoleSettingsFlag`. The non-override path (`ResolveRoleAgentConfig`) does.
- Fix: call `withRoleSettingsFlag` in `BuildStartupCommandWithAgentOverride` after all resolution paths, with an idempotency guard to prevent double-adding when the non-override path already includes it.

## Changes

- **`internal/config/loader.go`**: Add idempotency guard to `withRoleSettingsFlag` (scan for existing `--settings` before appending); call `withRoleSettingsFlag` in `BuildStartupCommandWithAgentOverride` after agent resolution.
- **`internal/config/loader_test.go`**: 5 new tests covering idempotency, Claude override gets `--settings`, polecat override scenario (the actual bug), non-Claude override skipped, and no double-add regression on the non-override path.

## Test plan

- [x] `TestWithRoleSettingsFlag_IdempotencyGuard` — calling twice doesn't double-add `--settings`
- [x] `TestBuildStartupCommandWithAgentOverride_SettingsFlagForClaudeOverride` — Claude custom agent override produces correct `--settings` path
- [x] `TestBuildPolecatStartupCommandWithAgentOverride_IncludesSettingsFlag` — the actual bug scenario
- [x] `TestBuildStartupCommandWithAgentOverride_NoSettingsFlagForNonClaude` — non-Claude override correctly omitted
- [x] `TestBuildStartupCommandWithAgentOverride_NoDoubleSettingsOnNonOverridePath` — non-override path still gets exactly 1 `--settings`
- [x] Full `go test ./internal/config/` passes (including integration tests)
- [x] `go build ./...` and `go vet` clean

### DIFF
--- a/internal/config/loader.go
+++ b/internal/config/loader.go
@@ -1428,6 +1428,13 @@ func withRoleSettingsFlag(rc *RuntimeConfig, role, rigPath string) *RuntimeConfi
 		return rc
 	}
 
+	// Guard against double-adding (ResolveRoleAgentConfig already calls this)
+	for _, arg := range rc.Args {
+		if arg == "--settings" {
+			return rc
+		}
+	}
+
 	settingsDir := RoleSettingsDir(role, rigPath)
 	if settingsDir == "" {
 		return rc
@@ -2311,6 +2318,13 @@ func BuildStartupCommandWithAgentOverride(envVars map[string]string, rigPath, pr
 			}
 		}
 	}
+
+	// Ensure Claude agents get --settings when their settings directory
+	// differs from the session working directory. This must run for ALL
+	// resolution paths (including agent overrides) — previously only the
+	// non-override ResolveRoleAgentConfig path included it, causing hooks
+	// to silently not fire for polecats launched with --agent.
+	rc = withRoleSettingsFlag(rc, role, rigPath)
 
 	// Apply exec wrapper from rig/town settings if not already set on the resolved config.
 	if len(rc.ExecWrapper) == 0 {
--- a/internal/config/loader_test.go
+++ b/internal/config/loader_test.go
@@ -5429,3 +5429,160 @@ func TestBuildStartupCommandWithAgentOverride_ExecWrapper(t *testing.T) {
 		t.Errorf("expected wrapper immediately before claude command, got: %q", cmd)
 	}
 }
+
+// --- Tests for GH#3153: --agent override skips --settings flag ---
+
+func TestWithRoleSettingsFlag_IdempotencyGuard(t *testing.T) {
+	t.Parallel()
+	rigPath := "/fake/town/myrig"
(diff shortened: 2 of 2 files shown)


steveyegge:
Fix-merged to main via e0ede7dc (rebased to 1faf78ee). Thanks @htristan — clean fix with solid test coverage. The idempotency guard and the withRoleSettingsFlag call after all resolution paths are exactly right.

PR #3159, opened 2026-03-22
fix(convoy): use Unix epoch instead of zero time for initial event poll

etrapeznikov:
## Summary

- Fixes recurring `Error 1366 (HY000): +Inf is not a valid value for double` in convoy event polling
- Root cause: when no previous high-water mark exists, `pollStore` passes Go's zero `time.Time` (0001-01-01) to `GetAllEventsSince`, which the Dolt SQL driver converts to `+Inf` as a float parameter
- Fix: initialize `highWater` to Unix epoch (1970-01-01 UTC) instead of zero time — safe for all SQL backends and semantically equivalent for event polling

## Test plan

- [ ] Start daemon with no previous convoy event state (fresh start)
- [ ] Verify no `+Inf` errors in daemon logs during initial event poll
- [ ] Verify convoy event polling works normally after fix
- [ ] Verify existing high-water marks (non-zero) continue to work

### DIFF
--- a/internal/daemon/convoy_manager.go
+++ b/internal/daemon/convoy_manager.go
@@ -261,8 +261,12 @@ func (m *ConvoyManager) pollStoresSnapshot(stores map[string]beadsdk.Storage) bo
 // The seen set deduplicates issueIDs across stores within a poll cycle.
 // Returns an error if the poll failed (used by caller for backoff decisions).
 func (m *ConvoyManager) pollStore(name string, store beadsdk.Storage, stores map[string]beadsdk.Storage, seen map[string]bool) error {
-	// Load per-store high-water mark
-	var highWater time.Time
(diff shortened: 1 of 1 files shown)


htristan:
For context, this looks like a residual edge case from the convoy UUID→time polling migration tracked in #3076. The main fix (switching from integer event IDs to time-based cursors) already landed on main via the beads v0.61 upgrade — see steveyegge's comment on #3079 confirming that.

This PR addresses a remaining gap: when there's no prior high-water mark (fresh start), the default `time.Time` zero value (0001-01-01) doesn't serialize through Dolt's SQL driver, producing the `+Inf` error. Initializing to Unix epoch closes that edge case.

Linking these together for anyone triaging: #3076 → beads v0.61 upgrade (on main) → this PR (initial poll edge case).

PR #3160, opened 2026-03-22
feat(handoff): prompt user before restarting session

quad341:
## Summary
- Adds confirmation prompt to `gt handoff` so the mayor cannot restart the session without user consent
- Adds `--yes/-y` flag for automation/scripting to skip the prompt
- Updates the `/handoff` skill template to ask "Ready to hand off?" before proceeding

Fixes gas-6z0

## Test plan
- [ ] Run `gt handoff` and verify it prompts before restarting
- [ ] Run `gt handoff --yes` and verify it skips the prompt
- [ ] Verify the `/handoff` skill template includes the confirmation step

### DIFF
--- a/internal/cmd/handoff.go
+++ b/internal/cmd/handoff.go
@@ -162,7 +162,7 @@ func runHandoff(cmd *cobra.Command, args []string) error {
 	// without warning the user who may be actively reading or working (gas-6z0).
 	if !handoffYes && !handoffDryRun {
 		if !promptYesNo("Ready to hand off? This will restart the session.") {
-			fmt.Println("Handoff cancelled.")
+			fmt.Println("Handoff canceled.")
 			return nil
 		}
 	}


quad341:
The `TestOutputRoleDirectives` Windows CI failure is pre-existing — fixed by #3141.

quad341:
## gas-keb Review (Opus)

### Summary

Migration of gt from shelling out to `bd` CLI to using the `beads` Go module directly. Two commits (dff20d3a4, 6e4492beb) touching 7 files across `internal/beads/` and `internal/cmd/`.

**What was migrated to Go module calls:**
- `List()` / `listEphemeral()` → `store.SearchIssues()` (eliminates both `bd list` and `bd query` shell-outs)
- `ListMergeRequests()` → `store.SearchIssues()` with `Ephemeral=nil` (replaces the SQL join shell-out)
- `Show()` → `store.GetIssue()` with CLI fallback
- `FindLatestIssueByTitleAndAssignee()` → `store.SearchIssues()`
- `ShowMultiple()` → `store.GetIssuesByIDs()`
- `Create()` → `store.CreateIssue()` (with CLI fallback for parent-child)
- `CreateWithID()` → `store.CreateIssue()`
- `Search()` → `store.SearchIssues()`
- `Update()` → `store.UpdateIssue()` + label operations
- `Close()` / `CloseWithReason()` → `store.CloseIssue()`
- `ReleaseWithReason()` → `store.UpdateIssue()`
- `AddDependency()` / `RemoveDependency()` → `store.AddDependency()` / `store.RemoveDependency()`
- `Stats()` → `store.GetStatistics()`
- `UpdateAgentState()` → `store.UpdateIssue()` (replaces `bd agent state` and `runWithRouting`)
- `AddDelegation()` / `RemoveDelegation()` / `GetDelegation()` / `ListDelegationsFrom()` → metadata-based via `store.GetIssue()` / `store.UpdateIssue()`
- `MergeSlot*` operations → `store.GetIssue()` / `store.UpdateIssue()` / `store.CreateIssue()`
- `SlingContext*` operations → `store.CreateIssue()` / `store.SearchIssues()` / `store.CloseIssue()`
- `CreateAgentBead` hook_bead set → `store.UpdateIssue()` (replaces `bd slot set`)
- `deacon.go` `updateAgentBeadState()` → `bd.UpdateAgentState()` (replaces `exec.Command("bd", "agent", "state", ...)`)
- `done.go` agent state update → `bd.UpdateAgentState()` (replaces `bd.Run("agent", "state", ...)`)

**What still shells out (intentionally):**
- `Ready()` / `ReadyForMol()` / `ReadyWithType()` — `bd ready` has complex blocking semantics
- `Blocked()` — `bd blocked`
- `ForceCloseWithReason()` — needs `--force` flag not in Storage API
- `CreateAgentBead()` — uses `bd create` with `--type=agent --force` flags
- `CreateOrReopenAgentBead()` — uses `bd reopen` and `bd update --type=agent`
- `ListAgentBeads()` / `ListAgentBeadsFromWisps()` / `ListWispIDs()` — uses `bd list --include-infra` and `bd mol wisp list`
- `Init()` — `bd init`
- `createViaShellout()` — parent-child ID generation needs DoltStorage
- `Run()` — public escape hatch

### Issues Found

**1. Cross-rig routing lost on `UpdateAgentState` (medium severity)**
The old code used `runWithRouting()` which strips `BEADS_DIR` to let bd's native prefix-based routing via `routes.jsonl` resolve cross-prefix agent beads (e.g., `wa-*` agent beads from hq context). The new code uses `b.openStore()` which resolves to the local rig's database. If `UpdateAgentState` is called for an agent bead that lives in a different rig's database, the update will either fail silently or update the wrong database. The `deacon.go` callsite (`updateAgentBeadState`) creates `beads.New(townRoot)` and then calls `UpdateAgentState` for arbitrary agent bead IDs — these may route to different rigs. Compare with `ResetAgentBeadForReuse` and `CreateOrReopenAgentBead` which both do explicit `ResolveRoutingTarget` + `NewWithBeadsDir` for cross-rig handling.

**2. `openStore` error caching is permanent (low-medium severity)**
`openStore()` caches `storeErr` on first failure and never retries. If the Dolt server is temporarily unavailable during init (e.g., server restart), all subsequent operations on that `Beads` instance will fail without retry. The `bd` CLI shell-out path doesn't have this problem — each call is independent. For long-lived `Beads` instances (e.g., in the deacon or witness), a transient Dolt hiccup at startup permanently disables the Go module path. Consider clearing `storeErr` on retry or adding a TTL.

**3. Duplicate comment on `Blocked()` (trivial)**
Line 1097-1098: `// Blocked returns issues that are blocked by dependencies.` appears twice.

**4. `MergeSlotAcquire` / `MergeSlotRelease` have no atomicity guarantees (pre-existing, worth noting)**
The acquire/release operations do separate GetIssue + UpdateIssue without transactions. Two callers could both see `Holder==""` and both "acquire" the slot. This is pre-existing (the old CLI had the same race), but now that the code uses the Go module directly, it would be possible to wrap these in a transaction. Not a regression, just an opportunity.

**5. `issueFromModule` doesn't map `Holder` or `Waiters` fields**
The `issueFromModule` converter doesn't copy `Holder`/`Waiters` to the gastown `Issue` type (which doesn't have those fields). This works because `beads_merge_slot.go` uses `beadsmod.Issue` directly from `store.GetIssue()`, not the converted gastown `Issue`. But it means the gastown `Issue` type diverges from `beadsmod.Issue` — any code that converts first and then tries to read merge slot fields will get zero values. Not a bug today but a latent issue.

**6. `Show()` has inconsistent fallback behavior**
When `openStore()` succeeds but `GetIssue()` returns a non-"not found" error, the method returns that error directly without trying the CLI fallback. But when `openStore()` itself fails, it falls through to the CLI. This means a Dolt query error (e.g., table corruption) is harder to recover from than a Dolt connection error. The comment says "Fallback: store unavailable (isolated mode or schema mismatch)" but connection errors aren't the only failure mode.

### Build & Test Status

- `go build ./...` — PASS
- `go test ./internal/beads/ -count=1` — PASS (0.354s)

### Verdict: approve (with one follow-up)

The migration is solid and well-scoped. The Go module calls are correct, the type conversions are handled properly, the `issueFromModule` converter covers the important fields, and the intentional shell-out residuals are clearly documented (parent-child creates, ready/blocked queries, force-close).

Issue #1 (cross-rig routing on `UpdateAgentState`) should be addressed in a follow-up since it's a functional regression for the deacon's `updateAgentBeadState` path. The fix is straightforward — add `ResolveRoutingTarget` + `NewWithBeadsDir` at the top of `UpdateAgentState`, matching the pattern already used by `ResetAgentBeadForReuse`.

Issue #2 (permanent error caching) is worth a follow-up ticket but not blocking.

quad341:
Note: The gas-keb review comment above is stale — those commits were moved to PR #3166 (and will be superseded by the gas-a1c comprehensive migration). This PR is only the handoff confirmation feature (gas-6z0).

steveyegge:
Fix-merged to main in a4611422. Added a TTY guard (term.IsTerminal on stdin) so the prompt only fires for interactive users — agent automation proceeds without blocking. Thanks for the contribution!

PR #3161, opened 2026-03-22
fix(beads): replace bd agent state with description update

quad341:
## Summary
- `bd agent state` command was removed in bd v0.62 (commit 0bd598ce), breaking `UpdateAgentState` in gt
- Every polecat sling retried 8 times with exponential backoff (~1 min delay + noisy warnings)
- Fix: `UpdateAgentState` now does Show→Parse→Modify→Update on the description field, matching how other agent field updates already work
- Includes cross-rig routing support for agent beads in different databases

Fixes gas-5a7

## Test plan
- [x] `go test ./internal/beads/ -run Agent` — all agent tests pass
- [ ] `gt sling` a bead — verify no more `bd agent state` retry warnings

### DIFF
--- a/internal/beads/beads_agent.go
+++ b/internal/beads/beads_agent.go
@@ -416,18 +416,30 @@ func (b *Beads) ResetAgentBeadForReuse(id, reason string) error {
 }
 
 // UpdateAgentState updates the agent_state field in an agent bead.
-// Uses `bd agent state` command for the database column directly.
+// Uses description field update since bd 0.62 removed the `bd agent state` command
+// and the agent_state DB column (commit 0bd598ce). The read path in GetAgentBead
+// already falls back to parsing agent_state from the description text.
 func (b *Beads) UpdateAgentState(id string, state string) (retErr error) {
 	defer func() { telemetry.RecordAgentStateChange(context.Background(), id, state, nil, retErr) }()
-	// Update agent state using bd agent state command
-	// Use runWithRouting so bd can resolve cross-prefix agent beads (e.g., wa-*
-	// agent beads from hq context) via routes.jsonl instead of BEADS_DIR.
-	_, err := b.runWithRouting("agent", "state", id, state)
-	if err != nil {
-		return fmt.Errorf("updating agent state: %w", err)
+
+	// Resolve routing for cross-rig agent beads (e.g., wa-* from hq context).
+	targetDir := ResolveRoutingTarget(b.getTownRoot(), id, b.getResolvedBeadsDir())
+	target := b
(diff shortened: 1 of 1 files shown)


internal/beads/beads_agent.go
```diff
@@ -416,18 +416,30 @@ func (b *Beads) ResetAgentBeadForReuse(id, reason string) error {
...
 	if err != nil {
-		return fmt.Errorf("updating agent state: %w", err)
+		return fmt.Errorf("updating agent state: show %s: %w", id, err)
 	}
 
-	// Hook slot no longer maintained (hq-l6mm5) — removed hook_bead parameter.
+	fields := ParseAgentFields(issue.Description)
+	fields.AgentState = state
+	description := FormatAgentDescription(issue.Title, fields)
+	if err := target.Update(id, UpdateOptions{Description: &description}); err != nil {
```
htristan, line 440: This Show→Parse→Modify→Update cycle isn't atomic. The old `bd agent state` was a single DB column write — if two callers update the same bead's state concurrently, one write will be silently lost here. Is this field only written by a single actor at a time in practice? If not, this is a behavioral regression from the old approach.

internal/beads/beads_agent.go
```diff
@@ -416,18 +416,30 @@ func (b *Beads) ResetAgentBeadForReuse(id, reason string) error {
...
+// and the agent_state DB column (commit 0bd598ce). The read path in GetAgentBead
+// already falls back to parsing agent_state from the description text.
 func (b *Beads) UpdateAgentState(id string, state string) (retErr error) {
 	defer func() { telemetry.RecordAgentStateChange(context.Background(), id, state, nil, retErr) }()
-	// Update agent state using bd agent state command
-	// Use runWithRouting so bd can resolve cross-prefix agent beads (e.g., wa-*
-	// agent beads from hq context) via routes.jsonl instead of BEADS_DIR.
-	_, err := b.runWithRouting("agent", "state", id, state)
+
+	// Resolve routing for cross-rig agent beads (e.g., wa-* from hq context).
```
htristan, line 425: The cross-rig routing path (where `targetDir != b.getResolvedBeadsDir()`) creates a new `Beads` instance with a different database. Is there test coverage for this path? The old `runWithRouting` encapsulated this logic — now that it's inline, a test asserting correct routing for cross-prefix agent beads (e.g. `wa-*` from hq context) would prevent this from breaking silently if `ResolveRoutingTarget` or `NewWithBeadsDir` change.

quad341:
Closing — superseded by #3166 which migrates agent state to the beads Go module directly, avoiding the description-parsing approach and its atomicity concerns.

Issue #885, opened 2026-01-23
gt doctor --fix writes gastown artifacts into external git worktrees (data loss & repo pollution)


taeunlee:
gt doctor --fix creates gastown artifacts inside git worktrees (mayor/rig/, crew/*/) that connect to
  external repositories. These artifacts become untracked files in the external repo and will leak to remote
  if accidentally committed with git add ..

  Affected checks:
  - priming - Overwrites/creates CLAUDE.md with gastown boilerplate inside worktrees
  - patrol-roles-have-prompts - Creates internal/templates/roles/*.md.tmpl inside worktrees
  - claude-settings - May place .claude/settings.json in worktrees

  Example of the leak:
  my_app/
  ├── .beads/              ← rig-level (safe, gitignored by gt)
  ├── mayor/
  │   └── rig/             ← git worktree → external remote
  │       ├── internal/    ← CREATED BY gt doctor --fix (leaks!)
  │       │   └── templates/roles/
  │       │       ├── deacon.md.tmpl
  │       │       ├── refinery.md.tmpl
  │       │       └── witness.md.tmpl
  │       └── CLAUDE.md    ← OVERWRITTEN (was 158 lines of project guidelines, now 6 lines of "Mayor
  Context")

  Proposed Solution

  gt doctor --fix should place artifacts at rig level, not inside worktrees:
 +----------------------+-------------------------------+-----------------------------------------------+
| Artifact             | Current (wrong)               | Correct                                       |
+----------------------+-------------------------------+-----------------------------------------------+
| Templates            | <rig>/mayor/rig/internal/     | <rig>/internal/                               |
+----------------------+-------------------------------+-----------------------------------------------+
| CLAUDE.md (gastown)  | <rig>/mayor/rig/CLAUDE.md     | <rig>/mayor/CLAUDE.md                         |
|                      |                               | (or skip if external repo)                    |
+----------------------+-------------------------------+-----------------------------------------------+
| PRIME.md             | <rig>/crew/*/.beads/PRIME.md  | Gitignored or placed at rig level             |
+----------------------+-------------------------------+-----------------------------------------------+

  Key principle: Worktrees (mayor/rig/, crew/*/, polecats/*/) are clones of external repos. Gastown should
  NEVER write files into them that aren't part of the external repo's codebase.

  Alternatives Considered

  1. Add gastown artifacts to external repos' .gitignore - Pollutes external repos with gastown knowledge;
  requires coordinating with all repo maintainers
[...]

PR #3164, opened 2026-03-22
fix: patrol molecules use TownRoot instead of rig WorkDir

sottnik:
## Summary

- Witness and refinery patrol molecules were created in the rig's beads database (`ctx.WorkDir`), causing them to appear as rig-prefixed issues (e.g. `wi-*`) mixed in with real project work
- Deacon already correctly used `ctx.TownRoot` (HQ beads) — this fix makes witness and refinery consistent
- 3 files changed, 6 lines: `WorkDir` → `TownRoot` in all witness/refinery `PatrolConfig.BeadsDir` fields

## Related

- #885 (gastown artifacts leaking into external worktrees — same class of bug)

## Test plan

- [ ] Verify witness/refinery patrol molecules get `hq-` prefix instead of rig prefix
- [ ] Verify `bd list --status=open` in a rig no longer shows patrol beads
- [ ] Verify patrol report/new cycle correctly finds and closes patrols in HQ beads

### DIFF
--- a/internal/cmd/patrol_new.go
+++ b/internal/cmd/patrol_new.go
@@ -60,14 +60,14 @@ func runPatrolNew(cmd *cobra.Command, args []string) error {
 		cfg = PatrolConfig{
 			RoleName:      "witness",
 			PatrolMolName: constants.MolWitnessPatrol,
-			BeadsDir:      roleInfo.WorkDir,
+			BeadsDir:      roleInfo.TownRoot,
 			Assignee:      roleInfo.Rig + "/witness",
 		}
 	case RoleRefinery:
 		cfg = PatrolConfig{
 			RoleName:      "refinery",
 			PatrolMolName: constants.MolRefineryPatrol,
-			BeadsDir:      roleInfo.WorkDir,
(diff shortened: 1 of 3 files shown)


internal/cmd/patrol_new.go
```diff
@@ -60,14 +60,14 @@ func runPatrolNew(cmd *cobra.Command, args []string) error {
 		cfg = PatrolConfig{
 			RoleName:      "witness",
 			PatrolMolName: constants.MolWitnessPatrol,
-			BeadsDir:      roleInfo.WorkDir,
+			BeadsDir:      roleInfo.TownRoot,
```
htristan, line 63: Nit: This is the kind of bug that could silently recur if someone adds a new role or copies an existing `PatrolConfig` block using `WorkDir` as a template. A small table-driven test that asserts `BeadsDir == TownRoot` for all non-deacon patrol roles would be cheap insurance.

sottnik: Fair point. I'll leave that for a follow-up, this PR is scoped to the fix itself.

PR #3165, opened 2026-03-22
feat: add vision pipeline with screenshot, OCR, ollama, and visual diff

gagan114662:
## Summary
- **internal/ollama**: Client for Ollama API (DescribeImage, Ping)
- **internal/vision**: CaptureScreenshot (macOS screencapture), OCR (macOS Vision framework via Swift), VisualDiff (pixel-by-pixel PNG comparison)
- **internal/vision/integration_test.go**: End-to-end tests for all 4 pipeline stages (build-tagged integration)

## Files Changed
- `internal/ollama/client.go` — Ollama HTTP client
- `internal/vision/vision.go` — Screenshot, OCR, and visual diff functions
- `internal/vision/integration_test.go` — Integration tests

## Test plan
- [x] All 4 integration tests pass locally (screenshot, OCR, ollama describe, visual diff)
- [ ] CI passes on PR

Executed-By: gastown/polecats/chrome (gt-0s0)

### DIFF
--- a/internal/ollama/client.go
+++ b/internal/ollama/client.go
@@ -1,0 +1,106 @@
+// Package ollama provides a client for the Ollama API.
+package ollama
+
+import (
+	"bytes"
+	"context"
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"os"
+	"time"
+)
+
+const defaultBaseURL = "http://localhost:11434"
+
+// Client communicates with a local Ollama instance.
+type Client struct {
+	BaseURL    string
+	HTTPClient *http.Client
+}
+
+// NewClient returns a Client pointing at the default local Ollama endpoint.
+func NewClient() *Client {
+	return &Client{
+		BaseURL: defaultBaseURL,
+		HTTPClient: &http.Client{
+			Timeout: 300 * time.Second,
+		},
+	}
+}
+
+// GenerateRequest is the payload for /api/generate.
+type GenerateRequest struct {
+	Model  string   `json:"model"`
+	Prompt string   `json:"prompt"`
+	Images []string `json:"images,omitempty"` // base64-encoded
+	Stream bool     `json:"stream"`
+}
+
+// GenerateResponse is the response from /api/generate.
+type GenerateResponse struct {
+	Response string `json:"response"`
+	Done     bool   `json:"done"`
+}
+
+// DescribeImage sends an image file to a vision model and returns the description.
+func (c *Client) DescribeImage(ctx context.Context, model, imagePath, prompt string) (string, error) {
+	imgData, err := os.ReadFile(imagePath)
+	if err != nil {
+		return "", fmt.Errorf("read image %s: %w", imagePath, err)
+	}
+	b64 := base64.StdEncoding.EncodeToString(imgData)
+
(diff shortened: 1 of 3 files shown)


htristan:
Thanks for the contribution! This is an ambitious addition, but I think we need to step back and discuss the approach before this can move forward. I'd recommend **closing this PR and opening an issue first** to align on design, per our [contributing guidelines](https://github.com/steveyegge/gastown/blob/main/CONTRIBUTING.md#what-to-contribute):

> For larger changes, please open an issue first to discuss the approach.

A 377-line new feature spanning two new packages qualifies. Here are the concerns that should be addressed in that discussion:

---

### 1. Hard macOS dependency with no build constraints

`vision.go` shells out to `screencapture` (macOS-only) and runs inline Swift using `import Vision` and `import AppKit` (Apple frameworks). However, the file has **no `//go:build darwin` tag** — meaning this compiles on Linux but fails at runtime. At minimum this needs a build constraint, but the broader question is whether a macOS-only package is the right approach here given that Gas Town agents can run on other platforms.

### 2. Security: command injection in OCR

The `OCR()` function interpolates `imagePath` directly into a Swift source string via `fmt.Sprintf`:

```go
script := fmt.Sprintf(`
...
let url = URL(fileURLWithPath: "%s")
...
`, imagePath)
```

A path containing `"` or `\` breaks the script. A crafted path could inject arbitrary Swift code. This should pass the path as a command-line argument rather than interpolating it into source code.

### 3. Ollama dependency is undocumented and unclear

Ollama is a local inference server (no API key), but this PR introduces it as a hard dependency with no explanation of:
- How to install and run Ollama
- Which models are required (the test hardcodes `gemma3:4b`)
- What happens when Ollama isn't available (no sentinel errors, no `IsAvailable()` pattern)
- How this relates to existing LLM integration paths in the project
- Performance expectations (the 300s timeout suggests this can be very slow)

### 4. No documentation

There's no README, design doc, or setup guide for either new package. The PR body describes *what* was added but not *why* — what problem does this "vision pipeline" solve? Who calls it? How does it fit into the existing architecture? Per our contributing guidelines, new functionality should include documentation.

### 5. ZFC alignment question

Per our [design philosophy](https://github.com/steveyegge/gastown/blob/main/CONTRIBUTING.md#zero-framework-cognition-zfc), Go code provides transport and agents provide cognition. The individual primitives (screenshot, OCR, pixel diff) look like transport, which is fine — but the PR frames this as a "pipeline." The issue discussion should clarify: who orchestrates the pipeline? If it's meant to be agent-driven, how do agents invoke these tools? There's no molecule/formula integration shown.

### 6. Missing tests

- No unit tests for either package — only integration tests behind a build tag
- The `internal/ollama` package has zero tests
- The diff logic and request marshaling could have straightforward unit tests with fixture data

### 7. Minor code issues

- `DiffResult.Identical` uses `meanDelta == 0` — floating-point equality is fragile; should use an epsilon
- `VisualDiff` only supports PNG; no handling for other image formats
- No error wrapping conventions consistent with the rest of the codebase

---

To be clear — the *idea* of giving agents vision capabilities is interesting and worth exploring. But the right next step is an issue where we can discuss the design (cross-platform strategy, architectural fit, Ollama vs alternatives) before writing the implementation. Happy to help shape that discussion.

PR #3171, opened 2026-03-23
fix(plugins): prevent plugin-run beads from accumulating as open wisps

A3Ackerman:
## Summary

- Add `gt plugin record` subcommand wrapping `RecordRun` (create+close atomically)
- Update plugin instructions to use `gt plugin record` instead of raw `bd create`
- Daemon-dispatched dog runs now properly close their beads after reporting

## Problem

Plugin instructions (`plugin.md` and `run.sh`) create beads via `bd create --ephemeral` but never close them. The Go `RecordRun` API properly creates+closes but was only called for manual `gt plugin run`, not daemon-dispatched dog runs. This caused ~145+ open wisps accumulating per cycle.

## Test plan
- [ ] Run a plugin via daemon dispatch and verify bead is created and closed
- [ ] Verify `gt plugin record` CLI works for manual invocation
- [ ] Check HQ inbox after a full daemon cycle — no accumulating open plugin wisps

Closes gt-qfq

### DIFF
--- a/internal/cmd/plugin.go
+++ b/internal/cmd/plugin.go
@@ -18,15 +18,19 @@ import (
 
 // Plugin command flags
 var (
-	pluginListJSON     bool
-	pluginShowJSON     bool
-	pluginRunForce     bool
-	pluginRunDryRun    bool
-	pluginHistoryJSON  bool
-	pluginHistoryLimit int
-	pluginSyncSource   string
-	pluginSyncClean    bool
-	pluginSyncDryRun   bool
+	pluginListJSON      bool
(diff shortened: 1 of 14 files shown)


A3Ackerman:
Closing — investigation (gt-qfq) found the accumulating beads are dispatch mails (gt:message, from:daemon), not plugin-run receipts (type:plugin-run). This PR targets the wrong bead population. Also introduces a regression in quality-review-result by dropping custom labels needed for trend queries. See gt-qfq bead for full root cause analysis.

PR #3187, opened 2026-03-23
fix: use filepath.Join in explain_mode test for Windows compatibility

daniel-jasinski:
## Summary

- Fixes `TestOutputRoleDirectives/explain_mode_shows_file_paths` failing on Windows CI
- The assertion used a hardcoded `"directives/polecat.md"` (forward slash) while `filepath.Join` produces backslashes on Windows
- Fix: use `filepath.Join("directives", "polecat.md")` to match the OS-appropriate separator

## Test plan

- [x] Test passes on Linux
- [ ] Test should now pass on Windows CI (backslash path separators)

### DIFF
--- a/internal/cmd/prime_output_test.go
+++ b/internal/cmd/prime_output_test.go
@@ -148,7 +148,8 @@ func TestOutputRoleDirectives(t *testing.T) {
 		if !strings.Contains(out, "[EXPLAIN]") {
 			t.Errorf("expected EXPLAIN output, got: %s", out)
 		}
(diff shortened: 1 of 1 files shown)


htristan:
This is already addressed in #3141, which is open and pending merge. That PR fixes the `filepath.Join` path separator issue here *and* the underlying data race in `outputRoleDirectives` (the `os.Stdout` swap + `primeExplain` global mutation under `t.Parallel()`).

See #3140 for the tracking issue.

Recommend closing this in favor of #3141.

daniel-jasinski:
makes sense. closing

PR #3192, opened 2026-03-23
fix: prevent cross-town orphan cleanup from killing agents on other towns' sockets

### LINKED ISSUE #3191: Multi-town: cross-town orphan cleanup kills agents from sibling towns
## Summary

When multiple Gas Town instances run on the same machine, the Deacon's orphan-process-cleanup patrol step can kill agents belonging to a sibling town. The Deacon AI agent sees all Claude processes system-wide via `ps`, but can only see its own town's tmux sessions via `tmux list-sessions` (single-socket). Processes from other towns appear as orphans and get SIGTERM'd.

## Reproduction

1. Run two towns on the same machine (Town A and Town B), each with their own daemon, tmux socket, and agents.
2. Town A's Deacon runs its patrol cycle and hits the `orphan-process-cleanup` step.
3. The Deacon AI runs `ps aux | grep claude` and sees Claude processes from both towns.
4. The Deacon AI runs `tmux list-sessions` which only lists Town A's sessions (because the default socket is Town A's socket).
5. Town B's Mayor (or any agent) does not appear in Town A's session list.
6. The Deacon AI concludes Town B's agents are orphans and sends SIGTERM.

## Expected behavior

Orphan cleanup should never kill agents belonging to a different town on the same machine.

## Actual behavior

- Mayor process exits with code 143 (SIGTERM) — confirmed via tmux `pane_dead_status`.
- Crash loop: auto-respawn hook restarts the Mayor, but the next Deacon patrol cycle kills it again within 1–2 minutes.
- Crash loop self-resolves when the offending Deacon session is replaced (new session happened to use the safe built-in `gt deacon cleanup-orphans` instead of raw kill).
- Intermittent: depends on which approach the Deacon AI chooses for the orphan cleanup step. Some sessions use the safe built-in command; others use raw `ps | grep | kill`.

## Root cause

Two independent safety gaps combine:

**1. Single-socket tmux visibility.** `getTmuxSessionPIDs()` in `internal/util/orphan.go` calls `tmux.BuildCommand("list-sessions")` which only queries the current town's socket. Cross-town sessions are invisible to the protection logic.

**2. AI agent autonomy.** The `mol-deacon-patrol` formula's `orphan-process-cleanup` step describes detection heuristics but does not mandate using the built-in `gt deacon cleanup-orphans` command. The AI agent may use raw `ps` + `tmux list-sessions` + manual `kill`, bypassing the TTY-based and workspace-based safety checks in `FindOrphanedClaudeProcesses()`.

## Evidence

- Patrol report from offending Deacon session: *"Cleaned 2 orphans (PIDs X, Y)"* where PID Y was the sibling town's Mayor.
- Exit code 143 (SIGTERM) confirmed for 3 consecutive Mayor deaths via a watcher polling `pane_dead_status`.
- Neither daemon logged zombie detection or orphan cleanup for those PIDs.
- Orphan/zombie state files (`/tmp/gastown-*-state`) were empty — the built-in cleanup functions were not the source.
- Crash loop stopped when the offending Deacon session was replaced by one that used the safe built-in commands.

htristan:
## Summary

Fixes #3191

When multiple Gas Town instances run on the same machine, each uses its own tmux socket. `getTmuxSessionPIDs()` only queried the **current town's socket** when building the protected PID set, so processes in other towns' tmux sessions were invisible to the protection logic — and could be SIGTERMed as "orphans" by the deacon.

This was confirmed by tracing a production failure: a deacon in Town A ran `ps aux | grep claude`, saw Town B's mayor process, checked `tmux list-sessions` (which only showed Town A's sessions), concluded the process was orphaned, and killed it.

### Changes

- **`getTmuxSessionPIDs()` now scans all sockets** in `/tmp/tmux-{uid}/` instead of only the default socket, matching the documented intent ("protect ALL tmux sessions"). Falls back to `BuildCommand` for non-standard socket locations.
- **Extracted `resolveTownRoot()`** from `isInGasTownWorkspace()` and added `TownRoot` field to `OrphanedProcess`/`ZombieProcess` structs. CLI output now shows `town=<path>` for each candidate so the deacon agent can verify ownership.
- **Updated deacon patrol formula** (`mol-deacon-patrol.formula.toml`) with multi-town awareness: only clean up processes belonging to this town's `$GT_ROOT`.
- **Updated deacon role template** (`deacon.md.tmpl`) with a "Multi-Town Awareness" section.
- **Added 5 tests** including a cross-socket test that directly reproduces the bug.

### ZFC alignment

All changes follow the Zero Framework Cognition principle:
- Go code is pure transport: scans sockets, resolves town roots, exposes data
- Agent cognition (formula + template) decides what to kill based on that data

### Test evidence

The P0 test (`TestGetTmuxSessionPIDs_CrossSocket`) creates two tmux sockets, spawns a session on each, sets the default to Socket A, and asserts both PIDs are protected. It was verified red-to-green against the old implementation:

**Before fix** (old `orphan.go`, test applied):
```
=== RUN   TestGetTmuxSessionPIDs_CrossSocket
    orphan_crosssocket_test.go:63: PID 47367 from socket B not in protected set — cross-town process would be killed (set size=1)
--- FAIL: TestGetTmuxSessionPIDs_CrossSocket (0.59s)
```

**After fix**:
```
=== RUN   TestGetTmuxSessionPIDs_CrossSocket
--- PASS: TestGetTmuxSessionPIDs_CrossSocket (2.43s)
```

## Test plan

- [x] `TestGetTmuxSessionPIDs_CrossSocket` — reproduces the exact bug (P0)
- [x] `TestGetTmuxSessionPIDs_SingleSocket` — regression test for single-town setups (P1)
- [x] `TestResolveTownRoot` — town root resolution from root, subdir, and outside (P1)
- [x] `TestResolveTownRoot_DistinguishesAdjacentTowns` — sibling towns not confused (P2)
- [x] `TestOrphanedProcess_TownRoot_Populated` — TownRoot field wired end-to-end (P2)
- [x] Full `go test ./internal/util/` suite passes
- [x] `go vet` clean on both `internal/util/` and `internal/cmd/`

### DIFF
--- a/internal/cmd/deacon.go
+++ b/internal/cmd/deacon.go
@@ -1360,15 +1360,19 @@ func runDeaconCleanupOrphans(cmd *cobra.Command, args []string) error {
 	// Report results
 	var terminated, escalated, unkillable int
 	for _, r := range results {
+		town := r.Process.TownRoot
+		if town == "" {
+			town = "unknown"
+		}
 		switch r.Signal {
 		case "SIGTERM":
-			fmt.Printf("  %s Sent SIGTERM to PID %d (%s)\n", style.Bold.Render("→"), r.Process.PID, r.Process.Cmd)
+			fmt.Printf("  %s Sent SIGTERM to PID %d (%s) town=%s\n", style.Bold.Render("→"), r.Process.PID, r.Process.Cmd, town)
 			terminated++
 		case "SIGKILL":
-			fmt.Printf("  %s Escalated to SIGKILL for PID %d (%s)\n", style.Bold.Render("!"), r.Process.PID, r.Process.Cmd)
+			fmt.Printf("  %s Escalated to SIGKILL for PID %d (%s) town=%s\n", style.Bold.Render("!"), r.Process.PID, r.Process.Cmd, town)
 			escalated++
 		case "UNKILLABLE":
-			fmt.Printf("  %s WARNING: PID %d (%s) survived SIGKILL\n", style.Bold.Render("⚠"), r.Process.PID, r.Process.Cmd)
+			fmt.Printf("  %s WARNING: PID %d (%s) survived SIGKILL town=%s\n", style.Bold.Render("⚠"), r.Process.PID, r.Process.Cmd, town)
 			unkillable++
 		}
 	}
@@ -1406,8 +1410,12 @@ func runDeaconZombieScan(cmd *cobra.Command, args []string) error {
 	if zombieScanDryRun {
 		for _, z := range zombies {
 			ageStr := fmt.Sprintf("%dm", z.Age/60)
-			fmt.Printf("  %s PID %d (%s) TTY=%s age=%s\n",
-				style.Dim.Render("→"), z.PID, z.Cmd, z.TTY, ageStr)
+			town := z.TownRoot
+			if town == "" {
+				town = "unknown"
+			}
+			fmt.Printf("  %s PID %d (%s) TTY=%s age=%s town=%s\n",
+				style.Dim.Render("→"), z.PID, z.Cmd, z.TTY, ageStr, town)
 		}
 		fmt.Printf("%s Dry run - no processes killed\n", style.Dim.Render("○"))
 		return nil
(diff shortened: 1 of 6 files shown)


steveyegge:
Fix-merged to main as cc666b54. All 5 new tests pass, go vet clean. The CI failures on the PR were pre-existing (TestBuildRefineryPatrolVars_FullConfig on main). Thanks @htristan!

PR #3211, opened 2026-03-23
fix(doltserver): fix thundering herd in doltserver.Start()

### LINKED ISSUE #3195: doltserver: concurrent gt dolt start calls bypass startup lock, causing thundering herd of competing processes
## Summary

When Dolt goes down and multiple Gas Town agents (witnesses, polecats, refineries) simultaneously detect it via failed `bd` calls, they all call `gt dolt start` at the same time. This causes a thundering herd of competing `dolt sql-server` processes that fight for exclusive database lock files — with the result that **none of them successfully start Dolt**.

## Root cause

The startup lock in `doltserver.Start()` uses a 3-second wait before declaring the lock stale and forcibly removing it:

```go
// internal/doltserver/doltserver.go ~line 1308
for i := 0; i < 6; i++ {
    time.Sleep(500 * time.Millisecond)  // 6 × 500ms = 3s total
    locked, err = fileLock.TryLock()
    if err == nil && locked {
        break
    }
}
if !locked {
    // Remove the lock file even if another process is actively holding it
    fmt.Fprintf(os.Stderr, "Warning: dolt.lock held for >3s — removing stale lock\n")
    _ = os.Remove(lockFile)
    // ... spawn a new dolt process anyway
}
```

With 7 databases, Dolt takes **~18 seconds** to initialize (PR #3059 fixed the health-check timeout, but not the startup lock timeout). After 3s, every waiting `gt dolt start` removes the lock file and spawns its own `dolt sql-server`. Each new process acquires exclusive database LOCKs for different databases, so:
- None can acquire all 7 locks
- None binds to port 3307  
- All fail with "database X is locked by another dolt process"
- LOCK files accumulate, requiring manual cleanup
- The circuit breaker trips, blocking all `bd` operations

## Observed in production

With 6 witnesses (one per rig) + refinery + polecats, we've seen 5–8 competing dolt processes appear simultaneously within seconds of Dolt going down.

## Related issues

- #2690: similar race during `gt dolt restart` — both share the same root cause (concurrent spawning without coordination)
- PR #3059 (merged): scales health-check timeout by DB count — the complementary fix for the startup lock

## Proposed fix

Two-part fix in `doltserver.Start()`:

**1. Scale the lock wait timeout by database count** (mirrors PR #3059):
```go
dbCount := len(databases)
if dbCount < 1 { dbCount = 1 }
lockWait := time.Duration(dbCount) * 5 * time.Second  // 5s per DB
if lockWait < 15*time.Second { lockWait = 15*time.Second }
if lockWait > 120*time.Second { lockWait = 120*time.Second }
```

**2. Check if Dolt started successfully before removing the lock** (eliminates thundering herd):
```go
if !locked {
    // Another process may have successfully started Dolt while we waited
    if running, _, _ := IsRunning(townRoot); running {
        return nil  // Dolt is up — we're done, no need to spawn another process
    }
    // Lock is held but Dolt is NOT running — genuinely stale
    fmt.Fprintf(os.Stderr, "Warning: dolt.lock stale (no running server) — removing\n")
    _ = os.Remove(lockFile)
    // ... retry once
}
```

This makes concurrent callers **wait for the first process to succeed**, then return immediately if Dolt is already running. No more thundering herd.

## Reproduction

1. Start a Gas Town with 6+ active witnesses
2. Kill Dolt: `kill $(cat ~/gt/.dolt-data/dolt.pid)`
3. Within 30 seconds, observe multiple `dolt sql-server` processes spawned by witnesses
4. Observe `ps aux | grep dolt` shows 3–8 processes, none listening on port 3307
5. Observe stale LOCK files in `~/gt/.dolt-data/`

baumgold:
Fixes #3195

## Problem

When Dolt fails and multiple agents call `gt dolt start` concurrently, they all competed for the startup lock with a 3s timeout — far too short when initializing 7 databases (~18s). All waiters removed the stale lock and spawned duplicate `dolt sql-server` processes, which fought over exclusive DB locks and left port 3307 unbound.

## Fix

1. **Scale lock timeout dynamically**: 5s per database (detected via `.dolt` subdirectory count), clamped between 15s–120s, giving all callers enough time to wait out the leader's startup.
2. **Check before evicting**: Before force-removing a stale lock, call `IsRunning()`. If Dolt is already up, return success instead of spawning a duplicate.

Adds `countDoltDatabases()` helper and a test.

### DIFF
--- a/internal/hooks/installer_test.go
+++ b/internal/hooks/installer_test.go
@@ -230,8 +230,10 @@ func TestSyncForRole_JSONWhitespaceInsensitive(t *testing.T) {
 		t.Fatalf("reading created file: %v", err)
 	}
 
-	// Add extra whitespace — structurally identical JSON, different bytes
-	reformatted := strings.ReplaceAll(string(original), ":", " : ")
+	// Add extra whitespace — structurally identical JSON, different bytes.
+	// Use commas (not colons) since Windows paths contain "D:" which would
+	// corrupt the JSON value when colons are naively replaced.
+	reformatted := strings.ReplaceAll(string(original), ",", " , ")
 	if string(original) == reformatted {
 		t.Fatal("reformatted content should differ from original bytes")
 	}
@@ -265,9 +267,12 @@ func TestSyncForRole_GeminiWithGTBinSubstitution(t *testing.T) {
 	if strings.Contains(string(got), "{{GT_BIN}}") {
 		t.Error("{{GT_BIN}} placeholder was not substituted")
 	}
-	// Verify the resolved binary path is present
+	// Verify the resolved binary path is present.
+	// JSON settings files have backslashes escaped (Windows paths), so check
+	// for the JSON-encoded form of the path rather than the raw path.
 	gtBin := resolveGTBinary()
-	if !strings.Contains(string(got), gtBin) {
+	gtBinInFile := strings.ReplaceAll(gtBin, `\`, `\\`)
+	if !strings.Contains(string(got), gtBinInFile) {
 		t.Errorf("expected resolved gt binary %q in output", gtBin)
 	}
 }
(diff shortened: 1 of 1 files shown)


steveyegge:
Fix-merged to main (rebased and pushed directly). Thank you for the thundering herd fix and hooks improvements!

PR #3213, opened 2026-03-24
fix(daemon): use config.yaml with timeouts in DoltServerManager (gt-ch5)

outdoorsea:
## Summary

- DoltServerManager was starting `dolt sql-server` with CLI flags only, which cannot set `read_timeout_millis`/`write_timeout_millis`
- This caused CLOSE_WAIT accumulation when clients disconnected mid-session, degrading Dolt stability over time
- Fix: write `config.yaml` before starting the server (same approach as `doltserver.Start`), then use `--config` flag
- Sets read/write timeouts to 30s and max_connections to 1000, matching the non-daemon `gt dolt start` behavior

## Test plan

- [ ] Verify daemon-managed Dolt server starts successfully with config.yaml
- [ ] Confirm CLOSE_WAIT connections are cleaned up after 30s timeout
- [ ] Check that port, host, and data-dir settings are correctly written to config.yaml

### DIFF
--- a/internal/cmd/hooks_sync.go
+++ b/internal/cmd/hooks_sync.go
@@ -127,17 +127,48 @@ func runHooksSync(cmd *cobra.Command, args []string) error {
 			if loc.Rig != "" {
 				rigPath = filepath.Join(townRoot, loc.Rig)
 			}
-			rc := config.ResolveRoleAgentConfig(loc.Role, townRoot, rigPath)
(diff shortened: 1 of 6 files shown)


outdoorsea:
Closing — the core fix (2028a898, writeDaemonDoltConfig with timeouts) was already landed upstream. Cherry-pick is empty against current main.

PR #3215, opened 2026-03-24
Fix convoy close dedup after reopen

seanwessmith:
## Summary
- allow convoy close detection to recover after an issue is reopened and closed again
- add a 1-second event poll overlap plus lifecycle event-id dedup to handle Dolt's second-precision timestamps without replaying old close events
- preserve warm-up semantics and add regressions for reopen/reclose across polls and within one poll

## Testing
- `go test ./internal/daemon`


steveyegge:
Fix-merged to main (rebased and pushed directly). Convoy close-reopen-close dedup fix is solid — good edge case handling.

PR #3219, opened 2026-03-24
fix(witness): honor role_agents model config

### LINKED ISSUE #3218: Witness ignores role_agents model config, always uses default model
## Bug Description

The witness agent ignores the `role_agents` model configuration from `settings/config.json`. When `role_agents` maps witness to `claude-sonnet`, the witness still launches with bare `claude --dangerously-skip-permissions` (defaulting to Opus) instead of `claude --model sonnet`.

## Root Cause

In `internal/witness/manager.go`, the `buildWitnessStartCommand` function checks if the role TOML has a `start_command`. The built-in `internal/config/roles/witness.toml` has `start_command = "exec claude --dangerously-skip-permissions"`. When the resolved agent is Claude (`IsResolvedAgentClaude`), this hardcoded command is used verbatim — bypassing `BuildStartupCommandFromConfig` which correctly applies `--model`, `--settings`, env vars, and the initial prompt from the `role_agents` config.

The refinery does NOT have this bug because its `Start()` method calls `BuildStartupCommandFromConfig` directly without going through a TOML start_command path.

## Steps to Reproduce

1. Configure `role_agents` in town or rig settings to map `witness` to a non-default model (e.g., `claude-sonnet`)
2. Start a witness via `gt witness start`
3. Observe the witness launches with the default model instead of the configured one

## Expected Behavior

The witness should launch with the model specified by the `role_agents` configuration (e.g., `--model sonnet`).

## Actual Behavior

The witness launches with bare `claude --dangerously-skip-permissions`, ignoring the `role_agents` model mapping.

## Fix

Two changes needed in `internal/config/loader.go` and `internal/witness/manager.go`:

1. Add `HasExplicitRoleAgent()` function to detect when `role_agents` explicitly maps a role
2. In `buildWitnessStartCommand`, when an explicit mapping exists, skip the TOML `start_command` and fall through to `BuildStartupCommandFromConfig` which properly resolves the agent model and flags

Cdfghglz:
## Summary

- Fixes witness ignoring `role_agents` model configuration from settings, always launching with default model instead of the configured one (e.g., `claude-sonnet`)
- Adds `HasExplicitRoleAgent()` to detect explicit `role_agents` mappings and skip the hardcoded TOML `start_command` in favor of `BuildStartupCommandFromConfig`

Fixes #3218

## Test plan

- [x] `go test ./internal/witness/... ./internal/config/... -run "TestBuild|TestResolve"` passes
- [ ] Configure `role_agents` to map witness to a non-default model and verify correct `--model` flag in launched command
- [ ] Verify witness still works correctly when no `role_agents` mapping exists (falls back to default behavior)

### DIFF
--- a/internal/cmd/compact_report_test.go
+++ b/internal/cmd/compact_report_test.go
@@ -10,25 +10,27 @@ import (
 func TestWispTypeToCategory(t *testing.T) {
 	tests := []struct {
 		wispType string
+		title    string
 		want     string
 	}{
-		{"heartbeat", "Heartbeats"},
-		{"ping", "Heartbeats"},
-		{"patrol", "Patrols"},
-		{"gc_report", "Patrols"},
-		{"error", "Errors"},
-		{"recovery", "Errors"},
-		{"escalation", "Errors"},
-		{"", "Untyped"},
-		{"unknown", "Untyped"},
-		{"default", "Untyped"},
+		{"heartbeat", "", "Heartbeats"},
+		{"ping", "", "Heartbeats"},
+		{"patrol", "", "Patrols"},
+		{"gc_report", "", "Patrols"},
+		{"error", "", "Errors"},
+		{"recovery", "", "Errors"},
+		{"escalation", "", "Errors"},
+		{"", "", "Untyped"},
+		{"unknown", "", "Untyped"},
+		{"default", "", "Untyped"},
+		{"", "Patrol report", "Patrols"},
 	}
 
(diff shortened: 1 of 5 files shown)


steveyegge:
This fix is already in `main` — someone landed an equivalent change concurrently, so when I rebased your branch it collapsed to empty and GitHub auto-closed the PR. Your contribution helped, even though the merge record isn't tied to this PR. Thanks!

PR #3232, opened 2026-03-24
fix: add safety guards to prevent accidental deletion of Dolt databases

outdoorsea:
## Summary

- **Quarantine instead of delete**: `Start()` now moves corrupted databases to `.quarantine/` instead of permanently deleting them. Large databases (>1MB) are never deleted, only warned about.
- **Offline safety check**: `RemoveDatabase()` now checks filesystem size when the Dolt server is offline. Previously, the `databaseHasUserTables` safety check only ran when the server was running.
- **Bulk deletion guard**: `gt dolt cleanup` now refuses to proceed when >50% of databases would be removed (and count >3), requiring `--force`. This catches metadata corruption causing false orphan detection.

## Test plan

- [ ] `TestRemoveDatabase_RefusesLargeDBWhenServerDown` — verifies large DB protection when offline
- [ ] `TestRemoveDatabase_AllowsSmallDBWhenServerDown` — verifies small orphans can still be cleaned
- [ ] `TestQuarantine_MovesInsteadOfDeleting` — verifies quarantine behavior for corrupted DBs
- [ ] Existing cleanup tests still pass

### DIFF
--- a/internal/cmd/dolt.go
+++ b/internal/cmd/dolt.go
@@ -944,6 +944,23 @@ func runDoltCleanup(cmd *cobra.Command, args []string) error {
 		return nil
 	}
 
+	// BALK: If orphans are a large fraction of all databases, something is likely
+	// wrong with the orphan detection (e.g., metadata files not found). Refuse to
+	// proceed without --force to prevent accidentally dropping production databases. (gt-xvh)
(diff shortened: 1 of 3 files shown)


outdoorsea:
Closing — superseded by upstream c816b359 (fix: add safety guards to prevent accidental deletion of Dolt databases).

PR #3239, opened 2026-03-24
fix: gt polecat remove -f now bypasses unpushed commit check (gt-xwr)

baumgold:
## Summary

- `gt polecat remove -f` was failing with 'polecat has uncommitted work: N unpushed commit(s)' even when the target polecat's own branch was fully synced with origin
- The `--force` flag is documented to 'bypass checks' but the unpushed commits check was not bypassed
- Root cause: `CanForceRemove()` returned `false` for `CleanupUnpushed`, and the fallback git check also blocked on `UnpushedCommits > 0` in force mode

## Changes

- `internal/polecat/types.go`: `CanForceRemove()` now returns `true` for `CleanupUnpushed`
- `internal/polecat/manager.go`: Fallback git check only blocks on stashes (not unpushed commits) when `force=true`
- `internal/polecat/types_test.go`: Updated test expectation for `CleanupUnpushed.CanForceRemove()` to `true`

Stashes still block force removal since they represent intentional work-in-progress. Use `polecat nuke --force` to bypass all checks including stashes.

Fixes: https://github.com/steveyegge/gastown/issues/3238

🤖 Generated with Claude Code

### DIFF
--- a/internal/polecat/manager.go
+++ b/internal/polecat/manager.go
@@ -394,7 +394,7 @@ func (m *Manager) getCleanupStatusFromBead(name string) CleanupStatus {
 
 // checkCleanupStatus validates the cleanup status against removal safety rules.
 // Returns an error if removal should be blocked based on the status.
-// force=true: allow has_uncommitted, block has_stash and has_unpushed
+// force=true: allow has_uncommitted and has_unpushed, block has_stash
 // force=false: block all non-clean statuses
 func (m *Manager) checkCleanupStatus(name string, status CleanupStatus, force bool) error {
 	// Clean status is always safe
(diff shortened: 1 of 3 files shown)


baumgold:
Closing this PR as superseded by #3221.

PR #3221 contains all the fixes that this PR was intended to provide (gt-xwr: `--force` bypasses unpushed commits check), plus additional fixes that resolve the two failing CI checks here:
- `TestBuildRefineryPatrolVars_FullConfig` — fixed by the patrol vars test update in #3221
- `TestRunHooksSyncNonClaudeAgent` — fixed by the hooks-sync `ResolveRoleAgentName` change in #3221

Rather than rebasing this branch on top of all those fixes, it's cleaner to carry this work forward through #3221 directly.

PR #3241, opened 2026-03-24
fix(EnsureAllMetadata): prevent ping-pong when two databases map to same rig (gt-cn4)

### LINKED ISSUE #3222: bug: EnsureAllMetadata ping-pong when two Dolt databases map to same rig
## Problem

When a rig has two Dolt databases that both resolve to the same rig directory, `EnsureAllMetadata` causes them to fight over `metadata.json` in an infinite loop every time it's called (e.g. during `gt up`).

**Observed symptom:**
```
Warning: metadata.json dolt_database was "gt", correcting to "gastown" (identity mismatch repair)
Warning: metadata.json dolt_database was "gastown", correcting to "gt" (identity mismatch repair)
```

## Root Cause

`buildDatabaseToRigMap` strips the trailing `-` from route prefixes:
- Route `{"prefix":"gt-","path":"gastown/mayor/rig"}` → maps `"gt"` → `"gastown"` rig

`EnsureAllMetadata` then processes each Dolt database independently:
1. Database `"gastown"` → no entry in routes map → `rigName = "gastown"` → calls `EnsureMetadata(townRoot, "gastown", "gastown")`
2. Database `"gt"` → found in routes map → `rigName = "gastown"` → calls `EnsureMetadata(townRoot, "gastown", "gt")`

Both calls write to the same `metadata.json`, causing them to flip back and forth on every invocation.

## How to reproduce

1. Have a rig where the Dolt database name differs from the rig directory name (e.g. gastown rig with database name "gt" per `--prefix gt`, but an older "gastown" database also exists from initial setup)
2. Run `gt up`

## Fix

PR #3221: When multiple databases map to the same rig, prefer the database that `metadata.json` already points to (if it's in the valid database list). Only use the route-derived name as a fallback when metadata has no current value or the current value isn't in the known database list.

baumgold:
Fixes ping-pong when two Dolt databases map to the same rig (e.g. 'gastown' and 'gt' both mapping to gastown rig). When multiple databases resolve to the same rig, prefer the one `metadata.json` already points to. Eliminates `Warning: metadata.json dolt_database was "X", correcting to "Y"` spam during `gt up`.

Fixes #3222

### DIFF
--- a/internal/doltserver/doltserver.go
+++ b/internal/doltserver/doltserver.go
@@ -39,6 +39,7 @@ import (
 	"path/filepath"
 
 	"runtime"
+	"sort"
 	"strconv"
 	"strings"
 	"sync"
@@ -2977,6 +2978,12 @@ func EnsureAllMetadata(townRoot string) (updated []string, errs []error) {
 		dbToRig[k] = v
 	}
 
+	// Group databases by rig name to detect collisions (e.g., "gastown" and
+	// "gt" both mapping to the "gastown" rig). Preserve encounter order so
+	// the result slice is deterministic.
+	rigToDbs := make(map[string][]string)
+	var rigOrder []string
+	seenRig := make(map[string]bool)
 	for _, dbName := range databases {
 		rigName := dbName
 		if mapped, ok := dbToRig[dbName]; ok {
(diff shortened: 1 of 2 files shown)


baumgold:
## Build Status Note

The CI failures on this PR are **pre-existing failures on `main`**, not introduced by this PR's changes:

- **Lint**: `hooksDir is unused` in `internal/hooks/installer.go` — fixed in #3242
- **`TestBuildRefineryPatrolVars_FullConfig`**: test expects 5 vars but formula already has 7 — fixed in #3242

**Merge order:** #3242 must land first to fix lint and the patrol test. Once #3242 is merged, this PR's CI should pass cleanly.

Changes in this PR are verified correct and all locally-relevant tests pass.

steveyegge:
Superseded by #3287 which was merged and fixes the same EnsureAllMetadata oscillation issue.

Issue #3254, opened 2026-03-24
gt sling broken: bd agent state removed in beads v0.62.0 but gt still calls it


htristan:
## Summary

The `go.mod` bump from beads v0.61.0 → v0.62.0 (`b9875f17`, March 23) is a breaking dependency upgrade. Beads v0.62.0 intentionally removed the entire agent-as-bead subsystem ([`0bd598ce` — "refactor: remove agent-as-bead subsystem from beads (bd-4vw2)"](https://github.com/steveyegge/beads/releases/tag/v0.62.0)), but the `gt` codebase still shells out to `bd agent state` in multiple places.

## Impact

`gt sling` cannot start polecats. `UpdateAgentState` calls `bd agent state <id> working` which fails immediately (`Error: unknown command "agent" for "bd"`). `SetAgentStateWithRetry` retries 10 times with exponential backoff (~2.5 minutes), blocking the spawn path. The tmux session dies before Claude Code starts.

This also breaks:
- `gt done` (agent state transition to idle/stuck)
- Daemon heartbeat GUPP checks and orphaned work detection (`listAgentBeadsJSON` returns "no agent beads found")
- Deacon enters crash loop and cannot recover

## Affected code paths

- `internal/beads/beads_agent.go:426` — `b.runWithRouting("agent", "state", id, state)`
- `internal/polecat/manager.go:339` — `SetAgentStateWithRetry`
- `internal/cmd/polecat_spawn.go:403` — called during sling
- `internal/cmd/done.go:1419` — called during done
- `internal/polecat/manager.go:1640` — called during `ReuseIdlePolecat` (added in `e58ac466`)

## Reproduction

```
$ bd --version
bd version 0.62.0 (Homebrew)

$ bd agent state test-id working
Error: unknown command "agent" for "bd"
```

## Root cause

Commit `b9875f17` bumped `go.mod` from beads v0.61.0 → v0.62.0 without updating `gt` code to account for the removed `bd agent` subcommand. The beads v0.62.0 release explicitly removed gastown-specific subsystems including agent-as-bead, GUPP, patrol molecules, and polecat/crew terminology.

## Suggested fix

Either:
1. Gate `bd agent state` calls behind a version check and fall back to description-based updates for bd ≥ 0.62.0
2. Replace all `bd agent state` calls with the pre-`deb58838` approach (update agent_state via description field)
3. Coordinate with beads to re-expose `bd agent state` or provide an equivalent

htristan:
[...]

Issue #3216, opened 2026-03-24
gt sling fails: calls nonexistent 'bd agent state' command


KingMob:
## Summary

`gt sling` (v0.12.1) calls `bd agent state <agent-id> working` during polecat startup, but `bd` (v0.62.0) has no `agent` subcommand. This causes SetAgentState to retry 10 times with exponential backoff (~3 minutes wasted), then the polecat session dies.

## Reproduction

```bash
gt sling fe-b0y swarmcdn_browser
# Output:
# ⚠ Warning: SetAgentState attempt 1 failed: bd agent state fe-swarmcdn_browser-polecat-slit working:
#   Error: unknown command "agent" for "bd"
# ... retries 10 times ...
# Error: starting polecat session: getting pane for fe-slit (session likely died during startup)
```

## Impact

3 of 8 wave-1 tasks in convoy hq-cv-5say9 failed to dispatch. The failing tasks all attempted to reuse a polecat slot (slit, dementus) where a previous task had already completed — suggesting the agent state tracking is involved in slot reuse.

## Environment

- gt v0.12.1
- bd v0.62.0 (Homebrew)
- macOS Darwin 25.3.0

## Expected

`gt sling` should either:
1. Use a `bd` subcommand that actually exists, or
2. Gracefully handle missing agent state support without killing the polecat session

PR #3257, opened 2026-03-24
fix: remove stale bd agent state call from UpdateAgentState

thejosephstevens:
## Problem

\`gt sling\` adds ~3 minutes of delay on every invocation due to a 10-attempt exponential backoff loop calling \`bd agent state <id> working\`. This command was removed in beads v0.62.0 as part of a standalone decoupling effort — the agent-as-bead subsystem was extracted as Gas Town infrastructure that belonged in GT, not in the beads schema.

Tracked in #3216 and #3254.

## Root cause

\`UpdateAgentState\` in \`internal/beads/beads_agent.go\` calls \`runWithRouting("agent", "state", id, state)\` which shells out to \`bd agent state\`. Since beads v0.62.0 removed that subcommand, every call exits with "unknown command" — which is not in \`isDoltConfigError\`'s fast-fail list, so all 10 retries fire (~3 min total).

## Fix

This PR fixes the delay by dropping the \`bd agent state\` call and updating the description text only. This is a **correct but incomplete** fix — see "Architecture note" below.

- **\`UpdateAgentState\`**: drop \`runWithRouting\` call, update description only via \`UpdateAgentDescriptionFields\`
- **\`GetAgentBead\`** + **\`lifecycle.go getAgentBeadInfo\`**: remove dead DB column preference; description is now the read path
- **\`deacon.go:1175\`**: remove second \`bd agent state\` call site (found during review of competing PR #3283)
- **\`TestCheckPolecatHealth_DBStateOverridesDescription\`**: rename and invert to \`TestCheckPolecatHealth_DescriptionOverridesDBColumn\`

Design reviewed and approved by crew before implementation.

## Architecture note (follow-on work)

During review of competing PR #3283 (which used \`bd set-state\`) we identified that the correct long-term approach is:

**Write**: \`bd set-state <id> agent_state=<state>\` — writes as a label (\`agent_state:working\`) plus an event bead, using beads' generic dimension:value label convention. Indexed, queryable (\`bd list --label agent_state:working\`), survives session death.

**Read**: scan issue labels for \`agent_state:*\` instead of the dead DB column or description text.

This PR ships the description-only path to eliminate the 3-minute delay immediately. The label-based read/write is tracked as a follow-on in gt-247.

### DIFF
--- a/internal/beads/beads_agent.go
+++ b/internal/beads/beads_agent.go
@@ -415,15 +415,22 @@ func (b *Beads) ResetAgentBeadForReuse(id, reason string) error {
 	return nil
 }
 
-// UpdateAgentState updates the agent_state field in an agent bead's description.
+// UpdateAgentState updates the agent_state dimension on an agent bead.
 //
-// bd agent state was removed in beads v0.62.0 as part of observable-states
-// deprecation (ZFC design gt-zecmc). The DB column is no longer maintained
-// by gt; tmux session presence is the authoritative source for liveness.
-// This function updates the description text only, for human-readable bd show output.
+// Uses `bd set-state` (beads 0.62.0+) which writes a dimension:value label
+// (agent_state:<state>) and an event bead for the audit trail. Labels are
+// indexed and queryable ("bd list --label agent_state:working") across rigs.
+//
+// bd agent state was removed in beads v0.62.0 when the agent-as-bead subsystem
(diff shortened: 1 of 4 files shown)


steveyegge:
Superseded by #3283 (merged). That PR already migrated `bd agent state` → `bd set-state`. Please rebase if additional changes are needed beyond what #3283 covered.

PR #3269, opened 2026-03-25
fix: witness falls back to mail when mayor nudge fails

seanbearden:
## Summary
- `notifyMayorSlotOpen` only used tmux nudge and channel events to notify the mayor of polecat completions
- If the mayor is in an ACP/Claude Code session (not tmux-based), the nudge silently fails and the mayor never learns a slot opened
- Adds mail fallback: try nudge first (lightweight, no Dolt commit), send mail only when nudge can't reach the mayor

## Context
Observed: two polecats completed work (beads closed, merged), witness patrol ran, but mayor received no notification. Root cause: mayor was in a Claude Code ACP session, not a tmux idle loop, so the tmux nudge had no target.

## Test plan
- [x] Build succeeds
- [x] Verified nudge path still preferred (no unnecessary mail when tmux is available)
- [x] Mail fallback triggers when `HasSession` returns false

### DIFF
--- a/internal/witness/handlers.go
+++ b/internal/witness/handlers.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"sort"
 	"strconv"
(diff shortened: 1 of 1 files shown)


seanbearden:
Fork main contamination — includes 2 unrelated commits from PRs #3265/#3266. Recreating as single-commit PR from upstream/main.

PR #3274, opened 2026-03-25
fix(mol): cap backoff before overflow in await-event and await-signal

natorious:
## Summary

- Fixes negative timeout bug in `gt mol step await-event` and `gt mol step await-signal` at high idle counts
- At idle ~30+, the exponential backoff calculation could overflow `int64` (nanoseconds), producing a huge negative duration that caused instant timeouts
- Root cause: `backoff-max` was parsed and applied **after** the loop, so `30s * 2^34` overflowed before the cap applied
- Fix: parse `maxDur` before the loop and check `timeout >= maxDur` at the **start** of each iteration, returning early when the cap is hit

## Test plan

- [x] `TestCalculateEventTimeout/backoff_overflow_guard:_idle=34_with_max_cap` — new test, passes
- [x] `TestCalculateEffectiveTimeout/backoff_overflow_guard:_idle=34_with_max_cap` — new test, passes
- [x] All existing backoff tests pass
- [x] Build passes

Fixes: ne-5oq (await-event negative timeout at high idle counts, refinery burning sessions)

### DIFF
--- a/internal/cmd/molecule_await_event.go
+++ b/internal/cmd/molecule_await_event.go
@@ -275,18 +275,28 @@ func calculateEventTimeout(idleCycles int) (time.Duration, error) {
 		if err != nil {
 			return 0, fmt.Errorf("invalid backoff-base: %w", err)
 		}
-		timeout := base
-		for i := 0; i < idleCycles; i++ {
-			timeout *= time.Duration(awaitEventBackoffMult)
-		}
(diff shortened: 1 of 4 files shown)


steveyegge:
Fix-merged to main (rebased and pushed directly). Overflow guard for exponential backoff is solid.

PR #3276, opened 2026-03-25
feat(formula): add --set flag and fix bead prefix for convoy formulas

ckumar1:
## Summary

- Added `--set key=value` flag to `gt formula run` for passing input variables to convoy formulas
- Fixed bead prefix: `executeConvoyFormula` used hard-coded `hq-` prefix; now resolves via `config.GetRigPrefix()` to match the target rig

## Context

Convoy formulas define `[inputs]` in their TOML (e.g., `problem`, `context`, `scope`) but `gt formula run` had no mechanism to pass values for them. The only parameterization was `--pr` and `--files`. Formulas like `design` and `mol-prd-review` that need `--problem` had no working execution path.

Separately, convoy/leg/synthesis beads were created with `hq-` prefix (town root) but `gt sling` dispatches to the target rig, which rejects mismatched prefixes.

## Usage

```bash
gt formula run design --set problem="Design a caching layer" --set context="Go microservice"
gt formula run mol-prd-review --set problem="Review auth flow" --dry-run
```

## Test plan

- [x] `go build ./cmd/gt`
- [x] `go vet ./internal/cmd/`
- [x] `--set` values injected into template context for both execution and dry-run
- [x] Bead IDs use rig-scoped prefix instead of `hq-`

### DIFF
--- a/internal/cmd/formula.go
+++ b/internal/cmd/formula.go
@@ -32,6 +32,7 @@ var (
 	formulaRunDryRun  bool
 	formulaRunAgent   string
 	formulaRunFiles   []string
+	formulaRunSet     []string
 	formulaCreateType string
 )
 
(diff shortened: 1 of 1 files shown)


ckumar1:
Friendly ping — is this approach heading in the right direction? Happy to adjust if the --set flag design or bead prefix fix doesn't fit the project's direction. Let us know if you'd prefer this structured differently.

ckumar1:
Superseded by #3497 (same fix, cleaner cherry-pick from carry/operational).

PR #3278, opened 2026-03-25
fix: graceful degradation when bd agent state is unavailable

ckumar1:
## Summary

- Added `sync.Once` probe to check whether `bd agent state` exists on first call
- When missing, skips the command silently and falls back to description field updates
- Eliminates ~2 minutes of retry noise per `gt sling` / polecat spawn

## Context

`bd agent state` was documented in the beads CHANGELOG v0.40.0 but was never shipped (steveyegge/beads#2828). The command does not exist in beads v0.62.0 or HEAD.

Gas Town calls `bd agent state` in `UpdateAgentState` on every polecat spawn, sling, and done. `SetAgentStateWithRetry` retries 10 times with exponential backoff before giving up, adding ~2 minutes of warning noise to every agent lifecycle event.

The description field update path (`UpdateAgentDescriptionFields`) still works and is now the sole update mechanism when `bd agent state` is unavailable. When beads ships the command, the probe will detect it and resume using the direct path automatically.

## Test plan

- [x] `go build ./cmd/gt`
- [x] `go vet ./internal/beads/`
- [x] Verified `bd agent --help` returns "unknown command" on beads 0.62.0
- [x] Probe caches result via `sync.Once` — no per-call overhead

### DIFF
--- a/internal/beads/beads_agent.go
+++ b/internal/beads/beads_agent.go
@@ -7,14 +7,33 @@ import (
 	"errors"
 	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
 	"strings"
+	"sync"
 
 	"github.com/gofrs/flock"
 
 	"github.com/steveyegge/gastown/internal/constants"
 	"github.com/steveyegge/gastown/internal/telemetry"
 )
+
+// bdAgentStateAvailable caches whether `bd agent state` exists.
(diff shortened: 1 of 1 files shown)


steveyegge:
Superseded by #3283 (merged). That PR already migrated `bd agent state` → `bd set-state` directly. The backward-compat probing in this PR is no longer needed.

PR #3279, opened 2026-03-25
fix: filter phantom escalations + operational plugin fixes

natorious:
## Summary

- **fix(escalate):** `gt escalate list` now cross-checks each entry against live Dolt and skips phantom IDs that no longer exist (ne-hkt). This happens when a rig's Dolt server dies and restarts fresh — stale label-based query results would surface entries that can't be acked or closed.
- **fix(plugins):** dolt-archive and dolt-backup auto-discover databases instead of using a hardcoded list that breaks when rigs are added/removed.
- **fix(dolt-backup):** Corrected PROD_DBS — was referencing nonexistent "bd" and "gt" databases, causing HIGH escalations on every backup cycle.
- **fix(compact-report):** Idempotency check now uses `--status=closed` so it correctly detects already-created audit beads.

## Test plan

- [ ] `gt escalate list` after Dolt restart shows no phantom entries, reports skipped count
- [ ] `gt escalate list` under normal conditions unchanged
- [ ] dolt-backup/dolt-archive run without hardcoded-database errors

### DIFF
--- a/internal/cmd/escalate_impl.go
+++ b/internal/cmd/escalate_impl.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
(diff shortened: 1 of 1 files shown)


steveyegge:
Fix-merged to main (rebased and pushed directly). Phantom escalation filtering is a good operational improvement.

PR #3290, opened 2026-03-25
fix: remove calls to removed `bd agent state` command (gt-3hn3)

seanbearden:
## Summary
- `bd agent state` was removed in bd upstream refactor (0bd598ce), but gt still called it causing 10-retry exponential backoff spam on every sling
- Makes the description field the authoritative source for `agent_state` instead of the stale DB column
- Deacon's `updateAgentBeadState` now uses `bd set-state` (label-based) for audit trail

## Changes
- **`internal/beads/beads_agent.go`**: `UpdateAgentState()` uses `UpdateAgentDescriptionFields` only; `GetAgentBead()` uses description-parsed state
- **`internal/cmd/deacon.go`**: Replace `bd agent state` with `bd set-state` (best-effort, label-based)
- **`internal/witness/handlers.go`**: `fetchAgentBeadSnapshot` and `getAgentBeadState` prefer description-parsed state with column fallback
- **`internal/daemon/lifecycle.go`**: `getAgentBeadInfo` prefers description-parsed state with column fallback
- **Tests**: Updated to match new description-authoritative semantics

## Test plan
- [x] `go build` passes (excluding unrelated gt-model-eval)
- [x] `go test ./internal/beads/` — all pass including updated agent state tests
- [x] `go test ./internal/witness/` — all pass
- [x] `go test ./internal/polecat/` — all pass
- [x] `go test ./internal/cmd/ -run TestDone` — all pass
- [ ] Verify no more retry spam in sling logs on live system

### DIFF
--- a/internal/beads/beads_agent.go
+++ b/internal/beads/beads_agent.go
@@ -415,25 +415,16 @@ func (b *Beads) ResetAgentBeadForReuse(id, reason string) error {
 	return nil
 }
 
-// UpdateAgentState updates the agent_state field in an agent bead.
-// Uses `bd agent state` command for the database column directly,
-// then syncs the description's agent_state field to match (gt-ulom).
+// UpdateAgentState updates the agent_state field in an agent bead's description.
+// The description is the authoritative source for agent_state (gt-3hn3).
+// Previously used `bd agent state` to update a DB column directly, but that
+// command was removed in bd upstream refactor (0bd598ce). The column is no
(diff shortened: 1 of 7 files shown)


steveyegge:
Superseded by #3283 (merged). That PR already migrated `bd agent state` → `bd set-state`. Your description-only approach diverges from the merged direction. Please rebase if additional changes are needed.

PR #3293, opened 2026-03-26
fix: skip orphan databases in EnsureAllMetadata to prevent metadata flip-flop

dannomayernotabot:
## Summary
- When both an orphan database (e.g., "gastown") and its canonical counterpart ("gt") exist in Dolt, `EnsureAllMetadata` processes both and writes different `dolt_database` values to the same `metadata.json`, causing confusing "identity mismatch repair" warnings that flip-flop on every `gt up`.
- Builds a reverse rig→canonical-DB map and skips databases that share a rig's directory name but aren't the rig's canonical database.
- Includes test covering the exact scenario.

## Test plan
- [x] `TestEnsureAllMetadata_SkipsOrphanDatabases` — new test passes
- [x] All existing `TestEnsureAllMetadata*` tests pass (5/5)
- [x] `gt up` no longer shows flip-flop warnings

### DIFF
--- a/internal/doltserver/doltserver.go
+++ b/internal/doltserver/doltserver.go
@@ -2977,11 +2977,31 @@ func EnsureAllMetadata(townRoot string) (updated []string, errs []error) {
 		dbToRig[k] = v
 	}
 
+	// Build reverse map: rig → canonical DB name. When an orphaned database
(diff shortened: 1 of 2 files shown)


steveyegge:
Superseded by #3287 (fixes the same EnsureAllMetadata oscillation with DB-to-rig grouping) which was merged.

PR #3323, opened 2026-03-26
feat: inject GitHub CI failure logs into polecat MERGE_FAILED nudges

gagan114662:
## Summary

- Adds `GetFailedStepLogs()` to `githubci.Client` — calls `gh run view --log-failed` and tails output to 60 lines / 3000 bytes
- `reliability.EnsureGitHubBranchCI` now embeds the failed step logs in the error when CI fails, so the error message contains the actual stack trace rather than just a run URL
- `refinery.ProcessMRInfo` now sets `TestsFailed: true` for GitHub CI failures so the polecat nudge says `type=tests` instead of `type=build`

**Before:** polecat gets `MERGE_FAILED: type=build error=github ci assurance failed: workflow run failed: conclusion=failure (https://github.com/.../runs/12345)`

**After:** polecat gets the actual failed step logs inline, fixing itself without human intervention

## Test plan

- [x] `go test ./internal/githubci/...` — new `TestGetFailedStepLogs_Success` and `TestGetFailedStepLogs_Error` pass
- [x] `go test ./internal/refinery/...` — all existing tests pass

### DIFF
--- a/internal/githubci/githubci.go
+++ b/internal/githubci/githubci.go
@@ -263,6 +263,31 @@ func (c *Client) resolveRepo(ctx context.Context, opts EnsureOptions) (string, e
 	return repo, nil
 }
 
+// GetFailedStepLogs returns the log output of failed steps for a workflow run.
+// It calls `gh run view --log-failed` and tails the output so it fits in a nudge.
+// Returns empty string on any error (best-effort — non-fatal).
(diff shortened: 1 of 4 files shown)


gagan114662:
Opened against wrong repo — this work belongs in gagan114662/gastown fork.

PR #3330, opened 2026-03-27
fix: formula lookup falls back to disk when not embedded in binary

totalslacker:
## Summary
- `GetEmbeddedFormulaContent` now checks disk paths when a formula isn't found in the embedded FS
- Search order: embedded → `.beads/formulas/` (CWD) → `~/.beads/formulas/` → `$GT_ROOT/.beads/formulas/`
- All callers (`showFormulaSteps`, `showFormulaStepsFull`, `resolveFormulaToTempFile`) get the fallback automatically

## Motivation
User-defined formulas (like `mol-evolve` for rig-seed evolution workflows) exist on disk but aren't compiled into the `gt` binary. When `gt prime --hook` renders formula steps for polecats, it only checked the embedded FS, causing "formula not found" warnings. Polecats then skipped all formula steps (journal entries, roadmap updates, `gt mq submit`), leading to incomplete work that never reached the merge queue.

## Test plan
- [x] Existing `TestGetEmbeddedFormulaContent` passes (embedded path unchanged)
- [ ] Manual: place a formula in `$GT_ROOT/.beads/formulas/`, verify `gt prime` surfaces it
- [ ] Manual: sling work with a disk-only formula, verify polecat sees the steps

### DIFF
--- a/internal/daemon/handler.go
+++ b/internal/daemon/handler.go
@@ -3,6 +3,7 @@ package daemon
 import (
 	"fmt"
 	"path/filepath"
+	"sort"
 	"time"
 
 	"github.com/steveyegge/gastown/internal/config"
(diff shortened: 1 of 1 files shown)


steveyegge:

PR #3340, opened 2026-03-27
feat(daemon): implement stuck-agent-dog auto-restart

AndrewSkyflo:
## Summary

- Implements the stuck-agent-dog plugin that was referenced but never built in `checkDeaconHeartbeat`
- Replaces three "Detection only" comment stubs with actual kill-and-restart logic
- Adds `restartStuckDeacon` method: kills tmux session, respawns via `ensureDeaconRunning`, verifies, notifies Slack
- Adds `notifySlack` helper: calls `gt-notify` script (zero token cost, best-effort)
- Uses existing `RestartTracker` for exponential backoff and crash-loop prevention
- Also fixes merge conflict in `internal/hooks/installer.go`

## Context

The daemon detected stuck Deacon sessions every 3 minutes but never acted on them. In production, a stuck Deacon went undetected for 9+ hours until a human noticed. This fix closes the loop.

## Test plan

- [x] Binary builds cleanly
- [x] Deployed to production Gas Town instance
- [x] Stuck-agent-dog fired on first heartbeat, killed stuck session, respawned successfully
- [x] Slack notification delivered to admin channel
- [ ] Crash-loop detection prevents runaway restarts (verified via code review, RestartTracker already tested)

### DIFF
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -1344,9 +1344,10 @@ func (d *Daemon) checkDeaconHeartbeat() {
 				return
 			}
 			// Grace period expired without any heartbeat - Deacon failed to start
-			// Detection only: stuck-agent-dog plugin handles context-aware restart
+			// Stuck-agent-dog: kill and restart
 			d.logger.Printf("STUCK DEACON: started %s ago but hasn't written heartbeat (session: %s)",
 				timeSinceStart.Round(time.Minute), sessionName)
+			d.restartStuckDeacon(sessionName, fmt.Sprintf("no heartbeat after %s", timeSinceStart.Round(time.Minute)))
 			return
 		}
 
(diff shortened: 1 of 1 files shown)


A3Ackerman:
This addresses a real gap — we've seen stuck deacons go undetected for hours. A couple of related issues we've tracked that this should help with:

- **gt-ydy**: The `stuck-agent-dog` `run.sh` diverged from the `plugin.md` spec — it parses `gt hook` output with `grep 'Hooked: [^ ]+'` but the output format changed. The `plugin.md` spec uses `bd show --json | jq` which is more robust. Worth checking whether this PR's implementation inherited the fragile grep parsing or uses the JSON approach.

- **gt-bex** (now closed): The plugin had invalid `gt hook` syntax when checking polecat health, causing it to fail silently — stuck polecats weren't being detected at all.

The `RestartTracker` + exponential backoff approach looks solid for crash-loop prevention. Main concern: if the underlying cause of the stuck deacon isn't fixed (e.g., deacon patrol formula issues — see gt-wjz where the patrol loop dies after 1-2 cycles), auto-restart might just churn through restarts without the deacon ever achieving sustained patrol.

PR #3347, opened 2026-03-27
scanner: validate plugin frontmatter delimiters and gate fields

totalslacker:
## Summary
- Scanner now returns clear errors when plugins use YAML-style `---` delimiters instead of TOML `+++`
- Scanner detects when cooldown gates use `interval` instead of the correct `duration` field
- Scanner warns when cooldown gates have no duration set (will fire every patrol cycle)
- Also includes `.disabled` suffix support for disabling plugins by directory name

## Context
The `evolve` plugin (daily autonomous evolution for rig-seed projects) was silently ignored since creation because of two bugs in its `plugin.md`:
1. Used `---` frontmatter delimiters — scanner only accepts `+++`, silently returned nil
2. Used `interval = "24h"` — the Gate struct field is `duration`, so the value was silently dropped

Both went undetected because the scanner's error handling returned `nil` (skip) rather than surfacing the parse failure. This meant the evolve plugin **never ran**, blocking autonomous evolution cycles entirely.

## Test plan
- [x] `TestParsePluginMD_YAMLDelimiterError` — verifies clear error for `---` delimiters
- [x] `TestParsePluginMD_IntervalInsteadOfDuration` — verifies clear error for `interval` field
- [x] `TestScanner_SkipsDisabledPlugins` — verifies `.disabled` suffix directories are skipped
- [x] Full plugin test suite passes (all existing tests unaffected)

### DIFF
--- a/internal/daemon/handler.go
+++ b/internal/daemon/handler.go
@@ -231,12 +231,15 @@ func (d *Daemon) dispatchPlugins(mgr *dog.Manager, sm *dog.SessionManager, rigsC
 		return
 	}
 
+	d.logger.Printf("Handler: discovered %d plugins with cooldown gates", countCooldownPlugins(plugins))
+
 	if len(plugins) == 0 {
 		return
 	}
 
 	recorder := plugin.NewRecorder(d.config.TownRoot)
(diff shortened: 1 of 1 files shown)


steveyegge:

PR #3351, opened 2026-03-27
handler: fix plugin dispatch starvation and stale session blocking

totalslacker:
## Summary
- Sort cooldown plugins by duration descending before dispatch, so long-cooldown plugins (evolve 24h) get dispatched before short-cooldown ones (github-sheriff 5m)
- Track dogs that fail session start and skip them in subsequent iterations, preventing a single stale tmux session from blocking all dispatch
- Add debug logging for discovered plugin count

## Context
Two bugs compounded to prevent the evolve plugin from ever being dispatched by the daemon:

1. **Stale session blocking**: When a dog has `state=idle` but a stale tmux session, `GetIdleDog()` returns it repeatedly (reads from disk each call). Every plugin in the dispatch loop fails on the same dog, blocking all dispatch.

2. **Iteration starvation**: Plugins were iterated from a Go map (non-deterministic order). Short-cooldown plugins (github-sheriff 5m, stuck-agent-dog) consumed all idle dogs on every ~3 min heartbeat cycle. The evolve plugin (24h cooldown) never got dispatched because it never won the iteration race.

## Test plan
- [x] Builds clean
- [x] Verified stale-session fix: after killing stale tmux session, dispatch resumed to multiple dogs
- [x] Verified sort order: `time.ParseDuration` correctly parses gate durations, longer durations sort first
- [ ] Manual verification: evolve plugin dispatches on next heartbeat cycle after daemon restart

### DIFF
--- a/internal/daemon/handler.go
+++ b/internal/daemon/handler.go
@@ -3,6 +3,7 @@ package daemon
 import (
 	"fmt"
 	"path/filepath"
+	"sort"
 	"time"
 
 	"github.com/steveyegge/gastown/internal/config"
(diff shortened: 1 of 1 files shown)


steveyegge:

PR #3352, opened 2026-03-27
feat: persist tmux socket mode per town

Henry-E:
## Summary
- add a persisted `tmux_socket_mode` town setting with `default` and `auto` modes
- make `InitRegistry` honor that setting when `GT_TMUX_SOCKET` is unset, while keeping explicit env overrides
- initialize the first installed HQ on a machine with `tmux_socket_mode=default`, and later installs with `tmux_socket_mode=auto`
- expose `tmux.socket_mode` through `gt config get/set`

## Why
Current `main` rewrites unset / `default` / `auto` into a derived per-town socket name, so there is no way for a town to intentionally use tmux's real default socket without patching code.

This keeps legacy towns unchanged unless they opt in via settings, while giving new installs an explicit persisted choice.

## Implementation notes
- existing towns without `tmux_socket_mode` keep the current per-town socket behavior for backward compatibility
- explicit `GT_TMUX_SOCKET` still wins for tests and one-off operator overrides
- install records a lightweight machine-local install state so only the first HQ defaults to the real tmux default socket

## Test plan
- `go test ./internal/session ./internal/state ./internal/config`
- `go test -tags=e2e ./internal/cmd -run 'TestInstallCreatesCorrectStructure|TestInstallSecondTownDefaultsToAutoTmuxSocketMode'`

### DIFF
--- a/internal/session/registry.go
+++ b/internal/session/registry.go
@@ -28,6 +28,12 @@ type PrefixRegistry struct {
 	rigToPrefix map[string]string // "gastown" → "gt"
 }
 
+type socketSessionLister interface {
+	ListSessions() ([]string, error)
+}
+
+var socketListerForTest func(socket string) socketSessionLister
+
 // NewPrefixRegistry creates an empty prefix registry.
 func NewPrefixRegistry() *PrefixRegistry {
 	return &PrefixRegistry{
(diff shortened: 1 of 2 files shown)


Henry-E:
Honestly not sure if this is the right way to solve the issues introduced by #3118 but gastown default should not be to put agents into an unknown `-L gt-fdasfe` tagged socket. Default should be to go to unnamed socket. This will also help to solve issue when people upgrade their gas town next. It won't cause split brain issues, at least it shouldn't ideally.

I try to avoid making PRs like this because I don't understand the internals too well of gas town. But not defaulting to the unnamed / default tmux socket for the very first instance of a gas town will be both frustrating for users used to manually checking active sessions using `tmux ls`, as well as causing issues to users that update their gastown with any active agents, since now there will be a split between the default socket and this new named socket with no clean migration in place. No migration should be needed though if it just stays at the default socket.

htristan:
Interesting approach — I think there's a real insight here about single-town users not needing to pay the cost of multi-town socket isolation. A few thoughts:

**What I like:**
- The upgrade path that sniffs which socket currently has active sessions and persists the right mode is thoughtful
- First-town-default is pragmatic for the majority of users who only run one town
- Making it user-configurable via `gt config set` is a nice escape hatch

**Concerns:**

1. **Complexity vs. the problem size.** This is ~1200 lines across 11 files for what is fundamentally a socket name decision. It adds install-time state tracking (`state.RecordInstall`), a new config field, upgrade migration logic, and registry changes. That's a lot of ongoing surface area.

2. **First-install detection is fragile.** It relies on `~/.local/state/gastown/state.json` existence. If that file gets deleted, or the user sets up a fresh machine, the heuristic breaks — install order permanently determines socket behavior with no way to recover automatically.

3. **Upgrade sniffing when idle.** If no sessions are running at upgrade time, `upgradeTmuxSocketMode` falls back to `default`. A multi-town user who upgrades while idle would silently lose per-town isolation. That's the kind of thing that bites people weeks later.

4. **The conflict case errors out.** If sessions exist on both the default and a named socket (the exact split-brain scenario), the upgrade tells the user to resolve it manually. That's the hardest case for users to deal with.

**The deeper question:**

I think the real fork in the road here is whether `tmux ls` (on the default socket) showing gastown sessions is a design goal, or whether gastown should own its own session discovery via `gt status` / a future `gt tmux` command that passes the right `-L` flag.

If bare `tmux ls` is a goal → Henry-E's approach (or something like it) makes sense as the default for single-town setups.

If gastown owns session visibility → per-town sockets can just be the default for everyone, though the upgrade/migration path would need to be more robust than what #3118 has today. As we discussed on that PR, users upgrading with active sessions on the default socket would hit a split-brain situation without proper detection and migration — so either way, the migration story needs to be tightened up regardless of which direction we go.

@steveyegge — curious what your preferred end-state is here. That would help figure out whether this PR, the simpler restore in #3118, or some hybrid is the right path forward.

Henry-E:
It's crazy that #3118 was allowed into the released version. I've had a lot of frustrations working with gas town the last two months but this is really one where I'd consider rage quitting over.

steveyegge:

Henry-E:
@steveyegge  Strongly recommend having gas town default to the unnamed socket as a default for v1. Having it default to an oddly named socket is a majorly inconvenient change meant to support a tiny fraction of users with multiple gas towns on the same machine.

PR #3353, opened 2026-03-27
fix(deacon): stop undocking rigs that Mayor intentionally docked

KeithWyatt:
## Summary
- Deacon health-scan patrol step already says to skip docked rigs, but the deacon ignores this and "restores" docked rigs by undocking them and restarting witness/refinery
- This creates an undock/dock fight loop when the Mayor intentionally docks a rig
- **Role template**: Added explicit "Docked Rigs Are Off-Limits" section making docked rigs invisible to the deacon -- no checks, no escalation, no action
- **Patrol formula**: Expanded the skip-docked instruction to explicitly prohibit undocking, escalating about, or "restoring" docked rigs

## Test plan
- [ ] Dock a rig with `gt rig dock <rig>`, verify deacon patrol skips it entirely
- [ ] Verify deacon does not send escalation mail about docked rigs
- [ ] Verify operational rigs still get health-scanned normally

### DIFF
--- a/internal/formula/formulas/mol-deacon-patrol.formula.toml
+++ b/internal/formula/formulas/mol-deacon-patrol.formula.toml
@@ -500,11 +500,14 @@ gt rig status <rig>
 # Check the Status: line - if DOCKED or PARKED, skip entirely
 ```
 
-DOCKED rigs are globally shut down - do NOT:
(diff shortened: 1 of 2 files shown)


Henry-E:
that deacon is a real sumabitch

PR #3324, opened 2026-03-26
fix(beads): support bd 0.62 agent state flow

junlov:
## Summary
- switch Gastown agent-state writes away from the removed `bd agent state` path
- prefer description-backed agent_state with legacy structured-column fallback
- override stale inherited Dolt connection env when targeting another beads dir

## Verification
- `go test ./internal/beads -run 'Test(GetAgentBead_PrefersDescriptionAgentState|GetAgentBead_FallsBackToDescriptionAgentState|UpdateAgentState_UsesUpdateDescriptionPath|BuildRunEnv_OverridesStaleDoltPortFromBeadsDir|BuildRoutingEnv_OverridesStaleDoltPortFromBeadsDir)$'`\n- `go test -c -o /tmp/pr-bd-daemon.test ./internal/daemon`\n- `go test -c -o /tmp/pr-bd-witness.test ./internal/witness`\n\n## Notes\n- kept draft because `internal/daemon` runtime tests require Docker/testcontainers in this environment

PR #3356, opened 2026-03-27
fix(beads): keep agent state updates on target db

Henry-E:
## Summary
- run `bd set-state` against the same explicit beads DB as the agent bead itself
- avoid stripping `BEADS_DIR` in `UpdateAgentState`, which could send the state update to a different database
- add a regression test that fails if agent state updates stop honoring the explicit target beads dir

## Testing
- go test ./internal/beads -run 'Test(UpdateAgentState_UsesExplicitBeadsDir|GetAgentBead_PrefersStructuredAgentState|GetAgentBead_FallsBackToDescriptionAgentState)$' -count=1


steveyegge:
Fix-merged to main in commit 30265a9c. Thank you @Henry-E\! Routing target resolution incorporated into the reconciled agent state flow (combined with PR #3324's description-only write approach).

PR #3373, opened 2026-03-28
fix(beads): detect --allow-stale support via output, not exit code

athosmartins:
## Summary

- bd v0.60+ exits with code 0 even when passed an unknown flag, printing the error to stderr instead
- `BdSupportsAllowStaleWithEnv` checked only `cmd.Run() == nil` → always returned `true` on bd v0.60+
- Every `bd` call was incorrectly prefixed with `--allow-stale`, causing them all to fail with "unknown flag"
- Fix: capture combined output and check for `"unknown flag"` string

## Root cause trace

```
bd --allow-stale version
# Output: "Error: unknown flag: --allow-stale\n..."
# Exit code: 0  ← bug in detection
```

This caused `gt patrol new` and `gt patrol report` to emit warnings on every call.

## Test plan
- [ ] `gt patrol new --role=witness` no longer emits `unknown flag: --allow-stale`
- [ ] `BdSupportsAllowStale()` returns `false` when bd exits 0 with "unknown flag" in output

### DIFF
--- a/internal/beads/beads.go
+++ b/internal/beads/beads.go
@@ -72,7 +72,13 @@ func BdSupportsAllowStaleWithEnv(env []string) bool {
 	if env != nil {
 		cmd.Env = env
 	}
-	supported := cmd.Run() == nil
+	var combinedOut bytes.Buffer
+	cmd.Stdout = &combinedOut
+	cmd.Stderr = &combinedOut
+	_ = cmd.Run()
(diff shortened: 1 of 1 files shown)


steveyegge:
Cherry-picked --allow-stale fix to main (commit 96e1eb86). Thank you @athosmartins!

PR #3378, opened 2026-03-28
fix: fail fast on daemon beads schema mismatch

### LINKED ISSUE #3377: Add daemon startup compatibility check for beads schema / binary mismatch
## Summary
`gt daemon run` and `gt daemon start` should fail fast with a clear error when the binary is too old for the workspace's beads schema, instead of starting successfully and degrading into repeated Convoy poll errors.

## Motivation
Issue #3076 identified a real operator hazard: an older `gastown` binary built against the legacy beads event API can run against a newer workspace schema and silently lose Convoy event polling.

The main UUID event-ID bug is already fixed on current `main`, but we still lack an explicit compatibility guard for future binary/schema drift.

## Desired behavior
- Detect incompatible beads event schema / API expectations during daemon startup.
- Exit with an actionable compatibility error instead of repeatedly logging poll failures.
- Point the user to the remediation path (upgrade `gt`, rebuild from source, or use a matching release).

## Acceptance criteria
- [ ] `gt daemon run` / `gt daemon start` performs a compatibility check before starting Convoy polling
- [ ] incompatible schema / binary combinations fail fast with a clear message
- [ ] add regression coverage for the incompatibility path
- [ ] documentation or release notes mention the compatibility requirement

## Related
- Follow-up from #3076

fusioncid:
## Summary
- add a daemon startup compatibility guard for opened beads stores before Convoy polling starts
- compare workspace `bd_version` against the binary's embedded beads module version and run a lightweight event-table probe
- surface startup mismatch errors directly from `gt daemon start`, with regression tests and changelog note

## Testing
- go test ./internal/daemon -run 'TestCheckBeadsStoreCompatibility|TestDaemon_StartsManagerAndScanner|TestDaemon_StopsManagerAndScanner' -count=1
- go test ./internal/cmd -run 'TestReadDaemonStartupFailure' -count=1
- go test ./internal/daemon -run ^ -count=1
- go test ./internal/cmd -run ^ -count=1

Closes #3377

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **Daemon beads compatibility guard** — `gt daemon run` now fail-fast checks
+  workspace beads schema compatibility before Convoy polling starts, and
+  `gt daemon start` surfaces the startup mismatch directly instead of only
+  telling operators to inspect logs.
+
 ## [0.12.1] - 2026-03-15
 
 ### Added
--- a/internal/cmd/daemon.go
+++ b/internal/cmd/daemon.go
@@ -6,6 +6,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"runtime"
+	"strings"
 	"time"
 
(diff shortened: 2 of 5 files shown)


steveyegge:
Fix-merged to main. Thank you @fusioncid!

PR #3379, opened 2026-03-28
fix: prevent source installs from being shadowed

### LINKED ISSUE #3077: Source install can stay shadowed by Homebrew gt because ~/.local/bin is not prioritized

fusioncid:
## Summary
- prepend `~/.local/bin` and `~/go/bin` from the shell hook so source installs win on fresh shells
- add a `gt doctor` / `gt upgrade` infrastructure check that warns when a canonical source install is shadowed on PATH
- make `make install` emit a post-install warning when another `gt` would still win, and update install docs to the canonical source flow

## Testing
- go test ./internal/doctor -run 'TestGTBinaryShadowCheck|TestIsLikelyHomebrewGT' -count=1
- go test ./internal/shell -run 'TestShellHookScript|TestAddRemoveFromRCFile|TestUpdateRCFile' -count=1
- go test ./internal/cmd -run ^ -count=1

Closes #3077

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **Source install PATH shadowing** — shell integration now prepends
+  `~/.local/bin`, `gt doctor` warns when a canonical source install is still
+  shadowed on PATH, and `make install` emits a post-install warning when an
+  older `gt` would still win.
+
 ## [0.12.1] - 2026-03-15
 
 ### Added
--- a/Makefile
+++ b/Makefile
@@ -106,6 +106,16 @@ install: check-up-to-date build
 		fi; \
 	done
 	@echo "Installed $(BINARY) to $(INSTALL_DIR)/$(BINARY)"
+	@resolved_gt=$$(command -v $(BINARY) 2>/dev/null || true); \
+	if [ "$$resolved_gt" != "$(INSTALL_DIR)/$(BINARY)" ]; then \
+		echo "WARNING: PATH still resolves gt to $${resolved_gt:-<not found>} instead of $(INSTALL_DIR)/$(BINARY)"; \
+		case "$$resolved_gt" in \
+			/usr/local/bin/$(BINARY)|/opt/homebrew/bin/$(BINARY)) \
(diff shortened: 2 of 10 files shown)


steveyegge:

PR #3380, opened 2026-03-28
fix(handoff): fatal on Dolt failure, log after persist

fernanja:
## Problem

When Dolt is down during `gt handoff`, two bugs cause silent data loss:

1. `LogHandoff()` writes to town.log BEFORE `sendHandoffMail()` persists to Dolt. If Dolt is down, town.log records a false handoff that never persisted.
2. `sendHandoffMail` failure is swallowed with `PrintWarning` and the session respawns anyway. The next session finds an empty hook — all handoff context is lost.

## Root Cause

Discovered during a 43-hour Dolt commit gap where 3 mayor handoffs were logged to town.log but never persisted. The successor sessions found empty hooks and lost all context.

## Fix

**`runHandoff()`:**
- Move `LogHandoff()` call AFTER `sendHandoffMail()` succeeds
- Make persistence failure fatal: return error, do NOT respawn
- Log `EventHandoffNoPersist` on failure so crash recovery can distinguish real from failed handoffs
- Extract `sessionToGTRole()` above both branches to avoid duplication

**`runHandoffCycle()`:**
- Same fix: fatal on persistence failure, log after persist
- Blocks compaction cycling when Dolt is down (correct — prevents context loss)

**`runHandoffAuto()`:**
- Intentionally left non-fatal: auto is a state-save-only path (no respawn), different failure semantics

## Audit Notes

- `LogHandoff()` has exactly 2 callers: `runHandoff` and `runHandoffCycle`. Both fixed.
- `EventHandoffNoPersist` handled in all 4 display/parsing functions: `logger.go` (formatLogLine), `log.go` (printEvent + formatEventDetail), `audit.go` (formatTownlogSummary).
- The `--yes` flag and `term.IsTerminal` confirmation prompt (added in a4611422) are preserved — this PR does not touch that code path.
- When `sendHandoffMail` fails, the tmux pane stays alive (no respawn). The human can fix Dolt and retry `gt handoff`.

## Test plan

- [x] `go build ./...`
- [x] `go test ./internal/townlog/... -run TestEventHandoffNoPersist` — 3 tests pass
- [x] `go test ./internal/cmd/... -run TestSling` — no regressions

### DIFF
--- a/internal/cmd/audit.go
+++ b/internal/cmd/audit.go
@@ -380,6 +380,8 @@ func formatTownlogSummary(e townlog.Event) string {
 		return "Completed work"
 	case townlog.EventHandoff:
 		return "Handed off session"
+	case townlog.EventHandoffNoPersist:
+		return "Handoff FAILED (Dolt persistence)"
 	case townlog.EventCrash:
 		if e.Context != "" {
 			return fmt.Sprintf("Crashed: %s", e.Context)
--- a/internal/cmd/handoff.go
+++ b/internal/cmd/handoff.go
@@ -267,15 +267,10 @@ func runHandoff(cmd *cobra.Command, args []string) error {
 	// Handing off ourselves - print feedback then respawn
 	fmt.Printf("%s Handing off %s...\n", style.Bold.Render("🤝"), currentSession)
 
-	// Log handoff event (both townlog and events feed)
-	if townRoot, err := workspace.FindFromCwd(); err == nil && townRoot != "" {
-		agent := sessionToGTRole(currentSession)
-		if agent == "" {
-			agent = currentSession
-		}
-		_ = LogHandoff(townRoot, agent, handoffSubject)
-		// Also log to activity feed
-		_ = events.LogFeed(events.TypeHandoff, agent, events.HandoffPayload(handoffSubject, true))
+	// Resolve agent identity once for both success and failure paths.
+	agent := sessionToGTRole(currentSession)
+	if agent == "" {
+		agent = currentSession
 	}
 
 	// Dry run mode - show what would happen (BEFORE any side effects)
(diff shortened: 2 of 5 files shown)


fernanja:
## Design Alternatives Considered

### Fatal vs non-fatal on persistence failure

The current implementation makes `sendHandoffMail` failure **fatal** (blocks respawn). Two alternative approaches were considered:

**Option A (chosen): Fatal — block respawn until persistence succeeds.**
Rationale: A respawned session with an empty hook has lost all context. Silent data loss is worse than a blocked session. The human can diagnose and retry.

**Option B: Retry with backoff, then respawn anyway with a warning.**
This preserves availability (the session always cycles) at the cost of potential context loss. The warning would need to be unmissable — `PrintWarning` was insufficient (the original bug). A more prominent mechanism (e.g., writing a marker file that the successor reads) could work but adds complexity.

**Option C: Retry with backoff, then fail fatal only after exhausting retries.**
A middle ground: transient Dolt failures recover automatically, persistent failures still block. This adds ~30 seconds of retry overhead to every handoff when Dolt is down. It may be preferred if Dolt instability is common in some deployments.

The fatal approach was chosen because it is the simplest to reason about and provides the strongest data integrity guarantee. If a less strict approach is preferred, Option C could be implemented as a follow-up without changing the log-ordering fix (which is correct regardless of error handling strategy).

### runHandoffAuto left non-fatal

`runHandoffAuto()` is intentionally not changed. It is a state-save-only path (no respawn) used by PreCompact hooks. If persistence fails, the marker file is still written, preserving at least some recovery signal. Making auto fatal would block compaction, which is a different failure mode with different trade-offs.

fernanja:
## Windows CI Failure — Pre-existing, Not From This PR

The Windows Build failure is in `internal/cmd/estop.go` (lines 258, 287, 345) which uses Unix-only syscalls (`syscall.SIGTSTP`, `syscall.SIGCONT`, `syscall.Kill`) without build tags. This file was not modified by this PR.

```
##[error]internal\cmd\estop.go:258:49: undefined: syscall.SIGTSTP
##[error]internal\cmd\estop.go:287:49: undefined: syscall.SIGCONT
##[error]internal\cmd\estop.go:345:17: undefined: syscall.Kill
```

All other checks pass: Test, Integration Tests, Lint, Reject go.mod replace directives, Reject issues.jsonl.

steveyegge:
Fix-merged to main. Thank you @fernanja!

PR #3381, opened 2026-03-28
feat(health): add commit freshness check to Dolt health metrics

fernanja:
## Problem

`gt health` monitors connections, latency, disk usage, and read-only state — but has no way to detect a server that is reachable but not committing. A 43-hour commit gap went undetected because no metric checked commit freshness.

## Fix

Add `GetLastCommitAge()` that queries `MAX(date) FROM dolt_log` across all databases and reports the age of the most recent commit.

**HealthMetrics (internal struct):**
- `LastCommitAge time.Duration` — age of most recent commit
- `LastCommitDB string` — which database had the most recent commit
- Warning added to `metrics.Warnings` when gap >1 hour

**ServerHealth (JSON output from `gt health`):**
- `last_commit_age_seconds float64` — converted to seconds for clean JSON
- `last_commit_db string` — database name

## Implementation Details

- Uses `database/sql` (like `MeasureQueryLatency`) for accurate timing, not subprocess
- Uses `config.EffectiveHost()` for remote Dolt support
- `ListDatabases()` filters system databases (information_schema, mysql) — only user databases queried
- 15-second total timeout across all databases; failures skip gracefully
- Dolt's `dolt_log.date` is DATETIME(6) (microsecond precision); Go's ".999" format correctly parses any fractional digit count
- `HealthMetrics.LastCommitAge` JSON tag is `last_commit_age_ns` (accurate — `time.Duration` serializes as nanoseconds). Noted pre-existing same issue on `QueryLatency` tag.

## Audit Notes

- SQL injection safe: `dbName` comes from `ListDatabases()` which reads filesystem entries / SHOW DATABASES (not user input), and is backtick-quoted
- NULL handling: `MAX(date)` on empty `dolt_log` returns NULL → Scan gets empty string → `time.Parse` fails → `continue` to next database (correct graceful degradation)
- 1-hour threshold justified by the 43-hour incident and Gas Town's commit frequency (agents commit every few minutes)
- Zero performance concern: O(N) queries where N = number of databases (typically 3-5), each query hits an index

## Test plan

- [x] `go build ./...`
- [x] `go test ./internal/doltserver/... -run "TestGetLastCommitAge|TestHealthMetrics"` — 3 tests pass
- [x] `go test ./internal/cmd/... -run TestSling` — no regressions

### DIFF
--- a/internal/cmd/health.go
+++ b/internal/cmd/health.go
@@ -35,14 +35,16 @@ type HealthReport struct {
 }
 
 type ServerHealth struct {
-	Running        bool          `json:"running"`
-	PID            int           `json:"pid,omitempty"`
-	Port           int           `json:"port,omitempty"`
-	LatencyMs      int64         `json:"latency_ms,omitempty"`
-	Connections    int           `json:"connections,omitempty"`
-	MaxConnections int           `json:"max_connections,omitempty"`
-	DiskUsageBytes int64         `json:"disk_usage_bytes,omitempty"`
-	DiskUsageHuman string        `json:"disk_usage_human,omitempty"`
+	Running            bool    `json:"running"`
+	PID                int     `json:"pid,omitempty"`
+	Port               int     `json:"port,omitempty"`
+	LatencyMs          int64   `json:"latency_ms,omitempty"`
+	Connections        int     `json:"connections,omitempty"`
+	MaxConnections     int     `json:"max_connections,omitempty"`
+	DiskUsageBytes     int64   `json:"disk_usage_bytes,omitempty"`
+	DiskUsageHuman     string  `json:"disk_usage_human,omitempty"`
+	LastCommitAgeSec   float64 `json:"last_commit_age_seconds,omitempty"`
+	LastCommitDB       string  `json:"last_commit_db,omitempty"`
 }
 
 type DatabaseHealth struct {
@@ -167,6 +169,10 @@ func checkServerHealth(townRoot string) *ServerHealth {
 	sh.MaxConnections = metrics.MaxConnections
 	sh.DiskUsageBytes = metrics.DiskUsageBytes
 	sh.DiskUsageHuman = metrics.DiskUsageHuman
+	if metrics.LastCommitAge > 0 {
+		sh.LastCommitAgeSec = metrics.LastCommitAge.Seconds()
+		sh.LastCommitDB = metrics.LastCommitDB
+	}
 
 	return sh
 }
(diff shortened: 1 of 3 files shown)


fernanja:
## Design Alternatives Considered

### Hardcoded vs configurable threshold

The 1-hour commit gap threshold is hardcoded. Two alternatives were considered:

**Option A (chosen): Hardcoded 1 hour.**
Rationale: This is a health monitor, not a user-facing setting. The threshold reflects a reasonable expectation that a production Gas Town instance commits at least once per hour (agents create beads, mail, and state updates continuously). A configurable threshold adds complexity without clear benefit — if the threshold needs changing, it is a single constant.

**Option B: Configurable via `town-settings.json` (operational.dolt.commit_freshness_threshold).**
This would follow the existing pattern for operational thresholds (`PressureCPUThreshold`, `PressureMemThresholdGB`, etc.). If deployments exist where hourly commits are not expected, this would prevent false warnings. Could be added as a follow-up if needed.

### JSON tag naming

`HealthMetrics.LastCommitAge` uses `json:"last_commit_age_ns"` because `time.Duration` serializes as nanoseconds. This is accurate but inconsistent with the pre-existing `QueryLatency` tag (`json:"query_latency_ms"`) which has the same Duration-as-nanoseconds issue.

The chosen approach is to be accurate on the new field and note the pre-existing inconsistency via a TODO comment, rather than fixing `QueryLatency`'s tag in this PR (which could break external consumers parsing that tag name). A follow-up PR could address the `QueryLatency` tag if desired.

### ServerHealth.LastCommitAgeSec (float64)

The `gt health --json` output uses `float64` seconds (via `.Seconds()` conversion) rather than exposing the raw `time.Duration` nanoseconds. This follows the existing pattern where `LatencyMs` is `int64` milliseconds (via `.Milliseconds()`), giving JSON consumers human-readable units.

### Scope of this PR

This PR could be split into two:
1. Core: `GetLastCommitAge()` + `HealthMetrics` fields + warning
2. Surface: `ServerHealth` JSON fields in `gt health`

They were kept together because the health output is the primary consumer and shipping the metric without surfacing it would make it invisible to operators. If a smaller initial scope is preferred, the `health.go` changes can be extracted into a follow-up.

fernanja:
## Windows CI Failure — Pre-existing, Not From This PR

The Windows Build failure is in `internal/cmd/estop.go` (lines 258, 287, 345) which uses Unix-only syscalls (`syscall.SIGTSTP`, `syscall.SIGCONT`, `syscall.Kill`) without build tags. This file was not modified by this PR.

```
##[error]internal\cmd\estop.go:258:49: undefined: syscall.SIGTSTP
##[error]internal\cmd\estop.go:287:49: undefined: syscall.SIGCONT
##[error]internal\cmd\estop.go:345:17: undefined: syscall.Kill
```

All other checks pass: Test, Integration Tests, Lint, Reject go.mod replace directives, Reject issues.jsonl.

PR #3386, opened 2026-03-28
fix(dolt): reduce Dolt server crash probability from SIGTERM races (gt-9bxzs)

athosmartins:
## Summary

- **doltserver.Stop()**: add pre-SIGTERM connection drain — waits up to 10s for active queries to complete before sending SIGTERM, shrinking the window where SIGTERM interrupts Dolt's NomsBlockStore nbs_manifest temp file writes
- **RigConfigSyncCheck.Fix()**: guard DB-rename restart — skip Stop/Start cycle if server has been running < 60s, preventing crash-triggering SIGTERM during Dolt startup churn (doctor_dog patrol)

## Root cause

Upstream Dolt bug: `NomsBlockStore.Close()` calls `Fatalf()` (panics) on missing `nbs_manifest_<N>` temp file during SIGTERM shutdown. Confirmed in 3 databases over 5 weeks. Upstream bug report needed separately.

## Also filed

- `gt-z6c4m`: missing `crystallizes` column migration in beads repo (unrelated to crashes but discovered during investigation)

## Test plan

- [x] `go build ./...` passes
- [x] Unit tests pass (`go test -short ./internal/doltserver/ ./internal/doctor/`)
- [ ] No regressions in Stop() behavior — drain is non-fatal (max 10s timeout), always proceeds
- [ ] doctor --fix with newly-started server (< 60s) no longer triggers unnecessary restart

### DIFF
--- a/internal/doctor/rig_config_sync_check.go
+++ b/internal/doctor/rig_config_sync_check.go
@@ -7,6 +7,7 @@ import (
 	"os/exec"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/steveyegge/gastown/internal/beads"
 	"github.com/steveyegge/gastown/internal/config"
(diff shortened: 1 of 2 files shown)


steveyegge:
Fix-merged to main. Thank you @athosmartins!

PR #3388, opened 2026-03-28
fix: inherit CLAUDE_CONFIG_DIR in mayor session startup

quad341:
## Summary

- `mayor.StartTMUX` never called `ResolveAccountConfigDir`, so `CLAUDE_CONFIG_DIR` was never set in the mayor tmux session environment — unlike `gt start crew` which resolves the account before starting
- The zombie-respawn path in `runMayorAttach` had the same gap
- Both paths now resolve via: `accounts.json` default → `GT_ACCOUNT` env var → `CLAUDE_CONFIG_DIR` env var fallback

## Test plan

- [ ] Run `gt mayor restart` with `CLAUDE_CONFIG_DIR` set in shell; verify mayor session picks it up
- [ ] Run `gt mayor restart` with a configured `mayor/accounts.json`; verify the account's config dir is used
- [ ] Trigger zombie-respawn path (kill mayor process, run `gt mayor attach`); verify `CLAUDE_CONFIG_DIR` is prepended to the respawned command

> **Note:** I'm not fully confident this is the right fix — open to feedback on whether the env-var fallback is the right approach vs. requiring explicit account config.

### DIFF
--- a/internal/cmd/mayor.go
+++ b/internal/cmd/mayor.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/spf13/cobra"
 	"github.com/steveyegge/gastown/internal/config"
+	"github.com/steveyegge/gastown/internal/constants"
 	"github.com/steveyegge/gastown/internal/daemon"
 	"github.com/steveyegge/gastown/internal/doltserver"
 	"github.com/steveyegge/gastown/internal/mayor"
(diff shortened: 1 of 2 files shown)


quad341:
Superseded by #3391 which covers the daemon restart path in addition to mayor cold start and zombie respawn.

PR #3392, opened 2026-03-28
perf(hook): lazy storePool eliminates bd subprocess overhead in hook status

fernanja:
## Problem

`gt hook` (status display) takes **22+ seconds** on workspaces with multiple rigs. Profiling with `strace` and `go tool pprof` reveals the root cause: `runMoleculeStatus()` and `runHookShow()` spawn **18 \`bd\` subprocesses** (~0.9s each) to query beads databases that are already accessible via the running Dolt SQL server on localhost.

The subprocess breakdown for a town-level agent (mayor/deacon) with N rigs:
- 1× `bd list --status=hooked` per rig (N calls)
- 1× `bd list --status=in_progress` per rig (up to N calls)
- 1× `bd show` for agent bead lookup
- 1× `bd version` for capability probe

Each `bd` invocation pays ~600ms of fixed overhead: process spawn, Go runtime init, Dolt connection setup, circuit breaker check, and JSON serialization.

## Root Cause

The `beadsdk.Storage` in-process path already exists — `Beads.SetStore()`, `storeList()`, `storeShow()`, and `OpenFromConfig()` are all implemented and used by the daemon. But zero callers in `internal/cmd/` use them. Every `beads.New(dir)` creates a `Beads` instance with no store, forcing all queries through `exec.Command("bd", ...)`.

## Solution

Introduce `storePool` — a lazy, cached connection pool that opens `beadsdk.Storage` connections on first access and closes them when the command exits.

### Why lazy, not eager

We considered three approaches:

1. **Eager: open all rig stores upfront** (rejected) — Opens O(N) connections for every `gt hook` call, even when only 1-2 are needed. A polecat checking its own hook would open connections to every rig in the workspace. Wasteful and scales poorly.

2. **Singleton: one shared store for all databases** (rejected) — `beadsdk.Storage` binds to a single Dolt database. There's no cross-database query path without changing the beads library.

3. **Lazy pool: open on first access, cache for reuse** (chosen) — Each `pool.get(dir)` call checks the cache, then opens a store only if needed. A polecat opens 1-2 stores; the mayor scanning all rigs opens N stores incrementally. Failed opens are cached to prevent retry storms.

### Design details

- **Lazy opening**: Stores opened on first `get()` or `inject()` call, not upfront
- **Resolve caching**: `beads.ResolveBeadsDir()` results cached to avoid repeated filesystem walks for redirect files
- **Failure caching**: Failed `os.Stat` or `OpenFromConfig` recorded to prevent repeated attempts
- **Context propagation**: Pool takes `cmd.Context()` via `cmdContext()` nil-safe wrapper so Ctrl+C cancels in-progress store opens
- **Graceful fallback**: If a store can't be opened, the `Beads` instance has no store and silently falls back to the `bd` subprocess path — performance degrades but correctness is preserved
- **Warning visibility**: Failed opens log to stderr with path and error
- **Sequential use**: `storePool` is designed for single-goroutine use (documented in type comment). `DOLT_CHECKOUT` is session-level, so concurrent queries could see inconsistent branch state. All callers are sequential.

### What this does NOT do

- Does not change `gt mail inbox` (different code path, same problem — follow-up)
- Does not parallelize `scanAllRigsForHookedBeads` — with the store path at ~50ms/query, sequential scan is fast enough

## Results

| Metric | Before | After |
|--------|--------|-------|
| `gt hook` wall time | 22s+ | **2.5s** |
| `bd` subprocess calls | 18 | **0** |
| `git` subprocess calls | 32 | 1 |

Measured on a workspace with 4 rig databases and 8 routes in `routes.jsonl`.

## Files changed

| File | Change |
|------|--------|
| `internal/cmd/molecule_status.go` | Add `storePool` type (lazy, cached, with resolve/failure caches), `cmdContext()` nil-safe helper, wire into `runMoleculeStatus` and `scanAllRigsForHookedBeads` |
| `internal/cmd/hook.go` | Wire `storePool` into `runHookShow` |
| `internal/cmd/molecule_status_test.go` | 7 new unit tests for `storePool`: resolve caching, missing dir, no beads dir, inject-no-store, empty close, cancelled context with stderr verification |

## Test plan

- [x] `make build` — compiles clean
- [x] `go test ./internal/cmd/ -run "StorePool|MoleculeStatus|HookShow"` — 9 tests pass (7 new + 2 existing)
- [x] `go test ./internal/cmd/ -run TestHookShowShorthand -tags integration` — passes (was panicking before `cmdContext` fix)
- [x] `time gt hook` — 2.5s (from 22s)
- [x] `strace -f -e trace=execve gt hook | grep bd` — 0 calls (from 18)
- [x] Output identical before and after (correctness preserved)
- [x] Ctrl+C during hook status exits promptly (context cancellation verified)
- [x] With Dolt server stopped: falls back to subprocess path (slower but functional)

### DIFF
--- a/internal/cmd/hook.go
+++ b/internal/cmd/hook.go
@@ -1,7 +1,6 @@
 package cmd
 
 import (
-	"context"
 	"encoding/json"
 	"fmt"
 	"os"
@@ -498,16 +497,12 @@ func runHookShow(cmd *cobra.Command, args []string) error {
 		}
 	}
 
-	// Open in-process stores to bypass bd subprocess overhead.
-	townRootForStores, _ := workspace.FindFromCwd()
-	if townRootForStores == "" {
-		townRootForStores, _ = findTownRoot()
-	}
-	stores, cleanupStores := openHookStores(context.Background(), townRootForStores, workDir)
-	defer cleanupStores()
+	// Lazy store pool bypasses bd subprocess overhead (~600ms/call).
+	pool := newStorePool(cmdContext(cmd))
+	defer pool.close()
 
 	b := beads.New(workDir)
-	injectStore(b, workDir, stores)
+	pool.inject(b, workDir)
 	// Query for hooked beads assigned to the target
 	hookedBeads, err := b.List(beads.ListOptions{
 		Status:   beads.StatusHooked,
@@ -528,7 +523,7 @@ func runHookShow(cmd *cobra.Command, args []string) error {
 			townBeadsDir := filepath.Join(townRoot, ".beads")
 			if _, err := os.Stat(townBeadsDir); err == nil {
 				townBeads := beads.New(townBeadsDir)
-				injectStore(townBeads, townBeadsDir, stores)
+				pool.inject(townBeads, townBeadsDir)
 				townHooked, err := townBeads.List(beads.ListOptions{
 					Status:   beads.StatusHooked,
 					Assignee: target,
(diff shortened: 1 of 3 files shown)


steveyegge:

PR #3407, opened 2026-03-29
feat: bd-route plugin — auto-route bd create to correct rig via PreToolUse hook

azanar:
Closes steveyegge/gastown#3404

Adds `plugins/bd-route/`: a Python PreToolUse hook that intercepts `bd create` commands missing `--rig` and auto-routes them based on keyword matching against a user-maintained YAML config.

- If exactly one rule matches → rewrites command with `--rig <rig>`
- If zero or multiple rules match → passes through unchanged (safe fallback)
- Config is user-customizable; the hook itself is generic


steveyegge:

PR #3415, opened 2026-03-29
fix: replace mapfile with while-read loop for bash 3.2 compat

sethbc:
## Summary

- dolt-archive and dolt-backup plugin scripts use `mapfile` (bash 4+ built-in) which fails on macOS default bash 3.2
- Replaces `mapfile` with POSIX-compatible `while read` loops

## Test plan

- [ ] Run dolt-archive plugin on macOS with system bash 3.2
- [ ] Run dolt-backup plugin on macOS with system bash 3.2
- [ ] Verify both plugins still work on bash 4+/5+


sethbc:
Closing — not needed, bash 4+ is available.

PR #3417, opened 2026-03-29
fix(doctor): detect gt binary shadowing and stale builds

### LINKED ISSUE #3416: Upgrade UX on existing installs is still fragile: docs drift, channel drift, stale-binary shadowing, and non-convergent doctor --fix
## Summary

On March 29, 2026 I upgraded an existing macOS/zsh Gas Town install from an older dev binary to the current community release.

The upgrade was recoverable, but the path there exposed several upgrade/recovery gaps that seem worth tracking together. Some pain points were local environment/operator mistakes, but several were clearly product-level UX or state-convergence issues.

Environment at the end of the recovery:
- `gt 0.12.1`
- `bd 0.62.0`
- `dolt 1.84.0`
- town root: `~/gt`

Related issues:
- #3077 covers one PATH-shadowing direction (Homebrew/source precedence), but this report also covers the reverse case: a stale `~/.local/bin/gt` shadowing a fresh Homebrew install.
- #2922 covers partial beads init / recovery fragility, which feels related to the broader theme that repair paths are still easy to get into and hard to reason about.

### 1. Install and upgrade guidance is internally inconsistent

Current guidance points in different directions:
- `README.md` recommends `brew install gastown`
- `docs/INSTALLING.md` still points users to `go install github.com/steveyegge/gastown/cmd/gt@latest`
- `Makefile` explicitly warns: `use make install, not go install`
- canonical runtime path appears to be `~/.local/bin/gt`, but the docs do not clearly explain upgrade behavior when multiple install channels coexist

Expected:
- one canonical upgrade path for end users
- one canonical path for source/dev installs
- explicit documentation for what happens when Homebrew + source/manual installs coexist

### 2. Release channels can disagree on what “latest” means

On March 29, 2026:
- GitHub latest release: `v0.12.1`
- Homebrew stable: `0.12.1`
- npm `@gastown/gt`: `0.12.0`

That makes "upgrade to the latest community version" channel-dependent.

Expected:
- either channels stay in lockstep
- or docs explicitly say which channel is authoritative and which channels may lag

### 3. Existing-install upgrades can silently keep running the wrong `gt`

This machine had an older dev binary at `~/.local/bin/gt`.

After installing/upgrading Homebrew `gastown 0.12.1`, the shell still resolved the stale `~/.local/bin/gt` until I manually moved it aside and re-pointed `~/.local/bin/gt -> /usr/local/bin/gt`.

Observed gaps:
- `gt doctor` did not warn that the active `gt` binary was stale / non-release / shadowing another installed `gt`
- the upgrade flow did not identify that another `gt` on PATH was still active
- users can think they upgraded successfully while still executing an old binary

This is related to #3077, but in the reverse direction.

Expected:
- install/doctor should enumerate discovered `gt` binaries and clearly identify the active one
- install/doctor should warn when a stale or unexpected `gt` shadows the intended binary
- there should be an explicit recommended migration path between Homebrew and source/manual installs

### 4. `gt doctor --fix` does not fully converge in one pass

Initial post-upgrade warnings included:
- daemon not running
- outdated formulas
- stale PID files
- stale worker lock
- `in_progress` beads with `NULL` assignee

First `gt doctor --fix` repaired those, but a follow-up `gt doctor` still reported:
- daemon not running again, which required a manual `gt daemon start`
- after daemon was running, new `zombie-sessions` and `stale-runtime-files` warnings appeared and required a second `gt doctor --fix`

Expected:
- if `doctor --fix` reports success, the immediately following `doctor` should not need another repair pass just because daemon state changed
- if daemon startup reveals additional checks, `doctor --fix` should either rerun those checks internally or report that another pass is required
- a fix should not be reported as successful if the relevant condition is still unstable on the next check

### 5. `NULL assignee` + `in_progress` beads are still possible and become invisible to `bd`

`gt doctor` reported:

> `2 in_progress bead(s) with NULL assignee — invisible to bd, blocking molecule progress`

`gt doctor --fix` repaired them, but this feels like a state invariant violation that should either be prevented at write time or surfaced much earlier.

Expected:
- system should prevent creating/transitioning beads into `in_progress` with `NULL` assignee
- if this state is intentionally representable, `bd` should still surface it instead of making it invisible
- daemon/doctor should not be the first place users discover it

## Non-product issues from this session (not filing these as GT bugs)

These were real problems during the recovery, but they look local/environmental rather than GT defects:
- I initially ran `gt doctor` from a source repo instead of from the Town root (`~/gt`)
- the local `gastown` checkout was dirty and far behind `origin/main`, so building from that checkout would have been a usage mistake
- Homebrew briefly hit network failures fetching `formula.jws.json`
- Homebrew also hit a local `/usr/local/var/log` permission problem while upgrading a `dolt` dependency

I am including these only to separate operator/environment issues from product issues.

## Suggested follow-up issue split

If this umbrella is useful, I would split it into focused issues/PRs roughly like this:

1. Unify install/upgrade docs and define canonical paths per install channel
2. Add install/doctor detection for active-binary shadowing across Homebrew/source/manual installs
3. Make `gt doctor --fix` converge after daemon-affecting fixes, or explicitly rerun affected checks
4. Decide and document release-channel policy for GitHub/Homebrew/npm version sync
5. Prevent or surface `NULL assignee` + `in_progress` bead states earlier

## PR candidates

- Add a doctor check that lists all discovered `gt` binaries, marks the active one, and warns on shadowing/stale builds
- Add a post-install / upgrade sanity check that verifies `command -v gt` matches the intended install target
- Align `README.md`, `docs/INSTALLING.md`, and source install guidance around one canonical story
- Add release automation or release docs that keep npm/Homebrew/GitHub expectations explicit
- Make `doctor --fix` rerun daemon-sensitive checks after starting/stopping daemon
- Harden bead state transitions so `in_progress` cannot coexist with `NULL` assignee, or make `bd` show those rows explicitly

fusioncid:
## Summary
- add a reusable `gt` binary inventory/probe layer that scans PATH plus the currently running executable
- extend the `stale-binary` doctor check so it lists discovered `gt` binaries, marks the active/PATH-selected one, and warns on shadowing, older installs, and stale dev builds
- add focused tests for both binary discovery and doctor warning/OK behavior

Checked against `origin/main` at `42f9d568fc1fcc8fc64c991936c16c71fc7bec7f`; the latest main branch still only had the old repo-vs-binary stale check and did not yet implement PATH inventory/shadowing detection.

Fixes #3416.
Implements the first PR candidate from #3416.

## Verification
- `go test ./internal/version`
- `go test ./internal/doctor`
- `go test ./internal/cmd -run TestDoesNotExist`

### DIFF
--- a/.github/workflows/windows-ci.yml
+++ b/.github/workflows/windows-ci.yml
@@ -45,8 +45,15 @@ jobs:
 
       - name: Install Dolt
         shell: pwsh
+        env:
+          GH_TOKEN: ${{ github.token }}
         run: |
-          $release = Invoke-RestMethod 'https://api.github.com/repos/dolthub/dolt/releases/latest'
+          $headers = @{
+            Authorization        = "Bearer $env:GH_TOKEN"
+            Accept               = "application/vnd.github+json"
+            "X-GitHub-Api-Version" = "2022-11-28"
+          }
+          $release = Invoke-RestMethod -Headers $headers 'https://api.github.com/repos/dolthub/dolt/releases/latest'
           $asset = $release.assets | Where-Object { $_.name -like 'dolt-windows-amd64.zip' }
           $zip = "$env:RUNNER_TEMP\dolt.zip"
           Invoke-WebRequest -Uri $asset.browser_download_url -OutFile $zip
--- a/internal/cmd/convoy_bd_routing_test.go
+++ b/internal/cmd/convoy_bd_routing_test.go
@@ -93,6 +93,7 @@ func TestRunConvoyList_UsesTownRootAndStripsBeadsDir(t *testing.T) {
 	scriptBody := fmt.Sprintf(`
 # Allow-stale version probe is exempt from BEADS_DIR check.
 if [ "$*" = "--allow-stale version" ]; then
+  echo "bd version 0.62.0"
   exit 0
 fi
 
@@ -166,6 +167,7 @@ func TestRunConvoyStatus_UsesTownRootAndStripsBeadsDir(t *testing.T) {
 	scriptBody := fmt.Sprintf(`
 # Allow-stale version probe is exempt from BEADS_DIR check.
 if [ "$*" = "--allow-stale version" ]; then
+  echo "bd version 0.62.0"
   exit 0
 fi
 
--- a/internal/cmd/convoy_external_tracking_test.go
+++ b/internal/cmd/convoy_external_tracking_test.go
@@ -67,6 +67,7 @@ func TestGetTrackedIssues_FallsBackToShowTrackedDependencies(t *testing.T) {
 	scriptBody := fmt.Sprintf(`
 case "$*" in
   "--allow-stale version")
+    echo "bd version 0.62.0"
     exit 0
     ;;
   "dep list hq-cv-ext --direction=down --type=tracks --allow-stale --json")


steveyegge:

PR #3425, opened 2026-03-29
fix: polecat settings.json placement + account switch onboarding

ckumar1:
## Summary

Two fixes for issues that caused a full polecat fleet failure (21+ polecats):

**1. Settings.json symlink into polecat worktrees**

When a customer repo has its own `.claude/` directory (e.g., with custom commands), Claude Code may look there for `settings.json` instead of honoring the `--settings` flag. Polecats' SessionStart hooks don't fire, and they sit idle at the prompt.

Fix: symlink the shared `polecats/.claude/settings.json` into each worktree's `.claude/` during setup. Belt-and-suspenders — the `--settings` flag should work, but the local symlink ensures hooks are always found.

**2. Pre-seed onboarding on account switch and session start**

When `gt account switch` creates a symlink to a new/empty config directory, Claude Code shows the first-run setup wizard (theme picker, ToS). Agents can't navigate interactive prompts, so every session freezes. The cascade: account switch → wizard → freeze → restart → hook empty → gt done → beads closed.

Fix: write `hasCompletedOnboarding: true` to `.claude.json` in the target config dir before symlinking (account switch path) and at session start when `RuntimeConfigDir` is set (`--account` flag path).

## Test plan

- [x] `go test ./internal/cmd/ -run TestAccountSwitch` — all pass including new onboarding pre-seed test
- [x] `go vet ./internal/cmd/ ./internal/polecat/` — clean
- [x] Manual: polecats spawned after fix find settings.json and fire SessionStart hook
- [x] Manual: account switch to empty dir no longer triggers first-run wizard

### DIFF
--- a/internal/polecat/session_manager.go
+++ b/internal/polecat/session_manager.go
@@ -304,6 +304,12 @@ func (m *SessionManager) Start(polecat string, opts SessionStartOptions) error {
 			return fmt.Errorf("building startup command: %w", err)
 		}
 	}
+	// Pre-seed onboarding in the runtime config dir so Claude Code doesn't show
+	// the first-run wizard. This covers the --account flag path (gt-1ob follow-up).
+	if opts.RuntimeConfigDir != "" {
+		ensureOnboardingComplete(opts.RuntimeConfigDir)
+	}
+
 	// Prepend runtime config dir env if needed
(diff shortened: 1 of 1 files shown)


steveyegge:

PR #3439, opened 2026-03-30
test(convoy): update DepFail test to match non-fatal tracking behavior

richardpospisil:
## Summary

- Commit 103b6aaa changed `createAutoConvoy` to treat `dep add` failure as a non-fatal warning (returns convoy ID, no error, no orphan close) to handle cross-rig beads where `bd` cannot validate both IDs in the same database
- The test `TestCreateAutoConvoy_DepFailCleansUpOrphan` was not updated at the time, leaving it asserting the old hard-error + orphan-close behavior
- This updates the test to verify the current behavior: dep failure logs a warning, convoy ID is returned, and `close` is not called

## Test plan

- [ ] `go test ./internal/cmd/... -run TestCreateAutoConvoy_DepFailCleansUpOrphan` passes
- [ ] `go test ./...` passes (no regressions)

### DIFF
--- a/internal/cmd/sling_batch_test.go
+++ b/internal/cmd/sling_batch_test.go
@@ -952,7 +952,8 @@ exit 0
 }
 
 // TestCreateAutoConvoy_DepFailCleansUpOrphan verifies that when the dep add
-// fails, the convoy is closed to prevent orphans.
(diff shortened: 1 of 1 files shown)


steveyegge:
Already fixed on main — DepFail test updated to match non-fatal tracking behavior.

PR #3237, opened 2026-03-24
feat: emergency stop (gt estop / gt thaw) — pause and resume all agent work

outdoorsea:
## Problem

When critical infrastructure fails (Dolt server crash, disk full, 2.1GB log files), all coding agents continue burning API tokens on work that will fail or need to be redone. There is no mechanism to immediately halt agent activity across the town.

This also affects operators who travel frequently — closing a laptop, moving to spotty internet, or needing to rotate API licenses requires manually killing sessions one by one, losing all agent context in the process.

## Solution

A factory-floor E-stop button for Gas Town. All agent sessions are frozen in place (SIGTSTP) — context preserved, no work is lost. One command to stop, one to resume.

### Commands

```bash
gt estop                              # Freeze all agent sessions
gt estop -r "closing laptop"          # Freeze with reason
gt estop --rig gastown                # Freeze only one rig
gt thaw                               # Resume all sessions + nudge agents
gt thaw --rig gastown                 # Resume only one rig
```

### Three layers of protection

1. **Manual E-stop** (`gt estop`): Creates `ESTOP` sentinel file at town root, sends SIGTSTP to all agent sessions. Mayor and overseer exempt. `gt thaw` sends SIGCONT and nudges all sessions to continue.

2. **Daemon auto-trigger**: Tracks consecutive Dolt health failures. After 3 failures (~90s), auto-creates ESTOP and freezes everything. When Dolt recovers, daemon auto-clears the E-stop, thaws sessions, and nudges them. Manual E-stops always require manual `gt thaw`.

3. **Agent-side hook check**: The `gt mail check --inject` hook (runs on every UserPromptSubmit) checks for ESTOP files and injects a system-reminder telling the agent to checkpoint and wait. Defense-in-depth for agents that survive the SIGTSTP freeze.

### Per-rig granularity

`gt estop --rig <name>` uses `ESTOP.<rigname>` sentinel files — freeze one rig while others keep working. Useful for rig-specific maintenance.

### Visual feedback

- `gt status` shows red E-stop banner with trigger type, duration, and reason
- tmux status bar shows red `ESTOP HH:MM` indicator (refreshes every 5s)

## Alternative uses beyond infrastructure failure

- **Travel**: `gt estop -r "flying to NYC"` → close laptop → land → `gt thaw`
- **Spotty internet**: freeze agents before they burn tokens on failed API calls
- **License rotation**: pause work while rotating to a new API key
- **Maintenance windows**: freeze a single rig while upgrading its dependencies
[...]

Issue #2716, opened 2026-03-13
Windows CI: ACP tests fail on Windows (process management assumes Unix)


jason-curtis:
## Problem

Windows CI has been failing consistently for 5+ days. All failures are in `internal/acp` and `internal/mayor` packages:

- `TestProxy_RunKeepAlive_Logic` — heartbeat timeout
- `TestWriteToAgent_Parity` (6 subtests) — "agent process is not running"
- `TestIsACPActive_CurrentProcess` — ACP detection returns false
- `TestCleanupVetoChecker_ShouldVetoCleanup` / `VetoIfActive` — depends on ACP detection

## Root Cause

ACP process detection relies on Unix-specific mechanisms (`/proc`, pipe semantics) that don't exist on Windows. The tests don't guard for this.

## Fix

Add `t.Skip` guards on Windows for tests that require Unix process management, or fix the cross-platform process detection in the ACP package.

## Evidence

Last 5 upstream Windows CI runs: all failures.
```
gh run list --repo steveyegge/gastown --workflow windows-ci.yml --limit 5
```

DreadPirateRobertz:
The ACP tests use Unix-specific process detection (`/proc`, pipes, signals). On Windows, `os.FindProcess` always succeeds (even for non-existent PIDs), and `/proc` doesn't exist.

Quick fix: add build tags or runtime skip guards:
```go
//go:build !windows

// OR at runtime:
if runtime.GOOS == "windows" {
    t.Skip("ACP process detection requires Unix")
}
```

Alternatively, abstract the process detection behind an interface with platform-specific implementations, but that's more effort and may not be worth it if Windows isn't a primary target for Gas Town.

jason-curtis:
Do you think we should just deliberately cut Windows support- WSL is the way to go?
I suppose folks might want to actually have GT build applications for Windows

DreadPirateRobertz:
@jason-curtis I'd lean toward keeping Windows support but being explicit that **WSL is the supported path for Windows users**, rather than native Win32.

The core problem is that Gas Town's process management (tmux, signals, `lsof`/`ss`, Unix sockets) is fundamentally Unix-shaped. Getting full parity on Win32 would require a large ongoing maintenance burden for a runtime that's fundamentally different. WSL2 gives Windows users a proper Linux environment with full compatibility.

[...]

PR #3441, opened 2026-03-30
test: remove Windows from cross-platform build targets

richardpospisil:
## Summary

- Removes `windows/amd64` from the `TestCrossPlatformBuild` platform list

Gas Town requires tmux and Unix process signals (`SIGTSTP`, `SIGCONT`, `kill(-pgid, sig)`) which have no native Windows equivalent. WSL2 is the supported path for Windows users (per the discussion in #2716).

Keeping Windows in the cross-compile matrix causes spurious build failures whenever Unix-only code is added — most recently `estop.go` (PR #3237), which uses `syscall.SIGTSTP`, `syscall.SIGCONT`, and `syscall.Kill`.

## Test plan

- [ ] `go test ./cmd/gt/... -run TestCrossPlatformBuild` passes

### DIFF
--- a/cmd/gt/build_test.go
+++ b/cmd/gt/build_test.go
@@ -31,7 +31,6 @@ func TestCrossPlatformBuild(t *testing.T) {
 		{"linux", "arm64", "0"},
 		{"darwin", "amd64", "0"},
 		{"darwin", "arm64", "0"},
-		{"windows", "amd64", "0"},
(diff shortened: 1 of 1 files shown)


steveyegge:
Closing — removing Windows from build targets is the wrong fix. The build should work cross-platform; CI issues should be fixed at the CI level.

PR #3444, opened 2026-03-30
fix(reaper): don't escalate on Dolt 'nothing to commit'

Jaredjaskot:
## Summary
- Stop treating Dolt "nothing to commit" as a `dolt_commit_failed` anomaly in the reaper
- With Dolt autocommit=1 (server default), DELETEs are committed immediately, making the explicit `DOLT_COMMIT` call return "nothing to commit" — this is expected, not an error
- Fixes both `purgeClosedWisps` and `AutoClose` code paths
- Prevents recurring HIGH severity false-positive escalations (3 occurrences so far)

## Test plan
- [ ] Verify reaper purge runs without escalation when autocommit is on
- [ ] Verify real commit failures still escalate


steveyegge:
Already fixed on main — both purgeClosedWisps and AutoClose already have the isNothingToCommit guard. Thanks for the catch though!

PR #3447, opened 2026-03-31
fix: stop removing/modifying Dolt-internal files (.dolt/noms/LOCK, manifest, sql-server.info)

coffeegoddd:
## Summary

Related: https://github.com/gastownhall/beads/issues/2933

- **Remove LOCK file cleanup**: `cleanupStaleDoltLock()` in `doltserver.go` deleted `.dolt/noms/LOCK` files before server start. Dolt manages its own lock files on startup — external removal risks data corruption.
- **Remove database directory quarantine**: Pre-startup code inspected `.dolt/noms/manifest` inside database directories and moved/deleted those missing it. A valid `.dolt` directory does not necessarily contain `noms/manifest` — this heuristic was wrong and destroyed legitimate databases. `ListDatabases()` already skips unloadable databases with a warning.
- **Make `sql-server.info` Fix a no-op**: `StaleSQLServerInfoCheck.Fix()` deleted `.dolt/sql-server.info`, which is written and managed by Dolt itself (`dolt/commands/sqlserver/creds.go`). Fix now returns nil; hint directs operators to restart the Dolt server instead.
- **Update dog formula**: `mol-dog-phantom-db` no longer instructs dogs to `rm -rf` database directories. Changed from quarantine to escalation-only.
- **Update templates**: Dog role template and `claude.md` now carry explicit warnings against modifying `.dolt/` internals.

All changes include a standardized warning comment:
```
// WARNING: DO NOT remove, delete, or modify files inside Dolt's .dolt/
// directory — including noms/LOCK files. These are Dolt-internal files.
// Removing them WILL cause unrecoverable data corruption and data loss.
// Dolt manages these files itself; external interference is never safe.
```

## Test plan

- [x] `go build ./internal/doltserver/` compiles clean
- [x] `go build ./internal/doctor/` compiles clean
- [x] `go vet` passes on both packages
- [x] `go test ./internal/doltserver/` passes (pre-existing unrelated socket test failure confirmed on main)
- [x] `go test ./internal/doctor/` passes — renamed test `TestStaleSQLServerInfoCheck_FixIsNoOp` verifies Fix no longer deletes files
- [ ] Verify Dolt server starts correctly with databases that lack `noms/manifest` (they should be skipped via `ListDatabases` warning)
- [ ] Verify `gt dolt cleanup` still works (uses `RemoveDatabase` which already does `DROP DATABASE` via SQL before filesystem removal — unchanged)

### DIFF
--- a/internal/doctor/stale_sql_server_info_check.go
+++ b/internal/doctor/stale_sql_server_info_check.go
@@ -96,17 +96,19 @@ func (c *StaleSQLServerInfoCheck) Run(ctx *CheckContext) *CheckResult {
 		Status:  StatusWarning,
 		Message: fmt.Sprintf("%d stale sql-server.info file(s) from dead Dolt servers", len(c.staleFiles)),
 		Details: details,
-		FixHint: "Run 'gt doctor --fix' to remove stale sql-server.info files",
+		FixHint: "Restart the Dolt server to clear stale sql-server.info files (Dolt writes and cleans these itself)",
 	}
 }
 
-// Fix removes all detected stale sql-server.info files.
+// Fix is a no-op. sql-server.info is a Dolt-internal file written and managed
+// by Dolt itself (see dolt/commands/sqlserver/creds.go). Restarting the Dolt
+// server will create a fresh one.
+//
+// WARNING: DO NOT remove, delete, or modify files inside Dolt's .dolt/
+// directory — including noms/LOCK files. These are Dolt-internal files.
+// Removing them WILL cause unrecoverable data corruption and data loss.
+// Dolt manages these files itself; external interference is never safe.
 func (c *StaleSQLServerInfoCheck) Fix(ctx *CheckContext) error {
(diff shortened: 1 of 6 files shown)


A3Ackerman:
Strong +1 on this. We've been tracking Dolt server crashes during merge conflict resolution (our internal bead gt-3dx). On 2026-03-26 the Dolt server crashed right after a manual `dolt conflicts resolve --theirs` on the q_core database, and the server has been restarted at least 3 times in that session alone. Each restart is disruptive — all bd/beads operations fail until the server comes back.

The `cleanupStaleDoltLock()` removal and the database directory quarantine removal both address plausible causes of these instabilities. The recurring crosstown-sync merge conflicts leave dirty state that triggers these code paths, so removing the dangerous heuristics here should significantly reduce crash frequency.

The `sql-server.info` change to no-op is also the right call — we've seen dogs attempt to "fix" this file which compounds the problem.

PR #3452, opened 2026-04-01
fix: use canonical deacon heartbeat in stuck-agent-dog

Bella-Giraffety:
## Summary
- switch `stuck-agent-dog` from the legacy `.deacon-heartbeat` file to the canonical `deacon/heartbeat.json` heartbeat source
- make the plugin tolerate older/runtime-copied layouts by falling back from `mayor/rigs.json` to town-root `rigs.json`
- fail safe when `rigs.json` is malformed instead of crashing the plugin
- use portable heartbeat timestamp parsing that accepts the RFC3339/RFC3339Nano values written by `internal/deacon/heartbeat.go`
- add focused plugin scanner regression coverage for the canonical heartbeat path and rigs.json fallback guidance

## Related Issue
- Fixes `gs-0ul`
- Related to `gs-ogf`

## Bug
During the upstream PR workflow, `stuck-agent-dog` repeatedly escalated a healthy live deacon because it was reading the legacy `$TOWN_ROOT/deacon/.deacon-heartbeat` file instead of the canonical `deacon/heartbeat.json`. It also assumed a single `rigs.json` location and brittle parse path.

That produced two classes of false positives:
- a healthy deacon could look stuck because the plugin ignored the canonical heartbeat writer
- malformed or relocated `rigs.json` content could fail the plugin instead of failing safe

## Why this fix
This PR keeps the change scoped to the plugin that is actually producing the bad signal:
- use the same heartbeat source the deacon writes today
- parse heartbeat timestamps portably from the JSON file instead of shelling out to GNU/BSD-specific `date` parsing
- add a canonical-then-fallback `rigs.json` lookup and safe parse failure behavior

It does not change broader deacon lifecycle semantics or patrol scheduling.

## Testing
Focused command run:
- `GOTOOLCHAIN=auto go test ./internal/plugin -run 'TestParsePluginMD_GitHubSheriff|TestParsePluginMD_StuckAgentDogUsesCanonicalHeartbeatPath'`

Coverage added/updated:
- canonical `deacon/heartbeat.json` path must appear in the shipped plugin
- legacy `.deacon-heartbeat` path must not appear
- the plugin instructions must include the `rigs.json` fallback path and fail-safe parse handling

## Review process
This branch went through multiple review-only passes before and after implementation. Incorporated findings included:
- BSD/macOS timestamp parsing must accept Go JSON fractional-second timestamps
- operator behavior should remain fail-safe when `rigs.json` is missing or malformed
- scope should stay inside the plugin path and not widen deacon lifecycle changes

### DIFF
--- a/internal/plugin/scanner_test.go
+++ b/internal/plugin/scanner_test.go
@@ -400,6 +400,28 @@ func TestParsePluginMD_GitHubSheriff(t *testing.T) {
 	}
 }
 
+func TestParsePluginMD_StuckAgentDogUsesCanonicalHeartbeatPath(t *testing.T) {
+	content, err := os.ReadFile(filepath.Join("..", "..", "plugins", "stuck-agent-dog", "plugin.md"))
+	if err != nil {
+		t.Skipf("stuck-agent-dog plugin not found (expected in plugins/): %v", err)
+	}
+
+	plugin, err := parsePluginMD(content, "/test/stuck-agent-dog", LocationRig, "gastown")
+	if err != nil {
+		t.Fatalf("parsePluginMD failed: %v", err)
+	}
+
+	if plugin.Name != "stuck-agent-dog" {
+		t.Fatalf("expected name 'stuck-agent-dog', got %q", plugin.Name)
+	}
+	if !strings.Contains(plugin.Instructions, "deacon/heartbeat.json") {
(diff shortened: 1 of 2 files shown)


Bella-Giraffety:
Rebuilt this PR on a clean upstream main base and force-pushed it so it now contains only the stuck-agent-dog plugin fix and focused tests. The previous branch had unrelated local workspace artifacts in history.

PR #3455, opened 2026-04-01
fix: align reaper scan with reap eligibility

Bella-Giraffety:
## Summary
- make `Scan()` exclude agent wisps using the same stale-wisp eligibility that `Reap()` already uses
- add focused regression coverage so the `Scan()` function body specifically must contain the agent-bead exclusion
- keep the change scoped to scan/reap eligibility alignment only; purge and stale-issue behavior are unchanged

## Related Issue
- Fixes `gs-mn3`

## Bug
A live reproduction on `hq` showed:
- `gt reaper scan --db hq --max-age=24h` reported 4 reap candidates
- `gt reaper reap --db hq --max-age=24h` closed 0

The root cause was a predicate mismatch:
- `Scan()` counted all stale open wisps
- `Reap()` explicitly excluded `issue_type='agent'`

So scan could report candidates that reap would never close, especially when the candidates were all agent wisps.

## Why this fix
This is the smallest safe fix: keep the existing `Reap()` behavior and make `Scan()` use the same eligibility predicate. That makes the operator-facing scan count truthful without widening the change into purge or stale-issue logic.

## Testing
Focused command run:
- `GOTOOLCHAIN=auto go test ./internal/reaper -run 'TestReap|TestScanExcludesAgentBeads'`

Coverage:
- `TestScanExcludesAgentBeads` now scopes its assertion to the `Scan()` function body specifically, so it cannot pass just because `Reap()` already had the agent exclusion.

## Review process
This branch went through pre-implementation review and a second review wave against the implementation branch. Incorporated findings covered:
- keeping the change limited to scan/reap eligibility alignment
- ensuring the regression test actually targets `Scan()` instead of matching the broader file text
- confirming operator-facing scan/reap counts now correspond for agent wisps

### DIFF
--- a/internal/reaper/reaper.go
+++ b/internal/reaper/reaper.go
@@ -234,9 +234,11 @@ func Scan(db *sql.DB, dbName string, maxAge, purgeAge, mailDeleteAge, staleIssue
 	parentJoin, parentWhere := parentExcludeJoin(dbName)
 
 	// Count reap candidates: open wisps past max_age with eligible parent status.
+	// Must match Reap() eligibility semantics exactly, including the exclusion of
+	// agent beads, otherwise scan can report candidates that reap will never close.
 	// Uses LEFT JOIN anti-pattern instead of correlated EXISTS to avoid O(n*m) cost (gt-jd1z).
 	reapQuery := fmt.Sprintf(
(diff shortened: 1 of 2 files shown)


Bella-Giraffety:
Rebuilt this PR on a clean upstream main base and force-pushed it so it now contains only the scan/reap eligibility alignment fix and focused tests.

PR #3457, opened 2026-04-01
fix: preserve routed bead context at town root

Bella-Giraffety:
## Summary
- make town-root `gt show` and sling/convoy helper lookups resolve non-HQ bead prefixes through the owning rig context instead of always forcing town root
- batch mixed-prefix convoy lookups by resolved execution directory so HQ and non-HQ issues no longer silently drop out of the same batch
- keep the fix scoped to town-root CLI/convoy helper paths and split web fetcher parity into follow-up `gs-mdh`
- add focused regression coverage for routed non-HQ `gt show`, routed bead existence checks, existing formula-routing behavior, and mixed-prefix convoy batches

## Related Issues
- Fixes `gs-60j`
- Follow-up: `gs-mdh`

## Bug
Direct routed bead commands could succeed from the owning rig root while higher-level town-root `gastown` paths still failed for the same non-HQ bead IDs. The concrete operator impact was that `gt sling`, `gt show`, and convoy detail fetchers still needed HQ wrapper workarounds for valid `do-`, `gs-`, or `bd-` issues even when `routes.jsonl` was correct.

A first implementation fixed single-prefix town-root lookups but regressed mixed-prefix convoy batches by forcing the entire batch into the first issue's rig context. This final branch fixes both problems while keeping the patch scoped.

## Why this fix
This branch keeps the change at the `gastown` town-root helper layer instead of widening into general beads routing semantics:
- `resolveBeadDir` now resolves registered non-HQ prefixes via town `routes.jsonl` (with `rigs.json` fallback) while keeping `hq-*` at town root
- `gt show` and the sling preflight helpers use that resolved context for non-HQ beads
- convoy batch lookups group issue IDs by resolved directory so mixed HQ + non-HQ batches preserve both contexts instead of dropping whichever IDs do not match the first issue's rig
- a sibling web fetcher parity gap is intentionally filed separately as `gs-mdh` to keep this PR surgical

## Testing
Focused command run:
- `GOTOOLCHAIN=auto go test ./internal/cmd -run 'TestResolveBeadDir_UsesRoutesForNonHQPrefix|TestVerifyBeadExists_RoutesNonHQPrefixFromTownRoot|TestSlingFormulaOnBeadRoutesBDCommandsToTargetRig|TestGetIssueDetailsBatch_GroupsMixedPrefixesByResolvedDir'`

Coverage added/updated:
- routed non-HQ prefix resolution from town root
- town-root bead existence verification for non-HQ prefixes
- existing formula-routing behavior remains intact
- mixed-prefix convoy batch now issues separate grouped lookups for HQ and non-HQ IDs

## Review process
This branch went through review-only passes for correctness, test coverage, scope, style, and operator diagnostics. The final branch incorporates the major review finding by grouping mixed-prefix convoy batches per resolved directory, and explicitly splits web convoy fetcher parity into follow-up `gs-mdh` to keep the scope tight.

### DIFF
--- a/internal/cmd/convoy.go
+++ b/internal/cmd/convoy.go
@@ -235,7 +235,7 @@ Examples:
   # Auto-discover issues from an epic's children:
   gt convoy create --from-epic gt-epic-abc
   gt convoy create --from-epic gt-epic-abc --owned --merge=direct`,
-	Args: cobra.ArbitraryArgs,
+	Args:         cobra.ArbitraryArgs,
 	SilenceUsage: true,
 	RunE:         runConvoyCreate,
 }
@@ -247,7 +247,7 @@ var convoyStatusCmd = &cobra.Command{
 
 Displays convoy metadata, tracked issues, and completion progress.
 Without an ID, shows status of all active convoys.`,
-	Args: cobra.MaximumNArgs(1),
+	Args:         cobra.MaximumNArgs(1),
 	SilenceUsage: true,
 	RunE:         runConvoyStatus,
 }
@@ -277,7 +277,7 @@ If the convoy is closed, it will be automatically reopened.
 Examples:
   gt convoy add hq-cv-abc gt-new-issue
   gt convoy add hq-cv-abc gt-issue1 gt-issue2 gt-issue3`,
-	Args: cobra.MinimumNArgs(2),
+	Args:         cobra.MinimumNArgs(2),
 	SilenceUsage: true,
 	RunE:         runConvoyAdd,
 }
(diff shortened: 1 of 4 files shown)


Bella-Giraffety:
Rebuilt this PR on a clean upstream main base and force-pushed it so it now contains only the routed-context fix and its focused tests. The previous branch had unrelated local workspace artifacts in history.

Bella-Giraffety:
Closing this one after re-auditing it against current upstream main. The original routing symptom was real in the local incident, but this PR changes a shared town-root helper too broadly, and current upstream already has direct town-root routing tests that argue for the opposite default. If there is still a live bug here, it should come back with a failing integration test against current main rather than this broader helper rewrite.

PR #3474, opened 2026-04-01
fix(daemon): deduplicate Dolt alerts and use nudge instead of mail

outdoorsea:
## Summary
- Deduplicates Dolt health alerts to prevent alert storms flooding the inbox
- Switches routine agent-to-agent Dolt alerts from `gt mail send` (permanent Dolt commit) to `gt nudge` (zero cost)
- Follows the communication hygiene principle: mail only for messages that must survive session death

## Test plan
- [x] Dolt alerts deduplicated — same alert not sent repeatedly
- [x] Nudge used for routine alerts, mail reserved for escalations
- [x] Builds clean

### DIFF
--- a/internal/daemon/dolt.go
+++ b/internal/daemon/dolt.go
@@ -112,6 +112,47 @@ type DoltServerStatus struct {
 	Error     string    `json:"error,omitempty"`
 }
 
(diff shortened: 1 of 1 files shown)


outdoorsea:
Already upstream — content landed on main.

Issue #3354, opened 2026-03-27
bug: FK violation in child_counters when agent beads are ephemeral (wisps table)


that0n3guy:
## Summary

When `gt sling` dispatches work to a polecat, the polecat session startup calls `SetAgentState` to transition from "spawning" to "working". This calls `bd set-state`, which creates a state-change event (a child bead). To generate the child bead ID, `GetNextChildIDTx` does:

```sql
INSERT INTO child_counters (parent_id, last_child) VALUES (?, ?)
```

The `child_counters` table has:
```sql
CONSTRAINT fk_counter_parent FOREIGN KEY (parent_id) REFERENCES issues(id) ON DELETE CASCADE
```

But agent beads (like `saa-solidactions_app-polecat-dementus`) are created with `ephemeral=1`, which routes them to the `wisps` table instead of `issues`. The FK only references `issues.id`, so the insert fails:

```
Error 1452 (HY000): cannot add or update a child row - Foreign key violation on fk:
fk_counter_parent, table: child_counters, referenced table: issues,
key: [saa-solidactions_app-polecat-dementus]
```

## Impact

- All polecats in the town are affected
- `gt sling` works but agent state is never updated
- Polecats show as "spawning" forever
- Witness/refinery can't accurately track polecat lifecycle
- Contributed to pipeline stalls — polecats completed work but the system couldn't track their state transitions

## Fix options

1. Skip child event creation in `bd set-state` when the parent is ephemeral (wisps table) — ephemeral beads shouldn't spawn persistent child events
2. Make `child_counters` FK reference both tables, or drop the FK
3. Don't make agent beads ephemeral (move them to issues table)

## Relevant code

- `beads/internal/storage/issueops/child_id.go` — `GetNextChildIDTx` (the INSERT)
- `beads/internal/storage/dolt/schema.go` — FK definition
- `beads/cmd/bd/state.go:183` — `bd set-state` calls `GetNextChildID`
- `gastown/internal/beads/beads_agent.go` — `CreateAgentBead` sets ephemeral
- `gastown/internal/polecat/manager.go` — `SetAgentState` calls `UpdateAgentState`

## Environment

- gt version 848fe20 (built from main)
- bd version 0.62.0
- Dolt 1.84.0

Bella-Giraffety:
[...]

PR #3478, opened 2026-04-01
fix: update migrated agent bead state directly in wisps

Bella-Giraffety:
## Summary
- update `UpdateAgentState` to write `wisps.agent_state` directly in the routed beads database
- keep the description `agent_state:` field in sync inside the same critical section
- treat `0 rows affected` as a failure unless a readback already shows the requested structured state
- add focused regression tests for the direct SQL path and the important error/no-op cases

## Why
Follow-up to issue #3354.

Migrated agent beads are wisp-backed, but the higher-level state-update path can still create child-event side effects against issue-backed tables. In practice that can fail on `child_counters` foreign keys during polecat reuse and leave agent-state transitions unreliable.

Writing `wisps.agent_state` directly matches where migrated agent-bead state actually lives today, while still keeping the description field aligned for `bd show` and dashboards.

## Validation
- `GOTOOLCHAIN=auto go test ./internal/beads -run 'TestUpdateAgentState_UsesDirectWispSQL|TestUpdateAgentState_ErrorsWhenNoRowsUpdated|TestUpdateAgentState_ErrorsWhenDescriptionSyncFails|TestUpdateAgentState_RetriesDescriptionSyncAfterZeroRowNoOp|TestEscapeSQLString|TestGetAgentBead_'`

## Notes
- This PR is intentionally limited to migrated/wisp-backed agent state updates.
- It does not broaden into a general rewrite of all agent-state write paths.

### DIFF
--- a/internal/beads/beads_agent.go
+++ b/internal/beads/beads_agent.go
@@ -415,15 +415,72 @@ func (b *Beads) ResetAgentBeadForReuse(id, reason string) error {
 // that contract with fallback to the legacy structured column via ResolveAgentState.
 //
 // Resolves the concrete target DB first so the update hits the correct database
-// when the agent bead routes to a different beads dir via routes.jsonl.
+// UpdateAgentState updates the agent_state field in an agent bead.
+// Agent beads live in wisps, so bd set-state is unsafe here: it creates a child
+// event bead and can fail its child_counters foreign key because the parent is
+// not stored in issues. Update the wisps column directly, then sync the
+// description's agent_state field to match.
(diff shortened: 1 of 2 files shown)


Bella-Giraffety:
Closing this after re-auditing it against current upstream main. The child_counters failure behind issue #3354 is real, but this PR bakes in an older storage assumption by hardcoding direct wisps updates as the primary path. That no longer matches current main closely enough for me to stand behind it as the right upstream fix. Leaving the issue open is better than merging the wrong storage model.

PR #3482, opened 2026-04-01
fix: keep daemon Dolt verification in TCP client mode

Bella-Giraffety:
## Summary
- update the daemon's `buildDoltSQLCmd` helper to always use explicit TCP client mode
- preserve inherited `DOLT_CLI_PASSWORD` only for remote checks when config does not specify a password
- keep forcing the empty-password path for local checks so inherited shell credentials cannot make a healthy local daemon look broken
- add focused helper tests for local TCP mode, remote no-password prompt suppression, local inherited-password blocking, and remote inherited-password preservation

## Why
PR #3470 fixed the shared `internal/doltserver` helper but intentionally left the daemon mirror helper unchanged.

That left daemon health, write-probe, and identity-query shellouts able to use the older helper behavior. This PR brings the daemon helper to the same explicit TCP/non-interactive model without broadening into unrelated Dolt verification cleanup.

## Validation
- `GOTOOLCHAIN=auto go test -c ./internal/daemon`
- `GOTOOLCHAIN=auto go test ./internal/doltserver -run 'TestBuildDoltSQLCmd.*|TestCheckServerReachable.*|TestIsRunning.*'`

## Notes
- Follow-up to #3470.
- This PR is intentionally limited to the daemon helper and its focused tests.
- The `internal/doltserver.VerifyServerDataDir` fallback path still needs the same explicit-TCP cleanup in a separate change.

### DIFF
--- a/internal/daemon/dolt.go
+++ b/internal/daemon/dolt.go
@@ -144,19 +144,19 @@ type DoltServerManager struct {
 	onRecoveryFn func()
 
 	// Test hooks (nil = use real implementations; set only in tests)
-	healthCheckFn      func() error
-	writeProbeCheckFn  func() error
-	identityCheckFn    func() error // nil = use real VerifyServerDataDir
-	startFn            func() error
-	runningFn          func() (int, bool)
-	stopFn             func()
-	sleepFn            func(time.Duration)
-	nowFn              func() time.Time
-	escalateFn         func(int)
-	unhealthyAlertFn   func(error)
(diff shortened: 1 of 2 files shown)


Bella-Giraffety:
Follow-up note: the remaining VerifyServerDataDir fallback path that still queried SHOW DATABASES through the older local helper is now addressed separately in #3485, so this PR can stay scoped to the daemon helper only.

PR #3485, opened 2026-04-01
fix: force VerifyServerDataDir fallback over TCP

Bella-Giraffety:
## Summary
- add a dedicated explicit-TCP `dolt sql` helper for verification paths
- switch `VerifyServerDataDir`'s fallback `SHOW DATABASES` query to that helper
- preserve inherited `DOLT_CLI_PASSWORD` only for remote fallback checks when config omits a password
- add focused tests for local and remote explicit-TCP command construction

## Why
Builds on the explicit-TCP helper work in #3470 and the daemon-helper parity follow-up in #3482.

`VerifyServerDataDir` still had one fallback path that could issue `SHOW DATABASES` through the older local `dolt sql` behavior. That left imposter detection able to drift back into embedded-mode / autodiscovery behavior instead of querying the live shared server.

This PR keeps that remaining fallback path on the same explicit TCP client path without broadening into unrelated Dolt shellout cleanup.

## Validation
- `GOTOOLCHAIN=auto go test ./internal/doltserver -run 'TestBuildDoltSQLTCPClientCmd_(Local|Remote|RemoteNoPasswordPreservesInheritedCredentials)$|TestBuildDoltSQLCmd.*|TestCheckServerReachable.*|TestIsRunning.*'`

## Notes
- Follow-up to #3482.
- Builds on the explicit-TCP rationale from #3470.
- Intentionally limited to `internal/doltserver` verification fallback behavior and focused tests.

### DIFF
--- a/internal/doltserver/doltserver.go
+++ b/internal/doltserver/doltserver.go
@@ -425,8 +425,46 @@ func buildDoltSQLCmd(ctx context.Context, config *Config, args ...string) *exec.
 	cmd.Dir = config.DataDir
 	setProcessGroup(cmd)
 
-	if config.IsRemote() && config.Password != "" {
+	if config.Password != "" {
 		cmd.Env = append(os.Environ(), "DOLT_CLI_PASSWORD="+config.Password)
+	} else if config.IsRemote() {
+		if inherited, ok := os.LookupEnv("DOLT_CLI_PASSWORD"); ok {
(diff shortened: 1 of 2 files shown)


Bella-Giraffety:
Closing this as redundant after re-auditing the stack. The VerifyServerDataDir fallback bug is real, but keeping #3470 plus #3482 is the cleaner split to carry forward, and this PR now reads as an extra follow-up layer in the same TCP-helper chain rather than an independent must-merge fix.

PR #3497, opened 2026-04-02
feat(formula): add --set flag and fix bead prefix for convoy formulas

ckumar1:
## Problem

1. **No way to pass input values to convoy formulas.** `gt formula run` only accepted `--pr` and `--files`, so convoy formulas with `[inputs]` variables had no mechanism to receive values.

2. **Bead prefix mismatch in convoy formulas.** The convoy formula executor used the wrong prefix when creating beads, causing lookup failures downstream.

## Changes

- Add `--set key=value` flag to `gt formula run` for passing arbitrary input values to formulas
- Fix bead prefix resolution in `executeConvoyFormula`

## Usage

```bash
gt formula run my-formula --set reviewer=alice --set priority=high
```

1 file changed, self-contained.

### DIFF
--- a/internal/cmd/formula.go
+++ b/internal/cmd/formula.go
@@ -32,6 +32,7 @@ var (
 	formulaRunDryRun  bool
 	formulaRunAgent   string
 	formulaRunFiles   []string
+	formulaRunSet     []string
 	formulaCreateType string
 )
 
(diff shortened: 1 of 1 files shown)


steveyegge:
Landed via #3613 (sheriff rebase with credit to @ckumar1). Thanks for the fix!

steveyegge:
Landed via #3613 (sheriff rebase with credit to @ckumar1). Thanks for the fix!

PR #3503, opened 2026-04-02
fix: qualify pane ID with session name in FindAgentPane verification

ckumar1:
## Summary

`FindAgentPane` and `NudgeSessionWithOpts` qualify pane IDs with the session name to avoid ambiguity in multi-session environments. The existing code used `session:%N` syntax (e.g., `hq-dog-alpha:%22`), but `tmux send-keys` interprets this as a **window** reference, failing with "can't find window: %22".

The correct tmux target syntax for panes is `session:.%N` (note the dot), which tells tmux to look up a pane rather than a window.

`display-message` accepts both forms, which masked the bug — `FindAgentPane`'s verification via `display-message` passed, but the subsequent `send-keys` delivery failed.

## What changed

- `internal/tmux/tmux.go`: Changed pane qualification from `session + ":" + pane` to `session + ":." + pane` in both `NudgeSessionWithOpts` (delivery) and `FindAgentPane` (verification)

## Reproduction

```
$ tmux send-keys -t "session:%22" ""
can't find window: %22

$ tmux send-keys -t "session:.%22" ""
(success)
```

## Test plan

- [x] `go build ./cmd/gt` passes
- [x] `go vet ./internal/tmux/` clean
- [x] `TestFindAgentPane_*` tests pass
- [x] Manual: `gt nudge` succeeds for target with pane ID `%22`

### DIFF
--- a/internal/tmux/tmux.go
+++ b/internal/tmux/tmux.go
@@ -1646,12 +1646,13 @@ func (t *Tmux) NudgeSessionWithOpts(session, message string, opts NudgeOpts) err
 	// running the agent rather than sending to the focused pane.
 	target := session
 	if agentPane, err := t.FindAgentPane(session); err == nil && agentPane != "" {
-		// Qualify the pane ID with the session name (e.g., "hq-dog-alpha:%1")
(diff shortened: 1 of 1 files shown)


steveyegge:

PR #3508, opened 2026-04-02
feat(crew): mail mayor before actions requiring human intervention

n43-ryan:
## Summary

- Crew agents now send `HUMAN_INPUT_REQUIRED` mail to mayor before attempting actions that will block on a permission prompt
- Covers `.claude/` directory edits, destructive ops, and any tool call that triggers a permission dialog
- Agent mails mayor, nudges, then proceeds immediately — mayor notifies the user to come unblock

## Context

Upstream already addresses permission prompts at the infrastructure layer (`--dangerously-skip-permissions`, `AcceptStartupDialogs` tmux auto-dismiss, witness dialog auto-dismiss in `DismissStartupDialogsBlind`). This PR targets the **residual case** those mechanisms don't cover: Claude Code's own `.claude/` configuration files, which remain protected even under `--dangerously-skip-permissions`.

When a crew agent attempts to edit `.claude/commands/` or `settings.json`, Claude Code presents an interactive confirmation that blocks indefinitely in an autonomous session. This protocol gives the human a path to unblock it.

## Protocol added

```bash
gt mail send mayor/ -s "HUMAN_INPUT_REQUIRED: <brief description>" --stdin <<'BODY'
About to: <describe the action>
Why: <why you need to do this>
What will happen: The agent will block on a permission prompt until the user approves.
BODY
gt nudge mayor "Check mail — human approval needed"
# Then proceed immediately — do NOT wait for reply
```

## Follow-up (out of scope for this PR)

A PreToolUse hook that auto-detects `.claude/` write attempts and fires the mail+nudge automatically would be more robust than an instructions-layer protocol. Tracked separately.

## Test plan

- [ ] Crew agent encountering a `.claude/` edit mails mayor before attempting
- [ ] Mayor receives `HUMAN_INPUT_REQUIRED` subject
- [ ] Agent proceeds without waiting for reply (does not deadlock)
- [ ] Non-crew agents (polecats, witnesses) unaffected


n43-ryan:
Closing in favor of a clean branch PR. The N43-Studio:main head mixed in 249 upstream sync commits; new PR from N43-Studio:pr-gt-7y7-gpy is clean.

PR #3511, opened 2026-04-03
fix: sling retry noise, convoy cross-rig routing, mail error detection

maloferriol:
## Summary
- **Sling retries**: Reduce `SetAgentStateWithRetry` from 10 retries (~2min) to 3 (~3.5s). Agent state is monitoring-only; caller treats errors as warn-only.
- **Convoy cross-rig routing**: Fix `bdListChildren` to use `resolveBeadDir` (returns parent of .beads) instead of `beadsDirForID` (returns .beads itself, causing bd to look for .beads/.beads/). Also strip BEADS_DIR and fix raw exec.Command in `createStagedConvoy`.
- **Mail error detection**: Add "no issue found" match to all `ContainsError` checks. bd returns "no issue found matching" but mail only checked for "not found", preventing cross-rig wisp fallback from triggering.

## Test plan
- [x] `bd show sh-wisp-xxx` works from both town and shipyard contexts
- [x] `gt mail read sh-wisp-xxx` returns wisp content
- [x] `bd create` from town root generates `hq-` prefix (not `sh-`)
- [x] `bd show`, `bd close` work for both prefixes
- [ ] `gt convoy create --from-epic` with cross-rig dependencies
- [ ] `gt sling` completes without 2-minute retry delay

### DIFF
--- a/internal/cmd/convoy_stage.go
+++ b/internal/cmd/convoy_stage.go
@@ -726,9 +726,11 @@ func createStagedConvoy(dag *ConvoyDAG, waves []Wave, status string, title strin
 	}
 
 	// Set the staged status.
-	statusCmd := exec.Command("bd", "update", convoyID, "--status="+status)
-	statusCmd.Dir = townBeads
-	if out, err := statusCmd.CombinedOutput(); err != nil {
+	// Strip BEADS_DIR so bd discovers the correct database from Dir()
+	// rather than using an inherited (possibly wrong) override.
+	if out, err := BdCmd("update", convoyID, "--status="+status).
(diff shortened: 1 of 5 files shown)


gerfalcon:
Can confirm this bug on `gt v1.0.0 (5f07285e)` with a two-prefix setup (`hq-` town + `po-` rig, routed via `routes.jsonl`). PR #3209 is in our build but `gt convoy status` still shows 0/0 for cross-rig beads. The `.beads/.beads/` double-nesting in `bdListChildren` matches what we're seeing.

jordanlewis:
Chiming in to agree that I can trivially reproduce this one. Thanks @maloferriol

Issue #2832, opened 2026-03-15
bug: Scheduler integration tests flaky on main — different test fails each run


rileywhite:
## Summary

Scheduler integration tests in `internal/cmd` fail intermittently on `main`. The specific test that fails varies between runs — it's not a single broken test but a class of flaky tests.

## Evidence

| Run | Branch | Failing Test | Integration Result |
|-----|--------|-------------|-------------------|
| [23120418617](https://github.com/steveyegge/gastown/actions/runs/23120418617) | PR #2806 | `TestSchedulerDirectConvoyDispatch` | fail |
| [23117077787](https://github.com/steveyegge/gastown/actions/runs/23117077787) | main `5b8dc2cc` | `TestSchedulerMultiRigConvoyAutoResolve` | fail |
| [23102887532](https://github.com/steveyegge/gastown/actions/runs/23102887532) | main `317c960c` | `TestSchedulerMultiRigConvoyAutoResolve` | fail |
| [23120335042](https://github.com/steveyegge/gastown/actions/runs/23120335042) | main `a3b672dc` | *(none)* | pass |
| [23095216845](https://github.com/steveyegge/gastown/actions/runs/23095216845) | main `f28273a1` | *(none)* | pass |

The same tests pass on some runs and fail on others, suggesting a timing or resource sensitivity rather than a logic bug.

## Affected Tests

- `TestSchedulerMultiRigConvoyAutoResolve`
- `TestSchedulerDirectConvoyDispatch`

Both are in `internal/cmd/scheduler_integration_test.go`.

## Impact

Flaky integration tests cause false CI failures on PRs, requiring contributors to investigate failures unrelated to their changes. Since different tests fail each run, it's difficult to tell at a glance whether a PR introduced a real regression or hit a flaky scheduler test.

DreadPirateRobertz:
Looking at the failing tests (`TestSchedulerMultiRigConvoyAutoResolve`, `TestSchedulerDirectConvoyDispatch`), both use `setupMultiRigSchedulerTown` which calls `requireDoltServer(t)` and `cleanStaleBeadsDatabases(t)` — they share a live Dolt server on port 3307.

The atomic counter (`schedulerTestCounter`) generates unique prefixes per test, but when tests run in parallel (Go's default for `t.Parallel()`), there can be race conditions in:

1. **`cleanStaleBeadsDatabases`** — if multiple tests clean simultaneously, one test's setup databases could get dropped by another test's cleanup
[...]

Issue #2624, opened 2026-03-09
fix(convoy): cross-database dependency resolution fails for multi-rig towns


Cdfghglz:
## Problem

In multi-rig Gas Town setups where each rig has its own Dolt database, convoy auto-dispatch is completely broken. Convoys are always `hq-cv-*` (stored in the HQ database), but tracked issues use rig-specific prefixes (e.g., `ds-*` in the dashboard database). The dependency resolution cannot cross this database boundary.

### Symptoms

- `gt convoy status hq-cv-xxx` → "Progress: 0/0 completed" despite tracking 14 issues
- `gt convoy launch hq-cv-xxx` → "tracks no beads"
- `bd dep list hq-cv-xxx -v` → finds 14 raw "external-deps" but reports "has no dependencies"
- Daemon logs: "Convoy hq-cv-xxx: empty but within grace period"
- `gt convoy stage <epic>` works (resolves deps at stage time) but the created convoy can't be launched or tracked

### Root Cause

Two code paths in `internal/convoy/operations.go` fail for cross-database deps:

1. **`getConvoyTrackedIssues()` (line ~293)**: Calls `store.GetDependenciesWithMetadata()` on the HQ store, which returns `external:ds:ds-xxx` refs. The fallback `fetchCrossRigBeadStatus()` (line ~336) resolves these via subprocess `bd show`, but the convoy status/launch commands don't use this path — they see 0 tracked beads.

2. **`isIssueBlocked()` (line ~193)**: Calls `store.GetDependenciesWithMetadata()` directly with NO cross-database fallback. Returns stale/empty metadata for cross-rig blocking deps, causing incorrect dispatch decisions.

### The Catch-22

- `hq-*` prefixed issues → same DB as convoy ✓ → but `rigForIssue()` returns empty (can't route) ✗
- `ds-*` prefixed issues → `rigForIssue()` routes correctly ✓ → but convoy can't resolve deps ✗

### Impact

Multi-rig towns cannot use convoy auto-dispatch at all. Manual `gt sling` per wave is required, defeating the purpose of the convoy system.

### Related Issues

Consolidates: #2605, #2606, #2604, #2423. Prior closed instances: #2373, #992, #915.

## Proposed Fix

The daemon already has `map[string]beadsdk.Storage` with all rig stores open (`daemon.go:openBeadsStores()`). The data is available — just not wired through to the convoy operations.

**Create a multi-store resolver** that:
1. Accepts the store map from the daemon
[...]

PR #3512, opened 2026-04-03
fix: convoy cross-rig tracking via direct SQL INSERT

maxinflection:
## Summary

Convoy tracking relations are 100% broken on bd >= v0.62 because `bd dep add`
validates that both issue IDs exist in the same database. Since convoys live
in hq while tracked issues live in rigs, every convoy tracking relation is
cross-database by definition.

- Replace `bd dep add` with direct SQL `INSERT IGNORE INTO dependencies` via
  `bd sql`, which writes to the hq dependencies table without resolving the
  foreign issue ID
- Extract `addTracksDep`/`removeTracksDep` helpers (5 call sites across
  convoy create, add, stage, and stage-update)
- Fix `getIssueDetails` to use `resolveBeadDir()` for cross-rig `bd show`,
  so `convoy status` correctly fetches rig issue details

## Context

- `bd dep add` validates both IDs exist in the local store (by design — bd
  is gt-unaware per maintainer policy)
- The read side (`bdDepListRawIDs`, `getTrackedIssues`) already uses raw SQL
  for the same cross-database reason (#2624, #2832)
- `bd sql` for writes has precedent in `doctor/null_assignee_check.go`,
  `doltserver/wisps_migrate.go`, and others
- Works across all Dolt modes (server, embedded) and multi-server configs
  since the INSERT only targets the hq database

## Test plan

- [x] `TestConvoyCreate_DepAddUsesTownRoot` updated — verifies SQL INSERT
  runs from town root with correct query
- [x] All existing convoy tests pass (`go test ./internal/cmd/ -run TestConvoy`)
- [x] Full suite passes (`go test ./...` — 0 regressions)
- [x] Manual e2e: create convoy tracking 3 rig issues, close them
  one by one, verify `convoy status` shows progress (0/3 → 1/3 → 2/3 → 3/3),
  `convoy check` auto-closes on completion

### DIFF
--- a/internal/cmd/convoy.go
+++ b/internal/cmd/convoy.go
@@ -470,6 +470,60 @@ func runBdJSON(dir string, args ...string) ([]byte, error) {
 	return stdout.Bytes(), nil
 }
 
+// runBdSQL executes a write query via bd sql with auto-commit enabled.
+// Used for direct dependency writes that bypass bd dep add's cross-database
+// validation limitation.
+func runBdSQL(dir, query string) ([]byte, error) {
+	cmd := BdCmd("sql", query).
+		WithAutoCommit().
+		Dir(dir).
+		StripBeadsDir()
+	built := cmd.Build()
+	var stdout, stderr bytes.Buffer
+	built.Stdout = &stdout
+	built.Stderr = &stderr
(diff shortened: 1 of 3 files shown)


steveyegge:

PR #3517, opened 2026-04-04
feat(opencode): close sidebar by default and add continuous mail checking

justintanner:
## Summary

- **Sidebar auto-close**: New TUI plugin (`gastown-tui.js`) that triggers `sidebar.toggle` on session creation, freeing screen real estate in tmux/Docker
- **Continuous mail checking**: Added debounced `gt mail check --inject` (30s interval) in `experimental.chat.system.transform` for autonomous roles — previously mail was only checked once at session start, making inter-agent messages invisible until compaction
- **Multi-file plugin support**: Added `AdditionalHooksFiles` to agent presets so `gt hooks sync` and `gt doctor` deploy and verify both plugin files

## Test plan

- [x] Docker build succeeds
- [x] `gt hooks sync` deploys both `gastown.js` and `gastown-tui.js` to all OpenCode locations
- [x] `gt doctor` detects missing `gastown-tui.js` and reports out of sync
- [x] `gt doctor --fix` restores missing plugin files
- [x] All unit tests pass (hooks, runtime, config, doctor)


justintanner:
Superseded — squashed into main and reworked the plugin approach (export rename to `server`, debug logging).

PR #3521, opened 2026-04-04
fix(adopt): run post-InitBeads finalization in rig adoption path

outdoorsea:
## Summary

- `gt rig adopt` calls `InitBeads` but skips the post-init steps that `AddRig` performs (EnsureMetadata, orphan cleanup, config re-set)
- This leaves `metadata.json` with `dolt_database=beads_<prefix>` instead of the rig name, causing all `bd` commands to fail with `database "<prefix>" not found`
- Adds `finalizeBeadsAfterInit()` helper called from both adoption code paths, mirroring the post-init sequence already in `manager.go` AddRig (lines 632-666)

## Reproduction

1. `gt rig adopt` a rig with a beads prefix that differs from the directory name (e.g. prefix `mm` for rig `meety_me`)
2. Run any `bd` command in the adopted rig
3. Fails with: `Error: database "mm" not found on Dolt server`

## Changes

Single file: `internal/cmd/rig.go` (+39 lines)
- New `finalizeBeadsAfterInit()` function performing: EnsureMetadata, orphan DB cleanup, issue_prefix + types.custom re-set
- Called from both adoption paths (existing `.beads` candidate reinit and fresh init fallback)

## Test plan

- [x] `go build ./cmd/... ./internal/...` passes
- [x] Existing `TestRigAdoptBeadsCandidateDetection` and `TestRigAdoptFallbackInitNeeded` pass
- [x] Manually verified: adopted rig's `bd doctor` shows 0 errors after fix

### DIFF
--- a/internal/cmd/rig.go
+++ b/internal/cmd/rig.go
@@ -15,6 +15,7 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/steveyegge/gastown/internal/beads"
 	"github.com/steveyegge/gastown/internal/config"
+	"github.com/steveyegge/gastown/internal/constants"
 	"github.com/steveyegge/gastown/internal/crew"
 	"github.com/steveyegge/gastown/internal/deps"
 	"github.com/steveyegge/gastown/internal/doltserver"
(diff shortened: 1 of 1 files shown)


outdoorsea:
Already upstream — FinalizeBeads was refactored into the manager on main.

PR #3535, opened 2026-04-05
fix(nudge): use correct tmux target syntax for pane IDs

### LINKED ISSUE #3534: Nudge broken on macOS/Linux: session:paneID target format is invalid tmux syntax
## Summary

`NudgeSessionWithOpts` in `internal/tmux/tmux.go` constructs a tmux target as `session + ":" + agentPane` (e.g., `co-obsidian:%1871`). In tmux target syntax, the segment after `:` is a **window** specifier, not a pane specifier. Pane IDs (`%NNN`) are not valid window identifiers (windows use `@NNN`). This causes all nudges to fail with:

```
can't find window: %1871
```

## Introduced

Commit `2781602a` (2026-03-30) — `feat(windows): cherry-pick Windows platform support (#3387)`.

Changed `target = agentPane` to `target = session + ":" + agentPane` to handle tmux 3.3 on Windows where pane IDs are not globally unique. The fix is correct in intent but uses invalid tmux target syntax.

## Repro

```bash
# 1. Create a session
tmux new-session -d -s test-session

# 2. Get its pane ID
PANE=$(tmux list-panes -t test-session -F '#{pane_id}')
echo "Pane: $PANE"  # e.g., %42

# 3. Set GT_PANE_ID (simulating what gastown does at session startup)
tmux set-environment -t test-session GT_PANE_ID "$PANE"

# 4. This works (bare pane ID):
tmux send-keys -t "$PANE" "# hello" Enter

# 5. This works (session name only):
tmux send-keys -t "test-session" "# hello" Enter

# 6. This FAILS (what gastown does):
tmux send-keys -t "test-session:$PANE" "# hello" Enter
# Error: can't find window: %42

# 7. This works (correct qualified format):
WINDOW=$(tmux list-panes -t test-session -F '#{window_id}')
tmux send-keys -t "test-session:${WINDOW}.${PANE}" "# hello" Enter

# Cleanup
tmux kill-session -t test-session
```

## Impact

All `gt nudge` calls fail on macOS/Linux. This breaks:
- Mail delivery notifications (polecats never see incoming mail)
- Inter-agent communication via nudge
- Patrol cycle signaling

## Suggested fix

For the non-Windows case, use bare pane ID (globally unique on macOS/Linux). For Windows, use `session:@windowID.%paneID` format. See PR (incoming).

## Workaround

Clear `GT_PANE_ID` from session environments so `FindAgentPane` returns empty for single-pane sessions, falling back to session-name targeting:

```bash
tmux set-environment -t <session> -u GT_PANE_ID
```

jordanlewis:
## Summary

- Fix `NudgeSessionWithOpts` using invalid tmux target format `session:%paneID` — the segment after `:` is a **window** specifier, so `%paneID` there causes "can't find window" errors
- On macOS/Linux, revert to bare pane ID (globally unique); on Windows, use proper `session:@window.%pane` syntax
- Regression introduced in #3387 (`2781602a`, 2026-03-30)

## Test plan

- [x] Verified repro: `tmux send-keys -t "session:%pane"` fails, bare `%pane` works
- [x] Verified fix compiles (`go build ./internal/tmux/`)
- [x] Verified workaround (clearing GT_PANE_ID) restores nudge delivery on live Gas Town instance
- [ ] Run `go test ./internal/tmux/...`
- [ ] Test on Windows with tmux 3.3 (non-unique pane IDs)

Fixes #3534

### DIFF
--- a/internal/tmux/tmux.go
+++ b/internal/tmux/tmux.go
@@ -1646,12 +1646,24 @@ func (t *Tmux) NudgeSessionWithOpts(session, message string, opts NudgeOpts) err
 	// running the agent rather than sending to the focused pane.
 	target := session
 	if agentPane, err := t.FindAgentPane(session); err == nil && agentPane != "" {
-		// Qualify the pane ID with the session name (e.g., "hq-dog-alpha:%1")
-		// to avoid ambiguity. On some tmux versions (e.g., 3.3 on Windows),
-		// pane IDs are NOT globally unique — every session may have "%1".
-		// A bare "send-keys -t %1" targets the attached session's pane,
-		// not necessarily this session's.
-		target = session + ":" + agentPane
+		if runtime.GOOS == "windows" {
+			// On Windows tmux 3.3, pane IDs are NOT globally unique —
+			// every session may have "%1". Qualify with session and window
+			// using proper tmux target syntax: session:@window.%pane.
+			// (The segment after ":" is a window specifier — bare pane IDs
+			// in that position cause "can't find window" errors.)
+			if winID, err := t.run("display-message", "-t", agentPane, "-p", "#{window_id}"); err == nil {
+				target = session + ":" + strings.TrimSpace(winID) + "." + agentPane
+			} else {
+				target = agentPane
+			}
+		} else {
+			// On macOS/Linux, pane IDs are globally unique — use them
+			// directly. Do NOT use "session:%pane" format: tmux interprets
+			// the segment after ":" as a window specifier, and pane IDs
+			// are not valid window identifiers. (GH#3534)
+			target = agentPane
+		}
 	}
 
 	// 0. Pre-delivery: dismiss Rewind menu if the session is stuck in it.


A3Ackerman:
+1 on this PR. We're running a multi-rig Gas Town setup and the broken nudge is cascading into several downstream failures for us:

1. **Mail delivery silently fails** — `gt mail send` auto-nudge fires but the tmux target is invalid, so recipients never wake to read mail. Wisp GC then reaps the unread messages. From the user's perspective, inter-agent mail just disappears.
2. **Deacon gets stuck** — deacon tries to nudge agents during patrol, hits the same broken targeting, and hangs. The daemon detects the stuck heartbeat but can't auto-restart (separate bug), so we get an escalation flood with nobody acting on it.
3. **All inter-agent coordination degrades** — the nudge system is the backbone of agent-to-agent comms in GT. With it broken, everything falls back to polling or manual intervention.

We evaluated all four competing PRs for #3534:

| PR | Approach | Windows? | Scope | CI |
|---|---|---|---|---|
| **#3535 (this one)** | `runtime.GOOS` branch + window ID resolution | ✅ Yes | 1 file, 35 lines | No full run |
| #3548 | Bare pane ID + WakePane fix | ❌ No | 1 file, 88 lines | ❌ All failing |
| #3556 | Bare pane ID + deacon startup fix | ❌ No | 4 files, 114 lines | No full run |
| #3566 | `%` prefix check + unrelated lint fixes | Partial | 6 files, 224 lines | ❌ All failing |

This PR is the best option because:
- **Only one that handles Windows correctly** — resolves the actual window ID via `display-message` rather than assuming bare pane IDs are globally unique (they aren't on tmux 3.3/Windows)
- **Smallest, most focused diff** — 35 lines in one function, trivial to review and revert if needed
- **No bundled extras** — the other PRs all mix in unrelated fixes (deacon startup, lint cleanup, WakePane) which increase merge risk

The fixes in #3548 (WakePane targeting) and #3556 (deacon startup prompt) are real bugs worth fixing separately, but shouldn't block landing this critical nudge fix.

We're cherry-picking this onto our carry/operational branch in the meantime. Happy to report back on production results.

steveyegge:

PR #3480, opened 2026-04-01
fix(tmux): use bare pane ID in NudgeSession, resolve window ID in WakePane (la-imkr)

seanbearden:
## Summary

- **NudgeSessionWithOpts** qualified pane IDs with session names (`session:%95`), but tmux parses `session:X` as `session:window` — failing with `can't find window: %95`. This broke all nudge delivery to crew agents via `FindAgentPane`, causing agents to sit idle despite having mail.
- **WakePane** passed pane IDs directly to `resize-window`, which requires a window target. Now resolves the window ID via `display-message` first.
- **WakePaneIfDetached** in `NudgeSessionWithOpts` now receives the resolved target (pane ID) instead of session name, correctly waking the agent's window in multi-window sessions.

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./internal/tmux/ -run Nudge` passes
- [x] Manual verification: `tmux send-keys -t %95` succeeds, `tmux send-keys -t "session:%95"` fails with the reported error
- [x] Manual verification: `tmux display-message -p -t %95 "#{window_id}"` correctly resolves `@95`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3548, opened 2026-04-06
fix(tmux): use bare pane ID in NudgeSession, resolve window ID in WakePane

seanbearden:
## Summary

- **NudgeSession**: `session + ":" + paneID` (e.g., `"session:%95"`) is parsed by tmux as `session:window`, causing `can't find window: %95`. Fix uses bare pane IDs which are globally unique.
- **WakePane**: `resize-window` requires a window target, not a pane target. Fix resolves `#{window_id}` from the target before resizing.
- **WakePaneIfDetached**: Pass resolved target (pane ID) instead of session name so multi-window sessions wake the correct window.

Supersedes #3480 (stale branch from fork reset).

## Test plan

- [x] `go build ./...` passes
- [x] `go test ./internal/tmux/ -run "Nudge|Wake"` — all pass
- [ ] Manual: `gt nudge` to a crew agent in a multi-pane session delivers successfully
- [ ] Manual: detached sessions receive SIGWINCH after nudge

### DIFF
--- a/internal/tmux/tmux.go
+++ b/internal/tmux/tmux.go
@@ -1289,6 +1289,18 @@ func (t *Tmux) IsSessionAttached(target string) bool {
 // Note: This always performs the resize. Use WakePaneIfDetached to skip
 // attached sessions where the wake is unnecessary.
 func (t *Tmux) WakePane(target string) {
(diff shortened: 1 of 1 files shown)


steveyegge:

PR #3556, opened 2026-04-08
Fix deacon patrol startup and stuck-session nudges

davidmwkim:
## Summary
- update Deacon startup prompts to use `gt sling mol-deacon-patrol deacon` instead of the nonexistent `wisp` flow
- fix tmux nudge targeting to use pane ids directly instead of invalid `session:%pane` targets
- add a tmux regression test covering the multi-pane `GT_PANE_ID` nudge path

## Validation
- `go test ./internal/tmux -run 'TestNudgeSession_WithRetry|TestNudgeSession_UsesPaneIDTarget' -count=1`
- `go test ./internal/deacon ./internal/cmd` currently fails on an unrelated pre-existing `internal/cmd/compact_report_test.go` compile error on `main`

### DIFF
--- a/internal/cmd/deacon.go
+++ b/internal/cmd/deacon.go
@@ -526,7 +526,7 @@ func startDeaconSession(t *tmux.Tmux, sessionName, agentOverride string) error {
 		Recipient: "deacon",
 		Sender:    "daemon",
 		Topic:     "patrol",
-	}, "I am Deacon. First run `gt deacon heartbeat`. Then check gt hook, if empty create mol-deacon-patrol wisp and execute it.")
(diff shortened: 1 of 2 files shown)


sottnik:
+1, hitting this in production. Manual gt deacon restart lands the deacon at an "Interrupted" prompt every time,  traced to the broken gt wisp reference in cmd/deacon.go:521 exactly as this PR describes. The bug has been on main since the mass revert in 3e67d3f7 on 2026-02-20 (~7 weeks). Happy to test the fix locally if it would help get this reviewed. Thanks for writing it up.

PR #3566, opened 2026-04-09
fix(tmux): use bare pane ID in send-keys target

baumgold:
## Problem

`gt nudge` was failing with `Error: nudging session: tmux send-keys: can't find window: %198` on tmux 3.6a. The bug was in how `NudgeSession` built the `send-keys` target: it concatenated the session name with the pane ID as `session:%N`. In tmux, the `:` separator is used to split *session* from *window* — it does not introduce a pane ID. So `session:%198` is parsed as "window named `%198` in `session`", which doesn't exist, causing the error.

## Solution

When the pane reference starts with `%`, it's a globally unique pane ID (tmux 3.4+ on Linux/macOS) and should be used bare as the `send-keys` target. Session qualification (`session:window`) is only needed for numeric window/pane indices that are not globally unique.

Also includes fixes for pre-existing CI failures inherited from the base branch:
- `compact_report_test.go`: test called `wispTypeToCategory` with 1 arg; function signature takes 2 (`wispType, title`)
- `engineer.go`: `doMergePR` accepted a `ctx context.Context` it never used (unparam lint)
- `capacity_dispatch.go` / `scheduler.go` / `sling_schedule.go`: `listAllSlingContexts` declared an error return that was always `nil`; removed it and updated all callers

### DIFF
--- a/internal/cmd/capacity_dispatch.go
+++ b/internal/cmd/capacity_dispatch.go
@@ -247,10 +247,7 @@ func beadsForContext(townRoot string, fields *capacity.SlingContextFields) *bead
 // cleanupStaleContexts closes invalid and stale sling context beads.
 // Called explicitly before the dispatch cycle to separate cleanup from querying.
 func cleanupStaleContexts(townRoot string) {
-	contexts, err := listAllSlingContexts(townRoot)
-	if err != nil {
-		return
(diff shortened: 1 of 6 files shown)


steveyegge:

PR #3578, opened 2026-04-10
fix(nudge): use correct tmux target syntax for pane IDs

ckumar1:
## Summary

Fixed tmux target syntax when nudging sessions by pane ID. The previous code used an incorrect target format that could fail to deliver nudges to the right pane.

## Test plan
- [x] `go test ./internal/tmux/` — passes
- [x] `go build ./cmd/gt` — builds clean


steveyegge:
This fix is already in `main` — a concurrent PR landed the same pane-ID normalization. When I rebased your branch onto main, it collapsed to empty (all your changes were already present). Closing as landed-via-other-PR. Thanks for the work!

PR #3590, opened 2026-04-11
fix(scheduler): accept positional bead-id arg in 'gt scheduler clear'

popandpeek:
## Summary

- `gt scheduler clear <bead-id>` now correctly removes only the specified bead (positional arg)
- Previously, cobra silently ignored the positional arg and cleared the entire queue
- The `--bead` flag still works for backward compatibility

## Root cause

`gt scheduler promote <bead-id>` and `gt scheduler demote <bead-id>` accept bead IDs as positional args, so users naturally type `gt scheduler clear <bead-id>` expecting the same UX. But `clear` only accepted `--bead` as a flag, and since `schedulerClearCmd` had no `Args` validator, cobra silently dropped the positional arg — resulting in the entire queue being cleared.

## Changes

- `internal/cmd/scheduler.go`: Add `Args: cobra.MaximumNArgs(1)`, resolve target bead from positional arg first (positional wins over `--bead`), update help text
- `internal/cmd/scheduler_queue_test.go`: Add regression tests for positional arg, flag backward compat, and not-found case

## Test plan

- `TestSchedulerClearPositionalArg`: verifies `gt scheduler clear bead1` removes only bead1, leaves bead2 intact
- `TestSchedulerClearFlagStillWorks`: verifies `--bead` flag still works
- `TestSchedulerClearNotFound`: verifies "No sling context found" message for unknown bead

Issue: be-2wdvi


steveyegge:
Thanks @popandpeek for the scheduler-clear UX improvement — apologies for the churn here. The underlying fix (accept positional bead-id on 'gt scheduler clear', matching 'gt scheduler promote/demote' UX) is a real papercut and we want it.

Unfortunately this branch has drifted quite far from main (79 commits, +5577/-401 across the gastown tree), and the fix commit depends on earlier scheduler-feature commits in the branch that overlap with work that has since landed via other PRs. Untangling which slices are still needed vs. which are superseded is the kind of archaeology that's simpler for the maintainers to reimplement internally than to put back on you.

Closing this as 'would-reimplement' — we'll open an internal bead for the scheduler-clear positional-arg fix and credit you there. No action needed from you. If there are other parallel fixes in this branch you still want us to look at, feel free to open narrower follow-up PRs and we'll land those quickly.

Thanks again for the work.

PR #3618, opened 2026-04-12
fix(npm): release HTTPS response stream in postinstall downloadFile (gh#3595)

### LINKED ISSUE #3595: `npm install -g @gastown/gt` fails on Windows — `postinstall.js` file handle leak prevents zip extraction
# Describe the bug

Running `npm install -g @gastown/gt` on Windows fails during the `postinstall` script.

The binary downloads successfully through redirects, but ZIP extraction fails with:

~~~text
New-Object : Exception calling ".ctor" with "3" argument(s): "The process cannot access the file
'C:\\Users\\...\\@gastown\\gt\\bin\\gastown_0.12.0_windows_amd64.zip' because it is being
used by another process."
~~~

The postinstall script then errors out with:

~~~text
Binary not found after extraction: ...\\bin\\gt.exe
~~~

# To reproduce

Run the below install in Windows:
~~~bash
npm install -g @gastown/gt
~~~

# Expected behavior

The ZIP downloads, extracts successfully, and `gt.exe` is placed in the `bin/` directory.

# Root cause

In `scripts/postinstall.js`, the `downloadFile()` function pipes the HTTP response to a file and resolves when `file.close()` completes, but it never destroys the HTTPS response stream.

On Windows, the readable stream still holds a file handle lock, so when `extractZip()` immediately calls `Expand-Archive`, PowerShell cannot open the ZIP file.

Relevant code (approximately lines 76–86):

~~~js
response.pipe(file);
file.on('finish', () => {
  file.close((err) => {
    if (err) reject(err);
    else resolve(); // resolves here, but `response` is never closed
  });
});
~~~

Additionally, redirect handling (approximately lines 66–70) has the same issue. It starts a recursive `downloadFile()` call without destroying the current response:

~~~js
if (response.statusCode === 301 || response.statusCode === 302) {
  const redirectUrl = response.headers.location;
  downloadFile(redirectUrl, dest).then(resolve).catch(reject);
  // `response` and `request` are never destroyed here
  return;
}
~~~

# Why this only affects Windows

Unix uses advisory file locking, so multiple processes can read a file simultaneously even if another process still has it open.

Windows uses mandatory file locking, so `Expand-Archive` cannot open the ZIP while Node.js still holds the handle.

# Suggested fix

Destroy the response stream before resolving, or use `stream/promises.pipeline`, which handles cleanup automatically.

Example using `pipeline`:

~~~js
const { pipeline } = require('stream/promises');

async function downloadFile(url, dest) {
  // ... redirect handling with response.destroy() before recursive call ...
  await pipeline(
    require('https').get(url),
    require('fs').createWriteStream(dest)
  );
}
~~~

Minimal fix:

~~~js
file.on('finish', () => {
  file.close((err) => {
    if (err) reject(err);
    else {
      response.destroy(); // release the HTTP stream handle
      resolve();
    }
  });
});
~~~

# Environment

- OS: Windows (`win32`)
- npm version: `11.x`
- `@gastown/gt` version: `0.12.0`
- Node.js: any version tested so far (confirmed on `25.x`)

# Additional context

Installing the same package on macOS and Linux works correctly because the file-locking issue does not manifest there.

steveyegge:
## Summary
Closes the HTTPS response stream in `postinstall.js` `downloadFile()` before rename, fixing the file-handle leak that prevented zip extraction on Windows.

## Scope
Partial fix for hq-j6hur.6 (install/build breakage). Covers:
- [x] gh#3595 — npm postinstall file-handle leak (Windows)
- [ ] gh#3451 — brew HEAD gastown on Linux (follow-up)
- [ ] gh#3502 — Docker build (follow-up)

## Context
Dispatched by mayor as part of release-gate **hq-j6hur**. Polecat: gastown/slit.

Opened by mayor because polecat pushed branch but did not open the PR (matches gh#3603, tracked in hq-j6hur.3).

Fixes #3595

### DIFF
--- a/npm-package/scripts/postinstall.js
+++ b/npm-package/scripts/postinstall.js
@@ -83,8 +83,15 @@ function downloadFile(url, dest) {
 
       file.on('finish', () => {
         file.close((err) => {
-          if (err) reject(err);
-          else resolve();
+          if (err) {
+            reject(err);
+            return;
+          }
+          // On Windows, the HTTPS response stream still holds a file handle
+          // lock after the write stream closes. If we don't destroy it here,
+          // PowerShell's Expand-Archive can't open the ZIP for extraction.
+          response.destroy();
+          resolve();
         });
       });
     });


outdoorsea:
**Context for reviewers** (added by GT refinery — outdoorsea lacks merge permissions)

## What this fixes
`npm install -g @gastown/gt` fails on Windows during postinstall ZIP extraction:
> _"The process cannot access the file because it is being used by another process."_

## Root cause
`downloadFile()` pipes the HTTPS response into a write stream and resolves after `file.close()`, but never calls `response.destroy()`. The readable stream keeps an open file handle until GC. Windows uses **mandatory** file locking, so PowerShell's `Expand-Archive` can't open the ZIP. macOS/Linux use advisory locking and are unaffected.

## Fix
One line: `response.destroy()` before `resolve()` in the `file.on('finish')` callback. The redirect path was already correct and is unchanged.

## Test
- Windows: `npm install -g @gastown/gt` → ZIP extraction succeeds ✓
- macOS/Linux: no regression (`destroy()` is a no-op after a stream has ended) ✓

Closes #3595

Bella-Giraffety:
Thanks for carrying this fix, @steveyegge. The underlying Windows postinstall file-handle fix is already present on current , so this original PR is now obsolete rather than still needing review. Closing it to keep the queue tidy, with credit to the original investigation and implementation.

PR #3630, opened 2026-04-13
fix: preserve canonical rig beads identity on adopt

Bella-Giraffety:
## Summary
- preserve canonical rig beads identity during rig adopt instead of leaving adopt paths pointed at orphaned beads databases
- add adopt-path regression coverage for beads candidate detection, fallback initialization, and bound-env finalization
- keep adopt finalization aligned with shared Dolt/beads bindings used by the non-adopt path

### DIFF
--- a/internal/cmd/beads_db_init_test.go
+++ b/internal/cmd/beads_db_init_test.go
@@ -304,9 +304,9 @@ func TestBeadsDbInitAfterClone(t *testing.T) {
 		}
 	})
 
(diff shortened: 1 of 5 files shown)


Bella-Giraffety:
Closing as superseded. I re-triaged this against current upstream/main and the old CI failures were stale drift in the adopt/beads path, not a still-rebaseable branch-only issue. The fix has been ported to a fresh current-main branch at Bella-Giraffety:gastown polecat/chrome-pr3630 (commit caa064df) for maintainer merge-queue landing instead of reviving this stale contributor branch.

PR #3632, opened 2026-04-13
fix: track push-failed cleanup and slot reopening

Bella-Giraffety:
Summary:
- create or reuse cleanup wisps for push-failed polecat completions instead of returning before recovery tracking
- mark push-failed cleanup wisps with an explicit state and still notify Mayor that the polecat slot is open
- cover both direct witness completion handling and bead-based completion discovery with targeted tests


Bella-Giraffety:
Thanks for the work here. This branch is now too stale and mixed to be the right path forward; the queue has moved on and the remaining recovery behavior should be re-cut on top of current main if we still need it. Closing it to keep the queue tidy rather than carrying an old branch with a red CI footprint.

PR #3642, opened 2026-04-14
fix(beads): route rig-targeted Create via cwd + BEADS_DIR (gh#3641)

### LINKED ISSUE #3641: gt done: MR bead creation fails with 'no database selected' — sibling of #3518
## Symptom

`gt done` (and `gt done --pre-verified`) fails when creating the merge-request bead with embedded-dolt error:

```
embeddeddolt init schema error: no database selected
```

Branch is pushed to origin successfully; only the MR bead creation fails. Main Dolt server is healthy (0s latency). Reproducible on v1.0.0-74-g677877bf across multiple polecats/rigs.

## Root cause

Same class of bug as #3518, different call site.

#3518 fixed `gt rig add`'s DDL path by switching `serverExecSQL`/`doltSQL` from `buildDoltSQLCmd` (embedded mode — writes to disk but server never learns) to `buildServerSQLCmd` (explicit `--host`/`--port` to the live sql-server).

The MR-bead-creation DDL path (invoked by `gt done`) still uses the old embedded helper. Schema init writes to disk; subsequent inserts via MySQL protocol on port 3307 fail because the server has no database selected for that connection.

`internal/doltserver/doltserver.go` still has multiple `buildDoltSQLCmd` callers; at least one is on the MR-bead creation path.

## Reproduction

1. Create a polecat: `gt sling <bead-id> <rig> --create`
2. Polecat completes work and calls `gt done --pre-verified`
3. Branch pushes OK; MR bead creation fails

## Build

`gt v1.0.0-74-g677877bf-dirty` — ancestors include #3518 (d792ec5e).

## Impact

All merge-queue automation broken: polecats complete work and push branches, but refinery never sees an MR bead to pick up. Every completion requires manual PR creation.

Workaround: Mayor opens PRs manually via `gh pr create`.

## Suggested fix

Apply the #3518 treatment to the remaining DDL call sites in `internal/doltserver/doltserver.go` — particularly the MR-bead schema init path. The five untouched `buildDoltSQLCmd` sites at lines 1964, 2106, 2835, 3549, 4024 should be reviewed; any that run DDL (CREATE/ALTER) need to switch to `buildServerSQLCmd`.

Will follow up with a PR if no one else picks it up in the next day or two.

---

## Also: bug is self-blocking

Confirming the scope of the regression: I cannot create a tracking bead for this fix in our local gastown rig because `bd create` itself hits the same error:

```
Error: failed to open target store: embeddeddolt: init schema: creating schema_migrations table: Error 1105: no database selected
```

This affects any rig where the database has not been pre-registered with the running sql-server (e.g. freshly-cloned rigs, or rigs where the embedded DDL created the DB on disk but never registered it). It is the same failure mode #3518 fixed for `gt rig add`, manifesting in every other DDL path in `internal/doltserver/doltserver.go`.

work2fly:
## Summary

Fixes gastownhall/gastown#3641.

PR #3540 (37c2b382) routed MR beads to the rig's DB via `--rig`/`--repo`
on `bd create`. This works for rigs with flat `.beads/metadata.json` but
fails for rigs using `.beads/redirect` (→ `mayor/rig/.beads/`): bd's
`--repo` flag does **not** follow the redirect, so it falls back to
embedded-dolt mode and fails schema init with:

```
failed to open target store: embeddeddolt: init schema:
creating schema_migrations table: Error 1105: no database selected
```

This broke `gt done`'s MR bead creation across every redirect-using rig,
taking the whole merge-queue automation offline.

## Fix

When `opts.Rig` is set, run bd via a new `runInRig` helper that:

1. Resolves `<townRoot>/<rig>` → actual beads dir via `ResolveBeadsDir`
   (which *does* follow `.beads/redirect`).
2. Clones the wrapper with `workDir = rigPath` and `beadsDir = resolved`.
3. Invokes `run()` — subprocess `cmd.Dir` + `BEADS_DIR` now both point at
   the rig's real beads directory.

cwd-based discovery already honours redirect, so bd finds the server-
backed store and MR bead creation succeeds. The `--repo` flag is no
longer passed for rig-targeted creates, sidestepping the bd bug entirely.

## Test plan

- [x] `go build ./internal/beads/` passes
- [x] `go test ./internal/beads/` passes (5 tests)
- [x] End-to-end: `b.Create(CreateOptions{Rig: "gastown"})` now succeeds
      against a redirect-style rig (previously failed with "no database
      selected")
- [ ] Verify downstream: `gt done` on a polecat branch creates MR bead
      (will exercise in refinery once deployed)

## Notes

- bd's `--repo` flag still has the underlying redirect-ignoring bug and
  should be fixed upstream in `gastownhall/beads`. This PR works around
  it in gastown rather than waiting for that.
- Filed as gastownhall/gastown#3641 by mayor this morning.

### DIFF
--- a/internal/beads/beads.go
+++ b/internal/beads/beads.go
@@ -482,6 +482,34 @@ func (b *Beads) run(args ...string) (_ []byte, retErr error) {
 	return stripStdoutWarnings(stdout.Bytes()), nil
 }
 
+// runInRig executes a bd command targeted at a specific rig's beads store.
+// This is a workaround for bd's --repo flag, which does not follow
+// .beads/redirect files. Rigs that store metadata.json behind a redirect
+// (e.g. <rig>/.beads/redirect → mayor/rig/.beads) hit "no database selected"
+// errors when bd falls back to embedded-dolt mode. Setting the subprocess
+// cwd to the rig directory triggers cwd-based discovery, which does follow
+// the redirect; we also set BEADS_DIR to the resolved path for belt-and-braces.
+func (b *Beads) runInRig(rigName string, args ...string) ([]byte, error) {
+	townRoot := b.getTownRoot()
+	if townRoot == "" {
+		return nil, fmt.Errorf("runInRig: cannot resolve town root from workDir %q", b.workDir)
+	}
+	rigPath := filepath.Join(townRoot, rigName)
+	resolved := ResolveBeadsDir(rigPath)
+	if resolved == "" {
+		return nil, fmt.Errorf("runInRig: cannot resolve beads dir for rig %q at %s", rigName, rigPath)
+	}
+	// Clone the wrapper with rig-targeted workDir + beadsDir so run() sets
+	// cmd.Dir and BEADS_DIR to the rig's resolved beads directory.
+	rigBeads := &Beads{
+		workDir:    rigPath,
+		beadsDir:   resolved,
+		isolated:   b.isolated,
+		serverPort: b.serverPort,
+	}
+	return rigBeads.run(args...)
+}
+
 // runWithRouting executes a bd command without setting BEADS_DIR, allowing bd's
 // native prefix-based routing via routes.jsonl to resolve cross-prefix beads.
 // This is needed for slot operations that reference beads with different prefixes
@@ -1226,9 +1254,6 @@ func (b *Beads) Create(opts CreateOptions) (*Issue, error) {
 	if opts.Ephemeral {
 		args = append(args, "--ephemeral")
 	}
-	if opts.Rig != "" {
-		args = append(args, "--rig="+opts.Rig)
-	}
 	// Default Actor from BD_ACTOR env var if not specified
 	// Uses getActor() to respect isolated mode (tests)
 	actor := opts.Actor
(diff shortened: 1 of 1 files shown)


popandpeek:
Tested locally with multi-rig setup (beacon + gantry, both using `.beads/redirect` layout) — cherry-picked commit `291ae684` onto gastownhall/main `677877bf`, rebuilt gt, ran end-to-end against active polecats.

**Results:**
- All `TestCreateOptions*`, `TestCreateOptionsRig`, `TestRigFields*`, `TestRigBeadID*` tests pass ✅
- Zero regressions from the cherry-pick
- 1 pre-existing failure (`TestCreateAgentBead_UsesTownRootForCrossRigRoutes`) confirmed present on `gastownhall/main` HEAD without this change — unrelated to `--rig` / `runInRig` logic, separate upstream issue

**Context we're hitting:** prior to this fix, every `gt done` in our multi-rig town fails at `bd create --rig=<rig>` (`unknown flag`), stranding polecat completions with no PR + no MR bead. 6+ recoveries needed today in our pipeline. Each failed `gt done` also leaks Dolt connections (464→939/1000 in 14min at peak) — this fix likely stops that side-effect too.

**Reference:** matches issue #3661. The `runInRig` helper correctly follows `.beads/redirect` for rigs with non-flat layouts — the key reason this is preferable to the simpler `--rig`→`--repo` rename in #3634.

Recommend merging. Happy to test any follow-ups.

mmlac:
Post-landing observation from exercising this change in a real workspace — possible follow-up bug in the `gt done` MR-bead creation path (not in the `bd create` routing this PR fixes, but adjacent to it).

After installing a binary built from this PR, two consecutive `gt done` runs on the same polecat showed inconsistent behavior in the **source-issue inference** step:

- `ws-pb8` on branch `polecat/capable-mo39t3ki` → `gt done` succeeded first try, MR-bead queued cleanly.
- `ws-49r` on branch `polecat/capable-mo3ajdyv` → `gt done` first attempt failed with `cannot determine source issue from branch polecat/capable-mo3ajdyv; use --issue to specify`. Retry with `gt done --issue ws-49r` succeeded (MR bead created via checkpoint resume).

Both branches share the same `polecat/<name>-<random>` format (no bead ID encoded in the branch name), and the hook had the same source-issue information available in both cases. The failure appears intermittent — my guess is a race in checkpoint/convoy state being read during the source-issue inference step, rather than a pure branch-name parser issue.

The branch push itself worked both times; the regression is isolated to source-issue inference during MR-bead creation. Non-blocking (polecat self-recovered on retry), but worth flagging as a follow-up since the same path is what this PR touches.

Happy to open a separate issue if useful.

Bella-Giraffety:
Thanks for carrying this fix. The core cross-rig/server-routing bug in this area has already been addressed by later merged work, so this original branch is now stale rather than the best path to land. Closing it to keep the queue tidy; if there is still an uncovered gap, we should reopen it as a smaller current-main replacement branch instead of reviving this older branch.

PR #3646, opened 2026-04-14
test(daemon): regression test for gh-3622 events.id scan type

### LINKED ISSUE #3622: convoy event poller: events.id scanned as int64, but column is CHAR(36) UUI
## Summary

  The convoy event polling path in `gt daemon` fails on every poll cycle with:

  Convoy: event poll error (): failed to scan event:
  sql: Scan error on column index 0, name "id":
  converting driver.Value type []uint8 ("65a7013c-ac1b-4829-89c7-c251b7652092")
  to a int64: invalid syntax

  Reproduces every ~5 seconds, across each rig database the daemon polls (locally, this is 5).

  ## Environment

  - `gt --version`: 0.12.0
  - Dolt: 1.86.1 (previously 1.84.0 — see "History" below)
  - OS: macOS 24.6.0 (Darwin / arm64)

  ## Root cause

  The `events` table schema has:

  id        char(36) NOT NULL PRIMARY KEY  DEFAULT (uuid())
  issue_id  varchar(255) NOT NULL
  event_type varchar(32) NOT NULL
  ...

  The `id` column stores UUID strings. The compiled daemon's event poller scans
  column index 0 into an `int64` destination (per the Go `sql.Rows.Scan` error
  message), which fails because a UUID string isn't an int64. The scan target
  should be `string` (or `[]byte` / `uuid.UUID` depending on conventions).

  Verified no user table in any rig DB has a `DOUBLE` or `FLOAT` column
  (via `information_schema.columns`), ruling out any stored-data explanation.

  ## Suggested fix

  Wherever the convoy event poller defines the scan destinations for `SELECT ... FROM events ...`, change the first destination from `*int64` to
  `*string`
  (or the equivalent UUID type). Likely a 1–2 line patch.

  Searching the binary's strings suggests the relevant call site is near the
  `failed to get events since %d: %w` error wrapper — that's the function that
  needs updating.

  ## History / how this was discovered

  Before Dolt 1.86, the same query produced:

  Error 1366 (HY000): error: '+Inf' is not a valid value for 'double'

  — a Dolt query-engine bug that returned `+Inf` for some computed internal
  value in `dolt_log`/`dolt_diff` system table queries. That error aborted the
  query pipeline before the Go scan step, masking this client-side bug for
  12+ days (102K daemon-log errors since 2026-03-28).

  Upgrading Dolt 1.84 → 1.86.1 resolved the `+Inf` issue (thanks upstream) and
  revealed the underlying scan mismatch.

  ## Impact

  - Convoy coordination is non-functional across all rigs.
  - Daemon event-driven actions that depend on convoy events (e.g. deacon
    restart on stale heartbeat — hypothesised by `au-zp4` V2 investigation)
    are blocked, contributing to deacon death cascade symptoms.
  - Non-blocking for polecat research work (sling/done flows are separate), but
    blocking for any cross-rig coordination that uses convoy events.

  ## Workarounds tried

  - `bd dolt cleanup` (no effect, no orphan DBs)
  - Dolt 1.84 → 1.86.1 upgrade (fixed +Inf, exposed this bug)

LightOfSeven:
## Summary

- Adds `TestEventPoll_EventIDString_Regression3622` to guard against regression of gh-3622
- The bug: convoy event poller scanned `events.id` (CHAR(36) UUID) into an `int64` destination, failing with `sql: Scan error on column index 0, name "id": converting driver.Value type []uint8 to a int64: invalid syntax`
- The fix was in beads SDK v1.0.0 (`scanEvents` correctly scans `event.ID` into `string`); gastown already depends on v1.0.0
- This test documents the fix and prevents future regression

## Test plan

- [ ] Test compiles cleanly (`go build ./internal/daemon/`)
- [ ] `go vet ./internal/daemon/` passes
- [ ] `TestEventPoll_EventIDString_Regression3622` passes in CI (requires Docker for Dolt container)
- [ ] Existing daemon tests unaffected

Closes #3622

### DIFF
--- a/internal/daemon/convoy_manager_test.go
+++ b/internal/daemon/convoy_manager_test.go
@@ -166,6 +166,53 @@ func TestEventPoll_DetectsCloseEvents(t *testing.T) {
 	}
 }
 
+// TestEventPoll_EventIDString_Regression3622 is a regression test for gh-3622:
+// the convoy event poller scanned events.id (a CHAR(36) UUID column) into an
+// int64 destination, producing:
+//
+//	sql: Scan error on column index 0, name "id":
+//	converting driver.Value type []uint8 to a int64: invalid syntax
+//
+// The fix is in beads SDK v1.0.0: scanEvents scans event.ID into string.
+// This test verifies pollStoresSnapshot returns no error when the store
+// contains events with UUID IDs.
+func TestEventPoll_EventIDString_Regression3622(t *testing.T) {
+	if runtime.GOOS == "windows" {
+		t.Skip("skipping on Windows")
+	}
+	store, cleanup := setupTestStore(t)
+	defer cleanup()
+
+	ctx := context.Background()
+	now := time.Now().UTC()
+	issue := &beadsdk.Issue{
+		ID:        "gt-reg3622",
+		Title:     "Regression 3622 — events.id type check",
+		Status:    beadsdk.StatusOpen,
+		Priority:  2,
+		IssueType: beadsdk.TypeTask,
+		CreatedAt: now,
+		UpdatedAt: now,
+	}
+	if err := store.CreateIssue(ctx, issue, "test"); err != nil {
+		t.Fatalf("CreateIssue: %v", err)
+	}
+
+	townRoot := t.TempDir()
+	var logged []string
+	logger := func(format string, args ...interface{}) {
+		logged = append(logged, fmt.Sprintf(format, args...))
+	}
+
+	m := NewConvoyManager(townRoot, logger, "gt", 10*time.Minute, map[string]beadsdk.Storage{"hq": store}, nil, nil)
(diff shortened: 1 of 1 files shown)


LightOfSeven:
**Requesting CI approval and review.**

This PR adds a regression test for gh-3622 (`events.id` scanned as `int64` when column is `CHAR(36)` UUID). The fix was already shipped in the beads SDK — this test guards against future regression.

Summary of changes:
- `TestEventPoll_EventIDString_Regression3622` in `convoy_manager_test.go` (47 lines)
- Creates an issue in a test store, calls `pollStoresSnapshot`, asserts no error
- Windows-skipped (requires Dolt container for Docker-based tests)

Build and `go vet` pass. The CI workflows are waiting for a maintainer to approve the workflow run. Would appreciate CI approval + review when you get a chance. Thank you!

PR #3662, opened 2026-04-16
fix: honor explicit --target in gt done contamination check

Bella-Giraffety:
## Summary
This fixes a regression where gt done with an explicit target branch still ran contamination validation against the default branch instead of the requested target branch. That blocked valid fork-reconciliation work targeting non-default branches such as upstream-rebuild-main.

## Related Issue
Closes #3650

## Changes
- normalize explicit target input for contamination preflight in gt done
- use the explicit target branch as the contamination base ref instead of hard-coding the default branch
- add focused regression coverage for explicit target contamination behavior
- keep broader target normalization concerns split into follow-up work instead of widening this fix

## Why This PR Exists
The reported bug was narrow: contamination preflight ignored the branch the user explicitly asked gt done to target. This PR fixes that specific behavior with the smallest possible code change. It does not attempt to solve broader target normalization across every downstream gt done path; that is tracked separately so this branch stays bug-sized and reviewable.

## Testing
- [x] Unit tests pass: go test ./internal/cmd -run TestDoneContamination
- [x] Manual validation of the preserved branch diff and targeted lane behavior
- [ ] Full go test ./...

Notes:
- Full-tree tests were not used as the acceptance gate for this PR because they include unrelated existing failures outside this bug scope.

## Checklist
- [x] No breaking changes
- [x] Follow-up work for broader normalization remains separate

### DIFF
--- a/internal/cmd/done.go
+++ b/internal/cmd/done.go
@@ -76,6 +76,15 @@ const (
 	ExitDeferred  = "DEFERRED"
 )
 
+func doneContaminationBaseRef(defaultBranch, explicitTarget string) string {
+	targetBranch := defaultBranch
+	if explicitTarget != "" {
+		targetBranch = strings.TrimPrefix(explicitTarget, "origin/")
+	}
+
+	return "origin/" + targetBranch
+}
+
 func init() {
 	doneCmd.Flags().StringVar(&doneIssue, "issue", "", "Source issue ID (default: parse from branch name)")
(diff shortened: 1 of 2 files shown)


Bella-Giraffety:
Rebased this branch onto current upstream/main after confirming the earlier red Integration run was stale branch-state, not a failure in the done contamination change itself. Fresh CI on the rebased head is now green, including Integration Tests and the main Test workflow.

Issue #3651, opened 2026-04-15
Incorrect install instructions on 1.0 release


justin-abrahms-parloa:
https://github.com/gastownhall/gastown/releases/tag/v1.0.0

That says we should install `npm install -g @anthropic/gastown` which appears to be incorrect.

Bella-Giraffety:
This sounds interesting, but Gastown is in maintenance mode and staying focused on infrastructure and reliability fixes only. If you want to pursue docs/release experience work like this, please check out Gas City instead: https://github.com/gastownhall/gascity

PR #3664, opened 2026-04-16
fix: restore provisioned OpenCode refusal auto-continue template

Bella-Giraffety:
## Summary
This restores the authoritative embedded/provisioned OpenCode `gastown.js` template so worker environments again receive the historical refusal auto-continue behavior. The restored template includes the `message.updated` refusal handler, `Continue` retry loop, and escalation after the retry limit.

## Related Issue
Part of #3651

## Changes
- restore refusal auto-continue logic in the embedded OpenCode template
- add regression coverage that the provisioned template contains the expected refusal-handling fragments
- keep worker upgrade-path and remaining behavioral gaps split into follow-up issues instead of widening this PR

## Why This PR Exists
The provisioning lane was fixed first, but review and worker validation showed the template being provisioned no longer contained the historical refusal auto-continue logic that acceptance depends on. This PR restores that template content as the smallest possible bug-fix slice.

## What This PR Does Not Claim
This PR intentionally does not solve every remaining follow-up in the broader refusal stream. Those are tracked separately so this branch stays narrowly bug-scoped:
- existing worker installs may still need explicit refresh/sync (`gt-9eh`)
- assistant-text fetch failure handling still needs a correctness fix (`gt-1wc`)
- role-scope alignment with historical behavior is still open (`gt-e5f`)
- stronger behavior-level regression coverage is still open (`gt-6p7`)
- direct nested command/doctor follow-ups remain separate (`gt-rku`, `gt-5hx`, `gt-jzl`)

## Testing
- [x] `go test ./internal/hooks ./internal/doctor ./internal/runtime`
- [x] `make build`
- [ ] `go test ./...`

Notes:
- Full-tree tests still include unrelated existing failures outside this bug scope.

## Checklist
- [x] Smallest practical bug-fix slice
- [x] Follow-up gaps kept separate
- [x] No new commands or features introduced

### DIFF
--- a/internal/hooks/installer_test.go
+++ b/internal/hooks/installer_test.go
@@ -178,6 +178,28 @@ func TestSyncForRole_CreatesNewFile(t *testing.T) {
 	}
 }
 
+func TestComputeExpectedTemplate_OpenCodeIncludesRefusalAutoContinue(t *testing.T) {
+	content, err := ComputeExpectedTemplate("opencode", "gastown.js", "polecat")
+	if err != nil {
+		t.Fatalf("ComputeExpectedTemplate: %v", err)
+	}
+
+	got := string(content)
+	checks := []string{
+		`const refusalTexts = new Set([`,
+		`const refusalPrompt = "Continue";`,
+		`const maxConsecutiveRefusalRetries = 4;`,
+		`if (event?.type === "message.updated") {`,
+		`await promptSession(sessionID, refusalPrompt);`,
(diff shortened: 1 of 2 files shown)


Bella-Giraffety:
Closing this draft because the refusal-template lane is mis-scoped for this repository. Keeping it open would create duplicate upstream noise instead of a clean project contribution. If any narrow gastown-owned slice remains after the current review and packaging pass, it should come back as a fresh focused bug-fix PR linked to the correct issue.

PR #3668, opened 2026-04-17
Fix doctor false warning for modular shell configs

### LINKED ISSUE #3667: gt doctor: false 'Shell integration not installed' with modular shell config
## Bug

`gt doctor` reports "Shell integration not installed" even when shell integration IS correctly installed, if the user has a modular shell config where `.zshrc` sources profile-specific files that in turn source the Gas Town hook script.

## Root Cause

`hasShellIntegration()` in `internal/doctor/global_state_check.go` only reads the top-level RC file (`~/.zshrc`) and greps for the marker string `"Gas Town Integration"`. It does not:

1. Follow `source` / `.` directives to check sourced files
2. Detect manual sourcing of `shell-hook.sh` (which may not contain the marker string)

## Reproduction

User has a modular zsh config:

```zsh
# ~/.zshrc
source ~/dotfiles/zsh/home.zsh
```

```zsh
# ~/dotfiles/zsh/home.zsh
[[ -f "$HOME/.config/gastown/shell-hook.sh" ]] && source "$HOME/.config/gastown/shell-hook.sh"
```

Running `gt doctor` produces:

```
⚠ global-state: Shell integration not installed
```

## Expected Behavior

`gt doctor` should detect the integration is installed regardless of sourcing indirection. Possible approaches:

- Recursively follow `source`/`.` directives (one or two levels deep) when scanning RC files
- Check whether `shell-hook.sh` exists and is referenced anywhere in the sourced file tree
- Run a shell subprocess that evaluates the RC file and checks for the hook's side effects

## Affected File

`internal/doctor/global_state_check.go` (both mayor and refinery rig variants)

8888:
## Summary
- Fixes #3667
- `gt doctor` reported "Shell integration not installed" even when integration was correctly sourced via profile-specific shell config files
- `hasShellIntegration()` now recursively follows source/. directives in RC files instead of only checking the top-level RC file
- Also detects manual sourcing of `shell-hook.sh` (not just the `gt shell install` marker)

## Test plan
- [ ] 14 new unit tests covering: direct marker, sourced files, variable expansion, glob fallback, tilde expansion, depth limits, circular references, conditional source patterns
- [ ] Full `go test ./internal/doctor/` suite passes
- [ ] Verified against real dotfiles setup with `.zshrc → profile.zsh → shell-hook.sh` chain

### DIFF
--- a/internal/doctor/global_state_check.go
+++ b/internal/doctor/global_state_check.go
@@ -6,6 +6,7 @@ package doctor
 import (
 	"os"
 	"path/filepath"
+	"sort"
 	"strings"
 
 	"github.com/steveyegge/gastown/internal/shell"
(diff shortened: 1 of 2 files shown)


8888:
Pushed additional tests to cover the reachable gaps. The remaining uncovered lines are OS-level error paths that aren't worth the fragile test setup:

- `filepath.Abs` returning an error (line 123) — only fails if `os.Getwd()` fails, which requires the working directory to be deleted mid-test
- `os.UserHomeDir()` error discard (line 142) — no reasonable way to make this fail in a unit test without OS-level mocking; function still works correctly with empty homeDir
- `filepath.Glob` returning an error or empty matches (line 290) — Glob only errors on malformed patterns which `replaceUnresolvedVars` won't produce; empty matches are the normal no-op fallthrough
- Partial branch on the `if err != nil` check for `filepath.Abs` (line 121) — always evaluates to false in practice since Abs only fails on relative paths when getwd fails

PR #3677, opened 2026-04-18
Fix SetAgentState failure: deterministic agentBeadID across spawn and session

### LINKED ISSUE #3676: SetAgentState fails on every polecat spawn: agentBeadID path mismatch
## Description

`SetAgentState` fails with "issue not found" on every polecat spawn, making polecats invisible to the dashboard (show as "stalled", 0 polecats in control center).

```
Starting session for gastown/rust...
⚠ Warning: SetAgentState attempt 1 failed, retrying in 466ms: issue not found
⚠ Warning: SetAgentState attempt 2 failed, retrying in 897ms: issue not found
⚠ Warning: could not update agent state after retries: setting agent state after 3 attempts: issue not found
```

## Root Cause

`agentBeadID()` in `internal/polecat/manager.go:381` computes the agent bead ID using `workspace.Find(m.rig.Path)`:

```go
func (m *Manager) agentBeadID(name string) string {
    townRoot, err := workspace.Find(m.rig.Path)
    if err != nil || townRoot == "" {
        return beads.PolecatBeadID(m.rig.Name, name)
    }
    prefix := beads.GetPrefixForRig(townRoot, m.rig.Name)
    return beads.PolecatBeadIDWithPrefix(prefix, m.rig.Name, name)
}
```

During **spawn** (`AddWithOptions`), the Manager uses the **main rig path** → `workspace.Find` resolves the town root → computes agent bead ID with the correct prefix → `createAgentBeadWithRetry` succeeds.

During **startPolecatSession** (`polecat_spawn.go:396-398`), a **new Manager** is created from the polecat's worktree rig path:
```go
polecatMgr := polecat.NewManager(r, polecatGit, t)
if err := polecatMgr.SetAgentStateWithRetry(s.PolecatName, "working"); err != nil {
```

If `workspace.Find()` resolves a different town root from the worktree (or fails), `agentBeadID` computes a **different ID** and can't find the bead created during spawn.

## Impact

- Polecats work but are invisible to the Gas Town Control Center dashboard
- `gt polecat list` shows polecats as "stalled" even when actively working
- `gt polecat nuke` reports "agent bead not found or already cleaned"
- Witness sees stalled polecats and escalates unnecessarily

## Reproduction

Every `gt sling <bead> <rig>` that spawns a polecat triggers this. 100% reproducible.

## Suggested Fix

Ensure `agentBeadID` is deterministic regardless of call site. Options:
1. Pass the town root explicitly to the Manager rather than resolving it from rig path
2. Cache the agent bead ID on the polecat config at spawn time so StartSession uses the same ID
3. Have `startPolecatSession` reuse the spawn Manager instead of creating a new one

8888:
## Summary
- Fixes #3676
- `SetAgentState` failed with "issue not found" on every polecat spawn because `agentBeadID()` computed different IDs from the main rig path vs the polecat worktree path
- The Manager now uses a deterministic town root so the agent bead ID is consistent regardless of whether it's called during spawn or session start

## Impact
- Polecats were working but invisible to the Gas Town Control Center dashboard
- `gt polecat list` showed working polecats as "stalled"
- Witness escalated unnecessarily on healthy polecats

## Test plan
- [ ] `gt sling` a bead — verify no `SetAgentState` warnings
- [ ] `gt polecat list` shows polecat as "working" (not "stalled")
- [ ] Dashboard shows polecat count > 0
- [ ] `gt polecat nuke` no longer reports "agent bead not found"

### DIFF
--- a/internal/polecat/manager_test.go
+++ b/internal/polecat/manager_test.go
@@ -381,7 +381,6 @@ func TestNewManager_NamepoolFromRigConfig(t *testing.T) {
 		t.Errorf("expected first name from rig config (alpha), got %q", name)
 	}
 }
-
 // Note: State persistence tests removed - state is now derived from beads assignee field.
 // Integration tests should verify beads-based state management.
 


internal/polecat/manager.go
```diff
@@ -378,14 +373,10 @@ func (m *Manager) assigneeID(name string) string {
 // agentBeadID returns the agent bead ID for a polecat.
 // Format: "<prefix>-<rig>-polecat-<name>" (e.g., "gt-gastown-polecat-Toast", "bd-beads-polecat-obsidian")
 // The prefix is looked up from routes.jsonl to support rigs with custom prefixes.
+// Uses the town root computed at Manager construction for deterministic IDs
```
utafrali, line 387: No regression test was added for the core bug. The failure mode was non-determinism across two Manager instances computing different IDs from the same rig path, and that class of bug is easy to silently reintroduce during a future refactor of `NewManager`. A simple test that constructs a Manager from a known path and asserts `agentBeadID("somecat")` returns the same string on repeated calls, regardless of process working directory, would lock this in. Without it, the fix relies entirely on code review to catch regressions.

internal/polecat/manager.go
```diff
@@ -290,12 +290,7 @@ func (m *Manager) CheckDoltHealth() error {
 // Fails closed if the check errors — a server that can't report capacity is likely
 // already under stress (gt-lfc0d).
 func (m *Manager) CheckDoltServerCapacity() error {
-	townRoot, err := workspace.Find(m.rig.Path)
-	if err != nil || townRoot == "" {
-		return nil // Can't determine town root, skip check
-	}
-
-	hasCapacity, active, err := doltserver.HasConnectionCapacity(townRoot)
+	hasCapacity, active, err := doltserver.HasConnectionCapacity(m.townRoot)
```
utafrali, line 304: This is a subtle behavior change worth calling out explicitly. The old code had `return nil // Can't determine town root, skip check` when `workspace.Find` failed, meaning a broken workspace lookup silently made every capacity check pass. The new code always calls `HasConnectionCapacity(m.townRoot)`, so if `HasConnectionCapacity` returns an error for a valid-but-under-load server, it now propagates that error where it previously would have been swallowed. This is almost certainly the right behavior (and was probably a latent bug in the old path), but it changes the failure mode from silent-pass to fail-closed. Worth a brief comment here noting that the old skip-on-error behavior was intentionally removed.

utafrali (approved):
Correct fix with clear reasoning. Replacing repeated `workspace.Find()` calls with a single `filepath.Dir(r.Path)` at construction time is the right approach given the rig path invariant, and the code is meaningfully simpler as a result. The two gaps are a missing regression test for a non-obvious non-determinism bug, and an undocumented behavior change in the capacity-check error path.

internal/polecat/manager.go
8888: Added `TestAgentBeadID_Deterministic` in commit 0d5634b — it constructs two Managers from the same rig path (including after changing the process working directory) and asserts `agentBeadID("somecat")` returns identical strings across both. This locks in the fix against future refactors of `NewManager`.

internal/polecat/manager.go
8888: Added an explicit comment at `CheckDoltServerCapacity` documenting this intentional behavior change. The old path silently returned nil when `workspace.Find` failed, making every capacity check pass. The new code propagates errors as fail-closed, which is the correct behavior — better to block a spawn than to overload Dolt. Comment calls this out directly so future readers understand it was deliberate.

popandpeek:
The Integration Tests failure on this PR is a pre-existing issue with `TestSchedulerAutoConvoyCreation` — it's been failing on main CI since Apr 15 ([run 24477479864](https://github.com/gastownhall/gastown/actions/runs/24477479864)).

Root cause: the test uses `bd show --json` which queries the `crystallizes` column (dropped in bd v0.63.3, but CI pins bd v0.57.0). The fix is in PR #3689 — replaces `bd show` with a direct SQL query.

Once #3689 merges, rebasing this PR should give you green CI.

8888:
Closing as redundant. The fix for #3676 already landed on `main` via two direct commits:

- 5602a89f — `fix: use deterministic townRoot in Manager for consistent agentBeadID (gt-lph)` — the actual fix (computes `townRoot` once at Manager construction, stored on the struct, used by `agentBeadID` for deterministic IDs regardless of call site)
- 0d5634b5 — `fix: add agentBeadID determinism test and capacity check comment (gt-27k)` — `TestAgentBeadID_Deterministic` regression test + capacity-check comment

After rebasing this branch onto `main`, the substantive changes collapsed (content is already present) and only a stray blank-line deletion remained, so there's nothing left for this PR to contribute.

Issue #23, opened 2026-01-02
Validate rig names to avoid hyphens and special characters


financialvice:
## Problem

When users create rigs with hyphenated names (e.g., `gt rig add my-project ...`), the subsequent agent bead creation fails because beads' agent ID validation incorrectly parses the hyphenated rig name.

See: https://github.com/steveyegge/beads/issues/854

## Example

```bash
gt rig add op-baby https://github.com/user/repo.git
# Later fails when creating agent beads:
# Error: invalid agent ID: invalid agent format: expected <prefix>-<rig>-<role>-<name> 
# where role is one of: crew, polecat (got "ob-op-baby-witness")
```

## Suggested Fix

Add validation in `gt rig add` to either:

1. **Reject** rig names containing hyphens (or other problematic characters) with a clear error message
2. **Warn** users and suggest an alternative name
3. **Auto-sanitize** the name (e.g., `op-baby` → `opbaby`) with user confirmation

The validation should happen early in `runRigAdd()` before any directories are created.

## Proposed Implementation

```go
// In internal/cmd/rig.go or internal/rig/manager.go
func validateRigName(name string) error {
    if strings.Contains(name, "-") {
        return fmt.Errorf("rig name %q contains hyphens which cause agent bead ID parsing issues; use %q instead", 
            name, strings.ReplaceAll(name, "-", ""))
    }
    // Could also check for other problematic chars: spaces, dots, etc.
    return nil
}
```

## Workaround

Users should avoid hyphens in rig names until this is fixed:
- Use `opbaby` instead of `op-baby`
- Use `myproject` instead of `my-project`

## Related

- https://github.com/steveyegge/beads/issues/854 - Agent bead ID validation fails for rig names containing hyphens

steveyegge:
Fixed in 1f72285 - rig names with hyphens, dots, or spaces are now rejected with a suggested sanitized alternative.

PR #3691, opened 2026-04-19
fix(env): remove ANTHROPIC_BASE_URL from parent process passthrough

popandpeek:
## Summary

`AgentEnv()` forwarded `ANTHROPIC_BASE_URL` from `os.Getenv()` into every spawned agent. When a non-Anthropic agent (e.g., MiniMax deacon) spawned Claude polecats, they inherited the MiniMax base URL but used an Anthropic API key — 401 on every API call.

## Root cause

`env.go:359` included `ANTHROPIC_BASE_URL` in the explicit passthrough list for cloud API credentials. This was added in `cbc08f21` (Feb 24) to support cloud provider configuration, but assumed all agents use the same provider as the parent process. That assumption breaks when the daemon runs on MiniMax and spawns polecats on Claude.

## Fix

Remove `ANTHROPIC_BASE_URL` from the passthrough list. Agents that need a custom base URL (MiniMax, Groq, etc.) already receive it from their agent config's `Env` block (`rc.Env` in `settings/config.json` `role_agents`), which is the correct mechanism. The passthrough was redundant for configured agents and harmful for cross-provider spawns.

## Also includes

- **dispatch_test.go**: add `TestDispatchCycle_Run_ZeroCapacity` — verifies `Skipped` count and `capacity` reason when available capacity is zero. Addresses coverage gap from PR #3689.

## Test plan

- [x] `TestAgentEnv_ExcludesAnthropicBaseURL` — verifies `ANTHROPIC_BASE_URL` is not in `AgentEnv()` output even when set in process env
- [x] `TestDispatchCycle_Run_ZeroCapacity` — verifies zero-capacity dispatch report
- [x] All CI checks passed on fork: [popandpeek/gastown PR #23](https://github.com/popandpeek/gastown/pull/23)

### DIFF
--- a/internal/cmd/capacity_dispatch.go
+++ b/internal/cmd/capacity_dispatch.go
@@ -194,6 +194,9 @@ func dispatchScheduledWork(townRoot, actor string, batchOverride int, dryRun boo
 	if report.Dispatched > 0 || report.Failed > 0 {
 		fmt.Printf("\n%s Dispatched %d, failed %d (reason: %s)\n",
 			style.Bold.Render("✓"), report.Dispatched, report.Failed, report.Reason)
+	} else if report.Skipped > 0 {
+		fmt.Printf("\n%s Skipped %d bead(s) — zero capacity (working: %d)\n",
+			style.Dim.Render("○"), report.Skipped, countWorkingPolecats())
 	}
 
 	return report.Dispatched, nil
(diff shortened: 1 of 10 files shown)


asblab (approved):

Bella-Giraffety:
Thanks for the original investigation and fix here, @popandpeek. The specific  passthrough bug-fix from this branch is already present on current , while the rest of the branch is a broader mixed bundle. Closing this original PR to keep the queue tidy rather than carrying the stale combined branch forward.

PR #3695, opened 2026-04-20
fix(daemon): dispatcher skips idle dogs with leaked tmux sessions

ckumar1:
## Summary

`findDispatchableDog` now skips idle dogs whose tmux sessions are still live, preventing the dispatcher from infinite-looping on a "session already running" error when an agent has exited but its tmux shell is leaked.

## Related issue

Internal tracking: `gt-o24`. Reproducible in fleets where an agent crashes inside tmux but the session persists — subsequent dispatch attempts would repeatedly try to start a new agent in that session, hit "session already running", return failure, and the daemon would retry the same dog forever.

## Changes

- `internal/daemon/handler.go`: in `findDispatchableDog`, check `tmux has-session` for each idle dog and advance past any whose session is live. Prior logic considered only the agent's claim state.
- `internal/daemon/handler_test.go`: adds tests for the live-session-skip branch, including mixed idle-with-session / idle-without-session / busy cases.

## Testing

- [x] `go test ./internal/daemon/...` — green, including the new cases.
- [x] `go test ./...` — green locally.
- [x] Manual: reproduced the infinite dispatch loop by `tmux new -d -s <name>` + killing the inner agent; before the patch, `gt daemon` wedged retrying; after the patch it advances to the next dog.

## Checklist

- [x] Code follows project style.
- [x] Tests added for the new skip behavior.
- [x] No breaking changes — existing single-dog paths behave as before when no tmux leak is present.

## Notes

Filing upstream because the loop is not fork-specific; any Gas Town deployment can hit a leaked tmux session the same way. Originated on our fork's `carry/operational`.

### DIFF
--- a/internal/daemon/handler.go
+++ b/internal/daemon/handler.go
@@ -2,6 +2,7 @@ package daemon
 
 import (
 	"fmt"
+	"log"
 	"path/filepath"
 	"time"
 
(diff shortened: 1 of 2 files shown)


asblab (changes requested):

PR #3704, opened 2026-04-20
fix: gt feed TUI tails appended events live

wdawson:
## Summary

Fix `gt feed` TUI event stream silently stops updating after initial render. A reused `bufio.Scanner` in `(*GtEventsSource).tail` latches at EOF and never re-reads the file on subsequent poll ticks, so the event panel shows only the startup backlog from `loadRecentEvents` — new events never appear.

## Related Issue

No existing upstream issue. Discovered while observing a live Gas Town workspace; `gt feed --plain --follow` tailed correctly from the same `.events.jsonl` while the TUI stayed frozen.

## Changes

- `internal/tui/feed/events.go` — move `bufio.NewScanner(s.file)` inside the ticker loop (fresh scanner per tick). `*os.File` preserves offset across scanner instances, so no manual bookkeeping is needed. Mirror the pattern already in `internal/tui/feed/print_events.go:92-123` (which calls out this exact pitfall in its comment — that's why `--plain --follow` works and the TUI doesn't).
- `internal/tui/feed/events.go` — add `<-ctx.Done()` branch inside the channel-send select to avoid a goroutine leak on shutdown when the buffered channel is full. Minor hardening, not the bug fix itself.
- `internal/tui/feed/events.go` — match `loadRecentEvents`'s 1 MiB scanner buffer for consistency on long JSONL lines; comment at the scanner site explains both the reuse pitfall and the buffer size.
- `internal/tui/feed/events_test.go` — new regression test that appends a JSONL event after the source is tailing and asserts it arrives on `Events()`.

## Testing

**Regression guard verified:** stashed the fix, new test fails with `timed out waiting for appended event; tail goroutine stopped reading after EOF`. Restored fix, test passes.

**E2E verified in real TUI:** built fixed + unfixed binaries from this tree, drove `gt feed` in a Python `pexpect` pty (140x40), appended a unique JSONL event mid-run, captured rendered output.

| binary | footer tick | live event visible? |
|---|---|---|
| fixed | `[tree] 2 events` → `[tree] 3 events` | ✅ yes |
| unfixed | `[tree] 2 events` (stuck) | ❌ no — reproduces the reported symptom |

Deterministic across two reruns each.

## Checklist

- [x] Code follows project style (`gofmt`, `go vet` clean)
- [x] Documentation updated (if applicable) — N/A, internal behavior fix

## Notes

Two commits on the branch; happy to squash on merge.

### DIFF
--- a/internal/tui/feed/events.go
+++ b/internal/tui/feed/events.go
@@ -264,10 +264,6 @@ func (s *GtEventsSource) tail(ctx context.Context) {
 	// regardless of the preload scanner's internal read-ahead buffer.
 	_, _ = s.file.Seek(0, 2)
 
-	// Poll for new lines using a fresh scanner each tick.
-	// bufio.Scanner sets an internal 'done' flag after EOF and won't retry,
-	// so we must create a new scanner each poll cycle while preserving the
-	// file offset (os.File tracks position across scanner instances).
 	ticker := time.NewTicker(100 * time.Millisecond)
 	defer ticker.Stop()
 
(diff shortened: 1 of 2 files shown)


Bella-Giraffety:
Closing this for the current infrastructure-only intake. This is primarily feed/TUI behavior expansion rather than a narrow infrastructure fix. If you still want to pursue it, feel free to reopen the work later when we broaden scope.

PR #3714, opened 2026-04-21
fix(config): resolve agent process names through wrapper commands

certivpaul:
## Summary

When an agent's `cmd` is a wrapper script (e.g., a shell script invoking `claude` with extra flags), `ResolveProcessNames` in `internal/config/agents.go` only looked at the literal first token. This meant tmux/process scans matched the wrapper script name rather than the underlying long-running process — breaking session detection for any user running Claude / Codex / etc. behind a launcher.

## Changes

- `internal/config/agents.go`: when an agent's `cmd` is detected as a wrapper (script, alias, or non-executable in PATH), peek inside it to discover the actual binary name(s) the process will appear as in `ps`/`tmux` output. Falls back to the literal `cmd` token when the wrapper can't be introspected.
- All call sites updated to thread the agent's `Args` through `ResolveProcessNames` so the wrapper invocation can be reconstructed accurately.
- Unit tests covering: literal binary, shell wrapper invoking `claude`, wrapper with `Args`, and unresolvable cases.

## Why

This is the root cause of the long-standing `gt status` mis-detection where a polecat session shows `[env]` instead of `[claude]` because the wrapper was launched but `ps` shows the underlying `claude` binary. ZFC-friendly: pure transport / introspection, no agent cognition added.

## Testing
- [x] Unit tests pass (`go test ./internal/config/...`)
- [x] Build clean (`go build ./cmd/gt`)
- [x] Verified against live town with multiple wrapped-agent configs

## Checklist
- [x] No breaking changes (additive — falls back to existing behavior when wrappers can't be peeked)

### DIFF
--- a/internal/cmd/handoff.go
+++ b/internal/cmd/handoff.go
@@ -1018,7 +1018,7 @@ func updateSessionEnvForHandoff(t *tmux.Tmux, sessionName, agentOverride string)
 			}
 			rc, _, err := config.ResolveAgentConfigWithOverride(townRoot, rigPath, currentAgent)
 			if err == nil {
-				resolved := config.ResolveProcessNames(currentAgent, rc.Command)
+				resolved := config.ResolveProcessNames(currentAgent, rc.Command, rc.Args...)
 				processNames = strings.Join(resolved, ",")
 			}
 		}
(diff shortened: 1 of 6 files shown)


Bella-Giraffety:
Thanks for the original investigation and fix here, @certivpaul. We re-cut this onto current main as a cleaner replacement PR with explicit credit: #3808. Closing this original PR as superseded by the refreshed branch so review and landing can happen on the conflict-free current-main diff.

PR #3716, opened 2026-04-21
fix(beads/mail): route cross-rig bead ID writes via prefix

certivpaul:
## Summary

`gt mail`, `gt escalate ack/close`, and refinery delivery acks all wrote to the **caller's** beads store regardless of which rig owned the target bead. When mayor (town-level) handed work to a rig polecat — or a polecat in rig A acked a message originating from rig B — the writes landed in the wrong database. The bead would appear "updated" in the source store but the canonical store (where the bead actually lives) saw nothing.

This caused two visible bugs:
- **au-ofe**: `gt escalate` ack/close + Update path didn't use ID-based routing.
- **au-b9d**: wisp/handoff beads became unqueryable from mail inboxes — the metadata writes were silently going to the wrong DB.

## Approach

Bead IDs are already prefixed by rig (e.g. `aa-`, `au-`, `si-`, `ma-`). We use that prefix as the routing key — same scheme upstream PR #3680 introduced for `--repo=<rig_dir>` lookup via `routes.jsonl`. This PR extends the same prefix→rig resolver to all mail / escalation / delivery write paths.

## Changes

- `internal/beads/beads.go` + `routes.go`: add prefix-based write routing helpers and apply them to channel, delegation, escalation, and sling-context paths.
- `internal/beads/beads_types.go`: thread routing context through the cross-rig call paths.
- `internal/mail/mailbox.go` + `bd.go`: route mail metadata writes via the recipient bead's prefix, not the caller's CWD.
- `internal/mail/delivery.go`: route refinery delivery acks the same way.

## Testing

- [x] Unit tests pass (`go test ./internal/beads/... ./internal/mail/...`)
- [x] Build clean (`go build ./cmd/gt`)
- [x] Verified live: `gt escalate ack` from mayor on a `au-*` bead now lands in `alleago_ui/.beads/`, not town-level.

## Checklist
- [x] No breaking changes — falls back to caller-CWD when the prefix doesn't resolve to a known rig (legacy behavior)

Fixes: au-ofe, au-b9d (Alleago internal tracker)

### DIFF
--- a/internal/beads/beads.go
+++ b/internal/beads/beads.go
@@ -518,6 +518,42 @@ func (b *Beads) runWithRouting(args ...string) (_ []byte, retErr error) { //noli
 	return stripStdoutWarnings(stdout.Bytes()), nil
 }
 
+// routesCrossRig reports whether id's prefix routes to a different beads dir
+// than b's own (per routes.jsonl). Returns false if no routing is needed or
+// the route can't be resolved (in which case the caller should use b.run).
+func (b *Beads) routesCrossRig(id string) bool {
+	own := b.getResolvedBeadsDir()
+	target := ResolveRoutingTargetQuiet(b.getTownRoot(), id, own)
+	return target != "" && target != own
+}
+
+// runBDForID executes a bd command that operates on a single bead id, choosing
+// between run (explicit BEADS_DIR for same-rig) and runWithRouting (strips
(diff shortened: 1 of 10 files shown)


Bella-Giraffety:
Thanks for the original investigation and fix here, @certivpaul. We re-cut this onto current main as a narrower replacement PR with explicit credit: #3796. Closing this original PR as superseded by the refreshed branch so review can happen on the cleaner current-main diff.

PR #3718, opened 2026-04-21
refactor: remove dead heartbeat:EPOCH label write on agent beads

ckumar1:
## Summary

Removes `updateAgentHeartbeat` and its three call sites. The function stamped a `heartbeat:<epoch>` label on agent beads whenever `gt mol step await-signal`/`await-event` observed a signal or timed out, intended to signal agent liveness to watchers.

However, nothing in the codebase reads that label. Agent liveness is determined from the file-backed heartbeats at:
- `deacon/heartbeat.json` (daemon liveness monitor, checked by `mol-gastown-boot.formula.toml`)
- `.runtime/heartbeats/<session>.json` (polecat session staleness, `stuck-agent-dog` plugin)

The label was write-only — a vestige of an older approach. Keeping it produces two consequences, both undesirable:

1. **False-positive divergence.** Operators who know the label exists sometimes trust it over the file, producing "stuck heartbeat" alerts that are actually stale-label artifacts rather than real agent stalls.
2. **Unnecessary bd update traffic.** Every await-signal/event cycle performs a read-modify-write bd call purely to rewrite the label.

This change leaves the file-backed heartbeats as the sole source of truth and removes the dead writer.

## Test plan

- [x] `go test ./internal/cmd/ -run 'TestAwait|TestMolecule|TestHeartbeat'` — pass
- [x] Manual: `gt mol step await-signal --agent-bead <bead>` still updates idle cycles and backoff window; no heartbeat-label writes observed on the agent bead

## Notes

The pre-existing `TestPersistentPreRunLoadsAgentRegistry` failure in `internal/cmd/root_test.go` (signed-binary check in `persistentPreRun`) is orthogonal to this change and was failing before this PR.

### DIFF
--- a/internal/cmd/molecule_await_event.go
+++ b/internal/cmd/molecule_await_event.go
@@ -204,12 +204,8 @@ func runMoleculeAwaitEvent(cmd *cobra.Command, args []string) error {
 	}
 	result.Elapsed = time.Since(startTime)
 
-	// Update agent bead idle cycles and heartbeat
+	// Update agent bead idle cycles
 	if awaitEventAgentBead != "" && beadsDir != "" {
-		// Always update heartbeat (both event and timeout) so witness doesn't
-		// think we're dead during long idle periods.
-		_ = updateAgentHeartbeat(awaitEventAgentBead, beadsDir)
-
 		if result.Reason == "timeout" {
 			newIdle := idleCycles + 1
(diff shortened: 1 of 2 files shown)


Bella-Giraffety:
Closing this for the current infrastructure-only intake. This is a refactor/dead-code cleanup rather than a concrete infrastructure bug fix. If you want to keep pursuing it, feel free to reopen the work later when we broaden scope.

PR #3722, opened 2026-04-21
fix: preserve built-in opencode preset fields in overrides

Bella-Giraffety:
## Summary
- preserve built-in OpenCode preset fields when settings/agents.json provides a partial override instead of replacing the whole preset
- keep ACP/subcommand resolution on the built-in runtime binary while still allowing wrapper-style town overrides for normal launches
- add regression coverage for inherited preset fields and override resolution behavior

## Validation
- go test ./internal/config


Bella-Giraffety:
Superseded by #3723. This PR was based on the wrong local branch and pulled in a large unrelated history stack; the replacement PR contains only the intended opencode preset inheritance fix.

PR #3727, opened 2026-04-22
fix: preserve polecat issue on session restart

vernon99:
## Summary
- Recover a polecat session issue when `gt session start` or `gt session restart` is called without an explicit issue.
- Prefer the active polecat issue, then an issue-scoped polecat branch, then the agent hook bead.
- Pass the recovered issue into session start/restart and log the recovered issue in wake events.

## Why
This is the focused version of the sentifold fix. Restarting a polecat should not silently lose the work item it was already attached to. Keeping the issue identity lets the existing upstream session branch logic make the right choice without introducing a broader branch-planning refactor in this PR.

## Tests
- `go test ./internal/cmd -run TestSelectSessionIssue`

### DIFF
--- a/internal/cmd/session.go
+++ b/internal/cmd/session.go
@@ -10,6 +10,7 @@ import (
 	"time"
 
 	"github.com/spf13/cobra"
+	"github.com/steveyegge/gastown/internal/beads"
 	"github.com/steveyegge/gastown/internal/config"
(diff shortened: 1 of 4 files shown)


vernon99:
Superseded by #3728 with the narrower session-only fix. The original PR included broader session branch-planning changes that are intentionally out of scope for this patch.

PR #3746, opened 2026-04-24
perf(daemon): parallelize compactorGetRowCounts to fix N+1 query pattern

plavering:
Serial loop issued one COUNT(*) per table after listing tables — 41+ round-trips for a 40-table DB. Dispatch all COUNT queries concurrently via goroutine-per-table with sync.WaitGroup + Mutex + error channel. Wall-clock time drops from Σ(query_i) to max(query_i).

information_schema.table_rows is not used because callers compare pre/post counts as a data-loss integrity gate; approximate statistics would either false-alarm or mask real row loss. COUNT(*) is preserved; only the execution order changes.

Adds rows.Err() check on the table-listing cursor (missed previously).

Adds compactor_dog_rowcounts_test.go with a fake database/sql driver covering: empty DB, single/multiple tables, list error, count error, parallel execution timing assertion, exact-count regression guard, and an end-to-end integrity gate simulation.

## Changes

-

## Testing

## Checklist

### DIFF
--- a/internal/daemon/compactor_dog.go
+++ b/internal/daemon/compactor_dog.go
@@ -7,6 +7,7 @@ import (
 	"math"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	_ "github.com/go-sql-driver/mysql"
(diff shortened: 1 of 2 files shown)


Bella-Giraffety:
Closing this for the current infrastructure-only intake. This is a performance optimization rather than a concrete infrastructure correctness fix. If you still want to pursue it, feel free to reopen the work later when we broaden scope.

PR #3747, opened 2026-04-25
feat(doltserver): add GT_DOLT_AUTO_GC env var to control auto garbage collection

jeffplourde:
## Summary

- Add `GT_DOLT_AUTO_GC` environment variable to control Dolt's `auto_gc_behavior.enable` setting in the generated `config.yaml`
- Defaults to `true` (backward-compatible). Set `GT_DOLT_AUTO_GC=false` to disable auto GC
- Follows the existing env var pattern (`GT_DOLT_HOST`, `GT_DOLT_PORT`, etc.) using `strconv.ParseBool`

## Problem

The generated `config.yaml` hardcodes `auto_gc_behavior.enable: true`, and the config is regenerated on every `gt dolt start`, so manual edits don't survive restarts.

In high-throughput environments (batch commit mode with many concurrent agents), Dolt's auto GC thrashes CPU at 70%+ and can contribute to OOM on memory-constrained nodes. There was no way to persistently disable it.

## Changes

**`internal/doltserver/doltserver.go`:**
- Add `AutoGC bool` field to `Config` struct (default `true`)
- Read `GT_DOLT_AUTO_GC` in `DefaultConfig()` via `strconv.ParseBool`
- Use `config.AutoGC` in the YAML template instead of hardcoded `true`

**`internal/doltserver/doltserver_test.go`:**
- `TestDefaultConfig_AutoGCDefault` — verifies default is `true`
- `TestDefaultConfig_AutoGCEnvVar` — verifies `GT_DOLT_AUTO_GC=false` disables it
- `TestDefaultConfig_AutoGCInvalidIgnored` — verifies invalid values are ignored
- `TestWriteServerConfig_AutoGCDisabled` — verifies YAML output contains `enable: false`
- Updated `TestWriteServerConfig_Defaults` to also check `enable: true`

## Usage

```bash
# Disable auto GC (recommended for high-throughput batch workloads)
export GT_DOLT_AUTO_GC=false
gt dolt start

# Verify
grep "enable:" ~/.gt/.dolt-data/config.yaml
# → enable: false
```

## Test plan

- [x] `go test ./internal/doltserver/ -run TestDefaultConfig_AutoGC -v` — all 3 pass
- [x] `go test ./internal/doltserver/ -run TestWriteServerConfig_AutoGC -v` — passes
- [x] `go test ./internal/doltserver/ -run TestWriteServerConfig_Defaults -v` — passes
- [x] Manual: `GT_DOLT_AUTO_GC=false gt dolt start` produces `enable: false` in config.yaml
- [x] Manual: `gt dolt start` (no env var) produces `enable: true` (backward-compatible)

### DIFF
--- a/internal/daemon/dolt.go
+++ b/internal/daemon/dolt.go
@@ -808,8 +808,15 @@ func writeDaemonDoltConfig(cfg *DoltServerConfig, configPath string) error {
 	if cfg.Host != "" {
 		hostLine = fmt.Sprintf("\n  host: %s", cfg.Host)
 	}
+	autoGC := true
+	if agc := os.Getenv("GT_DOLT_AUTO_GC"); agc != "" {
+		if v, err := strconv.ParseBool(agc); err == nil {
+			autoGC = v
+		}
+	}
 	content := fmt.Sprintf(`# Dolt SQL server configuration — managed by Gas Town daemon
 # Do not edit manually; overwritten on each daemon-managed server start.
+# To disable auto GC, set GT_DOLT_AUTO_GC=false
 
 log_level: info
 
@@ -824,12 +831,13 @@ data_dir: %q
 behavior:
   dolt_transaction_commit: false
   auto_gc_behavior:
-    enable: true
+    enable: %v
     archive_level: 1
 `,
 		cfg.Port,
 		hostLine,
 		cfg.DataDir,
+		autoGC,
 	)
 	return os.WriteFile(configPath, []byte(content), 0600)
 }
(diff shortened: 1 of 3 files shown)


Bella-Giraffety:
Closing this for the current infrastructure-only intake. This adds a new  feature/config knob rather than fixing a narrow infrastructure failure. If you want to keep pursuing it, feel free to reopen the work later when we broaden scope.

Bella-Giraffety:
Follow-up note: new config/env-var feature surface belongs in Gas City rather than Gastown. Gastown is staying focused on infrastructure and reliability fixes; if you still want to pursue this, a Gas City branch/PR is the better venue.

jeffplourde:
I respect the close but the fact that dolt routinely consumes enough memory to render its host unusable should qualify as an infrastructure issue. In case it helps anyone else I'm muddling through by applying cgroup constraints on dolt until I can find time to switch multi-agent stacks.

PR #3752, opened 2026-04-25
fix(rig): rig add rollback no longer deletes later successful re-add (gh-3683)

### LINKED ISSUE #3683: 'gt rig add' rollback can delete a successfully-completed re-add of the same rig (data loss)
# Symptom

`gt rig add <name> <url>` has a destructive race in its rollback logic: a stale rollback from a *prior* failed add can delete a *later, successful* add of the same rig name.

Sequence observed (real operator session, 2026-04-18):

1. `gt rig add gastown <url>` started in background (long clone — 2.4 GB).
2. Mid-clone, the parent shell session timed out / lost track of the background process; tool reported the foreground call as "completed" while the underlying process kept running.
3. Operator, seeing partial state on disk and the apparent failure, ran `rm -rf ~/gt/gastown` and successfully retried with `gt rig add gastown <url> --sparse-checkout cmd,internal,docs,scripts,plugins,templates`. The second add **succeeded cleanly** (logged `✓ Rig created in 20.9s`, registered in `mayor/rigs.json`, polecats spawned, work was successfully slung).
4. Several minutes later, the original background `gt rig add` finally exited — and **deleted the entire `~/gt/gastown` directory** as part of its rollback, including all the files the *second* add had created.

Outcome:
- `gt rig list` shows `! gastown` (registered-but-broken marker)
- `~/gt/gastown/` is gone
- Already-spawned polecats (`furiosa`, `nux`) had their working trees nuked under them; tmux sessions remained running but pointing at non-existent dirs
- Convoys (`hq-cv-aa5da`, `hq-cv-qoddu`) were left orphaned with assignees pointing to dead polecats
- Beads slung to those polecats stayed in `HOOKED` status with no live worker

# Root cause (suspected)

`gt rig add`'s failure-rollback path likely runs `rm -rf` on the rig's working directory unconditionally, without verifying it owns the directory it's deleting. That's safe for the single-add case but catastrophic when:

- A second `gt rig add <samename>` has succeeded between the first add's failure and its rollback (the case here)
- A user has manually placed files in that path (less common but possible)

# Fix candidates

**A. Mark and verify ownership.** When `gt rig add` creates a working dir, write a sentinel like `.gt-add-pid-<pid>` inside it. Rollback only deletes if the sentinel matches its own PID. If the sentinel is missing or differs (because a later add overwrote), abort the rollback with a warning.

**B. Atomic create-or-fail.** Refuse to start an add if `gt rig list` already has the entry registered. Currently the second add succeeds even though the first add's process is still running, which is the precondition for the race.

**C. Cancel-on-supersede.** If a new `gt rig add` for the same name runs while a prior one is in flight, kill the prior process before proceeding.

**D. Don't rm-rf in rollback.** Instead, move the partial dir to `~/gt/.failed-adds/<name>-<timestamp>/` so anything that was clobbered can be recovered manually. Trades disk for safety.

(A) is the smallest fix and addresses the specific failure mode.
(B) is more robust but harder (needs cross-process coordination).
(D) is defensive and worth doing alongside A.

# Repro recipe

(Hard to reproduce without the timing window. Approximate steps:)

1. `gt rig add testrig <large-repo-url>` — let it start, then `kill -STOP` the process to simulate a long stall
2. While stopped, `rm -rf ~/gt/testrig` and `gt rig add testrig <url>` again — let the second succeed
3. `kill -CONT` the original process and let it complete its rollback
4. Observe `~/gt/testrig` is gone but `mayor/rigs.json` still has the entry

# Operator-facing impact

- Several polecat sessions had their workspaces deleted under them, leaving zombie tmux processes
- Two beads stayed `HOOKED` in a state that can't be re-slung without `--force`
- Auto-convoy wisps (`hq-wisp-50yvmw`, `hq-wisp-iglwaz`) were created against polecats that no longer existed
- Recovery required manual: `tmux kill-session`, `gt rig remove --force`, `gt convoy close --force`, `bd update --status open --assignee ""`

# Versions

- gt 1.0.0 (Homebrew, `/opt/homebrew/Cellar/gastown/1.0.0`)
- macOS 25.4.0 (Darwin)

# Discovered

Operator session 2026-04-18. While trying to set up a local `gastown` rig to PR fixes for two other bugs (cross-rig convoy auto-close, dolt status truncation), the rig-add cycle itself produced a third, more destructive bug.

# Severity argument

Data loss potential. The rollback overwrites whatever happens to be at the path, including:
- A working rig the user explicitly created
- Polecat workspaces with uncommitted changes (in this session, the polecats were fresh — but in a longer-running scenario they could have had unmerged work)
- Anything else the user happened to put under `~/gt/<rigname>/` (unlikely, but possible)

The bug surface narrows after fix A is in place.

---

*Filed from local `gt/bd` workflow, discovered while attempting to PR fixes for two other gastown bugs.*

mmlac:
Closes #3683.

## Summary

\`gt rig add\` had a destructive race in its rollback logic: a stale background invocation whose parent shell had lost track of it would, on eventual exit, delete the entire rig directory — even when the operator had \`rm -rf\`'d the partial state and successfully re-added the same rig in the meantime.

This stamps each rig directory with a per-invocation random ownership token immediately after \`MkdirAll\`. The deferred rollback now refuses to delete unless the on-disk token still matches the rolling-back invocation's token. When tokens differ (a later successful add overwrote ours), or the token is missing on a non-empty directory (a successful add cleared its own token after registering), the rollback logs a warning and leaves the path intact. Empty directories with no token are still cleaned up — there's nothing to protect.

## How the protected race resolves

The scenario from the issue, step by step:

1. Add #1 starts, writes \`<rig>/.gt-add-token\` = T1, begins long clone in background.
2. Operator perceives failure, \`rm -rf <rig>\`, retries.
3. Add #2 starts, writes \`<rig>/.gt-add-token\` = T2, completes successfully, clears the token.
4. Add #1 finally exits with \`success=false\`. Its deferred cleanup reads the on-disk token: missing on a non-empty directory → skip with warning. Add #2's files survive.

If add #1 had instead crashed before \`MkdirAll\` returned, no token was ever written and the cleanup proceeds unconditionally — same as today's behavior.

## Test plan

- [x] \`go build ./...\` clean
- [x] \`go test ./internal/rig/\` passes — new \`add_token_test.go\` exercises all five branches (token match, token mismatch, token missing + non-empty, token missing + empty, no token issued)
- [x] \`go test ./... -short\` passes across all packages

### DIFF
--- a/internal/rig/add_token_test.go
+++ b/internal/rig/add_token_test.go
@@ -1,0 +1,96 @@
+package rig
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+)
+
+// TestRemoveRigPathIfOwned_TokenMatches verifies that when the on-disk token
+// matches the expected token, the directory is removed.
+func TestRemoveRigPathIfOwned_TokenMatches(t *testing.T) {
+	rigPath := t.TempDir()
+	if err := os.WriteFile(filepath.Join(rigPath, "some-file"), []byte("x"), 0644); err != nil {
+		t.Fatalf("write file: %v", err)
+	}
+
+	token, err := newAddOwnershipToken()
+	if err != nil {
+		t.Fatalf("token: %v", err)
+	}
+	if err := writeAddOwnershipToken(rigPath, token); err != nil {
+		t.Fatalf("write token: %v", err)
+	}
+
+	removeRigPathIfOwned(rigPath, token)
+
+	if _, err := os.Stat(rigPath); !os.IsNotExist(err) {
+		t.Fatalf("expected rig path to be removed, stat err=%v", err)
+	}
+}
+
+// TestRemoveRigPathIfOwned_TokenMismatch verifies that when the on-disk token
+// belongs to a different invocation, the directory is preserved (gh#3683).
+func TestRemoveRigPathIfOwned_TokenMismatch(t *testing.T) {
+	rigPath := t.TempDir()
+	preserved := filepath.Join(rigPath, "preserve-me")
+	if err := os.WriteFile(preserved, []byte("important"), 0644); err != nil {
+		t.Fatalf("write file: %v", err)
+	}
+
+	// Disk has a token from a *different* (later, successful) add.
+	if err := writeAddOwnershipToken(rigPath, "newer-token"); err != nil {
+		t.Fatalf("write token: %v", err)
+	}
+
+	// Stale rollback runs with its own (older) token.
+	removeRigPathIfOwned(rigPath, "older-token")
+
+	if _, err := os.Stat(preserved); err != nil {
+		t.Fatalf("preserved file was deleted: %v", err)
+	}
+}
+
+// TestRemoveRigPathIfOwned_TokenMissingNonEmpty verifies that a missing token
+// on a non-empty directory is treated as not-owned and skipped — covers the
+// case where a successful re-add has already cleared its token.
+func TestRemoveRigPathIfOwned_TokenMissingNonEmpty(t *testing.T) {
+	rigPath := t.TempDir()
+	preserved := filepath.Join(rigPath, "rig-content")
+	if err := os.WriteFile(preserved, []byte("important"), 0644); err != nil {
+		t.Fatalf("write file: %v", err)
+	}
+
+	removeRigPathIfOwned(rigPath, "stale-token")
+
+	if _, err := os.Stat(preserved); err != nil {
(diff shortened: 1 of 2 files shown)


Bella-Giraffety:
Thanks for the original investigation and fix here, @mmlac. We re-cut this onto current main as a narrower replacement PR with explicit credit: #3797. Closing this original PR as superseded by the refreshed branch so review can happen on the cleaner current-main diff.

PR #3762, opened 2026-04-26
fix(lint): correct 'recognises' misspelling in comments

certivpaul:
The misspell linter flags 'recognises' → 'recognizes' in two comments introduced by recent commits. Pure comment-only fix; no behavior change.

- `internal/polecat/session_manager.go:200`
- `internal/doctor/tmux_socket_check_test.go:32`

This is currently failing CI Lint on main: https://github.com/Alleago/gastown/actions/runs/24940404926


Bella-Giraffety:
Thanks for the cleanup here, @certivpaul. The exact  →  misspelling is already fixed on current , so this PR is now obsolete rather than needed. Closing it to keep the queue tidy, with credit for spotting and carrying the fix.

PR #3764, opened 2026-04-26
fix: add explicit gate=manual skip in dispatchPlugins

RyanMacMillanSoftware:
## Problem

`dispatchPlugins()` had no explicit guard for `gate=manual` plugins. It
only skipped non-cooldown types, but this relied on the cooldown check
being reached first. A misconfigured plugin.md using the wrong TOML
field name (`cooldown = "4h"` instead of `duration = "4h"`) caused
`p.Gate.Duration` to be empty, bypassing the cooldown check entirely and
triggering infinite re-dispatch loops.

Additionally, even with a correctly configured cooldown gate, the daemon
had no explicit semantics for `gate=manual` — the intent was implicit.

## Fix

Add an explicit guard at the top of the plugin loop in `dispatchPlugins()`:

```go
if p.Gate != nil && p.Gate.Type == plugin.GateManual {
    d.logger.Printf("Handler: skipping plugin %s (gate=manual, requires explicit trigger)", p.Name)
    continue
}
```

This makes the `gate=manual` contract explicit and logged, and correctly
short-circuits before any cooldown evaluation.

## Tests

Added `TestDispatchPlugins_SkipsManualGatePlugin` — creates a temp town
root with a `gate=manual` plugin.md, an idle dog, and verifies the dog
remains idle after `dispatchPlugins()` runs.

## Background

Tracked in bead hq-suin. Workaround already applied to `mol-dog-reaper`
plugin.md (`type = "manual"`) pending this daemon fix.

### DIFF
--- a/internal/daemon/handler.go
+++ b/internal/daemon/handler.go
@@ -239,6 +239,12 @@ func (d *Daemon) dispatchPlugins(mgr *dog.Manager, sm *dog.SessionManager, rigsC
 	router := mail.NewRouterWithTownRoot(d.config.TownRoot, d.config.TownRoot)
 
 	for _, p := range plugins {
+		// Never auto-dispatch manual-gate plugins — they require an explicit trigger.
+		if p.Gate != nil && p.Gate.Type == plugin.GateManual {
+			d.logger.Printf("Handler: skipping plugin %s (gate=manual, requires explicit trigger)", p.Name)
+			continue
+		}
+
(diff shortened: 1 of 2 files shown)


Bella-Giraffety:
Thanks for the original investigation and fix here, @RyanMacMillanSoftware. We re-cut this onto current main as a cleaner replacement PR with explicit credit: #3809. Closing the original PR as superseded so review and landing can happen on the conflict-free current-main diff.

PR #2205, opened 2026-02-28
feat(quota): proactive near-limit detection before hard 429

seanbearden:
## Summary

- Extends quota scanner with two-layer near-limit detection: pane warning patterns + Claude usage API
- Adds `gt quota watch` command for continuous monitoring with auto-rotation
- Proactive rotation before the hard 429 — detects approaching-limit signals early

## Problem

The current quota scanner only detects hard rate limits (429) after they've already blocked sessions. By then, agent work is interrupted and context may be lost. We need to detect *approaching* limits and rotate preemptively.

## Approach

**Two detection layers (best-effort, graceful degradation):**

1. **Pane patterns** — new `DefaultNearLimitPatterns` matched against tmux pane content: usage percentages ("85% of your daily usage"), approaching/nearing limit messages, remaining messages/requests counts
2. **Usage API** — queries `claude.ai/api/organizations/{orgId}/usage` for exact utilization % (five-hour and seven-day windows). Inspired by [erkantaylan/claude-multi-account-usage](https://github.com/erkantaylan/claude-multi-account-usage)

Accounts without API credentials silently skip the usage API and rely on pane patterns only. Zero-config degradation.

**Scanner extensions:**
- `Scanner.WithWarningPatterns()` — enables pane-based near-limit detection
- `Scanner.WithUsageChecker()` — enables API-based detection with configurable threshold
- `ScanResult` gains `NearLimit bool` and `Usage *UsageInfo` fields

**Planner extensions:**
- `PlanRotation` now takes `PlanOpts` struct with `IncludeNearLimit` flag
- `RotatePlan` separates `LimitedSessions` from `NearLimitSessions` for clear reporting
- When `IncludeNearLimit` is true, near-limit sessions are included as rotation candidates

**New `gt quota watch` command:**
```
gt quota watch                      # Default 5m interval, 80% threshold
gt quota watch --interval 2m        # Custom interval
gt quota watch --threshold 90       # Rotate at 90% utilization
gt quota watch --dry-run            # Detection only, no rotation
```

## Files changed

| File | Change |
|------|--------|
| `internal/constants/constants.go` | Add `DefaultNearLimitPatterns` and `DefaultUsageThreshold` |
| `internal/config/types.go` | Add optional `OrgID` and `SessionCookie` to `Account` |
| `internal/quota/usage.go` | **New**: HTTP usage client, `ReadOrgID`, `UsageInfo` types |
| `internal/quota/usage_test.go` | **New**: 6 tests for usage client and org ID extraction |
[...]

PR #3765, opened 2026-04-27
feat(quota): login-required signal detection in scanner

Prosoiler:
## Summary

Adds a third signal alongside `RateLimited` and `NearLimit`:
**`LoginRequired`**. When a session is blocked on `/login`, rotation
cannot resolve it — only interactive re-auth can. Surfacing the signal
separately lets watchers fire user-facing alerts (Telegram/email)
rather than attempting silent rotation.

Follows the pattern of #2075 (context-preserving keychain rotation)
and #2205 (proactive near-limit detection): minimal, opt-in additions
to the quota scanner that downstream watchers consume via the existing
JSON output of `gt quota scan`.

## Changes

- `internal/constants/constants.go`: `DefaultLoginRequiredPatterns`
  sibling list (7 patterns: `/login` prompts, "Authentication
  required", log-in prompts, re-auth instructions, session-expired,
  401-with-auth, invalid credentials).
- `internal/quota/scan.go`:
  - `ScanResult.LoginRequired` bool field (JSON tag `login_required`)
  - `Scanner.loginPatterns` field
  - `WithLoginPatterns(patterns []string) error` opt-in method
    matching the existing `WithWarningPatterns` ergonomics
  - Login check runs after hard-limit and near-limit checks (hard
    wins precedence). Opt-in via `WithLoginPatterns` — zero
    behavior change for callers that don't enable it.
- `internal/quota/scan_test.go`: 5 new test functions
  - `TestScanAll_DetectsLoginRequired` — basic detection
  - `TestScanAll_HardLimitTakesPrecedenceOverLogin` — hard-limit
    precedence guarantee
  - `TestScanAll_LoginNotEnabled_NoFalsePositive` — opt-in semantics
  - `TestWithLoginPatterns_InvalidPattern` — error path
  - `TestScanAll_LoginPatternVariations` — table-driven across all
    7 default patterns

## Use case

Gas Town crew daemons want to alert humans when `/login` is required
(rotation can't help) without conflating with rate-limit events. A
forthcoming Gas Town daemon (`gt-director-bottleneck-watch` in the
caller repo) consumes `ScanResult.LoginRequired` alongside its
bd-label-driven escalations.

## Test plan

- [x] `go test ./internal/quota/...` — passes (existing + new tests)
- [x] `go vet ./internal/quota/... ./internal/constants/...` — clean
- [ ] Manual verification on a real session showing `/login` prompt
      (deferred to reviewer; `WithLoginPatterns` is opt-in and tested)

## Notes

- All login-required messages chosen to be specific enough to avoid
  false positives (e.g., `Please (run|use) ['"]?/login['"]?` requires
  the literal `/login` token; `Authentication required` is the exact
  Claude Code prompt).
- Pattern list is conservative — additions can ship in follow-on
  PRs as new false-negative cases surface in production.
- Opt-in design preserves backward compatibility: callers must
  explicitly call `WithLoginPatterns(nil)` (or pass custom patterns)
  to enable detection.

### DIFF
--- a/internal/constants/constants.go
+++ b/internal/constants/constants.go
@@ -442,3 +442,19 @@ var DefaultNearLimitPatterns = []string{
 	`\d+\s*(messages?|requests?)\s*(left|remaining)`,               // "10 messages remaining"
 }
 
+// DefaultLoginRequiredPatterns are patterns that indicate a session needs
+// re-authentication (interactive /login). Distinct from rate-limit signals:
+// here the session is blocked on auth, not quota. Surfacing this as a separate
+// signal lets watchers fire user-facing alerts (e.g., Telegram/email) rather
+// than attempting silent rotation, since /login requires interactive input.
+// Matched with (?i) for case-insensitive matching.
+var DefaultLoginRequiredPatterns = []string{
+	`Please (run|use) ['"]?/login['"]?`,             // explicit /login prompt from Claude Code
+	`Authentication required`,                        // generic auth-required message
+	`Please log in to (Claude|continue)`,            // login prompt variants
+	`Run ['"]?/login['"]? to (re-?authenticate|sign in)`, // /login command instruction
+	`Your session has expired\.\s+Please log in`,    // session-expired prompt
+	`401 Unauthorized.*\b(login|auth(entication)?)\b`, // 401 with login/auth context
+	`Invalid (API key|credentials)`,                  // explicit credential rejection
+}
+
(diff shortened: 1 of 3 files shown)


Bella-Giraffety:
Closing this for the current infrastructure-only intake. This adds quota/login signal detection feature surface rather than fixing a narrow infrastructure failure. If you want to keep pursuing it, feel free to reopen the work later when we broaden scope.

Bella-Giraffety:
Follow-up note: expanded login/quota signal behavior is better directed to Gas City rather than Gastown. Gastown is staying focused on infrastructure and reliability fixes; if you still want to pursue this, a Gas City branch/PR is the better venue.

PR #3775, opened 2026-04-28
fix: use hook show in stuck-agent dog

gregorycarter:
## Summary

Fixes the stuck-agent-dog plugin's hook lookup for remote polecat agents.

The plugin was calling `gt hook <agent-address>`, which now treats the argument as a bead id and exits non-zero. Under `set -euo pipefail`, that caused the plugin to stop immediately before reporting or recovering stuck agents.

## Changes

- Adds `hook_bead_for_agent()` using `gt hook show <agent-address>`.
- Parses the compact hooked-bead output from `gt hook show`.
- Applies the helper in both crashed-session and zombie-session paths.

## Validation

- `bash -n plugins/stuck-agent-dog/run.sh`
- Verified `gt hook show bridge_town_core/polecats/jasper` parses `bt-yubx`.
- Patched and ran the live workspace plugin copy; it completed successfully and notified the Bridge Town witness for crashed jasper/quartz sessions.

### DIFF
--- a/plugins/stuck-agent-dog/run.sh
+++ b/plugins/stuck-agent-dog/run.sh
@@ -11,6 +11,13 @@ RIGS_JSON_PATH="${TOWN_ROOT}/mayor/rigs.json"
 
 log() { echo "[stuck-agent-dog] $*"; }
 
+hook_bead_for_agent() {
+  local agent="$1"
+  local output
+  output=$(gt hook show "$agent" 2>/dev/null || true)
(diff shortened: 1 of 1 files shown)


Bella-Giraffety:
Cleaning up a stale draft under the current infrastructure-only intake. If you still want to pursue this, feel free to reopen the work on a fresher branch and we’ll be happy to review it once it’s ready.

PR #3787, opened 2026-04-29
fix: verify pushed commits before bead closure

akng8:
## Summary

- adds exact remote-tip verification for pushed commits using the push target URL, including split fetch/push remotes
- blocks gt done MR creation, direct source-bead closure, and refinery source-bead closure when the pushed commit is not visible on origin/<target>
- records target_branch and commit_sha in source close reasons and adds --skip-verify audit comments for explicit non-code escapes
- adds regression coverage for success, missing branch, stale remote branch, and split push URL verification

## Verification

- git diff --check
- Not run: gofmt/go test; this host does not have go or gofmt installed

---
<a href="https://app.blacksmith.sh/gastownhall/codesmith/gastown/pr/3787"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/internal/cmd/done.go
+++ b/internal/cmd/done.go
@@ -53,6 +53,7 @@ Examples:
   gt done --target feat/my-branch      # Explicit MR target branch
   gt done --pre-verified --target feat/contract-review  # Pre-verified with explicit target
   gt done --issue gt-abc               # Explicit issue ID
+  gt done --skip-verify                # Audit-only escape hatch for non-code closes
   gt done --status ESCALATED           # Signal blocker, skip MR
   gt done --status DEFERRED            # Pause work, skip MR`,
 	RunE:         runDone,
@@ -67,6 +68,7 @@ var (
 	doneResume        bool
 	donePreVerified   bool
 	doneTarget        string
+	doneSkipVerify    bool
 )
 
 // Valid exit types for gt done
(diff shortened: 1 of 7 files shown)


akng8:
@steveyegge @julianknutsen — friendly ping. Posting context to make a case for review priority.

This PR implements a verified-push contract for bead closure. In one of our heavier nba_pims rig sessions today, we observed the bead-closure path firing CLOSED status without verifying the polecat's commit was actually pushed to origin. A single bead — np-e77 (a `/leaderboard` frontend page) — false-closed **four consecutive times** across both codex and claude polecats, losing all the work each time. Each respawn started fresh because the prior commits were never on any branch.

Supporting evidence: amber's 14-day closed-bead audit (`docs/incidents/closed_bead_audit_2026-04-29.md` in our rig) found a **1.4% systemic false-close rate** across 71 closed beads, but **100% on np-e77's specific code path**. The 5 concrete proposals in this PR are derived directly from that audit + the np-arb bead body.

Until this lands and ships in a Homebrew-pulled gastown release, our np-e77 work stays parked — the 5th respawn is gated on the verified-push contract being live.

Happy to address review feedback / run additional tests if there's blocker concern. Local Go toolchain isn't installed here so `gofmt + go test` were not run pre-push (auto-label is the only check that fired); we'd appreciate CI running on this PR in case there are quick fixes needed.

Thanks for taking a look.

PR #3793, opened 2026-04-30
fix(formula): substitute --set vars in workflow step descriptions

outdoorsea:
## Summary

`gt formula run --set key=value` documents itself as the way to fill `{{problem}}`, `{{context}}`, etc. on workflow-type formulas (e.g. `mol-idea-to-plan`), but `executeWorkflowFormula` was passing `step.Description` through verbatim into both the bead body and `gt sling -a`. Polecats received literal `{{problem}}` placeholders and bailed.

Two underlying issues:

1. The existing `renderTemplate` helper uses Go's `text/template`, which rejects bare `{{name}}` (it requires `{{.name}}`). Formulas use the bare form, so even on the convoy path `renderTemplate` silently fell back to the raw template via the error-fallback branch.
2. The workflow path (`internal/cmd/formula.go:803-805`) explicitly skipped substitution with a comment claiming placeholders were *"instructions for the executing AGENT, not Go template vars"* — but the executing agent is a polecat in a fresh session with no source for the value, so the variable was effectively dead.

## Changes

- New helper `substituteFormulaVars(text, vars)` that fills `{{key}}` placeholders from a parsed `--set` map. Handles whitespace variants (`{{ key }}`). Placeholders without a matching `--set` key (e.g. `{{review_id}}`, computed at runtime by the agent) are left intact, so existing semantics for agent-computed slugs are preserved.
- Apply substitution at both write sites of `executeWorkflowFormula`: bead body creation (`createCmd.Stdin`) and the subsequent `gt sling -a` args.
- Replace the misleading comment.
- Table-driven tests for `substituteFormulaVars` and `parseSetVars`, plus a round-trip integration test that confirms a value containing commas survives `parseSetVars` → `substituteFormulaVars`.

Scope is intentionally narrow: only the workflow path is fixed in this PR. The convoy path uses `renderTemplate` and could benefit from the same helper as a follow-up, but `convoy` is not blocked the way `workflow` is, so I kept the change surgical.

## Reproduction

Before:
```
$ gt formula run mol-idea-to-plan --set "problem=foo"
# polecat closes intake leg with reason:
#   "unfilled template variables ({{problem}}, {{context}}, {{review_id}});
#    no input to draft a PRD from"
```

After: the intake bead body and sling args contain the supplied problem text, and the workflow proceeds normally to the PRD review legs.

## Test plan

- [x] `go test ./internal/cmd/ -run "Formula"` — all formula tests pass (16 subtests across `TestSubstituteFormulaVars` and `TestParseSetVars`)
- [x] `make install` — clean build of `gt`
- [x] End-to-end: `gt formula run mol-idea-to-plan --set "problem=$(cat idea.txt)"` — verified the resulting workflow step bead body contains the substituted problem text instead of `{{problem}}`
- [x] `{{review_id}}` (intentionally not provided via `--set`) remains as a placeholder for the agent to fill

---
<a href="https://app.blacksmith.sh/gastownhall/codesmith/gastown/pr/3793"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/internal/cmd/formula.go
+++ b/internal/cmd/formula.go
@@ -9,6 +9,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"regexp"
 	"strconv"
 	"strings"
 	"text/template"
@@ -797,12 +798,16 @@ func executeWorkflowFormula(f *formula.Formula, formulaName, targetRig string) e
 	// Step 2: Create step beads and wire dependencies
 	stepBeads := make(map[string]string) // step.ID -> bead ID
 
+	// Parse --set key=value pairs so {{var}} placeholders in step descriptions
+	// (e.g., {{problem}}, {{context}}) are filled before the bead is written.
+	// Placeholders for keys not provided via --set (e.g., {{review_id}} which
+	// is computed by the executing agent at runtime) are left intact.
+	setVars := parseSetVars(formulaRunSet)
+
 	for _, step := range f.Steps {
 		stepBeadID := fmt.Sprintf("%s-wfs-%s", rigPrefix, generateFormulaShortID())
 
-		// Step descriptions contain {{var}} placeholders (e.g., {{problem}},
-		// {{context}}) that are instructions for the executing AGENT, not Go
-		// template vars. Do not render them — pass through verbatim.
+		stepDescription := substituteFormulaVars(step.Description, setVars)
 
 		// Use --body-file=- (stdin) for the description to avoid CLI arg
 		// length limits and quoting issues with large markdown descriptions.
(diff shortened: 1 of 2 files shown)


Bella-Giraffety:
Closing this for the current infrastructure-only intake. This changes formula/workflow rendering behavior rather than fixing a narrow infrastructure failure. If you still want to pursue it, feel free to reopen the work later when we broaden scope.

Bella-Giraffety:
Follow-up note: workflow/formula behavior expansion is better directed to Gas City rather than Gastown. Gastown is staying focused on infrastructure and reliability fixes; if you still want to pursue this, a Gas City branch/PR is the better venue.

PR #3814, opened 2026-05-01
fix(mail): clear satisfied reply reminders

Bella-Giraffety:
## Summary

- Clears queued reply reminders when a thread has already been answered successfully
- Avoids repeated `Remember to reply to deacon/ ...` churn after `gt mail send` succeeds
- Keeps the fix focused to mail-router and nudge-queue cleanup only

## Why

The mayor hit repeated reminder loops on the same deacon thread even after successful `gt mail send` replies. That is noisy control-plane behavior and makes real work harder to complete.

## Validation

- targeted mail/nudge tests on the touched paths
- `go build ./cmd/gt`

---
<a href="https://app.blacksmith.sh/gastownhall/codesmith/gastown/pr/3814"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>


Bella-Giraffety:
Thanks for the investigation here. The deacon reminder-loop fix now exists as a clean current-main change in , but this PR branch drifted under stale history and is still dirty in GitHub. I’m closing this original PR in favor of promoting the clean branch separately once we’re ready to land it.

PR #3823, opened 2026-05-01
fix: protect beads global dolt database

rbriski:
## Summary
- Protect beads_global from Dolt orphan detection and direct removal.
- Add regression coverage that keeps a real orphan database detectable while preserving beads_global.

## Validation
- go test ./internal/doltserver
- make build
- ./gt dolt cleanup --dry-run
- ./gt dolt status

Note: make test currently fails on origin/main before this branch because internal/daemon/handler_test.go contains a conflict marker at line 448.

---
<a href="https://app.blacksmith.sh/gastownhall/codesmith/gastown/pr/3823"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/internal/cmd/install_test.go
+++ b/internal/cmd/install_test.go
@@ -23,14 +23,17 @@ func TestBuildBdInitArgs_AlwaysIncludesServerPort(t *testing.T) {
 
 	args := buildBdInitArgs(townDir)
 
-	if len(args) != 6 {
-		t.Fatalf("expected 6 args, got %d: %v", len(args), args)
+	if len(args) != 7 {
+		t.Fatalf("expected 7 args, got %d: %v", len(args), args)
 	}
 	if args[4] != "--server-port" {
 		t.Fatalf("expected args[4] = --server-port, got %q", args[4])
 	}
 	if args[5] != "3307" {
 		t.Fatalf("expected default port 3307, got %q", args[5])
+	}
+	if args[6] != "--force" {
+		t.Fatalf("expected args[6] = --force, got %q", args[6])
 	}
 }
 
--- a/internal/doltserver/doltserver.go
+++ b/internal/doltserver/doltserver.go
@@ -2563,15 +2563,26 @@ type OrphanedDatabase struct {
 	SizeBytes int64
 }
 
+// protectedSharedServerDatabases returns the registry of databases that are
+// intentionally hosted by the shared Dolt server but are not referenced by any
+// rig's metadata. Mapped to a human-readable owner label used by
+// CollectDatabaseOwners so `gt dolt list` / `gt dolt status` annotate them as
+// protected rather than reporting them as orphans.
+//
+// Single source of truth for orphan-detection skipping (FindOrphanedDatabases,
+// RemoveDatabase) and owner-label reporting (CollectDatabaseOwners). Adding a
+// new protected database here automatically picks up all three surfaces.
+func protectedSharedServerDatabases() map[string]string {
+	return map[string]string{
+		"beads_global": "global shared beads database (protected)",
+	}
+}
+
 // isProtectedSharedServerDatabase reports databases that are intentionally
 // hosted by the shared Dolt server but are not referenced by rig metadata.
 func isProtectedSharedServerDatabase(dbName string) bool {
-	switch dbName {
-	case "beads_global":
-		return true
-	default:
-		return false
-	}
+	_, ok := protectedSharedServerDatabases()[dbName]
+	return ok
 }
 
 // FindOrphanedDatabases scans .dolt-data/ for databases that are not referenced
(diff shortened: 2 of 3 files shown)


steveyegge:
Note: #3827 is a duplicate recovery PR for the same gt-09kp from polecat/furiosa. @rbriski could you confirm which branch we should keep and close the other? Both are drafts, so low-stakes — just want to dedupe before either goes ready-for-review.

rbriski:
Confirmed — keeping this one, #3827 is now closed as the duplicate. #3823 has the broader fix (both `FindOrphanedDatabases` skip and the `RemoveDatabase` guard for defense in depth); #3827 had only the orphan-detection skip plus an owner label in `CollectDatabaseOwners`. The owner-label nicety can land as a small follow-up if we want `gt dolt list` to label `beads_global` explicitly.

rbriski:
The two failing CI jobs (`Test`, `Integration Tests`) on this PR are **not caused by this branch** — they're inherited from `main` and reproduce on every PR currently building against `main`.

**`Test` job** — fails on `internal/cmd.TestBuildBdInitArgs_AlwaysIncludesServerPort` (`install_test.go:27`). PR #3829 added `--force` to `buildBdInitArgs` (now 7 args), but the assertion still expects 6.

**`Integration Tests` job** — fails on the `TestBeadsDbInitAfterClone/*` (5 subtests) and `TestAgentWorktreesStayClean/*` (2 subtests). All collide on hardcoded Dolt port 32769, and underneath that have a deeper test-infrastructure issue: subtests share Dolt state and trip bd's "Refusing to destroy N issues" guard once the port conflict is removed.

This PR's diff only touches `internal/doltserver/doltserver.go` (+19/-3) and adds a new test file (+40/0). None of these failing tests exercise the changed code — same failures appear on PRs that don't touch any production code at all (e.g., I tried the test-infra fix in #3843 and confirmed the failures aren't related to either PR's contents).

Author already flagged this in the PR body — surfacing it again for whoever reviews. The substantive change in this PR (protecting `beads_global` from orphan detection) looks landable on its own merits despite the inherited red CI.

kab0rn:
CI fix: `install_test.go` expected 6 args but `buildBdInitArgs` returns 7 (includes `--force`). Restored the test to expect 7 args with `--force` assertion at args[6].

The fix is on the kab0rn fork at `fix/ci-3823-install-test` (commit 5b5a03dde). Cherry-pick to resolve the test failure.

Bella-Giraffety:
Thanks for the P0 data-loss protection work. I reviewed this during PR sheriff and ran local focused validation:

- `go test ./internal/doltserver -run 'TestFindOrphanedDatabases_ProtectsBeadsGlobal|TestRemoveDatabase_RefusesProtectedSharedServerDatabase'` passes.
- `go test ./internal/doltserver` currently fails on `TestCleanStaleSocket_RemovesStaleFile`, which appears unrelated to this PR.

I am holding this rather than merging immediately for two reasons:

- The PR checks are still red. If those are known baseline failures, please rerun/update so maintainers have a current signal, or note an explicit maintainer waiver.
- The patch protects `beads_global` from orphan cleanup/removal, but `gt dolt list` can still show it as `(orphan)` because `CollectDatabaseOwners` does not know about protected shared-server databases. That is operator-confusing for the exact class of cleanup workflows this PR is securing. Please either wire the protected database registry into owner/list reporting, or add a follow-up/maintainer note that this is intentionally left for a separate PR.

The duplicate situation with #3827 looks resolved: #3823 is the branch to keep.

rbriski:
Thanks @Bella-Giraffety — both points addressed in two new commits on this branch.

## CI (`eaf8e6ba`)

Cherry-picked @kab0rn's `5b5a03dde` from `kab0rn/gastown:fix/ci-3823-install-test`. Restores `install_test.go` to assert 7 args (the `--force` PR #3829 added). `Test` job should go green on the next run.

`Integration Tests` will still be red — those failures (`TestBeadsDbInitAfterClone/*` + `TestAgentWorktreesStayClean/*`) are the hardcoded-port + shared-state issue I detailed [in the May 4 comment](https://github.com/gastownhall/gastown/pull/3823#issuecomment-...) — pre-existing on `main`, not exercised by this PR's diff, tracked separately. Happy to take a maintainer waiver on that one or rebase if helpful.

## Owner labeling (`4791a515`)

Extracted the protected-DB list into a `protectedSharedServerDatabases()` registry — single source of truth for both orphan-detection skipping (existing) and owner-label reporting (new). `CollectDatabaseOwners` now labels `beads_global` as `"global shared beads database (protected)"` when it exists on disk, so `gt dolt list` / `gt dolt status` no longer render it as `(orphan)`. The on-disk check is intentional — we don't want to advertise a phantom owner for a DB that doesn't exist.

Two new tests added:
- `TestCollectDatabaseOwners_ProtectedSharedServerDatabaseLabeled` — on-disk `beads_global` gets the protected label.
- `TestCollectDatabaseOwners_ProtectedDatabaseNotPhantom` — absent `beads_global` stays out of the owners map.

Local validation: `go test ./internal/doltserver` clean (including the new tests and your two earlier checks `TestFindOrphanedDatabases_ProtectsBeadsGlobal` + `TestRemoveDatabase_RefusesProtectedSharedServerDatabase`).

This means the P0 data-loss protection AND the operator-visible labeling now share the same registry — adding a future protected DB (e.g. another global table) automatically picks up both surfaces.

Bella-Giraffety (approved):
Reviewed during PR sheriff. The updated branch addresses the previous concerns: beads_global is protected from orphan cleanup/removal and the protected database registry is also used for owner labeling, so operators no longer see the protected DB as an orphan. CI is green.

PR #3824, opened 2026-05-01
fix: remove daemon test conflict marker (gt-nese)

rbriski:
## Summary
- Remove the stray conflict residue from `internal/daemon/handler_test.go`.
- Restore the affected daemon tests so the package can compile again.

## Recovery Context
This branch was recovered after `gt done --target main` could not push to upstream because the authenticated `rbriski` account has read-only access to the upstream repository. The original local commit is `7f903d1d55dea07d8c6f2ebf8f247e0b5930f8e5` on `polecat/furiosa/gt-nese@mon7e2es`.

A verified local backup also exists at `/Users/bbriski/gt/recovery-backups/gt-nese-furiosa-7f903d1d-20260501T175419Z`.

## Validation
- Targeted `go test ./internal/daemon -run 'TestDispatchPlugins_SkipsManualGatePlugin|TestFindDispatchableDog_PicksFirstIdleWhenNoSessionsLive' -count=1 -timeout=30s -v` passed in the recovered work.
- Full `go test ./internal/daemon` was not completed by the polecat because it produced no output for several minutes and appeared to hang in unrelated package coverage.

---
<a href="https://app.blacksmith.sh/gastownhall/codesmith/gastown/pr/3824"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith</code> with what you need.</sup>

- [ ] Let Codesmith autofix CI failures and bot reviews

PR #3843, opened 2026-05-04
fix: bd 1.0 destroy-token migration + integration test port allocation (gt-eh81)

rbriski:
## Summary

Fixes the test failures currently red on `main`:

- `internal/cmd.TestBuildBdInitArgs_AlwaysIncludesServerPort` — stale 6-arg assertion after #3829 added `--force`
- `internal/cmd.TestBeadsDbInitAfterClone/*` (5 subtests) — Dolt port 32769 conflicts
- `internal/cmd.TestAgentWorktreesStayClean/*` (2 subtests) — same port-32769 collision

Reproduced on https://github.com/gastownhall/gastown/actions/runs/25268785313 and on every PR currently building against `main` (e.g., #3823, #3824).

## Why install.go also changes

PR #3829 added `--force` to `buildBdInitArgs` to make `bd init` reliably persist `issue_prefix` on bd >= 1.0. That's correct for the empty-database case, but `--force` is bd's deprecated alias for `--reinit-local` and only authorizes the **local data-safety guard**. When `bd init` runs over a tracked beads repo with existing issues — exactly the scenario `TestBeadsDbInitAfterClone` exercises — bd 1.0 trips a separate **destroy-issues guard**:

> Refusing to destroy 4 issues in non-interactive mode.
>   See 'bd help init-safety' for the required --destroy-token format.

The fix is to pass both flags explicitly: `--reinit-local --destroy-token DESTROY-hq`. This is the same authorization shape bd documents for non-interactive destructive re-init. `gt install` is already a destructive setup operation (see the existing 'use --force to reinitialize' guard at install.go:132), so the stronger token is consistent with intent.

## Test isolation

Each failing integration subtest was sharing hardcoded port 32769; the first subtest claimed it and the rest failed with 'Dolt port already in use'. Two changes:

1. **`getFreePort` helper** (`net.Listen` on `127.0.0.1:0`, read assigned port, close, return). Each subtest now passes a unique `--dolt-port` to `gt install`.
2. **`testutil.StartIsolatedDoltContainer(t)`** per subtest, so each gets its own fresh Dolt database without cross-subtest data leakage. The helper already existed; this PR just calls it from the integration tests that need isolation.

## Validation

- `go test ./internal/cmd/ -run TestBuildBdInitArgs` — 4/4 pass
- `go test -tags=integration ./internal/cmd/ -run 'TestBeadsDbInitAfterClone|TestAgentWorktreesStayClean' -count=2` — passes consecutively (~50s/run, no leaked-port flake)
- `go vet -tags=integration ./internal/cmd/` clean

## Bead

Closes gt-eh81 (port conflicts) and subsumes gt-e95i (test assertion update — the args list changed in this PR, so the gt-e95i fix is rolled in here rather than landed standalone).

## Test plan

- [ ] CI Test job passes
- [ ] CI Integration Tests job passes
- [ ] No regression on existing TestBuildBdInitArgs_RespectsGTDoltPortEnv / ConfigYAMLTakesPrecedence / PortMatchesDefaultConfig

---
<a href="https://app.blacksmith.sh/gastownhall/codesmith/gastown/pr/3843"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/internal/cmd/beads_db_init_test.go
+++ b/internal/cmd/beads_db_init_test.go
@@ -12,12 +12,27 @@ package cmd
 import (
 	"encoding/json"
 	"fmt"
+	"net"
 	"os"
 	"os/exec"
 	"path/filepath"
+	"strconv"
 	"strings"
 	"testing"
 )
+
+// getFreePort returns a free TCP port by binding to 127.0.0.1:0.
+// Used so each subtest's gt install gets a unique --dolt-port and they
+// don't collide on a shared default.
+func getFreePort(t *testing.T) int {
+	t.Helper()
+	ln, err := net.Listen("tcp", "127.0.0.1:0")
+	if err != nil {
+		t.Fatalf("getFreePort: failed to bind: %v", err)
+	}
+	defer ln.Close()
+	return ln.Addr().(*net.TCPAddr).Port
+}
 
 // extractJSON finds the first JSON object in output that may contain non-JSON warnings.
 // bd --json -q can still emit warnings to stdout before the JSON payload.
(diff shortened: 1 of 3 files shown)


rbriski:
Closing — punting on this. The unit-test assertion fix is straightforward, but fixing the integration tests properly requires real test-infrastructure surgery (per-subtest Dolt database isolation or pre-test catalog cleanup) that's beyond the scope of "unblock CI." The port-conflict patch I tried just exposed a deeper destroy-issues guard issue that was being masked.

Punting to a dedicated bead/PR for the test-infrastructure work. Branch left on the author's fork.

PR #3850, opened 2026-05-06
fix: use bd repo flag for MR creation

tommyatcomun:
## Summary
- map CreateOptions.Rig to bd create --repo instead of unsupported --rig
- add a focused regression test for merge-request bead creation args

## Verification
- go test ./internal/beads
- make build
- /opt/homebrew/bin/gt version
- gt mq list comunbackend shows co-wisp-vbf ready

Escalation: hq-wisp-ozt0

---
<a href="https://app.blacksmith.sh/gastownhall/codesmith/gastown/pr/3850"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/internal/beads/beads.go
+++ b/internal/beads/beads.go
@@ -298,7 +298,7 @@ type CreateOptions struct {
 	Parent      string
 	Actor       string // Who is creating this issue (populates created_by)
 	Ephemeral   bool   // Create as ephemeral (wisp) - not synced to git
-	Rig         string // Target rig database (e.g., "gastown"). When set, passes --rig to bd create.
+	Rig         string // Target rig database (e.g., "gastown"). When set, passes --repo to bd create.
 }
 
 // UpdateOptions specifies options for updating an issue.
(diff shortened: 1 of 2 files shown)


steveyegge:
Thanks for catching this, @tommyatcomun — you correctly identified that `--rig=X` is unsupported in `bd.Create`.

Closing as superseded: a different fix for the same underlying problem landed in #3771 (commit b7096d71, `fix(beads): route Create via BEADS_DIR instead of --repo to prevent pthread deadlock`, hq-1uf2). Main now routes Create through `bdForCreate` with `BEADS_DIR` instead of either `--rig` or `--repo`.

The reason we went with BEADS_DIR routing rather than `--repo`: when a polecat's own `BEADS_DIR` resolves to the same Dolt database that `--repo` would target (e.g., same-rig case like `gt done`), the two connections deadlock on `pthread_cond_wait`. The `--repo` path in this PR would re-introduce that deadlock for the same-rig scenario.

No homework on your end — the bug is fixed. Appreciate the diagnosis and the test case.

PR #3858, opened 2026-05-07
fix(agents): set EscapeCancelsRequest on Claude + Copilot presets to gate poller's Escape keystroke

jeremynorris:
## Summary

The nudge-poller fabricates `[Request interrupted by user]` events on Claude
sessions because its `NudgeSessionWithOpts` injection sends an unconditional
Escape keystroke (tmux.go step 5) on every drain — and Escape is Claude
Code's user-interrupt binding.

**Cadence-signature evidence**: interrupts cluster on 10-second poll
boundaries (`:07/:17/:27/:37/:47/:57`). 39 fabricated interrupts observed
across 6 sessions in 2 rigs in a single investigation pass. Operator
workaround was killing poller PIDs manually.

## Mechanism

The infrastructure to gate this was already 95% built under [gt-wasn]
for Gemini:

- `AgentPresetInfo.EscapeCancelsRequest` on the registry (`config/agents.go`)
- `tmux.NudgeOpts.SkipEscape`
- Wiring in `nudge_poller.go:79-87` and `nudge.go:271-278` that reads the
  preset and sets `SkipEscape` upstream
- Gating at `tmux.go:1737` that suppresses the Escape keystroke + the
  600ms readline timeout when `SkipEscape` is true

What was missing: **Claude's preset** (and `AgentGroqCompound`, which
routes through the Claude binary) had `EscapeCancelsRequest` defaulting to
`false`, so the poller's drain loop sent Escape every cycle into a Claude
pane that treats Escape as cancel.

## Fix

1. Set `EscapeCancelsRequest: true` on the **Claude** and **GroqCompound**
   presets. The existing plumbing propagates correctly with no other
   change.

2. **Bundled cleanup**: replace the hardcoded
   `if agentType == \"copilot\" { SkipEscape = true }` carve-out at
   `tmux.go:1727-1735` (added under [hq-isz]) with the registry-driven
   path. Set `EscapeCancelsRequest: true` on the **Copilot** preset and
   remove the per-agent `if` block.

   This is the same architectural pattern the operator's bar requires:
   registry truth, not per-rig carve-outs.

## Why no spawn-time gate

An earlier framing of this bug suggested skipping the nudge-poller spawn
entirely for Claude. Three spawn sites (`crew/manager.go:892`,
`witness/manager.go:264`, `refinery/manager.go:251`) explicitly justify
starting the poller for Claude — under [gt-dgf], to break the deadlock
where idle Claude sessions never submit a prompt and never trigger
`UserPromptSubmit` drain.

Skipping spawn for Claude would regress gt-dgf. The bug is the destructive
Escape during *delivery*, not the spawn.

## Tests

- New `TestAgentPresetEscapeCancelsRequest` in `internal/config/agents_test.go`
  covering the per-preset matrix (12 subtests: Claude/GroqCompound/Gemini/
  Copilot true; rest false).
- Existing `TestNudgeSession_*` integration coverage in
  `internal/tmux/tmux_test.go` continues to pass after the carve-out
  removal.

## Test plan

- [x] `go build ./...` clean
- [x] `go test ./internal/config/ ./internal/nudge/` green
- [x] `go test ./internal/tmux/ -run TestNudgeSession` green (NudgeSession
  delivery semantics unchanged for Claude path; carve-out removal does
  not regress existing behavior because callers route through the
  registry which now sets `SkipEscape` for Copilot)
- [ ] CI runs full suite (this PR)

Closes co-mhh5 (mirrors town-root ps-4vb).

---
<a href="https://app.blacksmith.sh/gastownhall/codesmith/gastown/pr/3858"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/internal/config/agents.go
+++ b/internal/config/agents.go
@@ -251,6 +251,7 @@ var builtinPresets = map[AgentPreset]*AgentPresetInfo{
 		InstructionsFile:       "CLAUDE.md",
 		EmitsPermissionWarning: true,
 		HasTurnBoundaryDrain:   true,
+		EscapeCancelsRequest:   true, // Claude Code: Escape cancels in-flight generation (ps-4vb / co-mhh5)
 	},
 	AgentGemini: {
 		Name:                AgentGemini,
@@ -403,10 +404,11 @@ var builtinPresets = map[AgentPreset]*AgentPresetInfo{
 		HooksProvider:      "copilot",
 		HooksDir:           ".github/hooks",
 		HooksSettingsFile:  "gastown.json",
-		HooksInformational: false,
-		ReadyPromptPrefix:  "",   // GA: no ❯ prompt; Copilot uses hint text, not a detectable prefix
-		ReadyDelayMs:       5000, // Delay-based readiness detection (no prompt prefix)
-		InstructionsFile:   "AGENTS.md",
+		HooksInformational:   false,
+		ReadyPromptPrefix:    "",   // GA: no ❯ prompt; Copilot uses hint text, not a detectable prefix
+		ReadyDelayMs:         5000, // Delay-based readiness detection (no prompt prefix)
+		InstructionsFile:     "AGENTS.md",
+		EscapeCancelsRequest: true, // Copilot CLI: Escape cancels in-flight generation (hq-isz)
 	},
 	AgentPi: {
 		Name:                AgentPi,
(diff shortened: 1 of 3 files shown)


jeremynorris:
Closing per operator direction; opened without consent. Sorry.

PR #3860, opened 2026-05-07
fix(mail): kill bd subprocess group on context timeout

athosmartins:
## Summary
`gt mail poll-and-nudge` (and any other `internal/mail` caller) was leaving orphaned `dolt sql-server` processes behind every time a per-bd-call context timeout fired. Switches the mail subprocess setup from `util.SetDetachedProcessGroup` (no Cancel hook installed) to `util.SetProcessGroup` (installs `cmd.Cancel = SIGKILL -PGID`), and bounds `cmd.WaitDelay` to 5s on both the primary and `--flat` retry paths.

## Why
On a Gas Town deployment with launchd-driven `mail poll-and-nudge`:
- A bd subprocess hits its 60s read timeout
- Go's `exec.CommandContext` SIGKILL'd bd directly — but with no `cmd.Cancel` hook installed, bd's child Dolt server (port 3307, working DB) survived as an orphan
- Next poll-and-nudge run inherited the contention and hung
- launchd retried every 5 minutes — the rig accumulated unkillable zombies (kill -9 required)

## Implementation
- `internal/mail/bd.go` — replace `util.SetDetachedProcessGroup` with `util.SetProcessGroup` for both bd invocations. The `SetProcessGroup` helper already installs the Cancel hook that signals the entire process group; `SetDetachedProcessGroup` intentionally omits it for fire-and-forget callers (documented in the new comment block).
- `cmd.WaitDelay = 5 * time.Second` on both invocations so `cmd.Run()` is bounded after Cancel even if a child ignores SIGKILL.
- No public API changes. The detached/non-detached distinction is preserved for the other call sites that rely on it.

## Tests
- `TestSetProcessGroup_KillsChildren` — spawns a shell that backgrounds a 600s sleep, captures the child PID, cancels the context, asserts the child PID is dead within a brief retry window.
- `TestSetDetachedProcessGroup_LeavesChildrenForContrast` — pins the opposite contract so a future refactor doesn't accidentally homogenize the two helpers.

## Local verification
Ran `gt mail poll-and-nudge --dry-run` 3 times post-install. All 3 timed out under load (a separate, bd-side slowness — `bd sql` re-runs auto-import from JSONL on every call), but `pgrep -c 'dolt sql-server'` stayed flat across all runs. Before this fix, each timed-out invocation grew the zombie set by one.

## Scope
This PR strictly addresses the subprocess-cleanup half of the bug: when bd hangs (whatever the reason), the mail plumbing now cleans up the entire process tree on context cancellation. The bd-side hang itself (auto-import re-running on every `bd sql` invocation despite Dolt already serving on port 3307) is a separate `steveyegge/beads` issue and stays out of this PR.

---
<a href="https://app.blacksmith.sh/gastownhall/codesmith/gastown/pr/3860"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>


athosmartins:
Closing — branch accumulated 48 commits of unrelated WIP. Re-cutting as a narrow single-commit PR against current main.

PR #3863, opened 2026-05-07
fix(mail): set BEADS_NO_AUTO_IMPORT=1 on bd subprocess env

athosmartins:
## Summary
Set `BEADS_NO_AUTO_IMPORT=1` on every bd subprocess spawned from `internal/mail`. The mail subprocess always targets an already-initialized database, so the JSONL-auto-import fallback added in bd 1.0.3 has nothing useful to do — it just re-reads multi-MB `issues.jsonl` on every call, races into transaction timeouts, and prints a noisy banner that operators see on every `gt mail send`.

## Why
On a Gas Town deployment with bd 1.0.3:
- Every `gt mail send` printed `auto-importing N bytes from issues.jsonl into empty database…` to stderr
- The mail subprocess timed out under load (poll-and-nudge particularly affected — companion fix in #3 of this fork, equivalent upstream PR)
- Crew workflow noise grew significantly without functional benefit

## Implementation
- Extracted `bdSubprocessEnv(baseEnv, beadsDir, extraEnv)` helper so env construction is unit-testable. Previously the entries were appended inline in `runBdCommand`; the helper keeps it cohesive and lets us add tests.
- `BEADS_NO_AUTO_IMPORT=1` is appended *after* the canonical env entries but *before* `extraEnv`, so a specific call site can still override it via `extraEnv` if needed (Go's `exec` honors the last occurrence of a key).
- Retry path (`--flat` fallback) reuses `cmd.Env`, so it inherits the override automatically.

Standalone `bd` invocations are unaffected — only mail subprocesses get this env override.

## Tests
- `TestBdSubprocessEnv_SuppressesAutoImport` — asserts the env var is in the constructed env, alongside `BEADS_DIR` and inherited `PATH`.
- `TestBdSubprocessEnv_ExtraEnvAppendedAfterCanonical` — verifies override ordering so per-call escape hatches work.
- `TestBdSubprocessEnv_DoesNotMutateBaseEnv` — regression guard against silently mutating `cmd.Environ()`'s backing array.

## Local verification
- `gt mail poll-and-nudge --dry-run`: 4.5s clean run (was 25-30s timeouts / noisy banner output)
- `gt mail send` to self: clean success line, no JSONL-import noise on stderr

---
<a href="https://app.blacksmith.sh/gastownhall/codesmith/gastown/pr/3863"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>


Bella-Giraffety (changes requested):
Thanks for tracking this down. I don't think this PR should land as-is because it is now merge-conflicting and carries a large set of unrelated changes beyond the `BEADS_NO_AUTO_IMPORT` mail subprocess fix.

The scoped env fix has been extracted locally as a small two-file change with tests. The extracted version sets `BEADS_NO_AUTO_IMPORT=1` for mail-spawned `bd` subprocesses, reuses that env on the `--flat` retry path, filters stale Beads target env, and preserves prefix routing when `beadsDir` is empty.

Recommendation: land the smaller fix separately, then close this PR as superseded or reopen with a freshly rebased, narrowly scoped branch if anything remains.

athosmartins:
Closing per reviewer request — the focused two-file BEADS_NO_AUTO_IMPORT env fix will land via the extracted change you mentioned. Thanks for triaging.

PR #3887, opened 2026-05-08
fix(convoy): set explicit sender on auto-notifications

athosmartins:
## Problem

Convoy completion notifications fired automatically (e.g. by refinery patrol auto-close) hit the recipient's inbox with `From: overseer`. The "overseer" identity is meant to represent the human user at the terminal; using it for automated convoy traffic silently impersonates them, making automated notifications indistinguishable from human-sent ones.

## Reproduce

Run any flow that lands a convoy without an explicit sender (e.g. `gt convoy land <id>` while the tracking convoy was set up by automation, or refinery patrol's auto-close):

```
$ gt mail inbox
  ● Convoy complete: <title>
      from: overseer    ← should be the convoy machinery, not the human
```

In our deployment this surfaced as 10+/min mayor-inbox notifications all from "overseer" after a v1.1.0 sync.

## Root cause

`internal/cmd/convoy.go` spawns `gt mail send` / `gt nudge` subprocesses without setting an explicit sender. `mail_identity.detectSender()` then walks:

1. `GT_ROLE` env var → not set in this code path
2. `detectSenderFromCwd()` → walks the cwd, doesn't match a known agent dir
3. Final fallback → `"overseer"`

So in any context where the parent process isn't itself an explicit-role agent (most automation entry points), the sender silently resolves to the human.

## Fix

Per call site in `internal/cmd/convoy.go` (5 total):

- **Mail** (`sendCloseNotification` + 2 in `notifyConvoyCompletion`): append `--from convoy/<id>`. The full-address syntax (with `/`) is honored directly by `detectSenderFromRole` (mail_identity.go:113-115), bypassing the cwd fallback that produces "overseer".

- **Nudge** (2 in `notifyConvoyCompletion` + `notifyMayorSession`): set `Cmd.Env = os.Environ() + "GT_ROLE=convoy/<id>"`. The nudge sender builder (nudge.go:431 default branch) uses `GT_ROLE` as a full address when it contains `/`.

Helpers `convoyNotifyFrom(convoyID)` and `convoyNudgeEnv(convoyID)` keep the format DRY.

## Why `convoy/<id>` (with the slash)

Any value without a slash (`system`, `convoy`) would fall into `detectSenderFromRole`'s `default` branch, re-trigger `detectSenderFromCwd`, and roll back to "overseer" — same bug. The slash makes the value a full address that `detectSenderFromRole` returns verbatim.

## Smoke test

```
$ ./gt mail send --self -s test -m smoke --from convoy/test-123
$ # received message: from = "convoy/test-123" ✓
```

## Backward compatibility

- Manual `gt convoy land/close` from a user terminal: the *automated notifications fired by these commands* now correctly attribute to the convoy machinery rather than the user. The user is still the one who initiated; only the machine-generated message body is correctly labeled.
- Other callers of `gt mail send` / `gt nudge` are untouched.

---
<a href="https://app.blacksmith.sh/gastownhall/codesmith/gastown/pr/3887"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>


athosmartins:
Closing — branch accumulated 53 commits of unrelated WIP. Re-cutting as a narrow single-commit PR against current main.

PR #3888, opened 2026-05-08
fix(convoy): persistent dedup on completion notifications

athosmartins:
## Problem

Convoy completion notifications fire repeatedly for the same convoy. Whenever a process re-runs the convoy close-check (e.g. refinery patrol on each cycle), `notifyConvoyCompletion` re-fires the full mail/nudge stack: same recipients, same content. Observed in our deployment: 3× notifications to mayor for one convoy, 2× for another, all within a few minutes.

Companion to the sender-impersonation fix (separate PR #3887 in this same repo) — that one fixed *who* the sender appears to be; this one fixes *how often* it fires.

## Reproduce

1. Land any convoy whose tracked issues all close (manually or via auto-close).
2. Run `gt convoy close-check` again (or wait for refinery patrol's next cycle).
3. Recipient inbox: same `Convoy complete: <title>` message arrives twice.

## Root cause

`notifyConvoyCompletion` only had an in-memory `notifiedAddrs` map for dedup, scoped to a single function call. Nothing on disk recorded that notifications had already fired for a given convoy, so each re-invocation started from a fresh empty map and re-sent everything.

## Fix

Stamp `NotificationSentAt` (RFC3339) on the convoy bead description the first time notifications fire. Subsequent calls early-return when the field is non-empty.

### Changes

**`internal/beads/fields.go`**
- Add `NotificationSentAt string` to `ConvoyFields`
- Parser cases: `notification_sent_at`, `notification-sent-at`, `notificationsentat` (tolerance for variants)
- Formatter line in `FormatConvoyFields`
- Entry in `convoyKeys` so `SetConvoyFields` replaces the line on update rather than duplicating

**`internal/cmd/convoy.go`**
- In `notifyConvoyCompletion`: after parsing fields, early-return if `NotificationSentAt != \"\"`
- After the mail/nudge sends, build new description via `SetConvoyFields(...NotificationSentAt = now)` and write via the existing `updateConvoyDescription` helper
- Persist failure is best-effort warned; worst case reverts to the previous (refire) behavior — strictly no worse than today

**`internal/beads/fields_test.go`**
- `TestConvoyFieldsNotificationSentAtRoundTrip` covers parse → format → parse and replace-on-update

## Test results

```
$ go test ./internal/beads/ ./internal/cmd/
ok  	github.com/steveyegge/gastown/internal/beads	7.885s
ok  	github.com/steveyegge/gastown/internal/cmd	102.053s
```

All existing convoy tests still pass.

## Race / failure modes

- **Concurrent close-check cycles**: two cycles could both observe an empty `NotificationSentAt`, both send, both write. Notifications duplicated within that race window only; subsequent cycles short-circuit. Refinery patrol is sequential in practice, so the window is tiny.
- **`updateConvoyDescription` fails**: warned via `style.PrintWarning`; behavior reverts to the previous (refire) bug for that convoy. Strictly no worse than today.
- **Existing convoys**: parser tolerates legacy convoys without the field; refire is suppressed once `NotificationSentAt` is stamped on the next cycle. No migration needed.

---
<a href="https://app.blacksmith.sh/gastownhall/codesmith/gastown/pr/3888"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>


athosmartins:
Closing — branch accumulated 54 commits of unrelated WIP. Re-cutting as a narrow single-commit PR against current main.

Issue #1400, opened 2026-02-13
How we triage: labels, priorities, and what to expect


julianknutsen:
We've introduced a label system to organize the issue tracker. Every triaged issue gets a **status**, **priority**, and **kind** label.

If you're wondering what a label on your issue means or how to label a new one, see the full guide:

**[Labels and Triage: How We Organize Issues and PRs](https://github.com/steveyegge/gastown/discussions/1399)**

### Quick reference

| Label | Meaning |
|-------|---------|
| `status/needs-triage` | Inbox — we haven't looked at it yet |
| `status/needs-info` | Waiting on you — please reply |
| `status/needs-repro` | We can't reproduce it — please provide steps |
| `status/accepted` | Confirmed and on our radar |
| `status/help-wanted` | Up for grabs — contributions welcome |
| `priority/p0` | Critical — data loss, security, completely blocked |
| `priority/p1` | High — core workflow broken |
| `priority/p2` | Medium — real problem, workaround exists |
| `priority/p3` | Backlog — good idea, reviewed when there's bandwidth |
| `kind/bug` | Broken behavior |
| `kind/feature` | New capability |
| `kind/docs` | Documentation only |
| `kind/chore` | Internal improvement (refactor, tests, CI, tooling) |

Questions? Comment here or start a thread in the [Discussion](https://github.com/steveyegge/gastown/discussions/1399).

PR #3890, opened 2026-05-08
fix(escalate): pass description via stdin to bd create

athosmartins:
## Summary

`gt escalate` is currently broken: every invocation fails because `CreateEscalationBead` builds a multi-line YAML metadata block (severity, reason, escalated_by, …) and passes it through `bd create --description=…`. `bd` 1.0.3+ rejects newline-containing flag values, so it prints help and exits non-zero. Reproducer:

```
$ gt escalate 'Test' -s high -r 'reason'
Error: creating escalation bead: bd create --json --title=Test
  --description=Test\n\nseverity: high\nreason: reason\n... (etc)
  --type=task --ephemeral ...
  Usage: bd create ...
```

This PR switches `CreateEscalationBead` to feed the description through stdin via `bd create --body-file=-`. To support that, `internal/beads/beads.go` adds a `runWithStdin([]byte, ...string)` helper. `run()` is a thin wrapper around it (`runWithStdin(nil, args...)`) so existing callers are unaffected. Stdin is re-attached on the `--flat` retry path so a re-exec doesn't lose the input.

Reported by `batista` (dc-wisp-zcb2, 2026-05-07); tracked downstream as `dc-1bxe` in the whatsapp_automation rig.

## Test plan
- [x] `go test ./internal/beads/...` — passes (10s)
- [x] New regression test `TestCreateEscalationBead_PassesDescriptionViaStdin` — fake bd stub asserts `--body-file=-` is set, `--description=…` is **not** in args, and the multi-line description reaches bd via stdin
- [x] `go test ./...` — only failures are pre-existing on `origin/main` (`internal/tmux/dialog_acceptance_test.go` timing flakes and `TestNudgeSession_*`); none touch beads/escalate
- [x] Confirmed end-to-end: stub bd subprocess receives the structured YAML metadata via stdin and returns valid JSON (real `gt escalate` from `~/gt` no longer fails at flag parsing — the only remaining failure mode is a separate, pre-existing bd auto-import 60s-timeout issue, untouched by this PR)

---
<a href="https://app.blacksmith.sh/gastownhall/codesmith/gastown/pr/3890"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/internal/beads/beads.go
+++ b/internal/beads/beads.go
@@ -456,7 +456,15 @@ func resolveBdSubprocessTimeout() time.Duration {
 }
 
 // run executes a bd command and returns stdout.
-func (b *Beads) run(args ...string) (_ []byte, retErr error) {
+func (b *Beads) run(args ...string) ([]byte, error) {
+	return b.runWithStdin(nil, args...)
+}
+
+// runWithStdin executes a bd command, optionally piping stdinData to bd's stdin.
+// When stdinData is nil, behaves identically to run. Use this for flags like
+// --body-file=- that read multi-line content from stdin (avoids embedding
+// newlines in --description, which bd 1.0.3+ rejects).
+func (b *Beads) runWithStdin(stdinData []byte, args ...string) (_ []byte, retErr error) {
 	start := time.Now()
 	// Declare buffers before defer so the closure captures them after cmd.Run.
 	var stdout, stderr bytes.Buffer
@@ -496,6 +504,9 @@ func (b *Beads) run(args ...string) (_ []byte, retErr error) {
 
 	cmd.Stdout = &stdout
 	cmd.Stderr = &stderr
+	if stdinData != nil {
+		cmd.Stdin = bytes.NewReader(stdinData)
+	}
 
 	err := cmd.Run()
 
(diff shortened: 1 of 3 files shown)


Bella-Giraffety (commented):
Final read-only sheriff review:

I do not see a code-level blocker in the PR diff. The change is tightly scoped: `gt escalate` now sends the multi-line escalation body through `bd create --body-file=-`, and the new regression test verifies both that stdin is used and `--description=...` is not used for the newline-containing body.

Merge readiness: not ready as-is because the PR still has failing CI checks (`Lint` and `Integration Tests`) in the current GitHub status. The failures I inspected look unrelated to this diff (`internal/polecat/manager.go` misspell, Dolt port/test harness failures, and `bd close --tombstone` support), so I would not ask the contributor for code changes based on this pass; I would rerun/fix the suite or have a maintainer override only if those failures are confirmed pre-existing.

Label audit: currently exactly one label in each required group is present: `status/needs-triage`, `priority/p2`, `kind/bug`. Count-wise this satisfies the #1400 scheme. Semantically, after triage/review I recommend replacing `status/needs-triage` with `status/accepted`; `kind/bug` is correct. `priority/p2` is acceptable, though `priority/p1` would also be defensible because `gt escalate` is a core workflow.

Review stance: comment only, not approve/request changes, until required checks are green or explicitly waived by a maintainer.

PR #3891, opened 2026-05-08
fix(git): preserve cherry-pick-equivalent commits during Rebase()

athosmartins:
## Summary
The refinery's merge pipeline (`internal/refinery/engineer.go::doMerge`) calls `Git.Rebase(onto)` on every MR before merging. The current implementation runs a bare `git rebase <onto>`, which under modern git (≥2.34, default `rebase.reapplyCherryPicks=false`) **silently drops** any commit whose patch-id matches a commit already on the target branch.

In a merge-queue context this is the wrong default. When a parallel MR has already landed a content-equivalent commit (`A'`), and an in-flight MR contains the original (`A`) plus downstream commits (`B`, `C`) that depend on `A`'s tree, the rebase removes `A` preemptively. `B` and `C` then conflict against main, the operator resolves by taking HEAD — and the MR author's actual changes are silently erased. The merge completes with no warning.

This was caught live in the `whatsapp_automation` rig (2026-05-08): refinery output showed `dropping 17a4fd8e... fix: v3.10 multi-owner lot click expande -- patch contents already upstream`, immediately followed by a conflict on a downstream commit that depended on the dropped commit's tree. Mayor had to bypass the queue manually for the affected MRs.

## Fix
Pass two flags to `git rebase`:

| Flag | Why it's needed |
| --- | --- |
| `--reapply-cherry-picks` | Skip the preliminary cherry-pick deduplication step that drops `A` outright. |
| `--empty=keep` | After `A` is reapplied on top of `A'`, the resulting tree is empty; without `--empty=keep` (default for non-interactive rebase is `drop`) the empty commit is removed and `A` is gone again. |

Both are required — using only one still drops the commit. The combination guarantees `git rebase --empty=keep --reapply-cherry-picks <onto>` produces a branch with the same commit count it started with. Genuine duplicates surface as visible empty commits the operator can resolve deliberately, never as silent loss.

## Test plan
Added `TestRebase_PreservesCherryPickEquivalentCommits` that constructs the exact failure scenario:

1. Initial commit + commit `A` ("fix" → `a.txt`)
2. Branch `feature`, add commit `B` (depends on `A`'s tree)
3. Reset `main` and recreate `A'` (same patch content, different SHA), then add an unrelated commit
4. Rebase `feature` onto `main` and assert both `A` and `B` are preserved

Without the fix the test fails with only `B` reaching HEAD; with the fix both commits land.

- [x] New test passes against the patched `Rebase()`
- [x] Full `go test ./internal/git/` suite passes (21.9s, no regressions)
- [x] `go build ./...` succeeds (no callers of `Rebase` need changes — the contract is strictly "more commits preserved", not "different signature")
- [x] Shell-level repro confirms the same flag combination preserves both commits in a fresh repo

## Related
- Diagnostic bead in the affected rig: `dc-41k3` (whatsapp_automation rig) with full investigation notes, smoking-gun production capture, and reproduction recipe.
- Affected rig conversation thread: `dc-wisp-71pq` (mayor → batista, 2026-05-07) where the bypass workaround is documented.

---
<a href="https://app.blacksmith.sh/gastownhall/codesmith/gastown/pr/3891"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>

### DIFF
--- a/internal/git/git.go
+++ b/internal/git/git.go
@@ -12,6 +12,7 @@ import (
 	"path/filepath"
 	"runtime"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/steveyegge/gastown/internal/util"
@@ -1256,8 +1257,49 @@ func (g *Git) ListPushRemoteRefsWithHashes(remote, prefix string) ([]RemoteRef, 
 	return g.ListRemoteRefsWithHashes(pushURL, prefix)
 }
 
+// rebaseSupportsNoSilentDrop is true when the local git binary supports
+// --reapply-cherry-picks (introduced in Git 2.34). Probed once at first use.
+var (
+	rebaseCapOnce       sync.Once
+	rebaseCapNoSilentDrop bool
+)
+
+func probeRebaseCapability() bool {
+	out, err := exec.Command("git", "rebase", "--help").CombinedOutput()
+	if err != nil {
+		// If --help fails, assume old git and fall back to safe default.
+		return false
+	}
+	return strings.Contains(string(out), "--reapply-cherry-picks")
+}
+
 // Rebase rebases the current branch onto the given ref.
+//
+// On Git ≥2.34, two flags prevent silent commit loss in the `gt done`
+// auto-rebase path:
+//
+//   --reapply-cherry-picks  Skip the cherry-pick dedup pass that silently
+//                           drops commits whose patch-id already exists upstream.
+//   --empty=keep            Preserve commits that become empty after rebase
+//                           (i.e. the patch was genuinely identical) so they
+//                           surface as visible empty commits, not silent loss.
+//
+// On older Git, these flags are unavailable and the function falls back to a
+// plain rebase. The README documents Git 2.25+ as the minimum.
+//
+// Without these flags on ≥2.34, when a `gt done` rebase encounters a commit
+// whose content already exists on the target (a parallel fix, a backport),
+// git silently removes it. Downstream commits then conflict against the target,
+// and the natural conflict resolution ("take HEAD") erases the author's changes
+// with no warning.
 func (g *Git) Rebase(onto string) error {
+	rebaseCapOnce.Do(func() {
+		rebaseCapNoSilentDrop = probeRebaseCapability()
+	})
+	if rebaseCapNoSilentDrop {
+		_, err := g.run("rebase", "--empty=keep", "--reapply-cherry-picks", onto)
+		return err
+	}
 	_, err := g.run("rebase", onto)
 	return err
 }
--- a/internal/git/git_test.go
+++ b/internal/git/git_test.go
@@ -486,6 +486,118 @@ func TestCheckConflicts_WithConflict(t *testing.T) {
 	}
 }
 
+// TestRebase_PreservesCherryPickEquivalentCommits verifies that Rebase passes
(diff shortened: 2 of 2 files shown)


Bella-Giraffety (changes requested):
Thanks for writing up the production incident and adding a focused regression test. I agree with the underlying concern that silent cherry-pick drops are dangerous in merge-queue workflows, but I don't think this PR is merge-ready yet.

Requesting changes for two blockers:

1. The fix does not appear to affect the refinery path described in the PR body. The PR says `internal/refinery/engineer.go::doMerge` calls `Git.Rebase(onto)`, but current refinery code checks conflicts and then squash-merges (`internal/refinery/engineer.go:555-652`), and batch processing also stacks via conflict checks and squash merges (`internal/refinery/batch.go:90-152`). The production caller I found for `Git.Rebase` is `gt done` auto-rebase (`internal/cmd/done_rebase.go:28-45`). Please either wire/test the actual refinery path being fixed, or retarget the PR body/comments/test names to the `gt done` auto-rebase behavior it currently changes.

2. This introduces `git rebase --empty=keep --reapply-cherry-picks`, but the README still documents Git 2.25+ support (`README.md:123-127`). Those flags are not available on the documented minimum Git version. Please add compatibility handling/capability detection, or intentionally raise/enforce the minimum Git version before this lands.

I would also trim the long `Rebase` comment once the scope is clarified. As written it states a broad refinery contract and same-commit-count guarantee on a thin git wrapper, but that does not match the current callers.

Verification I ran locally:

- `go test ./internal/git/` passed
- `go test ./internal/cmd/ -run 'TestAutoRebase|TestDone'` passed
- `go test ./internal/refinery/...` and broader `go test ./internal/git/ ./internal/cmd/` were blocked in this environment by missing rootless Docker/testcontainers setup

GitHub CI note: the current `Test` check passed. The failing `Lint` and `Integration Tests` checks appear unrelated/stale, but CI is still not fully green.

athosmartins:
Thanks @Bella-Giraffety — both blockers addressed in the updated commit.

**1. Scope retargeted to `gt done`**

You're right that `internal/refinery/engineer.go` squash-merges and doesn't call `Git.Rebase`. The actual caller is `autoRebaseOnTarget` in `internal/cmd/done_rebase.go`. Updated:

- Commit message now reads "preserve cherry-pick-equivalent commits during gt done rebase"
- `Rebase()` doc comment references the `gt done` auto-rebase path instead of refinery
- Test comment now says "`gt done` auto-rebase" instead of "merge-queue context"

**2. Git version compatibility — capability detection added**

Rather than bumping the README minimum to 2.34, I added a `probeRebaseCapability()` helper that runs `git rebase --help` and checks the output for `--reapply-cherry-picks`. Called once via `sync.Once` and cached in a package-level bool. If the flag is absent (Git <2.34) the function falls back to plain `git rebase <onto>`. README minimum stays at Git 2.25+.

**3. Long comment trimmed**

The `Rebase()` docstring is shorter and no longer makes broad claims about "same commit count" guarantees — it's scoped to the `gt done` fallout and the fallback behavior.

PR #3935, opened 2026-05-10
fix: bypass prefix routing for agent-bead operations (3 commits)

kolloch:
## Summary
Agent beads (issues with `gt:agent` label — one per polecat, dog, witness, refinery, mayor) live in the town/hq Dolt database, but their IDs carry a rig prefix (e.g. `<rig-prefix>-<rig>-polecat-<name>`). Default prefix routing therefore sends agent-bead lookups into the rig DB and returns "issue not found" — silently dropping cleanup-status, MR-pointer, and idle-state writes.

Symptom in production: `gt done` emits ~6 "issue not found" warnings per call; `agent_state` stays at `working` forever; the witness sees a "busy" polecat that's actually idle (zombie state requiring manual nuke).

This PR adds `Beads.ForAgentBead()` (sets `noRoute=true`) and threads it through the daemon's `getAgentBeadInfo` and `gt done`'s `UpdateAgentActiveMR`.

## Related Issue
Fixes the daemon getAgentBeadInfo lookup and gt done's agent-bead writes. (Tracking IDs in commit messages.)

## Changes
- `internal/beads/beads.go` + `beads_agent.go`: new `ForAgentBead()` helper with `noRoute` flag. `Show` / `forIssueID` / `UpdateAgentState` / `ResetAgentBeadForReuse` skip `ResolveRoutingTarget` when set.
- `internal/daemon/lifecycle.go`: `getAgentBeadInfo` uses `bdReadOnlyEnv()` + `BEADS_DIR=townRoot/.beads` override (subprocess equivalent of `ForAgentBead`).
- `internal/cmd/done.go`: switch all agent-bead writes (cleanup status, MR pointer, idle state, done-intent label, checkpoints) to the `agentBd := bd.ForAgentBead()` handle.
- `internal/polecat/manager.go`: same routing fix for `ResetAgentBeadForReuse`.

## Testing
- [x] `go build ./internal/...` clean
- [ ] Manual: run `gt done` in a polecat — should emit ZERO "issue not found" warnings; `agent_state` transitions `working` → `idle`.

## Checklist
- [x] No breaking changes (additive helper + opt-in usage)

### DIFF
--- a/internal/beads/beads.go
+++ b/internal/beads/beads.go
@@ -379,7 +379,7 @@ func NewWithBeadsDir(workDir, beadsDir string) *Beads {
 
 // ForAgentBead returns a Beads wrapper suitable for operating on agent beads.
 //
-// Agent beads (labelled gt:agent) live in the TOWN database, but their IDs
+// Agent beads (labeled gt:agent) live in the TOWN database, but their IDs
 // are prefixed with the rig prefix (e.g. "za-zack-polecat-furiosa"). The
 // default prefix routing in routes.jsonl maps "za-" → zack rig database, so
 // any agent-bead operation issued from a rig context (or any context that
@@ -394,7 +394,7 @@ func NewWithBeadsDir(workDir, beadsDir string) *Beads {
 //     ResolveRoutingTarget, forIssueID) do not redirect lookups by prefix.
 //
 // If the town root cannot be determined, returns the original wrapper to
-// preserve current behaviour.
+// preserve current behavior.
 func (b *Beads) ForAgentBead() *Beads {
 	townRoot := b.getTownRoot()
 	if townRoot == "" {
--- a/internal/beads/beads.go
+++ b/internal/beads/beads.go
@@ -341,6 +341,14 @@ type Beads struct {
 	// Populated on first call to getTownRoot() to avoid filesystem walk on every operation.
 	townRoot     string
 	townRootOnce sync.Once
+
+	// noRoute disables prefix-based routing for this Beads instance.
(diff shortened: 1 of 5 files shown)


Bella-Giraffety (changes requested):
Request changes.

Findings:

1. Agent bead create/reopen/reset still routes by prefix for manager lifecycle paths. `CreateAgentBead` and the existing-bead path in `CreateOrReopenAgentBead` still call `ResolveRoutingTarget` without honoring `noRoute`, so even using `ForAgentBead()` would route `gt-gastown-polecat-*` back to the rig DB. The manager reset/reuse callsites also still call `m.beads.ResetAgentBeadForReuse(...)` directly, so nuke/reuse/repair/rollback cleanup still misses the town agent bead and leaves stale `hook_bead`/`active_mr`/cleanup metadata.

2. Refinery merge cleanup still uses `e.beads.UpdateAgentActiveMR(mr.AgentBead, "")` from a rig-rooted `Beads` instance. That preserves the same `issue not found` routing failure after merge and leaves `active_mr` stale even though `gt done` now writes it correctly.

Verified:

- `go build ./internal/...` passes.
- `go test ./internal/beads` passed as part of a focused package run; the broader `go test ./internal/beads ./internal/cmd ./internal/polecat ./internal/daemon` command timed out after `internal/beads` completed.
- Manual read-only check: default `bd show gt-gastown-polecat-rust --json` from the rig misses the agent bead, while `BEADS_DIR=/home/coder/gt/.beads bd show ...` finds it in town DB, matching the PR's root cause.

PR #3942, opened 2026-05-11
bd-hlkrd: guard canonical Beads prefix writes

fengning-starsend:
Agent: codex

Feature-Key: bd-hlkrd

## Summary
- Guard canonical bdx create so new planning issues must use bd-* IDs and canonical prefix config must be consistent.
- Refuse Gastown town/rig prefix setup when it is accidentally pointed at ~/.beads-runtime/.beads.
- Strip inherited Beads routing env before local Gastown rig/town config writes.

## Why
A canonical epyc12 Beads runtime was contaminated with issue_prefix=bds, causing canonical bdx create to mint bds-* IDs that product repo commit hooks correctly reject. Local Gastown execution prefixes must not leak into canonical planning IDs.

## Verification
- go test ./internal/beads ./internal/rig
- Deployed locally on epyc6 and epyc12 as hotfix.
- Canonical bdx create now mints bd-* again after config repair.
- bdx create --id bds-guard-smoke2 fails with canonical_prefix_mismatch on epyc12.

### DIFF
--- a/internal/beads/beads_types.go
+++ b/internal/beads/beads_types.go
@@ -104,6 +104,9 @@ func EnsureCustomTypes(beadsDir string) error {
 	if beadsDir == "" {
 		return fmt.Errorf("empty beads directory")
 	}
+	if err := GuardNonCanonicalRuntime(beadsDir, "configure Gas Town custom types"); err != nil {
+		return err
+	}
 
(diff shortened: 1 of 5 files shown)


fengning-starsend:
Closing per owner instruction. This P0 guard remains deployed locally as an epyc6/epyc12 hotfix while we decide the correct upstream strategy.

PR #3950, opened 2026-05-12
fix: ignore stale nudge pane IDs from other sessions

### LINKED ISSUE #3563: gt nudge fails with stale tmux pane ID on cross-rig crew targets
## Bug

`gt nudge` fails when targeting a crew member on another rig after that crew member's session has been restarted. The error is:

```
Error: nudging session: tmux send-keys: can't find window: %164
```

The target session is confirmed running (`tmux has-session` succeeds), but the pane ID used by `send-keys -t` is stale.

## SSCE (Short, Self-Contained Example)

```bash
# Terminal 1: Start a crew member
gt crew start gilfoyle --rig k8s

# Terminal 2: Restart that crew member (simulates session cycling)
gt crew stop gilfoyle --rig k8s
gt crew start gilfoyle --rig k8s
# Gilfoyle's tmux session now has a NEW pane ID (e.g., %210),
# but something still references the OLD pane ID (e.g., %164)

# Terminal 3: From a different rig's crew, nudge gilfoyle
# (e.g., from www/crew/amos)
gt nudge k8s/crew/gilfoyle "test message"
# ERROR: nudging session: tmux send-keys: can't find window: %164
```

## Root Cause Analysis

The nudge delivery path is:

1. `runNudge()` in `cmd/nudge.go:558` calls `deliverNudge()`
2. `deliverNudge()` calls `NudgeSessionWithOpts()` in `tmux/tmux.go:1624`
3. At `tmux.go:1648`, `FindAgentPane()` is called to resolve the correct pane
4. `FindAgentPane()` first checks `GT_PANE_ID` from the tmux session environment (`tmux.go:1992`)
5. If `GT_PANE_ID` is set but the pane no longer exists, it falls through to scan (`tmux.go:1997`)
6. The scan at `findAgentPaneByScan()` calls `list-panes -s -t <session>` which should work

The likely issue is that `GT_PANE_ID` verification at line 1994 (`display-message -t <paneID>`) may be succeeding against a **different session's pane** with the same ID (tmux pane IDs like `%164` are global), or the session environment retains a stale `GT_PANE_ID` across `crew stop`/`crew start` cycles.

When the nudge then targets `session:%164`, tmux can't find that pane within the specified session's windows, producing the "can't find window" error.

## Expected Behavior

`gt nudge k8s/crew/gilfoyle "message"` should deliver successfully to any running crew session, regardless of whether it has been restarted since the last nudge.

## Environment

- Gas Town version: latest main
- tmux: 3.5a
- OS: Ubuntu 22.04 (Linux 6.8.0-107-generic)
- Trigger: cross-rig nudge (www/crew/amos → k8s/crew/gilfoyle)

## Suggested Fix

When constructing the `send-keys` target at `tmux.go:1654`, verify the resolved pane belongs to the target session by checking that `display-message -t <session>:<paneID>` succeeds (note: session-qualified), not just `display-message -t <paneID>` (globally scoped). If verification fails, fall through to `findAgentPaneByScan()` or use the session's default pane.

Bella-Giraffety:
## Summary
- Fixes #3563.
- Validates that a stored `GT_PANE_ID` still belongs to the target tmux session before `gt nudge` uses it.
- Falls back to the existing pane scan when the declared pane is stale or belongs to another session.
- Adds a regression test for the cross-session stale pane case.

## Bug
`gt nudge <rig>/crew/<name> ...` can fail after the target agent restarts because the target session may retain an old `GT_PANE_ID`. Since tmux pane IDs are global, validating `%pane` without checking its session can accept a pane from another session and later fail with errors like `can't find window: %164`.

## Why This Fix
This keeps the existing fast path for valid `GT_PANE_ID` values but makes the check session-aware. If the pane does not belong to the requested session, `FindAgentPane` uses the existing fallback scan instead of returning a stale pane.

## Scope / Duplicate Check
- Diff is limited to `internal/tmux/tmux.go` and `internal/tmux/tmux_test.go`.
- Checked open PRs for `GT_PANE_ID`, stale pane, nudge pane, cross-rig crew, and tmux pane overlap; no open duplicate found.
- Related prior work exists in merged/closed PRs, but current `main` still lacks this session-membership validation.

## Review Gate
Five independent review passes were run before submission:
- Reproduction/root-cause review: rework branch scope; use only focused commit.
- Minimality/ZFC review: approved focused commit only.
- Test coverage review: approved with non-blocking end-to-end coverage caveat.
- Duplicate/overlap review: no blocking duplicate found.
- PR readiness review: approved fork PR path with small diff.

## Testing
- `git diff --check origin/main...HEAD`
- `go test ./internal/tmux -run '^TestFindAgentPane_IgnoresPaneIDFromOtherSession$' -count=10 -timeout=2m -v`
- `go test ./internal/tmux -run 'TestFindAgentPane' -count=1 -timeout=3m -v`

## Notes
This PR intentionally does not change nudge delivery policy, add commands, or touch internal MQ behavior. It only tightens the tmux transport pane lookup.

### DIFF
--- a/internal/tmux/tmux.go
+++ b/internal/tmux/tmux.go
@@ -2040,11 +2040,19 @@ func (t *Tmux) FindAgentPane(session string) (string, error) {
 	// ZFC: read declared pane identity set at session startup (gt-qmsx).
 	// This replaces process-tree inference for sessions that record GT_PANE_ID.
 	if declaredPane, err := t.GetEnvironment(session, "GT_PANE_ID"); err == nil && declaredPane != "" {
-		// Verify the pane still exists in tmux (it may have been killed/respawned).
-		if _, verifyErr := t.run("display-message", "-t", declaredPane, "-p", "#{pane_id}"); verifyErr == nil {
+		targetSession := session
+		if sessionOut, sessionErr := t.run("display-message", "-t", session, "-p", "#{session_name}"); sessionErr == nil {
+			if resolved := strings.TrimSpace(sessionOut); resolved != "" {
+				targetSession = resolved
+			}
+		}
+
+		// Verify the pane still exists in the target session. Pane IDs are tmux-global,
+		// so a stale GT_PANE_ID may still resolve in a different restarted session.
+		if paneSession, verifyErr := t.run("display-message", "-t", declaredPane, "-p", "#{session_name}"); verifyErr == nil && strings.TrimSpace(paneSession) == targetSession {
 			return declaredPane, nil
 		}
-		// Declared pane is gone — fall through to scan.
+		// Declared pane is gone or belongs to another session — fall through to scan.
 	}
 
 	// Fallback: scan all panes for legacy sessions without GT_PANE_ID.
--- a/internal/tmux/tmux_test.go
+++ b/internal/tmux/tmux_test.go
@@ -1324,6 +1324,40 @@ func TestFindAgentPane_SinglePane(t *testing.T) {
 	}
 }
 
+func TestFindAgentPane_IgnoresPaneIDFromOtherSession(t *testing.T) {
+	tm := newTestTmux(t)
+	suffix := fmt.Sprintf("%d", time.Now().UnixNano()%10000)
+	targetSession := "gt-test-findagent-target-" + suffix
+	otherSession := "gt-test-findagent-other-" + suffix
+
+	_ = tm.KillSession(targetSession)
+	_ = tm.KillSession(otherSession)
+	if err := tm.NewSession(targetSession, ""); err != nil {
+		t.Fatalf("NewSession target: %v", err)
+	}
+	defer func() { _ = tm.KillSession(targetSession) }()
+	if err := tm.NewSession(otherSession, ""); err != nil {
+		t.Fatalf("NewSession other: %v", err)
+	}
+	defer func() { _ = tm.KillSession(otherSession) }()
+
+	otherPane, err := tm.GetPaneID(otherSession)
+	if err != nil {
+		t.Fatalf("GetPaneID other: %v", err)
+	}
+	if err := tm.SetEnvironment(targetSession, "GT_PANE_ID", otherPane); err != nil {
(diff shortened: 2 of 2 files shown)


Bella-Giraffety (commented):
Code review complete. I did not find blocking issues in the PR diff.

What I checked:
- The `FindAgentPane` change correctly keeps the fast path for a valid `GT_PANE_ID` while rejecting stale pane IDs that resolve in a different tmux session.
- The fallback scan remains in place when the declared pane is gone or belongs elsewhere.
- The new regression test covers the cross-session stale pane case.
- Local verification on the PR branch passed:
  - `go test ./internal/tmux -run '^TestFindAgentPane_IgnoresPaneIDFromOtherSession$' -count=10 -timeout=2m -v`
  - `go test ./internal/tmux -run 'TestFindAgentPane' -count=1 -timeout=3m -v`

Holding off on approval because current CI is red:
- `Lint` fails on `internal/polecat/manager.go:526` for `behaviour`, outside this PR's changed files.
- `Integration Tests` fail in `internal/cmd` due Dolt port reuse and `bd close --tombstone` CLI mismatch, also outside this PR's changed files.

This looks code-ready to me once CI is rerun/fixed or maintainers explicitly classify those failures as unrelated to this PR.

Bella-Giraffety (commented):
Independent final review 2 after the branch update onto current `main`.

Blocking findings: none found in the current diff.

Test coverage review:
- The stale-pane regression test still covers the important failure mode after the base update: `GT_PANE_ID` on the target session points at a live pane in a different tmux session, which the old code would have accepted and returned.
- The implementation now resolves the requested target session and compares the declared pane's actual `#{session_name}` before returning it, then falls back to scanning when the pane is gone or cross-session.
- The test's assertion that `FindAgentPane` must not return the other session's pane is sufficient for this regression and avoids over-specifying fallback behavior for a single-pane target session.

Readiness recommendation: code-ready from this review, but wait for the currently pending CI checks to finish green before merging. Current metadata shows the PR is mergeable and has the expected `kind/bug`, `status/accepted`, and `priority/p2` labels.

Issue #875, opened 2026-01-22
Daemon subprocess calls fail to find gt/bd executables due to missing PATH inheritance


DyrtyJax:
## Summary

The daemon's subprocess calls (`exec.Command`) don't explicitly set `cmd.Env`, causing them to fail when the daemon process doesn't have the expected `PATH` environment variable. This manifests as:

```
Warning: failed to fetch deacon inbox: exec: "gt": executable file not found in $PATH
```

## Root Cause

In Go, when `exec.Command` is called without setting `cmd.Env`, the subprocess inherits from the parent. However, when the daemon is started by certain mechanisms (launchd, systemd, or shells with minimal environments), the `PATH` may not include directories like `/opt/homebrew/bin` where `gt` and `bd` are installed.

The fix is simple: explicitly set `cmd.Env = os.Environ()` on all subprocess calls to ensure they inherit the full environment from the running daemon process.

## Affected Code

All `exec.Command` calls in `internal/daemon/lifecycle.go` are missing environment inheritance:

| Line | Command | Impact |
|------|---------|--------|
| 41 | `gt mail inbox` | Breaks lifecycle request processing |
| 576 | `git fetch origin` | Breaks workspace pre-sync |
| 592 | `git pull --rebase` | Breaks workspace pre-sync |
| 608 | `bd sync` | Breaks beads sync |
| 626 | `gt mail delete` | Breaks message cleanup |
| 663 | `bd show` | Breaks agent bead state lookup |
| 800 | `bd list` | Breaks GUPP violation checks |
| 875 | `gt mail send` | Breaks witness notifications |
| 898 | `bd list` | Breaks orphaned work checks |
| 971 | `gt mail send` | Breaks orphaned work notifications |

## Reproduction Steps

1. Start the daemon from a minimal shell environment (or via launchd)
2. Observe daemon logs showing `exec: "gt": executable file not found in $PATH`
3. Heartbeat completes but deacon inbox fetch fails every cycle

## Workaround

Start the daemon from a shell that has the correct PATH:

```bash
# Ensure /opt/homebrew/bin (or wherever gt/bd are) is in PATH
gt daemon start
```

This works because the daemon then inherits the shell's PATH.

## Proposed Fix

Add `cmd.Env = os.Environ()` to all `exec.Command` calls in `lifecycle.go`. This ensures subprocesses inherit the daemon's environment, including PATH.

Example diff:
```go
// Before
[...]

PR #3954, opened 2026-05-12
fix(daemon): recover launchd Homebrew PATH

### LINKED ISSUE #3738: Daemon inherits launchd's minimal PATH — can't find /opt/homebrew/bin, breaks Boot→Top recovery
### Summary

The Gastown launchd daemon inherits launchd's minimal default PATH (`/usr/bin:/bin:/usr/sbin:/sbin`), which **excludes `/opt/homebrew/bin`** — where `gt`, `tmux`, and `bd` all live on Apple Silicon Homebrew installs. The daemon starts, logs `Warning: gt not found in PATH`, and then runs in a permanently-degraded state: every `tmux`/`gt`/`bd` invocation it attempts fails `ENOENT`.

The most damaging downstream effect: **Fire Watch (Boot) fires on every heartbeat but can't actually restart Top (Deacon) when Top dies**, because the `gt`/`tmux` it needs to do that aren't reachable. Top-down incidents therefore never recover on their own — the supervisor is crippled by PATH alone.

### Evidence (from a real incident)

On a recent 24-minute Top outage:

- Daemon PID at the time: `823`
- `ps eww 823` showed PATH: `/usr/bin:/bin:/usr/sbin:/sbin` (no `/opt/homebrew/bin`)
- `daemon.log` contained `Warning: gt not found in PATH` from startup
- `events.jsonl` showed Boot firing every ~3 minutes for the full 24 minutes, **zero successful restart invocations**

Root cause is not a Boot bug — Boot is doing exactly what it's supposed to. The supervisor tree just can't execute anything because `launchctl` doesn't inherit the user's shell PATH.

### Repro

1. Install `gt` via Homebrew on Apple Silicon (binary lands in `/opt/homebrew/bin/gt`)
2. Install the daemon as a LaunchAgent (`~/Library/LaunchAgents/com.gastown.daemon.plist`)
3. Load it: `launchctl load ~/Library/LaunchAgents/com.gastown.daemon.plist`
4. Observe in `daemon.log`: `Warning: gt not found in PATH`
5. `ps eww $(cat ~/gt/.dolt-data/dolt.pid 2>/dev/null || pgrep -f com.gastown.daemon)` — no Homebrew bindir

### Proposed fixes (any of these; c is belt-and-suspenders)

**a) Daemon self-heals its PATH at startup.** Something like:

```go
// At daemon entrypoint, before spawning any subprocesses
func ensurePath() {
    paths := strings.Split(os.Getenv("PATH"), ":")
    seen := map[string]bool{}
    for _, p := range paths { seen[p] = true }

    // Add dir of this binary, plus conventional Homebrew/local bindirs.
    exe, _ := os.Executable()
    for _, p := range []string{
        filepath.Dir(exe),
        "/opt/homebrew/bin",
        "/usr/local/bin",
    } {
        if !seen[p] {
            paths = append(paths, p)
            seen[p] = true
        }
    }
    os.Setenv("PATH", strings.Join(paths, ":"))
}
```

Looking up the binary's own directory is the most robust piece — it works regardless of whether the install came from Homebrew, `go install`, a tarball, or npm.

**b) Homebrew formula installs a plist with `EnvironmentVariables.PATH` already populated**, e.g.:

```xml
<key>EnvironmentVariables</key>
<dict>
    <key>PATH</key>
    <string>/opt/homebrew/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin</string>
</dict>
```

**c) Both.** (a) protects every install path including `go install` and manual installs; (b) makes the Homebrew-specific story clean and visible in the plist.

### Current workaround (local only)

Hand-edit `~/Library/LaunchAgents/com.gastown.daemon.plist` to add an `EnvironmentVariables` block with the correct PATH, then `launchctl unload && launchctl load`. This fixes one machine but doesn't help anyone else.

### Why this matters

This bug silently disables the entire self-healing story Gas Town pitches — Fire Watch can't restart Top, Top-down incidents don't recover, and operators see heartbeat noise in the log that looks like "everything's fine" because Boot is firing on schedule. It's a bad failure mode because it's invisible until you actually need the supervisor to do its job.

Happy to follow up with a PR for option (a) if maintainers agree on the shape.

Bella-Giraffety:
## Summary
- Fixes #3738 by augmenting the daemon process PATH before any subprocess lookup, including the running binary directory, `/opt/homebrew/bin`, and `/usr/local/bin`.
- Moves PATH repair before early tmux setup so launchd-started daemons can resolve `tmux`, `gt`, and `bd` consistently.
- Adds coverage for the launchd/Homebrew PATH candidates.

## Triage
- Duplicate check: #875 was closed by merged PR #876, but that only inherited the daemon's existing environment and did not add `/opt/homebrew/bin` under launchd.
- PR check: no open or merged PR found specifically for #3738 or launchd `/opt/homebrew/bin` PATH recovery.
- Internal same-repo PR #3953 was auto-closed by policy; this PR is from a fork as required.

## Testing
- `go test -c ./internal/daemon -o /tmp/opencode/daemon.test`
- `go test ./internal/daemon -run TestDaemonPathCandidatesIncludesLaunchdToolDirs -count=1` currently fails before running tests because `internal/daemon.TestMain` panics in testcontainers with `rootless Docker not found` in this environment.

### DIFF
--- a/internal/daemon/daemon.go
+++ b/internal/daemon/daemon.go
@@ -153,6 +153,56 @@ const beadsModulePath = "github.com/steveyegge/beads"
 
 var semverPattern = regexp.MustCompile(`v?(\d+\.\d+\.\d+)`)
 
+func daemonPathCandidates(home, exePath string) []string {
+	candidates := make([]string, 0, 5)
+	if exePath != "" {
+		candidates = append(candidates, filepath.Dir(exePath))
+	}
+	if home != "" {
+		candidates = append(candidates,
+			filepath.Join(home, ".local/bin"),
+			filepath.Join(home, "bin"),
+		)
+	}
+	return append(candidates,
+		"/opt/homebrew/bin",
+		"/usr/local/bin",
+	)
+}
+
+func augmentDaemonPath(logger *log.Logger) {
+	exePath := ""
+	if exe, err := os.Executable(); err == nil {
+		exePath = exe
+	}
+	extras := daemonPathCandidates(os.Getenv("HOME"), exePath)
+	if len(extras) == 0 {
+		return
+	}
+
+	current := os.Getenv("PATH")
+	parts := strings.Split(current, string(os.PathListSeparator))
+	seen := make(map[string]struct{}, len(parts)+len(extras))
+	for _, p := range parts {
+		seen[p] = struct{}{}
+	}
+	augmented := parts
+	for _, extra := range extras {
+		if _, ok := seen[extra]; ok {
+			continue
+		}
+		if info, statErr := os.Stat(extra); statErr == nil && info.IsDir() {
+			augmented = append([]string{extra}, augmented...)
+			seen[extra] = struct{}{}
+		}
+	}
+	newPath := strings.Join(augmented, string(os.PathListSeparator))
+	if newPath != current {
+		_ = os.Setenv("PATH", newPath)
+		logger.Printf("PATCH-007: augmented daemon PATH with user/local bin dirs (was=%q, now=%q)", current, newPath)
+	}
+}
+
 // New creates a new daemon instance.
 func New(config *Config) (*Daemon, error) {
 	// Ensure daemon directory exists
@@ -172,6 +222,12 @@ func New(config *Config) (*Daemon, error) {
 
 	logger := log.New(logWriter, "", log.LstdFlags)
 	ctx, cancel := context.WithCancel(context.Background())
+
+	// PATCH-007 (hq-olcb): Augment PATH with common user/local bin
+	// directories before any subprocess lookup. The daemon is often launched
+	// from systemd / login shells / launchd without user-installed tool dirs
+	// such as ~/.local/bin or /opt/homebrew/bin.
+	augmentDaemonPath(logger)
 
 	// Initialize session prefix and agent registries from town root.
 	if err := session.InitRegistry(config.TownRoot); err != nil {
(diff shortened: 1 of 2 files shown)


Bella-Giraffety:
CI triage after fork PR creation:

- `Test` and `Windows Smoke Test` passed on the fork PR run.
- `Lint` failed on an unrelated pre-existing misspell in `internal/polecat/manager.go:526` (`behaviour` in a comment). This PR does not touch that file.
- `Integration Tests` failed on unrelated infrastructure/tooling issues: repeated `Dolt port 32769 is already in use` during `gt install`, plus `bd close --tombstone` rejected by the installed `bd` CLI. This PR only changes daemon PATH setup and its unit coverage.
- Recent `main` CI runs are already failing (latest checked run on `2eafac97` is failure), so the red checks do not appear introduced by this patch.

Bella-Giraffety:
PR sheriff review update:

- Current `main` still needs this: daemon PATH recovery on `main` did not include `/opt/homebrew/bin`, and it ran after early daemon setup that can touch subprocess-dependent paths.
- This PR was stale and conflicted with current `origin/main` in `internal/daemon/daemon.go` and `internal/daemon/daemon_test.go`; I merged `origin/main` into the PR branch and resolved those conflicts.
- The updated branch is now GitHub-mergeable, and the PR diff remains scoped to daemon PATH recovery plus unit coverage.
- Local verification: `go test -c ./internal/daemon -o /tmp/opencode/daemon-pr3954.test` passes. `go test ./internal/daemon -run TestDaemonPathCandidatesIncludesLaunchdToolDirs -count=1` still fails before running tests because `internal/daemon.TestMain` panics in testcontainers with `rootless Docker not found` in this environment.
- Labels applied per #1400: `status/accepted`, `priority/p0`, `kind/bug`.

Recommendation: squash-merge after the new CI run completes, or waive the existing repo-wide CI infrastructure failures if they match the prior unrelated lint/integration failures.

PR #1470, opened 2026-02-15
fix(formula): add squash/create-wisp/hook cycle to deacon loop-or-exit

seanbearden:
## Summary

Fixes #1371 — deacon patrol formula's `loop-or-exit` step was missing the squash/create-wisp/hook cycle.

**Problem:** The deacon's loop step just said "Return to inbox-check step" without squashing the current wisp and creating a new one. This contradicts:
1. `prime_molecule.go` WorkLoopSteps (which already document the correct cycle)
2. Witness and refinery formulas (which both include the full cycle)

Without this cycle, closed step beads accumulate, `bd ready` returns nothing, and `findActivePatrol` can't find the wisp on restart.

**Fix:** Updated the deacon `loop-or-exit` step description to match the witness/refinery pattern:
1. Squash: `bd mol squash <mol-id> --summary "<summary>"`
2. Create: `bd mol wisp mol-deacon-patrol --json`
3. Hook: `bd update "$NEW_WISP" --status=hooked --assignee=deacon`

Also synced the `.beads/formulas/` copy (local, gitignored).

## Test plan

- [x] `go test -race ./internal/formula/...` passes
- [x] `go vet ./internal/formula/...` clean
- [x] `TestPatrolFormulasHaveSquashCycle` — regression test verifies all 3 patrol formulas (witness, deacon, refinery) contain squash/create-wisp/hook instructions in their loop step
- [x] `TestPatrolFormulasHaveBackoffLogic` — existing test still passes

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #3957, opened 2026-05-13
fix(deacon): block patrol cycle batching

### LINKED ISSUE #2634: Deacon (Sonnet) ignores anti-loop instructions, batches patrol cycles
## Summary

The deacon agent (running on `claude-sonnet`) ignores explicit formula instructions not to loop and batches patrol cycles into `for i in $(seq 1 500)` constructs.

## Reproduction

1. Deacon formula (`mol-deacon-patrol`) final step says "execute one cycle then exit"
2. Template (`deacon.md.tmpl`) has an explicit ANTI-BATCH RULE: "NEVER use for/while/seq"
3. Sonnet reads the instructions, executes one proper cycle, then rationalizes batching as "efficient" and launches `for i in $(seq 2 50)`
4. Even after removing all loop language from the formula (v15, exit-only), a fresh session announces "Running batches of 400 cycles" before reading the formula steps

## Root cause

Sonnet pattern-matches "patrol loop" from the role template and defaults to batching regardless of formula instructions. The anti-loop text is treated as advisory. The model actively rationalizes violating the constraint ("context is fresh so running efficient batches").

Likely needs structural enforcement in Go rather than prompt-level guardrails.

Bella-Giraffety:
## Summary
- Fixes #2634 by adding Deacon-only `PreToolUse` guards that block `for ... seq` and open-ended `while` Bash patrol batching.
- Tightens the Deacon role template to describe one patrol wisp at a time, with scheduling delegated to `gt patrol report`/daemon instead of prompting an LLM-managed loop.
- Adds regression coverage that the Deacon hook config includes the anti-batch guards.

## Triage
- Duplicate issue check found only #2634 for this exact Sonnet anti-loop batching report.
- PR check found related patrol-loop fixes (#1470, #2812, and other deacon health PRs), but none add a structural guard against `for i in $(seq ...)` patrol batching.
- Current `main` remains susceptible from code inspection: Deacon prompt/template still described a patrol loop and only existing Deacon hooks block persistent patrol molecules.

## Testing
- `go test ./internal/hooks ./internal/formula`
- `go build ./cmd/gt`

### DIFF
--- a/internal/hooks/config.go
+++ b/internal/hooks/config.go
@@ -284,6 +284,27 @@ func DefaultOverrides() map[string]*HooksConfig {
 		"deacon": {
 			PreToolUse: []HookEntry{
 				{
+					Matcher: "Bash(*for *seq*)",
+					Hooks: []Hook{{
+						Type:    "command",
+						Command: "echo '❌ BLOCKED: Deacon must not batch patrol cycles with for/seq loops.' && echo 'Run one patrol cycle, then use gt patrol report or gt handoff.' && exit 2",
+					}},
+				},
+				{
+					Matcher: "Bash(*while true*)",
+					Hooks: []Hook{{
+						Type:    "command",
+						Command: "echo '❌ BLOCKED: Deacon must not run open-ended patrol loops.' && echo 'Run one patrol cycle, then use gt patrol report or gt handoff.' && exit 2",
+					}},
+				},
+				{
+					Matcher: "Bash(*while :*)",
+					Hooks: []Hook{{
+						Type:    "command",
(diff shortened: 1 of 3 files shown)


Bella-Giraffety:
CI triage after fork PR creation:

- `Test` and `Windows Smoke Test` passed on the fork PR run.
- `Lint` failed on the same unrelated pre-existing misspell in `internal/polecat/manager.go:526` (`behaviour` in a comment) that also failed recent PR #3954. This PR does not touch that file.
- `Integration Tests` failed on unrelated infrastructure/tooling issues: repeated `Dolt port 32769 is already in use` during `gt install`, plus `bd close --tombstone` rejected by the installed `bd` CLI. These match the failures observed on #3954 and are outside this Deacon anti-batch change.
- Recent `main` CI runs are already failing, so the red checks do not appear introduced by this patch.

Issue #3949, opened 2026-05-12
Integration tests fail from Dolt port reuse and unsupported bd tombstone flag


Bella-Giraffety:
Several PR runs are failing the Integration Tests job for reasons unrelated to branch changes.

Observed on PR #3947 and PR #3948:

- `TestBeadsDbInitAfterClone/*` fails with `Error: Dolt port 32769 is already in use`.
- `TestAgentWorktreesStayClean/*` fails with `Error: Dolt port 32769 is already in use`.
- `TestScheduleBead_RefusesTombstone` fails because `bd close --tombstone` is not supported by the installed bd CLI.

The failures appear to be CI/test harness drift rather than PR-specific behavior. Relevant failed job: https://github.com/gastownhall/gastown/actions/runs/25753678063/job/75636381628

PR #3964, opened 2026-05-13
fix(rig): detect dead tmux panes in status and start

### LINKED ISSUE #3146: gt rig status reports running for dead tmux panes
## Bug

`gt rig status` shows agents as "already running" when the tmux session exists but the pane is dead (exit code 1). This means dead agents go undetected.

## Example

Reli refinery died at 02:52 (Gemini CLI AbortError/quota). The tmux session `re-refinery` still existed with a dead pane. `gt rig start reli` reported "Skipped: refinery (already running)" — but it wasn't running.

## Expected

`gt rig status` and `gt rig start` should check if the tmux pane is alive (`tmux list-panes -t <session> -F '#{pane_dead}'`), not just if the session exists. If the pane is dead, the agent should be reported as stopped and `gt rig start` should restart it.

## Impact

Dead agents go undetected until a human or the next mayor session manually checks. The health loop and witness can't catch what the status command doesn't report.

## Kaizen

The witness should also periodically check pane health for its rig's agents and auto-restart dead ones.

Bella-Giraffety:
## Summary
- Treat retained tmux sessions with dead/non-agent panes as stopped in `gt rig status` for polecats and crew.
- Let `gt rig boot`, `gt rig start`, and restart start phases delegate to witness/refinery managers so zombie sessions are recreated instead of skipped as already running.
- Add a regression test that creates a retained `pane_dead=1` tmux session and verifies it is not reported healthy.

## Triage
- Fixes #3146.
- Duplicate/open PR checks: no PR found for #3146, \"gt rig status reports running\", or \"dead tmux panes\".

## Tests
- `go test ./internal/cmd -run 'TestIsAgentSessionHealthy_DeadPane|TestFindRigSessions|TestIsGitRemoteURL' -count=1`
- `go test ./internal/tmux -run 'TestCheckSessionHealth|TestEnsureSessionFresh|TestZombieStatus' -count=1`
- `go build ./cmd/gt`

## Notes
- `go test ./internal/cmd` and `go test ./internal/cmd ./internal/tmux ./internal/witness ./internal/refinery` both timed out after 5 minutes without output; targeted changed-path tests passed.

### DIFF
--- a/internal/cmd/rig.go
+++ b/internal/cmd/rig.go
@@ -296,31 +296,31 @@ Examples:
 
 // Flags
 var (
-	rigAddPrefix       string
-	rigAddLocalRepo    string
-	rigAddBranch       string
-	rigAddPushURL      string
-	rigAddUpstreamURL  string
-	rigAddAdopt           bool
+	rigAddPrefix         string
+	rigAddLocalRepo      string
+	rigAddBranch         string
+	rigAddPushURL        string
+	rigAddUpstreamURL    string
+	rigAddAdopt          bool
 	rigAddAdoptURL       string
 	rigAddAdoptForce     bool
 	rigAddFilter         string
 	rigAddSparseCheckout []string
-	rigResetHandoff    bool
-	rigResetMail       bool
-	rigResetStale      bool
-	rigResetDryRun     bool
-	rigResetRole       string
-	rigShutdownForce   bool
-	rigShutdownNuclear bool
-	rigRebootForce     bool
-	rigRebootNuclear   bool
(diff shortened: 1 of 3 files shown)


Bella-Giraffety:
Pushed ca2bfd14 to fix the hosted lint failure (repo-wide misspell in internal/polecat/manager.go). Current CI status after rerun: Lint, Test, Windows Smoke, and static guard checks pass. Integration Tests still fail with the same unrelated failures tracked in #3949: Dolt port 32769 reuse and unsupported `bd close --tombstone`.

PR #3966, opened 2026-05-13
test: cover Dolt-first wisp checks

### LINKED ISSUE #2003: Doctor checks read issues.jsonl with no Dolt-first path (2 checks are no-ops post-migration)
## Problem

After migration to Dolt, 2 doctor checks in gastown still read `issues.jsonl` directly with **no Dolt-first alternative**. Since nothing exports Dolt data back to JSONL, these checks scan an empty or stale file and silently report all-clear.

## Affected checks

1. **`findMisclassifiedWisps`** in `internal/doctor/misclassified_wisp_check.go:124` — scans JSONL for wisps missing the ephemeral flag
2. **`getWispIDs`** in `internal/cmd/ready.go:381` — reads JSONL to build a wisp exclusion set for `gt ready` output

## Why this is a problem

- **No code exports Dolt → JSONL.** The only writes to `issues.jsonl` are: `InitBeads` (creates empty file) and `createAgentBeadViaJSONL` (last-resort fallback when `bd create` fails twice).
- `getGitActivity` in beads `status.go:148` is dead code that confirms this: *"Previously calculated from git log of issues.jsonl; now returns nil as activity tracking has moved to Dolt-native queries."*
- These checks return nil on an empty file, masking real problems that exist only in Dolt.
- Notably, other gastown checks already have proper Dolt-first paths: `checkStuckWispsDolt` in `patrol_check.go` and `countAbandonedWisps` in `wisp_check.go` both query Dolt first and fall back to JSONL.

## Suggested fix

Port these 2 checks to query Dolt first, falling back to JSONL only if the database is unavailable — matching the pattern in `patrol_check.go` and `wisp_check.go`.

## Related

- Companion issue: https://github.com/steveyegge/beads/issues/2101 (4 checks in beads with the same problem)

Bella-Giraffety:
## Summary
- Add regression coverage that the misclassified-wisps doctor check does not use stale issues.jsonl when Dolt is unavailable.
- Add regression coverage that gt ready gets wisp IDs from `bd mol wisp list --json` instead of issues.jsonl.

## Test plan
- `go test ./internal/doctor -run TestRunIgnoresJSONLWhenDoltUnavailable -count=1`
- `go test ./internal/cmd -run TestGetWispIDsUsesBdMolWispList -count=1`

Closes #2003

### DIFF
--- a/internal/cmd/ready_test.go
+++ b/internal/cmd/ready_test.go
@@ -135,6 +135,35 @@ func TestFilterFormulaScaffolds_EmptyIssues(t *testing.T) {
 	}
 }
 
+func TestGetWispIDsUsesBdMolWispList(t *testing.T) {
+	beadsPath := t.TempDir()
+	if err := os.WriteFile(filepath.Join(beadsPath, "issues.jsonl"), []byte(`{"id":"stale-jsonl-wisp"}`+"\n"), 0644); err != nil {
+		t.Fatal(err)
+	}
+
+	binDir := t.TempDir()
+	bdPath := filepath.Join(binDir, "bd")
+	bdScript := `#!/bin/sh
+if [ "$1" = "mol" ] && [ "$2" = "wisp" ] && [ "$3" = "list" ] && [ "$4" = "--json" ]; then
+  printf '{"wisps":[{"id":"dolt-wisp-1"},{"id":"dolt-wisp-2"}],"count":2}\n'
+  exit 0
+fi
+exit 1
+`
+	if err := os.WriteFile(bdPath, []byte(bdScript), 0755); err != nil {
+		t.Fatal(err)
+	}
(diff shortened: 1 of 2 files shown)


Bella-Giraffety:
Sheriff review: no findings in the PR diff. Current main still has the Dolt-first behavior, but it lacks these regression tests, so this test-only PR is still useful. Targeted verification passed locally: go test ./internal/doctor -run TestRunIgnoresJSONLWhenDoltUnavailable -count=1; go test ./internal/cmd -run TestGetWispIDsUsesBdMolWispList -count=1. Existing red CI appears unrelated: lint reports internal/polecat/manager.go:526 spelling, and integration failures are Dolt port/bd CLI harness issues outside this diff. Recommendation: squash-merge when maintainers are comfortable accepting the unrelated red CI, or rerun after the existing test-harness issues are fixed.

PR #3975, opened 2026-05-13
fix(polecat): verify beacon delivery for hook+prompt agents on fresh spawn

pbazydlo:
## Summary

- Fresh Claude Code spawns show the splash screen with `❯ ` at the bottom, causing `WaitForRuntimeReady` to return prematurely and leaving the beacon CLI arg pre-filled but not auto-submitted
- `verifyStartupNudgeDelivery` was only called for Mode B agents (startup nudge path), leaving Mode A (hook+prompt / Claude) agents with no retry safety net
- Add Mode A verification: when neither `SendBeaconNudge` nor `SendStartupNudge` applies, call `verifyStartupNudgeDelivery` to re-deliver work instructions via nudge if the agent is still idle after startup

## Changes

- `internal/polecat/session_manager.go`: call `verifyStartupNudgeDelivery` for hook+prompt agents after the existing Mode B block
- `internal/polecat/session_manager_test.go`: add `TestModeABeaconVerificationCondition` with three cases covering Claude/non-hook/hook-noprompt combinations

## Test plan

- [x] All polecat tests pass (`go test ./internal/polecat/...`)
- [x] Mode A condition (Claude hook+prompt) triggers verification
- [x] Mode B (non-hook) and Mode C (hook, no prompt) are unaffected

Ref: hi-y44

### DIFF
--- a/internal/polecat/session_manager.go
+++ b/internal/polecat/session_manager.go
@@ -575,9 +575,11 @@ func (m *SessionManager) Start(polecat string, opts SessionStartOptions) error {
 	// Fresh spawns may show the Claude Code splash screen with the CLI beacon
 	// pre-filled but not auto-submitted. If the agent is still idle after startup,
 	// re-deliver the work instructions via nudge to kick it into action.
-	// Non-fatal: if verification fails the witness zombie patrol will recover.
+	// Runs asynchronously: verifyStartupNudgeDelivery sleeps before checking, so a
+	// synchronous call would add ~25s to every successful polecat startup on the
+	// common gt sling path. Non-fatal: the witness zombie patrol handles unrecovered stalls.
 	if !fallbackInfo.SendBeaconNudge && !fallbackInfo.SendStartupNudge {
-		m.verifyStartupNudgeDelivery(sessionID, runtimeConfig, startupNudgeContent)
+		go m.verifyStartupNudgeDelivery(sessionID, runtimeConfig, startupNudgeContent)
 	}
 
 	// Legacy fallback for other startup paths (non-fatal)
--- a/internal/polecat/session_manager.go
+++ b/internal/polecat/session_manager.go
@@ -571,6 +571,15 @@ func (m *SessionManager) Start(polecat string, opts SessionStartOptions) error {
 		m.verifyStartupNudgeDelivery(sessionID, runtimeConfig, verifyContent)
 	}
 
+	// Verify beacon delivery for hook+prompt agents (Mode A, hi-y44).
+	// Fresh spawns may show the Claude Code splash screen with the CLI beacon
+	// pre-filled but not auto-submitted. If the agent is still idle after startup,
+	// re-deliver the work instructions via nudge to kick it into action.
+	// Non-fatal: if verification fails the witness zombie patrol will recover.
+	if !fallbackInfo.SendBeaconNudge && !fallbackInfo.SendStartupNudge {
+		m.verifyStartupNudgeDelivery(sessionID, runtimeConfig, startupNudgeContent)
+	}
+
 	// Legacy fallback for other startup paths (non-fatal)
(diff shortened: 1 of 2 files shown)


Bella-Giraffety:
Thanks for the focused fix. I reviewed this during PR sheriff and found one blocking concern before this should merge:

- `verifyStartupNudgeDelivery` currently sleeps for the verify delay before checking idleness, so adding it to the normal hook+prompt/Claude path appears to add ~25s to every successful Claude polecat startup. That is likely too expensive for the common `gt sling` path.

Could you make the verification non-blocking, or otherwise avoid delaying successful startups? A stronger regression test that exercises `SessionManager.Start` invoking the recovery path would also help; the current test only validates the boolean condition.

Holding this for now, but the PR is small and does not look duplicated by another open PR.

Bella-Giraffety (changes requested):
Findings:

- **Blocking:** `internal/polecat/session_manager.go:579-580` adds `verifyStartupNudgeDelivery` to the normal hook+prompt path, but `verifyStartupNudgeDelivery` sleeps at `internal/polecat/session_manager.go:929-931` before checking whether the agent is idle. That makes every successful Claude hook+prompt polecat startup wait for `StartupNudgeVerifyDelayD()` before `Start` can finish, even when the beacon was delivered correctly. This is on the common `gt sling` path and is a stability/throughput regression. Please make the Mode A verification asynchronous or otherwise avoid blocking successful startups.

Recommendation: request changes. The issue looks real and the targeted condition test passes, but this should not merge until the startup path remains non-blocking or has equivalent latency protection.

Verification run:

- `go test ./internal/polecat -run 'TestModeABeaconVerificationCondition|TestVerifyStartupNudgeDelivery_NilConfig|TestPromptlessFallbackIncludesPrimeAndWorkInstructions'`

pbazydlo:
Thanks for the thorough review. Both issues addressed:

**Blocking fix**: Added `go` to the Mode A call in `Start` (`session_manager.go:579`). The comment now explains why: `verifyStartupNudgeDelivery` sleeps before its first idle check, so a synchronous call adds ~25s to every successful startup on the common `gt sling` path. The goroutine is safe to fire-and-forget — it's a non-fatal safety net and the witness zombie patrol handles any surviving stalls.

**Regression test**: Added `TestModeAStartupVerifyIsNonBlocking` (tmux-required). It replicates what `Start` does on the Mode A path — launches `verifyStartupNudgeDelivery` as a goroutine for a hook+prompt agent with an idle session — and asserts the caller returns in <500ms while the goroutine completes the full idle-retry loop. Output from a passing run:
```
[startup-nudge] attempt 1/2: agent gt-test-modeA-1 idle at prompt, retrying nudge
[startup-nudge] attempt 2/2: agent gt-test-modeA-1 idle at prompt, retrying nudge
[startup-nudge] WARNING: agent gt-test-modeA-1 still idle after 2 nudge retries
--- PASS: TestModeAStartupVerifyIsNonBlocking (56.97s)
```

Also rebased onto current `main` (branch was 82 commits behind).

Bella-Giraffety (approved):
Post-update sheriff review: no remaining blockers found. The Mode A verification is now launched asynchronously, so it no longer adds the startup verify delay to the common hook+prompt Claude path, and the condition stays scoped to agents with hooks and prompt support. Current main still lacks this Mode A safety net, and I did not find an open or recently merged duplicate that supersedes it.

Verification:
- PASS: go test ./internal/polecat -run 'TestModeABeaconVerificationCondition|TestModeAStartupVerifyIsNonBlocking|TestVerifyStartupNudgeDelivery_NilConfig|TestPromptlessFallbackIncludesPrimeAndWorkInstructions'
- FULL PACKAGE BLOCKED: go test ./internal/polecat/... fails in TestAddWithOptions_NoPrimeMDCreatedLocally because this environment lacks rootless Docker for testcontainers; failure is unrelated to this PR path.

Recommendation: squash-merge.

PR #3977, opened 2026-05-13
Fix daemon legacy tmux socket cleanup

### LINKED ISSUE #3570: gt daemon doesn't clean up legacy tmux sockets on upgrade → dual agents (split-brain)
## Summary

Upgrading `gt` from v0.12.1 to v1.0.0 changes the tmux socket naming strategy, but the daemon does not automatically clean up sessions on the legacy socket. This produces split-brain dual-agent conditions (dual mayor, dual deacon, duplicated witnesses/refineries) whenever a user upgrades the binary and restarts the daemon without running `gt down && gt up`.

## Root cause

In v0.12.1 (`internal/session/registry.go:131`):
```go
case "", "default", "auto":
    socket = ""  // empty → use default tmux socket (no -L flag)
```

All tmux sessions created by v0.12.1 live at `/tmp/tmux-<UID>/default`.

In v1.0.0 (`internal/session/registry.go:131`):
```go
case "", "default", "auto":
    socket = townSocketName(townRoot)  // → "gt-<sha256[:6]>"
```

New sessions go to `/tmp/tmux-<UID>/gt-<hash>`.

The daemon's `ensureDeaconRunning` (`internal/daemon/daemon.go:893`) calls `deacon.Manager.Start`, which calls `tmux.HasSession` with the **new** socket only — it never checks the legacy default socket. Result: daemon thinks the deacon is missing and spawns a new one, while the old deacon is still alive on the default socket.

This produces dual mayor, dual deacon, duplicated witnesses/refineries across both sockets. Each agent competes on shared state (`state.json`, dolt, PID files).

## Existing cleanup code is not reached

`internal/cmd/down.go:348-358` calls `cleanupLegacyDefaultSocket()` and `cleanupLegacyBaseSocket()` which correctly clean ghost sessions on the legacy sockets. **But these are only invoked from `gt down`.** A user who upgrades the binary and restarts the daemon (without running `gt down && gt up`) never reaches this code path.

## Suggested fix

**Option 1 (preferred):** Call `cleanupLegacyDefaultSocket()` + `cleanupLegacyBaseSocket()` from `ensureDeaconRunning` or daemon startup, not only from `gt down`. Guard with a `sync.Once` so it only runs at most once per daemon lifetime.

**Option 2:** At `gt up` or binary version change detection, emit a warning and require the user to run `gt down` first.

## Reproduction

1. Install `gt` v0.12.1, start the town.
2. Upgrade to v1.0.0.
3. Restart the daemon (`gt daemon stop && gt daemon start`).
4. Observe: dual mayor, dual deacon, duplicated witnesses/refineries across two tmux sockets.

## Workaround

```bash
tmux -S /tmp/tmux-<UID>/default kill-server
rm /tmp/tmux-<UID>/default
```

## Related

The dual-agent condition this produces can recur even after cleanup if a self-handoff re-triggers the spawn race. Worth investigating whether the handoff queue has a similar legacy-socket blindness.

## Environment

- gt v1.0.0 (upgraded from v0.12.1)
- macOS Darwin 25.4.0

Bella-Giraffety:
## Summary
- Move legacy tmux socket cleanup into `internal/session` so daemon startup and `gt down` share the same implementation.
- Run cleanup after daemon startup has acquired the lock and passed fatal preflight, before patrol agents can be spawned on the current socket.
- Tighten cleanup targets for automatic startup cleanup and add repeat passes to catch watchdog respawns.

Fixes #3570

## Tests
- `go test ./internal/session`
- `go test -c ./internal/daemon`

## Notes
- `go test ./internal/daemon -run TestCleanupLegacySocketSessionsRunsOnce` cannot run in this environment because `internal/daemon` package `TestMain` panics before selected tests when rootless Docker/testcontainers is unavailable.

### DIFF
--- a/internal/cmd/down.go
+++ b/internal/cmd/down.go
@@ -796,130 +796,29 @@ func findOrphanedClaudeProcesses(townRoot string) []int {
 	return orphaned
 }
 
-// legacySocketTmux is the subset of tmux.Tmux used by the legacy socket
-// cleanup functions, extracted to allow test injection.
-type legacySocketTmux interface {
-	ListSessions() ([]string, error)
-	KillSessionWithProcesses(name string) error
-}
-
-// Test hooks — nil in production, set by tests to avoid real tmux calls.
-var (
-	legacyTmuxForTest   func(socket string) legacySocketTmux
-	legacySocketForTest func() string // overrides tmux.GetDefaultSocket()
-)
-
-func getDefaultSocket() string {
-	if legacySocketForTest != nil {
-		return legacySocketForTest()
-	}
-	return tmux.GetDefaultSocket()
-}
-
-func newLegacyTmux(socket string) legacySocketTmux {
-	if legacyTmuxForTest != nil {
-		return legacyTmuxForTest(socket)
-	}
-	return tmux.NewTmuxWithSocket(socket)
-}
-
 // cleanupLegacyDefaultSocket removes Gas Town sessions left on the "default"
 // tmux socket by old binaries. Returns the number of sessions cleaned.
 func cleanupLegacyDefaultSocket() int {
-	currentSocket := getDefaultSocket()
-	if currentSocket == "" || currentSocket == "default" {
-		return 0 // Already on the default socket, nothing to clean up
-	}
-
-	legacyTmux := newLegacyTmux("default")
-	sessions, err := legacyTmux.ListSessions()
-	if err != nil {
-		return 0 // No server on default socket
-	}
-
-	var cleaned int
-	for _, sess := range sessions {
-		if session.IsKnownSession(sess) {
-			if err := legacyTmux.KillSessionWithProcesses(sess); err == nil {
-				cleaned++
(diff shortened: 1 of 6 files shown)


Bella-Giraffety:
CI status after rerun:

- Passing: Lint, Test, Windows Smoke Test, Reject go.mod replace directives, Reject issues.jsonl.
- Failing: Integration Tests only.

The Integration failures match current main-branch CI failures and appear unrelated to this PR:

- Multiple `gt install` integration tests fail with `Dolt port 32769 is already in use`.
- `TestScheduleBead_RefusesTombstone` fails because the installed `bd close` does not recognize `--tombstone`.

I reran the failed job once; it failed the same way. Local verification for this branch:

- `go test ./internal/session`
- `go test -c ./internal/daemon`

Full daemon tests cannot run locally here because `internal/daemon` package `TestMain` panics before selected tests when rootless Docker/testcontainers is unavailable.

PR #3982, opened 2026-05-14
Add gt hook ack command

### LINKED ISSUE #2635: gt hook ack — polecat-side work acknowledgment
## Problem

When a polecat is slung work, the infra sets state to "working" immediately on session start (`polecat_spawn.go`). But the agent hasn't actually parsed its hook yet — it still needs to boot, run `gt prime`, read `gt hook`, and understand the formula before real work begins.

If the agent crashes or gets confused during startup, the state still says "working" and the witness assumes everything is fine. There's no signal from the agent itself confirming it understood its assignment.

## Proposal

Add `gt hook ack` — an explicit command the polecat runs after reading its hook to confirm:
- It knows what bead/molecule is hooked
- It understands the formula/assignment
- It's starting execution now

This would be logged as a timestamped event on the bead, giving infra and witnesses a reliable "agent actually started" signal distinct from "session started."

## Why

- Closes the gap between session-alive and work-actually-happening
- Witnesses can distinguish "booting" from "working" without guessing
- Creates an audit trail: hook time → ack time → completion time
- Enables timeout detection on the startup phase (hooked but never acked)

Bella-Giraffety:
## Summary
- Add `gt hook ack` as a first-class acknowledgment command for hooked work.
- Record a timestamped `hook_ack` note on the current hooked/in-progress bead.
- Emit a `hook_ack` activity event with bead, molecule, formula, and session metadata.

## Verification
- `go build ./cmd/gt`
- `go test ./internal/cmd -run TestFormatHookAckNote -count=1`
- `go test ./internal/events`
- `go test ./internal/cmd -run 'TestHook|TestNormalizeHookShowTarget' -count=1`

Full `go test ./...` was attempted but did not pass in this sandbox due unrelated/environmental failures: rootless Docker unavailable, Dolt circuit-breaker/timeouts, and tmux command-name expectations (`coreutils` vs `sleep`).

Closes #2635

### DIFF
--- a/internal/cmd/hook.go
+++ b/internal/cmd/hook.go
@@ -38,6 +38,7 @@ you left off.
 
 Examples:
   gt hook                                    # Show what's on my hook
+  gt hook ack                                # Acknowledge current hooked work
   gt hook status                             # Same as above
   gt hook gt-abc                             # Attach issue gt-abc to your hook
   gt hook gt-abc -s "Fix the bug"            # With subject for handoff mail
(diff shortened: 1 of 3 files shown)


Bella-Giraffety:
This looks like a feature/enhancement rather than a bug fix. Per current sweep policy, please do not merge without explicit maintainer approval. Bug/stability fixes can continue through normal review, but new commands/features need separate maintainer sign-off.

Bella-Giraffety:
Closing this owned enhancement PR during PR burndown cleanup. We are not merging new features without explicit maintainer intent, and this can be re-filed later if accepted.

PR #4004, opened 2026-05-14
Fix dashboard rig options

fengning-starsend:
## Summary

- Use `gt rig list --json` when populating dashboard rig options.
- Keep the existing text-output parser as a fallback for compatibility.
- Add focused tests for JSON parsing and `/api/options?type=rigs` behavior.

## Why

The dashboard ready-work Sling dropdown fetches `/api/options?type=rigs`. The ready panel can render work from rigs, but the Sling dropdown cannot offer target rigs unless this endpoint returns rig names. The JSON rig list is the stable machine-readable surface for that data.

## Testing

- `go test ./internal/web -run 'TestParseRigListJSON|TestHandleOptionsUsesRigListJSON'`
- `go test ./internal/web`

### DIFF
--- a/internal/web/api.go
+++ b/internal/web/api.go
@@ -725,12 +725,19 @@ func (h *APIHandler) handleOptions(w http.ResponseWriter, r *http.Request) {
 	// Fetch rigs
 	go func() {
 		defer wg.Done()
-		if output, err := h.runGtCommand(r.Context(), 3*time.Second, []string{"rig", "list"}); err == nil {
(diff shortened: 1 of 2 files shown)


Bella-Giraffety (approved):
Reviewed during PR sheriff. The change switches rig options to the stable JSON surface, keeps the text parser as fallback, and adds focused coverage for JSON parsing and the options endpoint. Local validation passed: go test ./internal/web and the focused tests. CI rerun is green.

PR #4008, opened 2026-05-14
fix(stuck-agent-dog): skip restart loops when arrays are empty (gt-05d)

adensur:
## Summary
- Guards `${CRASHED[@]:-}` / `${STUCK[@]:-}` loops in `stuck-agent-dog/run.sh` against the bash empty-array-with-default-fallback gotcha (one phantom iteration over the empty string), which caused dogs to emit `RESTART_POLECAT: /` notifications every 5min gate firing.
- Replaces `:-` fallback with `${arr[@]+"${arr[@]}"}` proper empty-array guard.

Fixes gastownhall/gastown bead `gt-05d`. Reported by gastown/witness; escalated by deacon as high-noise (alpha + bravo dogs both affected).

## Test plan
- [ ] Trigger a gate firing with no CRASHED/STUCK candidates — confirm no notification
- [ ] Trigger with one candidate — confirm exactly one RESTART_POLECAT

### DIFF
--- a/plugins/stuck-agent-dog/run.sh
+++ b/plugins/stuck-agent-dog/run.sh
@@ -135,7 +135,10 @@ fi
 # --- Take action --------------------------------------------------------------
 
 # Crashed polecats: notify witness to restart
-for ENTRY in "${CRASHED[@]:-}"; do
+# Note: `"${arr[@]:-}"` expands an empty array to a single empty string under
+# `set -u`, which would fire a phantom `RESTART_POLECAT: /` notification. The
+# `${arr[@]+"${arr[@]}"}` form expands to nothing when the array is empty.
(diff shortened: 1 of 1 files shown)


Bella-Giraffety (approved):
Reviewed as PR sheriff for gt-pr4008. Recommendation: squash-merge. origin/main still has the buggy "${ARRAY[@]:-}" loop form; under bash nounset that expands an empty declared array to one empty element and can emit phantom RESTART_POLECAT notifications. I verified the PR form expands to zero iterations for empty arrays and preserves non-empty entries, including entries containing spaces. Targeted checks run: bash -n plugins/stuck-agent-dog/run.sh on main, plus bash -uc snippets for current vs PR loop behavior. shellcheck is unavailable in this worktree. Note: plain "${ARRAY[@]}" would also be safe here because both arrays are declared, but this patch is behaviorally correct and small.

PR #4011, opened 2026-05-14
fix: annotate max_polecats with deferred dispatch state in gt rig config show

### LINKED ISSUE #3866: ux: max_polecats activates capacity scheduler with no visibility — gt rig config show should annotate deferred dispatch state
## Summary

The capacity scheduler uses \`max_polecats > 0\` as the predicate for deferred dispatch mode. The system default is \`10\`, so deferred dispatch is active on every rig out of the box — but nothing in \`gt rig config show\`, the CLI help, or the changelog tells users this. There is no way to tell at a glance whether deferred dispatch is active or what \`max_polecats\` controls.

## Repro

```bash
gt rig config show <rig> --layers
# max_polecats   10   system
# ↑ no annotation, no mention of deferred dispatch
```

## Expected

- \`gt rig config show\` (or \`--layers\`) should annotate \`max_polecats\` to indicate whether deferred dispatch is currently active, e.g.:
  ```
  max_polecats   10   system   [deferred dispatch: ON — set to -1 to disable]
  ```
- Help text or docs should explain the relationship between \`max_polecats\` and the capacity scheduler (introduced in v0.8.0).

## Context

The scheduler predicate was confirmed in the commit message for \`fix(scheduler): guard scheduleBead against closed/tombstone beads\` (v1.1.0, #3840). The behaviour itself may be correct and intentional — the gap is purely visibility and documentation.

## Status

Not present in v1.1.0.

## Related

- #3840 — ghost convoy growth in deferred dispatch mode
- #3865 — \`gt rig config set\` writes ephemeral values silently

esciara:
## Summary

- `gt rig config show` now appends an annotation to `max_polecats` showing whether deferred dispatch is active, e.g. `[deferred dispatch: ON — set to -1 to disable]`
- When `max_polecats <= 0` (direct dispatch mode) the annotation reads `[deferred dispatch: OFF]`
- Closes the observability gap reported in #3866: the system default of 10 silently enables deferred dispatch on every rig with no indication in the UI

## Changes

- `internal/cmd/rig_config.go` — adds `configAnnotation()` helper; both `--layers` and plain `show` paths call it for `max_polecats` and append the annotation when non-empty
- `internal/cmd/rig_config_test.go` — adds `TestConfigAnnotation` (unit, 8 cases covering int/int64/float64/string types) and `TestRigConfigShow_MaxPolecatsAnnotation` (integration, 3 scenarios: system default ON, `--layers` mode, wisp override to -1)

## Testing

- [x] `go build ./...` passes
- [x] `TestConfigAnnotation` — all 8 subtests pass
- [x] `TestRigConfigShow_MaxPolecatsAnnotation` — all 3 subtests pass
- [x] `go vet ./...` passes
- [x] Pre-existing test failures (`TestRunPrimeExternalTools_BoundsSlow*`, `TestCheckPendingEscalations_BoundsSlowBdList`) confirmed present on upstream/main — unrelated to this change

Closes #3866


Bella-Giraffety (changes requested):
Thanks for the PR. I don't think this should merge as-is.

Blocking finding: this annotates the rig-layer `max_polecats` shown by `gt rig config show`, but current dispatch mode is controlled by the town setting `scheduler.max_polecats` in `shouldDeferDispatch`. On current `main`, I verified:

```text
$ go run ./cmd/gt config get scheduler.max_polecats
-1

$ go run ./cmd/gt rig config show gastown --layers
max_polecats              10              system
```

That means this PR would display the rig config row as `[deferred dispatch: ON ...]` even when the actual scheduler is in direct-dispatch mode. That would make the operator UX actively misleading.

Recommendation: request changes / likely close or retarget. If this visibility gap still needs work, it should be attached to the scheduler config surface (`scheduler.max_polecats`) rather than `gt rig config show`'s legacy rig config key. Current main also already documents `scheduler.max_polecats` in `gt config set/get` help and defaults it to `-1` direct dispatch, so the original issue premise appears at least partly superseded.

Targeted verification run:

```text
go test ./internal/rig ./internal/scheduler/capacity ./internal/cmd -run 'TestRigConfig|TestSchedulerConfigSetZero|TestSchedulerDisabledMode|TestSchedulerDirectModeNoQueue'
```

esciara:
@Bella-Giraffety Addressed in 22a05dc — retarget per your review. Details in the commit message.

Bella-Giraffety:
Re-review after 22a05dc:

The retarget from `gt rig config show` to `gt config get scheduler.max_polecats` addresses the prior blocking concern: the annotation is now attached to the scheduler setting that actually controls deferred dispatch.

Current blocker is CI-only: `Integration Tests` fails in `TestSchedulerConfigSetZero` because that existing integration assertion still expects exact bare output `0`, while this PR intentionally emits `0  [deferred dispatch: OFF]`.

I verified the label taxonomy from #1400 and added the missing `status/in-progress` label; the PR now has status, priority, and kind labels.

I also opened a small fix PR against the contributor branch: https://github.com/esciara/gastown/pull/6

Targeted checks I ran locally on the fix branch:

```text
go test ./internal/cmd -run 'TestConfigGetSchedulerMaxPolecatsAnnotation|TestSchedulerConfigSetZero'
go test ./internal/rig ./internal/scheduler/capacity ./internal/cmd -run 'TestRigConfig|TestSchedulerConfigSetZero|TestSchedulerDisabledMode|TestSchedulerDirectModeNoQueue|TestConfigGetSchedulerMaxPolecatsAnnotation'
```

Both passed. Once the fix PR is merged into this branch and CI reruns green, I don't see a remaining blocker from the previous review.

Bella-Giraffety (changes requested):
Thanks for retargeting the annotation away from `gt rig config show`; that addresses the earlier correctness issue about the wrong config surface.

I re-reviewed this under the PR sheriff policy and I still need to request changes before this can move forward.

Blocking concern: `gt config get scheduler.max_polecats` is a value getter. This PR changes that output from a bare scalar (`-1`, `0`, `N`) to annotated human text:

```text
0  [deferred dispatch: OFF]
```

That creates a CLI compatibility risk for scripts/tests/operators that parse `gt config get ...` as the raw configured value. The failing integration test is a signal of that contract, not just a stale assertion.

Recommended path:

- Keep `gt config get scheduler.max_polecats` machine-readable and exact.
- Put the deferred-dispatch annotation on a human/status surface instead, e.g. `gt scheduler status`, a future `gt config show`/verbose mode, or another explicitly human-readable command.
- Keep or restore an exact-output test for `gt config get scheduler.max_polecats` so this does not regress.

Because this is labeled as an enhancement rather than a bug/stability/docs/test fix, I should not push the behavior change through with only a test adjustment. I opened https://github.com/esciara/gastown/pull/6 earlier for the CI symptom, but after this final review I do not recommend merging that follow-up as the fix; it masks the compatibility issue instead of preserving the getter contract.

esciara:
@Bella-Giraffety Addressed in 1d23fe0e.

- `gt config get scheduler.max_polecats` is restored to bare scalar (machine-readable, no annotation). `TestSchedulerConfigSetZero` passes unchanged.
- Annotation removed from `gt rig config show` (wrong surface — rig-layer key does not control dispatch).
- Deferred-dispatch visibility moved to `gt scheduler status` (human-readable surface), which now shows a `Dispatch: deferred/direct (scheduler.max_polecats=N)` line.
- `TestSchedulerStatusDispatchMode` integration test added to cover both modes.

esciara:
All CI fixes are now pushed (commit c2ed00ba) — this should be ready to merge.

**Integration Tests** — `TestSchedulerStatusDispatchMode` invoked `gt config set scheduler.max_polecats -1`, which cobra rejects as `unknown shorthand flag: '1' in -1`. Both subtests now use the `configureScheduler` helper (writes TownSettings JSON directly), the existing convention used elsewhere in this file for the `-1` direct-dispatch case and already green in CI via three sibling tests. The deferred assertion is tightened to `scheduler.max_polecats=5` to match the exact format string in `scheduler.go`.

**Lint** — removed a stray trailing blank line accidentally added to `rig_config_test.go` (a `gofmt` violation); that file is now byte-identical to `main`.

Verified locally:
- `golangci-lint run --timeout=5m` (v2.11.4 — the exact CI Lint command) → **0 issues**
- `gofmt -l` on changed files → clean
- `go vet ./internal/cmd/` → OK
- `go build ./...` → OK

CI has been re-triggered and is running on the fix commit.

julianknutsen (changes requested):
Preflight found a scope mismatch: the PR summary/body describes changes to  and  for max_polecats annotations, but the live PR file list shows  and . That is a clear drift from the stated change. Please realign the branch, diff, and description before this can proceed.

julianknutsen (changes requested):
Preflight found a scope mismatch: the PR summary/body describes changes to `internal/cmd/rig_config.go` and `internal/cmd/rig_config_test.go` for max_polecats annotations, but the live PR file list shows `internal/cmd/scheduler.go` and `internal/cmd/scheduler_integration_test.go`. That is a clear drift from the stated change. Please realign the branch, diff, and description before this can proceed.

Bella-Giraffety:
Closing stale enhancement PR in favor of existing issue #3866. We will revisit from the issue rather than trying to fix up this stale PR.

PR #4012, opened 2026-05-14
fix(config): warn when prompt_mode: "none" drops a non-empty startup prompt

### LINKED ISSUE #3803: Custom agent override with prompt_mode: "none" silently drops daemon's startup beacon → looks like deacon crash-loop
### Repro

Source ref: `b4a45cc` on `main`.

In `~/dev/gt/settings/agents.json`, define a `claude` agent override (e.g. cloned from the existing `codex` entry) that includes `"prompt_mode": "none"`:

```json
{
  "version": 1,
  "agents": {
    "claude": {
      "name": "claude",
      "command": "claude",
      "args": ["--dangerously-skip-permissions"],
      "prompt_mode": "none",
      "tmux": { "process_names": ["claude"], "ready_prompt_prefix": "❯ " },
      "instructions": { "file": "CLAUDE.md" }
    }
  }
}
```

Set `default_agent: claude` in `settings/config.json`. Restart daemon and watch the deacon spawn.

### Expected

Deacon agent receives the daemon's startup beacon (`[GAS TOWN] deacon <- daemon • ... • patrol\n\nI am Deacon. Start patrol: run gt deacon heartbeat...`) as the positional prompt arg, claude auto-executes it, deacon emits `gt deacon heartbeat`, the system enters its normal patrol cycle.

### Actual

Deacon's claude pane sits idle at `❯ ` with no echoed prompt. After ~6 min the daemon's stuck-agent-dog kills the session for being heartbeat-stale ("STUCK DEACON: started 6m0s ago but heartbeat still pre-restart"). After 5 retries the daemon flips to permanent crash-loop and stops respawning. From the daemon log it looks identical to a real crash; the actual process is alive and healthy, just idle.

Inspecting `/proc/<pid>/cmdline` shows the daemon DID build a beacon prompt. But the launched claude command was `claude --dangerously-skip-permissions` — the prompt arg was never appended.

### Root cause

`internal/config/types.go:842` `BuildCommandWithPrompt`:

```go
func (rc *RuntimeConfig) BuildCommandWithPrompt(prompt string) string {
    resolved := normalizeRuntimeConfig(rc)
    base := resolved.BuildCommand()
    p := prompt
    if p == "" { p = resolved.InitialPrompt }
    if p == "" || resolved.PromptMode == "none" {
        return base   // ← drops a non-empty prompt silently
    }
    ...
}
```

When `PromptMode == "none"`, a non-empty caller-provided prompt is silently discarded. No log, no warning, no metric.

This is dangerous in combination with `loader.go:1986`:

```go
if result.PromptMode == "" && preset != nil && preset.PromptMode != "" {
    result.PromptMode = preset.PromptMode
}
```

— which means the preset's PromptMode (e.g. claude's `"arg"`) is **only** inherited if the user's `agents.json` override leaves PromptMode empty. Any explicit value, including `"none"`, wins. Combined with the fact that the existing `codex` preset/template ships with `prompt_mode: "none"`, users who clone the codex entry to make a claude entry inadvertently set `prompt_mode: "none"` and break startup-beacon injection.

The codex preset gets away with `"none"` because codex's spawn path doesn't depend on positional-arg injection (it has its own startup mechanics). Claude's preset uses `PromptMode: "arg"` because it relies on the positional prompt arg to bootstrap the startup beacon — the daemon (`internal/deacon/manager.go:109-141`, `internal/daemon/lifecycle.go:534`) builds the prompt via `BuildStartupPrompt(...)` and passes it through `BuildCommandWithPrompt` and never falls back to a post-spawn `tmux send-keys` for the initial bootstrap.

### Symptom matrix

This silent drop is hard to debug because three layers conspire:

1. The agent process is alive and the tmux session is alive — looks healthy.
2. The daemon log shows "Deacon started successfully" and then 6 minutes later "STUCK DEACON" → "Deacon is in crash loop". Looks like a real crash, not a config bug.
3. There's no log line saying "BuildCommandWithPrompt dropped prompt because PromptMode=none".

### Suggested fixes (any subset)

1. **Log a warning when a non-empty prompt is dropped.** In `BuildCommandWithPrompt` (and `BuildArgsWithPrompt`), when `prompt != "" && PromptMode == "none"`, emit a slog/log warn with agent name. This single change would have made the bug self-diagnosing.
2. **Reject the override at config-load time.** In `loader.go`, when an agent override sets `prompt_mode: "none"` and the resolved binary is one whose preset uses `PromptMode: "arg"`, return an error or warning (configurable). Agents that genuinely need `"none"` (codex, etc.) can keep the explicit value.
3. **Document the implication in the agents.json schema/template.** The current `internal/config/agents.go:294` codex preset has `PromptMode: "none"` but there's no sentence in the user-facing doc explaining what this disables and when not to copy it.
4. **Inherit prompt_mode from the resolved binary's preset when the override's command matches.** If the override's `command` resolves to `claude`/`gemini`/`copilot` etc., default `prompt_mode` from the matching preset rather than letting an explicit `"none"` from a copied template silently win.

Tip-of-the-spear fix is #1 — it's the smallest patch and would have saved hours of debugging.

### Adjacent observation

While debugging I also noticed the daemon's stuck-agent-dog interrupts the deacon mid-bash (the deacon's `gt sling mol-deacon-patrol deacon` was interrupted with "What should Claude do instead?") apparently because the deacon hadn't emitted `gt deacon heartbeat` yet when the kill timer fired. Possibly worth a separate issue: the kill threshold should reset on observable progress (any tool call, any bash exec) rather than only on `gt deacon heartbeat`.

esciara:
Closes #3803

## Summary

When `BuildCommandWithPrompt` or `BuildArgsWithPrompt` is called with a non-empty prompt but the resolved `PromptMode` is `"none"`, the prompt was silently discarded with no log, no warning, no metric.

This made the bug nearly impossible to diagnose: the agent process stayed alive, the daemon logged "started successfully", and only 6 minutes later the stuck-agent-dog killed the session with what looked like a crash-loop.

Now both functions emit a warning to stderr naming the agent command, so the misconfiguration is immediately visible in the daemon log.

The most common trigger: copying the codex agent entry (which ships with `prompt_mode: "none"`) to create a claude override in `agents.json`. Codex has its own startup mechanics and doesn't need the positional-arg beacon; claude does. The silent drop broke beacon injection without any indication.

## Changes

- `internal/config/types.go`: emit warning when `prompt_mode: "none"` silently drops a non-empty startup prompt
- `internal/config/loader_test.go`: add tests covering the warning path

🤖 Merged by [Gas Town Refinery](https://github.com/gastownhall/gastown) — esciara/gastown fork

### DIFF
--- a/internal/config/loader_test.go
+++ b/internal/config/loader_test.go
@@ -1,7 +1,9 @@
 package config
 
 import (
+	"bytes"
 	"encoding/json"
+	"io"
 	"os"
 	"os/exec"
 	"path/filepath"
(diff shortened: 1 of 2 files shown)


Bella-Giraffety (approved):
PR sheriff review: approved. Current main still silently drops non-empty startup prompts when prompt_mode is none; this PR adds a clear stderr warning in both BuildCommandWithPrompt and BuildArgsWithPrompt and covers the warning path with targeted tests. CI is green, merge state is clean, and I also verified locally with go test ./internal/config -run 'TestBuild(Command|Args)WithPrompt(Warns|NoWarn|Respects)|TestRuntimeConfigBuildCommandWithPrompt' on the PR branch. Recommendation: squash-merge.

PR #4014, opened 2026-05-14
fix: gt install provisions shared commands into both town root and role subdirs, causing duplicates in Claude Code

### LINKED ISSUE #3874: gt install provisions shared commands into both town root and role subdirs, causing duplicates in Claude Code
## Summary

`gt install` provisions `done`, `handoff`, and `review` slash commands into two locations simultaneously:

1. **Town root**: `<townRoot>/.claude/commands/` (via `templates.ProvisionCommands(absPath)` in `install.go:422`)
2. **Role subdirectories**: `<townRoot>/mayor/.claude/commands/` and `<townRoot>/deacon/.claude/commands/` (via `EnsureSettingsForRole(mayorDir, mayorDir, ...)` in `install.go:294/307`)

Claude Code loads slash commands from **all `.claude/` directories in the path hierarchy**. Since `mayor/` and `deacon/` are nested inside the town root, an agent running from `mayor/` inherits the town root's `.claude/commands/` AND also loads its own `mayor/.claude/commands/`. Each of the three commands appears twice in the agent's command list.

## Repro

1. `gt install` on a fresh town
2. Open a Claude Code session from within `<townRoot>/mayor/`
3. Type `/` and observe `done`, `handoff`, and `review` each listed twice

## Expected

Each command appears exactly once. The town root `.claude/commands/` copy is sufficient — role subdirectories inherit it automatically and should not receive a duplicate copy.

## Root Cause

In `internal/cmd/install.go`:
- Line 422 provisions commands at the town root (`ProvisionCommands(absPath)`)
- Lines 294 and 307 call `EnsureSettingsForRole` with `workDir = mayorDir / deaconDir`, which in turn calls `commands.ProvisionFor(workDir, provider)` writing the same files into each role directory

`EnsureSettingsForRole` in `internal/runtime/runtime.go` at line 51 runs `ProvisionFor` unconditionally for known agents. For polecats/crew the `workDir` is the customer repo (outside the town root hierarchy), so there's no clash. For mayor/deacon, `workDir` IS inside the town root — the duplication only manifests there.

## Fix options

**Option A (minimal):** In `EnsureSettingsForRole`, skip `ProvisionFor` if `workDir` is nested inside a directory that already has the commands provisioned.

**Option B (preferred):** Don't provision shared commands into mayor/deacon subdirs at all during install, since they're covered by town-root inheritance. Only `EnsureSettingsForRole` for roles whose `workDir` is outside the town root (polecats, crew).

**Option C:** `gt doctor --fix` detects and removes duplicate command files from role subdirectories.

## Workaround

Delete the command files from the role subdirectories manually:

```bash
rm mayor/.claude/commands/{done,handoff,review}.md
rm deacon/.claude/commands/{done,handoff,review}.md
```

The town root copies remain and are inherited correctly.

# searched

esciara:
## Summary

- Adds `commandsInherited(workDir)` check in `EnsureSettingsForRole` to skip provisioning slash commands when `workDir` is nested inside a Gas Town workspace root
- Implements the fix by walking up the directory tree to find the nearest git root, then checking if that root is a Gas Town workspace — if so, Claude Code's path-hierarchy traversal already delivers the town root's `.claude/commands/` to the agent
- Prevents duplicate `done`, `handoff`, and `review` commands appearing in mayor/deacon Claude Code sessions

## Changes

- `internal/runtime/runtime.go`: Added `commandsInherited(workDir)` guard to the commands provisioning call in `EnsureSettingsForRole`; added `commandsInherited` and `gitRootOf` helper functions
- `internal/runtime/runtime_test.go`: Added 6 tests covering the `commandsInherited` logic and the integration behaviour of `EnsureSettingsForRole` for both nested (mayor/deacon) and standalone (crew/polecat) workDirs

## Testing

- [x] `go build ./...` passes
- [x] `go test ./internal/runtime/...` passes (all new tests pass)
- [x] `go vet ./...` passes

Closes #3874

### DIFF
--- a/internal/runtime/runtime.go
+++ b/internal/runtime/runtime.go
@@ -14,6 +14,7 @@ import (
 	"github.com/steveyegge/gastown/internal/hookutil"
 	"github.com/steveyegge/gastown/internal/templates/commands"
 	"github.com/steveyegge/gastown/internal/tmux"
+	"github.com/steveyegge/gastown/internal/workspace"
 )
 
 // EnsureSettingsForRole provisions all agent-specific configuration for a role.
@@ -52,8 +53,11 @@ func EnsureSettingsForRole(settingsDir, workDir, role string, rc *config.Runtime
 	}
 
 	// 2. Slash commands (agent-agnostic, uses shared body with provider-specific frontmatter)
-	// Only provision for known agents to maintain backwards compatibility
-	if commands.IsKnownAgent(provider) {
+	// Only provision for known agents to maintain backwards compatibility.
+	// Skip provisioning when workDir is nested inside a town root: Claude Code's
+	// path-hierarchy traversal already delivers the town root's .claude/commands/
+	// to the agent, so a duplicate copy in workDir causes each command to appear twice.
+	if commands.IsKnownAgent(provider) && !commandsInherited(workDir) {
 		if err := commands.ProvisionFor(workDir, provider); err != nil {
 			return err
 		}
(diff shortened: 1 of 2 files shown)


Bella-Giraffety (changes requested):
Thanks for the fix. The issue is real and current `main` still needs a guard here, but this implementation misses the fresh-install path.

`gt install --git` calls `EnsureSettingsForRole` for `mayor/` and `deacon/` before `InitGitForHarness` creates the town `.git` directory (`internal/cmd/install.go`). In that state, `gitRootOf(workDir)` returns empty, so `commandsInherited` returns false and the duplicate role-local `.claude/commands/` files are still provisioned during the install flow that this PR is meant to fix.

I opened #4019 with the same core fix plus coverage for the pre-git-init install ordering and the nested crew/polecat git-repo boundary. Recommendation: close/supersede this PR in favor of #4019, or update this branch with the same install-order handling before merge.

Validated locally: `go test ./internal/runtime` passes on this PR and on the replacement branch.

Bella-Giraffety:
Superseded by replacement PR #4019, which covers the fresh install ordering case and nested repo boundaries. PR #4019 is merged; closing this PR to avoid duplicate fix paths.

PR #4020, opened 2026-05-15
Fix Ready Across Rigs sling resolution

fengning-starsend:
## Summary

This fixes the Ready Across Rigs dashboard path so rows shown in the UI are sling-resolvable by the same visible ID.

The change keeps the scope narrow:

- pin routed Beads commands to the selected workspace database when resolving/showing routed work
- filter Ready Across Rigs rows that cannot be resolved through the same visible ID and route
- resolve `type=rigs` dropdown options from the town rig registry without requiring the slower full options fanout
- request a longer `/api/run` timeout for dashboard Sling actions, matching the real launch path

## Why

The dashboard can display ready work across rigs, but the UI should only show actionable rows: if a user sees a row ID in Ready Across Rigs and clicks Sling, that same ID should resolve through Gastown's routing model and dispatch successfully.

Without this, Ready Across Rigs can surface rows whose visible ID and route context disagree, or the Sling dropdown can fail before the user reaches the launch step.

## Validation

Automated checks:

- `go vet ./...`
- `umask 022; go test ./...`

Manual browser smoke test against a local routed-rig fixture:

- installed `gt` from this branch: `gt version v1.1.0-145-g1dd1248c`
- restarted the dashboard from a local town workspace
- confirmed `/api/options?type=rigs` returns the configured rig
- clicked a visible Ready Across Rigs row
- clicked the stock `Sling` action
- selected the target rig from the dropdown
- observed `/api/run` return HTTP 200 with `success: true`
- confirmed the backend moved the same visible row ID to `hooked`
- unslung/reset the fixture afterward

## Notes

This does not add new dashboard concepts; it makes the existing ready-work and Sling surfaces agree on the visible work ID and routing context.

Internal tracking: bd-0xv78
Agent: codex

### DIFF
--- a/internal/cmd/bd_helpers.go
+++ b/internal/cmd/bd_helpers.go
@@ -101,6 +101,32 @@ func filterEnvKey(env []string, key string) []string {
 	return result
 }
 
+func filterBdTargetEnv(env []string) []string {
+	for _, key := range []string{
+		"BEADS_DIR",
+		"BEADS_DB",
+		"BEADS_DOLT_SERVER_DATABASE",
+		"BEADS_DOLT_SERVER_HOST",
+		"BEADS_DOLT_SERVER_PORT",
+		"BEADS_DOLT_PORT",
+	} {
+		env = filterEnvKey(env, key)
+	}
+	return env
+}
+
+func pinBeadsDirEnv(env []string, beadsDir string) []string {
+	env = filterBdTargetEnv(env)
+	if beadsDir == "" {
+		return env
+	}
+	env = append(env, "BEADS_DIR="+beadsDir)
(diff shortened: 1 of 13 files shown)


Bella-Giraffety (approved):
Reviewed end-to-end. The Ready Across Rigs sling-resolution fix is focused, covered by targeted tests for Beads env pinning, route filtering, rig option lookup, and dashboard run timeout behavior, and CI is green. I also ran targeted local checks: changed-code tests pass for internal/cmd, internal/web, and internal/config; broad package tests hit unrelated environment-sensitive failures (missing rootless Docker for testcontainers and local agent preset overrides), not regressions from this PR.

PR #4022, opened 2026-05-15
fix: filter '(not set)' sentinel from bd config output (gt-kbi)

adensur:
## Summary
- Fixes `gt convoy create` (and any other caller of `EnsureCustomStatuses` / `EnsureCustomTypes`) failing on workspaces where `bd config get status.custom` returns the literal unset sentinel `status.custom (not set)`.
- The beads package's helpers were doing naive `strings.TrimSpace` on bd's stdout, which treated the sentinel string as a valid value. The doctor package already had the correct parser (`internal/doctor/config_check.go:parseConfigOutput` filters `(not set)` and `Note:` lines) — fix unifies the parsing.

Reported by `tongue_sentences2/crew/Teacher` (mail `hq-wisp-ugai`). Tracking bead: `gt-kbi`.

## Test plan
- [ ] Unit test: feed `status.custom (not set)\n` through the filter; expect empty
- [ ] Regression: fresh town with unset `status.custom` → `gt convoy create` succeeds
- [ ] Existing workspaces (status.custom configured) unaffected

### DIFF
--- a/internal/beads/beads_types.go
+++ b/internal/beads/beads_types.go
@@ -234,7 +234,7 @@ func EnsureCustomStatuses(beadsDir string) error {
 
 	// Build merged set: existing + required
 	statusSet := make(map[string]bool)
-	if existing := strings.TrimSpace(string(existingOutput)); existing != "" {
+	if existing := ParseConfigOutput(existingOutput); existing != "" {
 		for _, s := range strings.Split(existing, ",") {
 			s = strings.TrimSpace(s)
 			if s != "" {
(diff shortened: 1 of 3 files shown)


Bella-Giraffety (approved):
LGTM. The shared parser matches the existing doctor behavior and fixes the `status.custom (not set)` regression path without broadening the config-write surface.

Verified locally:
- `go test ./internal/beads ./internal/doctor`
- `go test ./internal/cmd -run 'TestConvoy|TestResolveBeadsDir'`

Non-blocking note: the PR body mentions `EnsureCustomTypes`, but that helper already writes the canonical list directly; the actual regression fixed here is the status/custom parser path and the shared doctor parser.

PR #4041, opened 2026-05-16
fix(tmux): guard NewSession against socket clobber on unresponsive servers (gt-h9z)

Rome-1:
## Summary

Closes gt-h9z. When the configured \`-L\` socket path is held by a server that tmux's internal liveness probe fails to reach inside its tight timeout, tmux's recovery path unlinks the socket and binds a fresh server at the same path. The original server's listener stays on the kernel fd table — so **two \`tmux: server\` processes end up simultaneously bound to the same socket path with different inodes**, and any client invoked from a non-tmux shell routes to the new server (which sees none of the old sessions). The captured evidence is in the bead: \`ss -lx\` showing two listeners on \`/tmp/tmux-1000/gt-0a7d4e\`, and a user describing how they "lost" their existing pane because the new server didn't see the old sessions.

## Approach

Two-stage health probe in front of every \`new-session\` call:

1. **Kernel-level Unix-socket dial.**
   - \`ECONNREFUSED\` → file is a benign leftover (tmux exited cleanly, left the inode). tmux will recreate. **Proceed.**
   - Dial succeeds → listener is alive at the kernel layer; advance to stage 2.
   - Other dial error (\`ENOTSOCK\`, timeout, etc.) → file at the socket path is unsafe. **Bail.**
2. **App-level \`list-sessions\` probe.**
   - Succeeds → server is responsive. **Proceed.**
   - \`ErrNoServer\` → server shut down between stages 1 and 2. **Proceed.**
   - Other error → app-layer wedge; this is the gt-h9z window. **Bail.**

Gates all three NewSession entry points (\`NewSession\`, \`NewSessionWithCommand\`, \`NewSessionWithCommandAndEnv\`). Default-socket Tmux (no \`-L\`) is left untouched — that path is shared with the user's interactive tmux and outside the \`gt crew start\` workflow that produced the captured evidence.

## Why both stages

The kernel-level dial is the load-bearing piece. An earlier draft of this fix ran only the app-level probe and incorrectly bailed in the common case where a shared test server's last session was killed, tmux exited cleanly, and the socket file persisted — every shared-socket test in the package started failing. The kernel dial returns \`ECONNREFUSED\` in that case and lets tmux recreate the server.

The app-level probe then catches the gt-h9z window specifically: kernel-level connect succeeds (so the listener is alive in the abstract sense), but tmux's own protocol probe wouldn't.

## Test plan

- \`TestProbeServerHealth\` (new) covers the matrix:
  - default socket → nil
  - absent socket file → nil
  - stale Unix socket (\`net.Listen\`/\`Close\` — mirrors tmux's clean exit) → nil
  - live tmux server → nil
  - non-socket file at the socket path → error mentioning \`gt-h9z\` (the bail path)
- \`TestNewSessionAllowsStaleUnixSocket\` (new) is the regression guard for the earlier-draft failure mode — NewSession must succeed against an unbound Unix socket so tmux can recreate the server cleanly.
- Full \`go test ./internal/tmux/\` passes (the two pre-existing dialog-timing flakes are unrelated and were failing before this PR too).

## Note on venue

Filed from a Gas City crew worker (\`marlowe\`) because the bead was misrouted there. \`internal/tmux/tmux.go\` is gastown-side; the analogous gascity file (\`internal/runtime/tmux/tmux.go\`) has a structurally similar pattern but is not the venue described in the gt-h9z evidence, so I've left gascity alone for now. Happy to mirror this fix into gascity if maintainers agree it's worth prophylactic coverage.


Rome-1:
Closing in favor of #4150, which is the clean cherry-pick of the tmux fix commit (8f9ebab2) onto current `main`. This PR's branch had accumulated 76,791 lines of accidentally-committed `.beads/backup/*.jsonl` exports plus 2,398 lines of unrelated polecat-reconcile/config-loader work from sibling commits and was structurally unmergeable. #4150 contains only `internal/tmux/tmux.go` (+85) and `internal/tmux/tmux_test.go` (+139). Source branch will be deleted from Rome-1/gastown to remove the contaminated artifacts.

PR #4058, opened 2026-05-19
fix(release): age-gate npm self-update before publish

puzanov:
## Summary

Add `--min-release-age=7` to the release workflow's npm self-update step before publishing the npm package.

## Why

The `publish-npm` job updates npm from the public registry immediately before `npm publish --provenance`. Using npm's release-age gate avoids selecting a very fresh npm release in this privileged publish job while preserving the existing workflow behavior.

## Change

```diff
-npm install -g npm@latest
+npm install -g npm@latest --min-release-age=7
```

## Verification

- `git diff --check`
- Verified npm accepts `--min-release-age=7` with a dry-run install through a temporary prefix.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -251,7 +251,7 @@ jobs:
           registry-url: 'https://registry.npmjs.org'
 
       - name: Update npm for provenance support
-        run: npm install -g npm@latest
+        run: npm install -g npm@latest --min-release-age=7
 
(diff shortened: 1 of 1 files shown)


Bella-Giraffety:
Incorporated into main for v1.2.0 in release commit lineage ending at 2e2122b4. Closing this PR as merged/incorporated; thank you for the contribution.

PR #4061, opened 2026-05-19
fix: gt polecat prune --remote misses squash/rebase merges (uses IsAncestor instead of patch-equivalence)

### LINKED ISSUE #3779: gt polecat prune --remote misses squash/rebase merges (uses IsAncestor instead of patch-equivalence)
## Problem

`gt polecat prune --remote` is meant to be the operator's safety net for cleaning up stale polecat branches on `origin`. The code path at `internal/cmd/polecat.go:1635` checks merged-ness via:

```go
merged, mergeErr := repoGit.IsAncestor(branch, "origin/"+defaultBranch)
```

For a polecat branch that was merged via **squash** or **rebase** (the GitHub default for `gh pr merge`, and the merge mode many faultline-style repos use), the original commits no longer exist on main — the merged commit has a different SHA. `IsAncestor` returns `false`, and prune silently skips the branch. End result: stale polecat branches accumulate even when their work is fully merged.

## Reproduction

Repo with 29 polecat branches whose commits were all rebase-merged into main:

```
$ gt polecat prune <rig> --remote --dry-run
…
Pruning remote polecat branches...
No stale remote polecat branches found.
```

Manual patch-equivalence check confirms every one of those 29 branches has zero unique commits vs `origin/main`:

```bash
for b in $(git ls-remote --heads origin 'polecat/*' | awk '{print $2}' | sed 's|refs/heads/||'); do
  unique=$(git cherry origin/main "origin/$b" | grep -c '^+')
  [ "$unique" -eq 0 ] && echo "MERGED (patch-equivalent): $b"
done
```

…matched all 29. Deleting them with `git push origin --delete` cleared the backlog cleanly.

## Why the Refinery fix (#2435 / 2a0a8c760) doesn't cover this

`2a0a8c760` made the **Refinery** unconditionally delete polecat remote branches when an MR merges. That handles the happy path — but only when the merge actually flows through the MQ. Polecat branches that landed via:

- `gh pr merge --squash`
- `gh pr merge --rebase`
- Direct `git push` to main (crew-worker style)
- Manual CLI merge that bypassed Refinery

…never trigger the Refinery cleanup, so they need the operator-driven `gt polecat prune --remote` to clean up. That's exactly the path with the `IsAncestor` bug.

## Proposed fix

Replace `IsAncestor` at `internal/cmd/polecat.go:1635` with a patch-equivalence check. `git cherry` is the standard tool for this:

```go
// Determine whether `branch` is fully merged into `target` even if the
// merge was a squash or rebase (which rewrites SHAs). git cherry prints
// `+` for unique commits on `branch`, `-` for patch-equivalent commits
// already on `target`. Zero `+` lines → the branch is fully merged.
out, err := g.run("cherry", "origin/"+defaultBranch, branch)
if err != nil { continue }
unique := 0
for _, line := range strings.Split(strings.TrimSpace(string(out)), "\n") {
    if strings.HasPrefix(line, "+ ") { unique++ }
}
merged := unique == 0
```

This subsumes `IsAncestor` (a fast-forward merged branch has 0 unique commits and is also an ancestor) and additionally handles squash/rebase merges. Performance is fine for the prune use case — `git cherry` runs once per polecat branch, not per commit.

If preserving `IsAncestor` as a fast path is desired, gate `git cherry` only on the `!IsAncestor` case to keep the common direct-merge path unchanged.

## Refs

- `internal/cmd/polecat.go:1620` — `runPolecatPrune` `--remote` block
- `internal/cmd/polecat.go:1635` — the `IsAncestor` call
- Issue #2435 — Refinery happy-path fix (already shipped)
- `internal/cmd/prune_branches.go:20-32` — comment that *assumed* post-merge deletion would handle this

## Discovered while

Reviewing 30 stale polecat branches on `outdoorsea/faultline` (oldest 2026-03-24, all patch-merged but `IsAncestor`-orphaned). Cleared manually with the workaround above.

esciara:
## Summary

- `gt polecat prune --remote` used `IsAncestor` to detect merged branches, which returns `false` for squash/rebase merges because the original commits no longer appear on main with the same SHA.
- Replace with `git cherry` (patch-id comparison): commits whose patches are already on upstream show as `-` lines; zero `+` lines means fully merged regardless of merge strategy.
- Add `TestCherry_SquashMergedBranch`, `TestCherry_UnmergedBranch`, and `TestCherry_NormalMergedBranch` to `internal/git` to lock in the behavior across all three merge paths.

## Changes

- `internal/cmd/polecat.go`: swap `IsAncestor` call for `Cherry` with `+`-line counting; add explanatory comment for the empty-output edge case (normal merges produce no output, not `-` lines).
- `internal/git/git_test.go`: three new tests covering squash-merged, unmerged, and normal-merged branches.

## Testing

- [x] `go build ./...` passes
- [x] `go test ./internal/git/... -run TestCherry` — all three tests pass
- [x] `go test ./internal/git/...` — full package passes
- [x] `go vet ./internal/cmd/... ./internal/git/...` passes

Closes #3779

### DIFF
--- a/internal/cmd/polecat.go
+++ b/internal/cmd/polecat.go
@@ -1779,13 +1779,23 @@ func runPolecatPrune(cmd *cobra.Command, args []string) error {
 				continue
 			}
 			branch := strings.TrimPrefix(ref.Name, "refs/heads/")
-			// Use the listed remote tip, not the short branch name, so remote-only
-			// branches can be classified without a local branch.
-			merged, mergeErr := repoGit.IsAncestor(ref.Hash, "origin/"+defaultBranch)
-			if mergeErr != nil {
+			// Use git cherry (patch-id comparison) instead of IsAncestor so that
+			// squash/rebase merges—where the SHA is rewritten—are also detected.
+			// "+" lines are unique commits not yet on upstream; zero means merged.
+			cherryOut, cherryErr := repoGit.Cherry("origin/"+defaultBranch, ref.Hash)
+			if cherryErr != nil {
 				continue
 			}
-			if !merged {
+			// Count "+" lines: patches not yet on upstream. "-" lines are squash/rebase
+			// equivalents; empty output means all commits are reachable ancestors.
+			// Both "-" and empty → unique==0 → branch is merged.
+			unique := 0
+			for _, line := range strings.Split(strings.TrimSpace(cherryOut), "\n") {
+				if strings.HasPrefix(line, "+ ") {
+					unique++
+				}
+			}
+			if unique > 0 {
 				continue
 			}
 
--- a/internal/git/git_test.go
+++ b/internal/git/git_test.go
@@ -1067,6 +1067,124 @@ func TestListPushRemoteRefsWithHashesClassifiesRemoteOnlyMergedBranch(t *testing
 	}
 }
 
+func TestCherry_SquashMergedBranch(t *testing.T) {
+	localDir, _, mainBranch := initTestRepoWithRemote(t)
+	g := NewGit(localDir)
+	branch := "polecat/squash-merged"
+
+	runGit(t, localDir, "checkout", "-b", branch)
+	if err := os.WriteFile(filepath.Join(localDir, "squash-work.txt"), []byte("work"), 0644); err != nil {
+		t.Fatalf("write: %v", err)
+	}
+	runGit(t, localDir, "add", "squash-work.txt")
+	runGit(t, localDir, "commit", "-m", "squash work")
+	branchSHA, err := g.Rev("HEAD")
+	if err != nil {
+		t.Fatalf("Rev branch: %v", err)
+	}
+	runGit(t, localDir, "push", "origin", branch)
+
+	// Squash-merge into main: different SHA from the branch commit.
+	runGit(t, localDir, "checkout", mainBranch)
+	runGit(t, localDir, "merge", "--squash", branch)
+	runGit(t, localDir, "commit", "-m", "squash merge polecat/squash-merged")
+	runGit(t, localDir, "push", "origin", mainBranch)
+
+	// IsAncestor returns false for squash merges (SHA mismatch) — that's the bug.
+	isAncestor, ancestorErr := g.IsAncestor(branchSHA, "origin/"+mainBranch)
+	if ancestorErr != nil {
+		t.Fatalf("IsAncestor: %v", ancestorErr)
+	}
+	if isAncestor {
+		t.Fatal("test setup broken: IsAncestor returned true for a squash merge (expected false)")
+	}
+
+	// Cherry must report 0 unique commits — the fix for the bug.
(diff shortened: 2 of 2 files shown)


Bella-Giraffety:
PR sheriff workflow record for #4061.

Research legs (15):
1. Current `gt polecat prune --remote` uses ancestry and misses squash/rebase-equivalent branches.
2. Current fork base lacks `Git.Cherry`; a small wrapper is needed when porting.
3. Tests need squash/rebase-equivalent and unmerged coverage.
4. Remote prune should use `ls-remote` hashes, not local branch names.
5. `git cherry`: `+` unique, `-` equivalent, empty ancestor/no unique; aggregate squash needs a second proof.
6. Split fetch/push remotes must list and fetch from the push target.
7. Keep true ancestry checks elsewhere; add narrow prune-oriented helpers.
8. Preserve prune UX and fail closed on inspect errors.
9. Restrict deletion to `polecat/*` refs and skip on uncertainty.
10. `origin/main` and upstream PR base diverge heavily; port only relevant hunks.
11. Existing `internal/git` temp-repo helpers are the right test surface.
12. Centralized semantic helpers are cleaner than command-local parsing.
13. `git cherry` failures should skip/warn, not delete.
14. PR #4061 needed candidate ref fetching because `ls-remote` alone does not make objects local.
15. Labels: kept `kind/bug`, `priority/p2`; added `status/needs-review`.

Pre-implementation reviews (5):
1. Conditionally sound; add multi-commit squash coverage/aggregate no-op proof.
2. Block unless multi-commit squash is handled or explicitly scoped.
3. Approved with changes: test helper behavior and split push-url ref fetch.
4. Approved: `git cherry` is conservative but has squash false negatives alone.
5. Block unless implementation uses remote SHA and leaves local pruning unchanged.

Final post-implementation reviews (5):
1. No findings after leased delete, remote hash fetch, and no-op merge proof.
2. Tests cover multi-commit squash, cherry-pick, unmerged, split push-url fetch/list, and leased delete; CLI e2e gap is low/non-blocking.
3. No scope/minimality issues; logic converges into git helpers.
4. Git semantics approved: exact ref fetch, patch-id check, no-op merge proof, and force-with-lease delete.
5. No compile/API compatibility issues; existing `ListRemoteRefs`/`ListPushRemoteRefs` APIs are preserved.

Implementation has been ported on the sheriff branch for merge-queue submission: remote prune now lists push-remote refs with hashes, fetches the exact candidate ref, verifies the inspected tip, classifies landed work through ancestry/cherry/no-op merge, and deletes with an expected-tip lease.

Verification run:
- `go test ./internal/git`

Bella-Giraffety:
Closed as superseded by replacement PR #4447, merged as 2f6f14071afb49b71c0a41e03545b14158e7012c. Original contribution attribution was preserved in the squash merge with Co-authored-by: Emmanuel Sciara <emmanuel.sciara@gmail.com>.

PR #4062, opened 2026-05-19
fix(compact): add progress logging + hard timeout to prevent silent 10min+ hangs

certivpaul:
## Summary

- `gt compact` now imposes a hard 5m timeout (configurable via `--timeout` flag or `GT_COMPACT_TIMEOUT_SEC`) over the whole run, with `ctx.Err()` checks between wisps so a backed-up bd queue short-circuits cleanly.
- A heartbeat goroutine prints a one-line progress update to stderr every 15s (configurable via `GT_COMPACT_PROGRESS_SEC`, `0` disables for tests), showing current step, current wisp id, processed/total, and elapsed time. Stderr keeps `--json` output clean.
- On timeout, a one-shot diagnostic block dumps last step, current wisp, processed/total/remaining, partial result counts, and a remediation hint before returning `errCompactTimeout`.

## Root cause (bug aa-c6a)

During deacon patrol cycle 2026-05-09 16:17, `gt compact` produced zero stdout/stderr for 11+ minutes, alive but unresponsive (PID 488924), and required SIGTERM to recover. The per-bd-subprocess timeout (60s in `internal/beads/beads.go`) bounded individual `bd` calls, but `runCompact` had:

1. No top-level deadline — N wisps × slow Dolt = unbounded wall-clock.
2. No progress output in non-verbose, non-JSON mode — only one "Compacting N wisps..." line before the silent loop.

So a slow Dolt response made the run indistinguishable from a hang.

## Fix shape

- Added `context.WithTimeout` around the whole flow.
- Refactored the wisp processing into `runCompactLoop(ctx, ...)` which checks `ctx.Err()` between iterations.
- Added `compactProgress` (atomics-based) + `startProgressHeartbeat` goroutine.
- Added `dumpCompactDiagnostics` for the timeout path so silent hangs leave a forensic trail.
- New tests for timeout resolution precedence (flag > env > default), progress interval, format line behavior, and the timeout error message.

## Related

`aa-sct` (mail inbox hung ~30s mid-patrol) is being fixed in parallel — same morning, same Dolt server early-shutdown event noted in handoff `aa-4xg`. May share a root cause; this PR is scoped to `gt compact` only and does not conflict with the inbox fix.

## Test plan

- [x] `go build ./...` clean
- [x] `go vet ./internal/cmd/... ./cmd/gt/...` clean
- [x] `gofmt -l` clean on touched files
- [x] `go test ./internal/cmd/...` passes (incl. 8 new tests)
- [ ] Manual: invoke `gt compact` from deacon role and confirm a progress line every 15s
- [ ] Manual: invoke `gt compact --timeout 1s` and confirm the timeout dump + non-zero exit
- [ ] Manual: invoke `gt compact --json` and confirm stdout is still pure JSON (heartbeat suppressed)

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>

### DIFF
--- a/internal/cmd/compact.go
+++ b/internal/cmd/compact.go
@@ -2,10 +2,14 @@ package cmd
 
 import (
 	"bytes"
+	"context"
 	"encoding/json"
 	"fmt"
 	"os"
+	"strconv"
 	"strings"
+	"sync"
+	"sync/atomic"
 	"time"
 	"unicode/utf8"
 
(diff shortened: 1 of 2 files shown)


Bella-Giraffety:
Closing this PR as stale/conflicting. The compact progress/timeout problem is real and now tracked in #4399, but the current PR adds incomplete timeout behavior and touches stale compaction code. We should rebuild against current main with cancellation/progress at the actual blocking boundaries.

PR #4064, opened 2026-05-19
fix(refinery): resolve agent-bead from rig DB to avoid legacy hq.wisps stranding (aa-b2tm)

certivpaul:
## Summary

Refinery patrols on rigs whose agent beads still live in `hq.wisps` (the
town DB) get stuck in a 30s backoff loop and never reach reduced-effort
idle. `gt mol step await-event --agent-bead <legacy-id>` fails with
\"no issue found\" from the rig worktree because the worktree's beads
dir resolves to the rig DB, while the legacy bead is in the town DB.

Two complementary fixes, no data migration required:

- **Cross-DB fallback for agent-bead lookups** (`internal/cmd/agent_state.go`,
  `internal/cmd/molecule_await_signal.go`): all reads and writes against
  an agent bead now probe the passed beads dir first and, on miss, fall
  back to the town `.beads` (resolved via `FindTownRoot`). Legacy
  hq.wisps-stranded refinery beads become reachable transparently from
  rig worktrees.
- **`gt agents resolve --role <role> --rig <rig>`** (new command,
  `internal/cmd/agents_resolve.go`): searches BOTH the issues and the
  wisps tables in BOTH the rig and the town DBs, with preference
  ordering (rig-wisps > rig-issues > town-wisps > town-issues), and
  optional `--json` provenance. Replaces the patrol formula's
  `bd list --label=gt:agent` resolver, which only sees the issues table
  in a single DB and so misses both wisps-migrated and town-stranded
  beads.
- The refinery and witness patrol formulas now call `gt agents resolve`
  for their self-ID lookup. The legacy `bd list` query is kept as a
  documented fallback for older builds.
- Formula versions bumped (refinery 14→15, witness 12→13).

## Root cause (from aa-b2tm)

1. Legacy refinery agent beads `au-wisp-0ti` (alleago_ui) and
   `aa-wisp-a4v` (alleago_api) live in `hq.wisps` (town DB).
2. Modern wisps for the same agents (`au-alleago_ui-refinery`,
   `aa-alleago_api-refinery`) live in their respective rig DBs.
3. The refinery worktree's `findLocalBeadsDir` + `ResolveBeadsDir`
   resolve to the rig DB. `bd show` against the rig DB doesn't find the
   legacy ID, and the running refinery still passes the legacy
   `--agent-bead` arg.
4. Compounding: `bd list --label=gt:agent` only queries the issues table,
   not the wisps table, so the formula's bead-resolution step finds zero
   matches even when modern wisps exist.

## Test plan

- [x] `go build -o /tmp/gt-ref-test ./cmd/gt` succeeds
- [x] `go vet ./...` clean
- [x] `go test ./internal/cmd/ ./internal/formula/ ./internal/beads/` passes
- [x] New unit tests for `townBeadsDirFrom`, `pickBestAgentBead`,
      `entryMatchesRole`, `entryMatchesRig`
- [x] Updated `TestPatrolFormulasUseDynamicBeadResolution` accepts either
      legacy or new resolver
- [ ] Manual verification on the affected refineries:
      `gt agents resolve --role refinery --rig alleago_ui` should print
      the bead ID it can actually open; `gt mol step await-event
      --agent-bead <that-id>` should advance idle/backoff labels.

## Operator notes

No data migration is required. Legacy hq.wisps beads remain valid; the
runtime simply finds them again. Operators do not need to take any
action on Paul's host beyond restarting the affected refineries so they
pick up the new `gt agents resolve` resolver step (or wait for the next
handoff cycle).


Bella-Giraffety:
Closed as superseded by #4300, which implements the accepted cleanup-first replacement for agent-bead resolution by role and rig. Thank you for the original report and patch direction.

PR #4065, opened 2026-05-19
fix(test): reap leaked dolt sql-server processes

certivpaul:
## Summary
Fixes a bug where gt test runs leaked `dolt sql-server` child processes that ended up reparented to `systemd` and lived forever. On one developer host, 77+ orphaned dolt sql-servers accumulated over ~11 days, each holding open `.dolt` LOCK fds on `/tmp` dirs the test framework had already removed (`cwd` showed as `(deleted)`).

Two distinct leak sources, both fixed:

### 1. Unit tests using `newTestDAG` (the larger leak)
Tests like `TestLaunchAsAlias_EpicInput`, `TestLaunchAsAlias_TaskListInput`, and `TestCreateStagedConvoy_CleanReady` flowed through `createStagedConvoy` -> `addTrackingRelationFn` -> `mutateTrackingRelationViaStore` -> `beadsdk.OpenFromConfig`, and the beads Go SDK auto-spawned a `dolt sql-server` inside the test's `t.TempDir()`. The `bd` shell stub intercepted only the bd-subprocess path; the Go SDK path slipped through.

Fixed by:
- Stubbing `addTrackingRelationFn` / `removeTrackingRelationFn` inside `testDAG.Setup()` so the SDK path is never reached.
- Setting `BEADS_DOLT_AUTO_START=0` via `t.Setenv` as a belt-and-braces guard against any other store-open paths.
- Mirroring the bd stub's `CMD:...` log line shape so existing assertions on `bd.log` continue to pass.

**Verified**: running the three named tests now spawns **zero** new dolt sql-servers (was: one per test invocation). All 130+ tests in `internal/cmd` still pass.

### 2. e2e integration tests that genuinely need a real dolt
`TestInstallCreatesCorrectStructure`, `TestInstallBeadsHasCorrectPrefix`, `TestInstallFormulasProvisioned`, `TestInstallDoctorClean`, `TestInstallWithDaemon`, and `TestFreshInstallRigPolecatHookIntegration` spawn a real `dolt sql-server` via `gt install` and rely on either `gt dolt stop` or `pkill -f "dolt sql-server"` at `t.Cleanup` time. Both cleanups are fragile.

Added `testutil.ReapDoltOnCleanup(t, townRoot)`:
- Reads `<townRoot>/daemon/dolt.pid` (Dolt's own `.dolt/sql-server.info` as fallback).
- Sends SIGTERM -> waits 2s -> escalates to SIGKILL.
- On Linux also scans `/proc` for `dolt sql-server` processes whose `--config`/`--data-dir` args live inside `townRoot`, catching cases where the PID file write itself failed.
- Process-name verified (won't reap unrelated PIDs that happen to be reused).
- Registered **before** `gt install` runs, so cleanup fires even if a later assertion fatals.

The two fixes are layered: the unit-test stub **prevents** the leak from happening (best), and the reaper is the hard backstop for tests that genuinely need a real dolt sql-server.

## Test plan
- [x] `go build ./...` clean
- [x] `go vet ./...` clean (also with `-tags=e2e` and `-tags=integration`)
- [x] `go test ./internal/cmd/...` passes (all 130+ tests, ~46s)
- [x] `go test ./internal/testutil/...` passes (new helper tests included)
- [x] Confirmed `TestLaunchAsAlias_EpicInput` (and siblings) now spawn 0 new dolt sql-servers per run (was: 1 per run)
- [x] Cross-platform: helper uses build tags (`_linux.go`, `_other.go`) for `/proc` scan vs `ps` fallback
- [x] Pre-existing tmux test failures (`TestNudgeSession_WithRetry`, `TestNudgeSession_WithStoredPaneID`) confirmed unrelated to this change (failed on `gastownhall/main` before the diff)

Closes aa-7f9r.

### DIFF
--- a/internal/cmd/convoy_test_helpers_test.go
+++ b/internal/cmd/convoy_test_helpers_test.go
@@ -424,6 +424,39 @@ func (d *testDAG) Setup(t *testing.T) (townRoot, logPath string) {
 	// Inject bin/ into PATH.
 	t.Setenv("PATH", binDir+":"+os.Getenv("PATH"))
 
+	// Prevent the embedded beads SDK from auto-spawning a real `dolt
+	// sql-server` if any test code path slips past our stubs and hits
+	// b.OpenStore / mutateTrackingRelationViaStore. Without this, a
+	// previously-passing test like TestLaunchAsAlias_EpicInput could
+	// silently fork a dolt process whose cwd is inside this test's
+	// t.TempDir(); when the temp dir was removed at test exit the dolt
+	// kept running (cwd "(deleted)"), reparented to systemd. That was
+	// the bug behind aa-7f9r — 77+ orphaned dolt sql-servers piled up
+	// on a developer host over ~11 days.
+	t.Setenv("BEADS_DOLT_AUTO_START", "0")
+
+	// Stub the tracking-relation helpers so calls from runConvoyStage /
+	// createStagedConvoy don't reach mutateTrackingRelationViaStore (which
+	// would open a beads-SDK store and, if no server were running, fork a
+	// fresh dolt sql-server in the test temp dir). The stub records each
+	// call into bd.log using the same "CMD:dep add ..." line shape that
+	// the bd shell stub uses, so existing assertions that grep bd.log for
+	// `dep add <convoy> <bead>` continue to pass.
+	oldAddTracking := addTrackingRelationFn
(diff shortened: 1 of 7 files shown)


certivpaul:
Closing — upstream already has equivalent dolt server cleanup via `ReapOwnedDoltOnCleanup` in `internal/testutil/dolt_cleanup.go`. Our approach was superseded.

PR #4040, opened 2026-05-16
feat(dolt-backup): retention policy — prune .darc older than N days, keep floor of 3 (dc-2ajt)

athosmartins:
## Summary

Adds retention policy to dolt-backup plugin (Step 3 in run.sh):
- After backup sync loop, prune .darc files older than `BACKUP_RETENTION_DAYS=7` (env-overridable)
- Keep `BACKUP_SAFETY_FLOOR=3` most-recent archives regardless of age
- Non-fatal: errors logged, never block sync reporting
- Portable: macOS (`stat -f`) and Linux (`stat -c`)

## Why

Issue surfaced 2026-05-16: `~/gt/.beads/backup` grew to 43GB with 190 .darc files since April. Manual cleanup freed ~52GB but root cause was no retention. Analog to existing `gdrive_uploader.cleanup_old_files()` (dc-zto8/dc-fek1).

## Testing

- Dry-run: correctly identifies 8 backup dirs across rigs
- Functional: 5 files (3 new + 2 old@10d) → keeps 3, deletes 2

## Bead

dc-2ajt — implemented by gastown/crew/batista

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4068, opened 2026-05-20
feat(dolt-backup): retention policy — prune .darc older than N days, keep floor of 3

athosmartins:
## Summary

Adds a retention policy to `plugins/dolt-backup/run.sh`. After the backup loop,
each DB's configured backup URL is read from `repo_state.json` and
`retention_cleanup()` is called for `file://` remotes (cloud remotes manage
their own retention).

- **Safety floor**: never delete below `BACKUP_SAFETY_FLOOR` (default `3`) most-recent `.darc` archives
- **Age policy**: delete `.darc` older than `BACKUP_RETENTION_DAYS` (default `7`)
- **Scope**: only `file://` remotes; cloud remotes are skipped
- **Non-fatal**: errors are logged but never propagated — sync reporting is never blocked
- **Cross-platform**: macOS `stat -f` and Linux `stat -c` both supported

Both `BACKUP_SAFETY_FLOOR` and `BACKUP_RETENTION_DAYS` are env-overridable.

## History

Originally submitted as #4040 from `crew/batista/dc-2ajt`, but that branch was
contaminated with ~80 unrelated files (data, mail, beads, etc.) and the PR was
closed. This PR is a clean cherry-pick of just the `plugins/dolt-backup/run.sh`
commit (`0dee05c0`) onto a fresh branch from `main`.

## Test plan

- [x] Bash syntax: `bash -n plugins/dolt-backup/run.sh` passes
- [x] Original author dry-run: correctly identifies 8 backup dirs
- [x] Original author functional test: 5 files (3 new + 2 old@10d) → keeps 3, deletes 2
- [ ] CI on this PR

## Pre-existing test failures (not caused by this PR)

`internal/cmd` fails on `go test ./...` with "This binary was built with 'go build' directly. macOS will SIGKILL unsigned binaries. Use 'make build' instead." This failure exists on `upstream/main` (verified by running the test there). This PR only modifies a shell script and cannot affect Go tests.

### DIFF
--- a/plugins/dolt-backup/run.sh
+++ b/plugins/dolt-backup/run.sh
@@ -4,6 +4,8 @@
 # Syncs production databases to filesystem backups via `dolt backup sync`.
 # Skips databases that haven't changed since last backup (hash check).
 # Only escalates when actual backup operations fail — not on ping failures.
+# After each run, prunes old .darc files keeping at least BACKUP_SAFETY_FLOOR
+# most-recent archives and deleting any older than BACKUP_RETENTION_DAYS.
 #
 # Usage: ./run.sh [--databases db1,db2,...] [--dry-run]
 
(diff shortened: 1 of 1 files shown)


Bella-Giraffety:
Closing this PR as stale and not safe to carry forward directly. The retention need is real and is tracked in #4412, but backup retention should be designed in the current backup path with explicit safety rules, dry-run candidate reporting, invalid-config handling, and tests. We should not merge shell-side deletion of backup archives as-is. If we implement this, we will preserve attribution for the original idea.

PR #4082, opened 2026-05-21
fix(reaper): update depends_on_id refs after column split

drewtyler:
## Summary
- Updates 8 reaper queries in `internal/reaper/reaper.go` and 1 in `plugins/dolt-snapshots/main.go` to use the new typed dependency columns (`depends_on_issue_id`, `depends_on_wisp_id`, `depends_on_external`)
- Splits the single `DELETE FROM wisp_dependencies WHERE depends_on_id IN ...` into two deletes (one per typed column) since the args binding is per-statement

## Why
`wisp_dependencies` and `dependencies` had their `depends_on_id` column split into three typed columns by an earlier migration. The reaper queries still referenced the old name, so every cycle failed with `Unknown column 'depends_on_id'` and blocked all deacon cleanup. Repro from before the fix:

```
gt reaper scan --db=gt2
gt2: scan error: count stale candidates: Error 1054 Unknown column 'depends_on_id'
```

After the fix, `gt reaper scan` runs cleanly on all 5 town databases and surfaces real cleanup work (98 dangling parent refs across gt2 + pyloracle).

## Test plan
- [x] `go build ./...` clean
- [x] `go test ./internal/reaper/...` passes
- [x] `gt reaper scan --db={beads,discordia,gt2,gtvoice,pyloracle}` succeeds on all 5 dbs
- [ ] Maintainer review of the `dependencies` semantics (PR uses `depends_on_issue_id` since `dependencies` always references issues, but please double-check)

### DIFF
--- a/internal/cmd/compact.go
+++ b/internal/cmd/compact.go
@@ -281,9 +281,14 @@ func runCompact(cmd *cobra.Command, args []string) error {
 // but leaves behind its dependency records (bd delete has no cascade logic for
 // the wisp-level tables). Runs as a post-compact sweep.
 func cleanOrphanedWispDeps(bd *beads.Beads, result *compactResult) {
+	// depends_on_id was split into typed columns (depends_on_issue_id,
+	// depends_on_wisp_id, depends_on_external). A row is orphaned if its
+	// wisp-side target ID points at a wisp that no longer exists; rows
+	// targeting issues or external refs are out of scope for this cleanup.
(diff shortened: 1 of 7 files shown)


Bella-Giraffety:
Closing as superseded/contaminated after #4266 and #4277 landed. The reaper/compact typed dependency-column pieces are now on main via focused replacements; this older branch also contains unrelated operational changes and stale assumptions, so any still-needed non-reaper pieces should be rebuilt as fresh focused PRs.

PR #4092, opened 2026-05-21
fix: converge routing sling safeguards

Bella-Giraffety:
## Summary
- Adds same-rig guarded route registration with flock serialization and rollback so tracked source `.beads/config.yaml` prefixes cannot hijack another rig's route.
- Keeps existing source-of-truth routing semantics and direct target-rig sling validation; no repo-specific `do-` or `coder_dotfiles` hardcoding.
- Replaces the narrow formula-on-bead route test with a smoke that creates a rig-prefixed bead and immediately slings it, asserting create, target DB check, formula wisp/bond, hook, and metadata updates all use the route-resolved target rig DB.

## Supersedes
- Supersedes #4085, #4086, and #4088 with one collapsed fork PR.
- Cherry-picked/reimplemented useful pieces from #4086 and #4088.
- Dropped #4085's doctor/design layer from this implementation because it is diagnostic-only and not needed for the failed do-prefix sling class.

## RCA Evidence
- Completed 15 read-only research passes across current target/main, open PR metadata/files/diffs, merged predecessors #3979/#3968/#4020/#3987, and current route/sling/AddRig code.
- Completed pre-implementation reviews for route ownership, AddRig tracked prefix drift, target rig DB guards, bd create routing semantics, and duplicate validation risks.
- Rejected duplicate/fragile paths: no stale `--allow-stale` compatibility changes, no bd create bandaid, no duplicate sling validation layer, no hardcoded repo/prefix handling.

## Validation
- PASS: `make build`
- PASS: `go test ./internal/beads -run 'TestAppendRouteIfPrefixAvailable|TestCheckPrefixAvailable|TestResolveBeadsDirForID|TestGetPrefixForRig' -count=1`
- PASS: `go test ./internal/rig -run 'TestAddRig_TrackedBeadsPrefixCollisionDoesNotRewriteRoute|TestAddRig_RollsBackRouteWhenRegistrationFails|TestAddRig_TrackedBeads|TestAddRig_UpstreamURL|TestAddRig_BranchFlag' -count=1`
- PASS: `go test ./internal/cmd -run '^TestSlingNewlyCreatedRigBeadRoutesBDCommandsToTargetRig$' -count=1 -v`
- NOTE: `go test ./...` was attempted and still fails on existing unrelated suite failures across `internal/cmd`, `internal/doltserver`, `internal/polecat`, `internal/rig`, and `internal/tmux`.

Issue: gt-rca-routing-convergence

PR #4088, opened 2026-05-21
test: cover newly-created rig bead sling routing

Bella-Giraffety:
## Summary
- Add smoke coverage for creating a rig-prefixed bead and immediately running normal sling to a polecat target.
- Assert create, target-rig DB validation, formula wisp/bond, hook update, and attached molecule/formula metadata all use the route-resolved target rig DB.
- Keep tmux/session side effects out of the smoke by stubbing target-agent resolution.

## Test Plan
- PASS: `go test ./internal/cmd -run '^TestSlingNewlyCreatedRigBeadRoutesBDCommandsToTargetRig$' -count=1 -v`
- NOTE: `go test ./internal/cmd -count=1` still fails on existing unrelated package failures (agents/convoy/install/sling stub cases), recorded on bead `gt-rca-canon-new-bead-sling-smoke`.

Issue: gt-rca-canon-new-bead-sling-smoke

PR #4086, opened 2026-05-21
fix: block rig add prefix route hijacks

Bella-Giraffety:
## Summary
- Reject tracked source `.beads/config.yaml` prefixes when they collide with another rig before route registration can rewrite `routes.jsonl`.
- Add checked route registration that scans duplicate prefix entries, serializes checked updates with a route flock, and rolls back AddRig route writes on later registration failure.
- Cover second-rig prefix drift, route rollback, checked append, and duplicate-prefix behavior with targeted tests.

## RCA Evidence
- Confirmed current route values: `gt- -> gastown/mayor/rig`, `do- -> coder_dotfiles/mayor/rig`, `hq-`/`hq-cv- -> .`.
- Confirmed drift path: initial derived prefix check passes, tracked source config overrides to `gt`, and old route append semantics can update the existing `gt-` route to a second rig.
- Persisted full 15-pass research and review evidence on bead `gt-rca-canon-routing-rig-add`.

## Testing
- [x] `make build`
- [x] `go test ./internal/beads -run 'TestAppendRouteIfPrefixAvailable|TestAppendRouteIfPrefixAvailableScansDuplicatePrefixes' -count=1`
- [x] `go test ./internal/rig -run 'TestAddRig_TrackedBeadsPrefixCollisionDoesNotRewriteRoute|TestAddRig_RollsBackRouteWhenRegistrationFails|TestAddRig_TrackedBeads|TestAddRig_UpstreamURL|TestAddRig_BranchFlag' -count=1`

## Known Existing Failures
- `go test ./internal/rig -count=1` fails `TestInitBeadsWritesConfigOnFailure` because config output includes `export.auto: "false"`.
- `go test ./internal/beads -count=1` fails `TestRunEnv_StripsPollutedDoltEnvAndUsesRigMetadata` because polluted env selected prefix `hq`, want `gt`.
- Follow-up filed: `gt-existing-package-test-failures`.

PR #4085, opened 2026-05-21
RCA canonical: design routing repair and migration guard

Bella-Giraffety:
## Summary
- Defines canonical routing/prefix invariants, workspace repair separation, command agreement tests, and fork-PR rollout guidance.
- Adds diagnostic-only `routes.jsonl` shape validation in doctor before `LoadRoutes` can silently skip malformed entries.
- Refuses doctor auto-fix on malformed routes and guards HQ prefixes from being routed away from town root or masking missing rig routes.

## Validation
- `go test ./internal/doctor`
- `go test ./internal/beads -run 'Test(GetPrefixForRig|ResolveBeadsDirForID|ValidateRigPrefix|CreateRoutes|Build.*BDEnv|ArgsAreReadOnly)'`
- `go test ./internal/cmd -run 'Test(NormalizeHookShowTarget|OutputMoleculeStatus|DoneUsesResolveBeadsDir|MRBeadCreationUsesRig|MakeTestMR|VerifyBranch)'`

## RCA Evidence
- Completed 15 independent read-only research passes.
- Completed 5 pre-implementation reviews.
- Completed 5 post-implementation reviews and fixed the HQ route ownership finding before submission.
- Broader sling-targeted `internal/cmd` run exposed existing sling stub/rollback failures outside the changed doctor/doc paths.

PR #4096, opened 2026-05-21
fix: rebuild routing convergence for main

Bella-Giraffety:
## Summary
- Rebuilds the routing convergence fix as a clean `main`-target fork PR.
- Guards rig route registration so tracked source `.beads/config.yaml` prefixes cannot rewrite another rig's route.
- Adds route rollback for failed rig registration and smoke coverage for newly-created rig bead sling routing to the target rig DB.

## Supersedes
- Clean main-target rebuild of #4092.
- Supersedes and replaces #4092, #4086, and #4088.
- #4085 is intentionally not included; its doctor/design diagnostic route validation should remain a separate follow-up unless maintainers decide to drop that diagnostic slice.

## Validation
- PASS: `make build`
- PASS: `go test ./internal/beads -run 'TestAppendRouteIfPrefixAvailable|TestCheckPrefixAvailable|TestResolveBeadsDirForID|TestGetPrefixForRig|TestValidateRigPrefix' -count=1`
- PASS: `go test ./internal/rig -run 'TestAddRig_TrackedBeadsPrefixCollisionDoesNotRewriteRoute|TestAddRig_RollsBackRouteWhenRegistrationFails|TestAddRig_TrackedBeads|TestAddRig_UpstreamURL|TestAddRig_BranchFlag' -count=1`
- PASS: `go test ./internal/cmd -run '^TestSling(FormulaOnBeadRoutesBDCommandsToTargetRig|NewlyCreatedRigBeadRoutesBDCommandsToTargetRig)$' -count=1 -v`

## Closure Recommendation
- Close #4092, #4086, and #4088 as superseded after this lands.
- Keep #4085 open as follow-up, or close it separately as intentionally deferred/dropped if that is the desired product decision.

Issue: gt-pr-main-routing-rebuild

### DIFF
--- a/internal/beads/routes.go
+++ b/internal/beads/routes.go
@@ -77,7 +77,7 @@ func LoadRoutes(beadsDir string) ([]Route, error) {
 }
 
 // AppendRoute appends a route to routes.jsonl in the town's beads directory.
-// If the prefix already exists, it updates the path.
+// If the prefix already exists for the same rig, it updates the path.
 func AppendRoute(townRoot string, route Route) error {
 	beadsDir := filepath.Join(townRoot, ".beads")
 	return AppendRouteToDir(beadsDir, route)
@@ -169,13 +169,16 @@ func RestoreRouteInDirIfCurrent(beadsDir string, route Route, previous *Route) e
 }
 
 // AppendRouteToDir appends a route to routes.jsonl in the given beads directory.
-// If the prefix already exists, it updates the path.
+// If the prefix already exists for the same rig, it updates the path.
 func AppendRouteToDir(beadsDir string, route Route) error {
 	return withRoutesLock(beadsDir, func() error {
 		// Load existing routes
 		routes, err := LoadRoutes(beadsDir)
 		if err != nil {
 			return fmt.Errorf("loading routes: %w", err)
+		}
+		if err := checkPrefixAvailableInRoutes(routes, route.Prefix, route.Path); err != nil {
+			return err
 		}
 
 		// Check if prefix already exists
--- a/internal/beads/routes.go
+++ b/internal/beads/routes.go
@@ -23,6 +23,23 @@ type Route struct {
 // RoutesFileName is the name of the routes configuration file.
 const RoutesFileName = "routes.jsonl"
 
+func routesLockPath(beadsDir string) string {
+	return filepath.Join(beadsDir, RoutesFileName+".flock")
+}
+
+func withRoutesLock(beadsDir string, fn func() error) error {
+	if err := os.MkdirAll(beadsDir, 0755); err != nil {
+		return fmt.Errorf("creating beads directory: %w", err)
+	}
+	unlock, err := gtlock.FlockAcquire(routesLockPath(beadsDir))
+	if err != nil {
+		return fmt.Errorf("acquiring routes lock: %w", err)
+	}
+	defer unlock()
+
+	return fn()
+}
+
 // LoadRoutes loads routes from routes.jsonl in the given beads directory.
 // Returns an empty slice if the file doesn't exist.
 func LoadRoutes(beadsDir string) ([]Route, error) {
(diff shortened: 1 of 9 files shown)


Bella-Giraffety:
Mayor triage: this is the right main-target replacement path for the routing convergence work. Keep #4096 as the replacement for integration-target #4092/#4086/#4088; do not retarget #4092 directly.

Notes while checks are still running: the initial lint failure reran green. If the remaining checks pass, use this PR as the basis for closing/superseding #4092/#4086/#4088. Keep #4085 as a separate doctor/design diagnostics follow-up unless intentionally dropped.

Bella-Giraffety:
CI triage update: Lint passed on rerun. Test and Integration are red, but the visible failures match the known current-main/baseline failures already seen in #4081 review:

- `TestFilterAndSortSessions_SortOrder`
- `TestGuessSessionFromWorkerDir/different_rig`
- `TestJSONOutput_NoHumanReadableText`
- `TestJSONOutput_ErrorsReturnNonZeroExit`

No routing-branch-owned failure is visible from the integration log. Windows is still running at this check. Treat #4096 as blocked on baseline CI unless Windows reveals a branch-owned issue.

Bella-Giraffety:
CI update after latest push:\n\n- Lint now passes after commit `7a191530` cleaned up existing statusline `unparam` findings.\n- Windows Smoke Test passes.\n- Local validation passes: `make build`, full `golangci-lint v2.11.4`, targeted beads/rig/sling/statusline tests.\n- CI `Test` and `Integration Tests` still fail on existing unrelated broad-suite cases outside this PR's touched files: `agents_test.go`, `convoy_stage_test.go`, and `internal/polecat/manager_test.go`. Follow-up bead filed: `hq-8v6`.

Bella-Giraffety:
Mayor follow-up after root-cause routing fan:

- Collapsed duplicate sling routing coverage while preserving explicit `--on` formula routing checks.
- Fixed formula convoy root IDs to be town-scoped `hq-cv-*` instead of rig-prefixed `gt-cv-*` created from town/HQ context.
- Kept executable legs rig-scoped (`gt-leg-*`) and verified town convoy -> rig leg tracking uses town root.
- Replaced raw convoy `bd create` / failure-comment paths with `BdCmd` using explicit beads dirs.

Validation passed:
- `go test ./internal/cmd -run 'TestFormulaConvoyIDUsesTownConvoyPrefix|TestExecuteConvoyFormulaCreatesTownConvoyAndRigLegs|^TestSlingNewlyCreatedRigBeadRoutesBDCommandsToTargetRig$' -count=1 -v`
- targeted `internal/beads` route tests
- targeted `internal/rig` add/rollback tests
- `go build ./cmd/gt`
- `git diff --check`

Post-review gate: 5 approvals after fixes. Net diff in this follow-up commit reduces LoC.

esciara:
**Review: routing convergence fix**

The core fix is correct and worth landing — the silent route hijack (old `AppendRouteToDir` treating a prefix collision as "update the path") is a data-corruption-grade bug, and replacing overwrite-semantics with a same-rig guard + hard-fail + rollback is the right call. A few things I'd like addressed, two of them ideally before merge.

**What's good**
- Same-rig guard is the right invariant, and turning the old "Warning: could not update routes.jsonl" into a hard error + rollback closes a real hole where a rig could end up registered with no route.
- `RestoreRouteIfCurrent` is nicely concurrency-aware — it only rolls back if the route still matches what this call wrote, so it won't clobber a concurrent repair.
- Good test coverage (collision-doesn't-rewrite, rollback-on-failure, duplicate-prefix dedup).
- Convoy-ID town-scoping (`hq-cv-` vs `gt-leg-`) is conceptually right, and routing the leg-comment path to the rig DB is a genuine secondary fix.

**Please address (blocking-ish)**
1. **Partial locking.** Only `AppendRouteToDirIfPrefixAvailable` / `RestoreRouteInDirIfCurrent` take the flock. `AppendRouteToDir`, `RemoveRoute`, and `WriteRoutes` still lock nothing — so the old append path and rig-removal can still race the new path, since they don't share a lock. A lock only some writers honor gives false confidence. Either route all mutations through the lock or document explicitly why the others are safe.
2. **The invariant is now implemented three times** — `CheckPrefixAvailable`, the new `AppendRouteToDirIfPrefixAvailable`, and the explicit `AddRig` pre-check all carry the same cross-rig loop and the same error string verbatim. Worth collapsing to one source of truth before it drifts.

**Nits / follow-ups (non-blocking)**
3. **Contradictory error text.** The collision error suggests "use `--prefix` to specify a different prefix," but on the tracked-source path the adjacent mismatch guard *forces* `--prefix` to match the source repo — so the suggested remedy is impossible for the path that most often triggers it.
4. **Asymmetric guarding.** The fail-fast `CheckPrefixAvailable` is only wired into the tracked-source branch; the derived-prefix path relies solely on the late append-time guard. Two paths, two different error experiences.
5. **Scope.** The `statusline.go` `unparam` cleanup is pure lint appeasement unrelated to routing, and the convoy-ID change is a separate bug — both would be cleaner as their own PRs. Not worth re-cutting at this point, just flagging for next time.

**One worth considering:** the rollback machinery only exists because the route is written *before* rigs.json registration. If route registration were the *last* fallible step in `AddRig`, there'd be nothing to compensate — simpler than the deferred-closure rollback. Reasonable as a future refactor.

**On CI:** I verified the baseline-failure claim. All 7 flagged tests fail identically on a clean `upstream/main` checkout with none of this PR's changes applied — `TestFilterAndSortSessions_SortOrder`, `TestGuessSessionFromWorkerDir/different_rig`, `TestJSONOutput_NoHumanReadableText`, `TestJSONOutput_ErrorsReturnNonZeroExit` (internal/cmd), and `TestReuseIdlePolecat_KillsLiveSession` / `TestReuseIdlePolecat_KillsStaleSession` (internal/polecat). They are pre-existing and not introduced by this branch, so they shouldn't block merge — but they do mask any real regression behind a red suite, which is its own problem worth tracking (hq-8v6).

Bella-Giraffety:
Review disposition update after commit `665a764c`:

Pre-implementation review passes confirmed one remaining branch-owned blocker: `AppendRoute`/`AppendRouteToDir` had gained the shared route lock but still preserved the old cross-rig overwrite semantics, so adoption/install callers could bypass the same-rig prefix invariant.

Post-implementation review passes:
- Route mutation locking: all exported route writers that modify `routes.jsonl` now use the shared `routes.jsonl.flock` path.
- Prefix invariant: public `AppendRoute`/`AppendRouteToDir` now call the same `checkPrefixAvailableInRoutes` helper under the lock; same-rig path variants are still allowed, different-rig prefix rewrites now fail.
- AddRig/adopt rollback: AddRig rollback behavior is unchanged; adoption/install callers using public append now get the same collision protection.
- Convoy/sling scope: hq-cv town roots, gt-leg rig leaves, town-root tracking, rig-scoped failure comments, and route-resolved sling visibility remain unchanged.
- Closure: the two blocking review concerns (partial locking and duplicated/bypassed invariant) are addressed on the PR branch.

Validation:
- PASS: `make build`
- PASS: `git diff --check origin/main...HEAD`
- PASS: `go test ./internal/beads -run 'TestAppendRouteRejectsCrossRigPrefixRewrite|TestAppendRouteIfPrefixAvailable|TestCheckPrefixAvailable|TestWriteRoutesWaitsForRoutesLock' -count=1`
- PASS: `go test ./internal/rig -run 'TestAddRig_TrackedBeadsPrefixCollisionDoesNotRewriteRoute|TestAddRig_RollsBackRouteWhenRegistrationFails' -count=1`
- PASS: `go test ./internal/cmd -run 'TestFormulaConvoyIDUsesTownConvoyPrefix|TestExecuteConvoyFormulaCreatesTownConvoyAndRigLegs|^TestSlingNewlyCreatedRigBeadRoutesBDCommandsToTargetRig$' -count=1`
- BLOCKED locally: `go test -tags=integration ./internal/cmd -run '^TestBeadsAppendRoute$' -count=1` aborts in integration `TestMain` before the target test because this sandbox has no rootless Docker. Equivalent unit coverage was added in `internal/beads`.
- NOT RUN locally: `golangci-lint run` because `golangci-lint` is not installed in this sandbox.

CI note: existing PR `Test` / `Integration Tests` failures remain the documented current-main baseline failures outside this routing branch; tracked separately as `hq-8v6`.

Bella-Giraffety:
Delivery update per Mayor strategy correction:

- No new upstream PR was opened for this leaf.
- Leaf branch remains pushed at `fork/polecat/guzzle/main-routing-convergence` (`665a764c`).
- Leaf branch was merged into the routing identity aggregation branch `fork/integration/gt-1.2/routing-identity`, now pushed at `81324253`.
- Mayor will package the single upstream-main PR for the routing-identity subepic.

Aggregation branch validation after merge:
- PASS: `make build`
- PASS: `git diff --check fork/integration/gt-1.2/routing-identity...HEAD`
- PASS: `go test ./internal/beads -run 'TestAppendRouteRejectsCrossRigPrefixRewrite|TestAppendRouteIfPrefixAvailable|TestCheckPrefixAvailable|TestWriteRoutesWaitsForRoutesLock' -count=1`
- PASS: `go test ./internal/rig -run 'TestAddRig_TrackedBeadsPrefixCollisionDoesNotRewriteRoute|TestAddRig_RollsBackRouteWhenRegistrationFails' -count=1`
- PASS: `go test ./internal/cmd -run 'TestFormulaConvoyIDUsesTownConvoyPrefix|TestExecuteConvoyFormulaCreatesTownConvoyAndRigLegs|^TestSlingNewlyCreatedRigBeadRoutesBDCommandsToTargetRig$' -count=1`

Latest #4096 CI status remains consistent with baseline documentation: Lint and Windows pass; Test/Integration fail on existing current-main cases outside this routing branch.

PR #4117, opened 2026-05-23
fix: add dashboard pacing config

fengning-starsend:
## Summary

Adds two optional dashboard pacing knobs under the existing `web_timeouts` config:

- `dashboard_cache_ttl` controls the minimum interval between full dashboard fetches.
- `dashboard_sse_interval` controls the dashboard SSE change-detection interval.

Defaults preserve the current behavior (`10s` cache TTL and `2s` SSE interval). Empty, invalid, zero, and negative values fall back to the existing defaults.

## Why

Busy dashboards can generate repeated subprocess-backed fetch bursts when several refresh paths overlap. This keeps the default live-dashboard behavior unchanged while giving larger or busier towns an operator-controlled way to reduce dashboard polling/fetch pressure without patching source.

This is intentionally transport/configuration only: it does not add BD Symphony-specific behavior, heuristics, or agent decision logic.

## Tests

- `go test ./internal/config ./internal/web` ✅
- `go build -o /tmp/gt-dashboard-pacing ./cmd/gt` ✅
- `go test ./internal/cmd -run 'TestFilterAndSortSessions_SortOrder|TestGuessSessionFromWorkerDir|TestJSONOutput_NoHumanReadableText|TestJSONOutput_ErrorsReturnNonZeroExit' -count=1` ❌ reproduces unrelated existing/local failures outside this patch:
  - `TestFilterAndSortSessions_SortOrder`
  - `TestGuessSessionFromWorkerDir/different_rig`
  - `TestJSONOutput_NoHumanReadableText`
  - `TestJSONOutput_ErrorsReturnNonZeroExit`
- `go test ./...` ❌ attempted, but local full-suite is blocked by unrelated `internal/cmd`/`cmd/gt` failures/timeouts outside the touched dashboard config/web packages.

Agent: codex
Feature-Key: bd-a6olj.12.1

### DIFF
--- a/docs/examples/town-settings.example.json
+++ b/docs/examples/town-settings.example.json
@@ -81,8 +81,10 @@
         "gh_cmd_timeout": "10s",
         "tmux_cmd_timeout": "2s",
         "fetch_timeout": "8s",
+        "dashboard_cache_ttl": "10s",
+        "dashboard_sse_interval": "2s",
         "default_run_timeout": "30s",
-        "max_run_timeout": "60s"
+        "max_run_timeout": "120s"
     },
 
     "worker_status": {
--- a/internal/cmd/dashboard.go
+++ b/internal/cmd/dashboard.go
@@ -81,10 +81,12 @@ func runDashboard(cmd *cobra.Command, args []string) error {
 			return fmt.Errorf("creating convoy fetcher: %w", fetchErr)
 		}
 
-		// Load web timeouts config (nil-safe: NewDashboardMux applies defaults)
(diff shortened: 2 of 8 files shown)


Bella-Giraffety:
Closing this PR as stale while we decide the right dashboard direction. The pacing/config idea is now tracked in #4411, including a request to research the Gas City dashboard and determine whether we should wire that model into Gastown instead of adding permanent tuning knobs. If we decide explicit pacing controls are still the right answer, we should rebuild a small current-main PR and preserve attribution.

Issue #3955, opened 2026-05-12
gt sling: nudge-pollers spawned for Claude sessions + convoy close not durable (bd jsonl re-import loop causes notification flood)


prowe214:
## Bug Description

  Two related bugs that together make the mayor session unusable: a **notification flood** caused by non-durable convoy close operations, and **auto-injection of text into the operator's input** caused by `gt sling` spawning nudge-pollers for Claude Code sessions that already have the `UserPromptSubmit` hook.

  ---

  ## Bug 1: Convoy close is not durable — overseer floods inbox with duplicate "Convoy complete" notifications

  When a convoy's tracked issue is closed, the overseer sends repeated "Convoy complete" notifications to the mayor indefinitely (hundreds per session), rather than once. Root cause: both the `hq` and rig (e.g. `projectname`) Dolt databases are initialized but contain no data chunks. On every `bd` invocation, `bd` detects an empty database and re-imports from `.beads/issues.jsonl`. Since the jsonl is a periodic export, it lags behind real state. Any `bd close` call writes to the freshly-imported in-memory database, but on the next `bd` invocation the database is empty again and the stale jsonl re-imports, reverting the close. The overseer monitors convoy state via `bd`, not gt's internal convoy registry — so it sees the convoy bead as perpetually open and re-notifies on every patrol cycle.

  Note: `gt convoy close` alone is not sufficient. It updates gt's internal convoy registry but not the `bd` record. The overseer reads `bd`, so the flood continues until the `bd` record is also closed durably.

  ---

  ## Bug 2: `gt sling` spawns `gt nudge-poller` for Claude Code sessions that already have `UserPromptSubmit` hook configured

  After running `gt sling <bead> <rig>`, background processes `gt nudge-poller gr-refinery` and `gt nudge-poller gr-witness` are spawned even though both sessions have `UserPromptSubmit: gt mail check --inject` configured in their `.claude/settings.json`. The nudge-poller help text explicitly states it is "for non-Claude agents" and "not intended for direct user invocation." These pollers inject text directly into the operator's tmux pane and auto-submit it as user input — interrupting typing and hijacking the session. The pollers are respawned on every subsequent `gt sling`, and killing them manually only helps until the next sling.

  ---

  ## Steps to Reproduce

  **Bug 1:**
  1. Dispatch a bead to a rig polecat: `gt sling <bead> <rig>` (creates an auto-convoy)
  2. Polecat completes work and the tracked bead is merged/closed
[...]

PR #4118, opened 2026-05-24
fix(config): set BEADS_NO_AUTO_IMPORT=1 for all agents in AgentEnv

KayoticSully:
## Disclaimer
I'm opening this since I needed to do this in my gastown to fix an issue I was having with zombie tasks. However, I could be using something wrong which leads to this setting being set.  

I was essentially trying to solve Bug 1 from this issue since I ran into it as well: https://github.com/gastownhall/gastown/issues/3955 but am not sure if this is the correct way to go about it.  Curious about opinions on this change and if there are any reasons for `issues.jsonl` to be re-imported?

## Problem

Agents' **direct** `bd` calls auto-import the per-repo `.beads/issues.jsonl` into the working DB on every invocation (`auto-importing N bytes ... into empty database`). The export back to JSONL is throttled, so the file frequently lags the live Dolt server. The stale re-import then **reverts server-side status changes and resurrects deleted beads**.

Observed in production: `hooked` handoff beads that refuse to clear — `bd close`/`gt unsling` flips them to closed on `:3307`, but the next `bd` call re-imports the stale JSONL (still showing `hooked`) and re-hooks them. Independently reproduced for the deacon and a rig refinery; closing a deleted probe bead, then watching a later `bd` call re-insert it, confirmed the mechanism (the resurrected row kept its *old* `updated_at` — the import signature).

## Root cause

`gt` already knows this is harmful — `beads.SuppressBDSideEffects()` sets `BEADS_NO_AUTO_IMPORT=1` (+ export/push off) for the `bd` subprocesses **gt itself** spawns, with the comment that JSONL export/import from high-frequency callers "re-invalidates Beads' import freshness checks and can create a self-feeding Dolt load loop." But this env was **not** applied to agents' own interactive/direct `bd` calls.

## Fix

Set `BEADS_NO_AUTO_IMPORT=1` in `AgentEnv()` — the single source of truth for agent env — right next to the existing `BD_BACKUP_ENABLED=false`, so every role (mayor, deacon, witness, refinery, polecat, crew, dog) inherits the same policy. Dolt (`:3307`) is the authoritative data plane; `issues.jsonl` is a passive export and must not be silently re-imported over it.

## Test

Added `TestAgentEnv_SuppressesBeadsAutoImport` asserting the var is set across all roles. `go test ./internal/config/` passes; `go build ./cmd/gt` clean.

### DIFF
--- a/internal/beads/exec_test.go
+++ b/internal/beads/exec_test.go
@@ -1,0 +1,61 @@
+package beads
+
+import (
+	"strings"
+	"testing"
+)
+
+// TestEnvForSubprocessMode_SuppressesAutoImport guarantees that every bd
+// subprocess mode used by ConfigureCommand sets BEADS_NO_AUTO_IMPORT=1 — and
+// strips any inherited value that would re-enable it. The daemon dog patrols
+// (dog_molecule.runBd) and other migrated raw bd call sites route through this
+// helper, so this is the regression guard for the self-feeding "create N
+// issue(s)" loop where watchdog/compaction dogs auto-imported a stale
+// .beads/issues.jsonl over the live Dolt server.
+func TestEnvForSubprocessMode_SuppressesAutoImport(t *testing.T) {
+	t.Parallel()
+	// Hostile base: explicitly turns auto-import ON, to prove we override it.
+	base := []string{"BEADS_NO_AUTO_IMPORT=0", "PATH=/usr/bin"}
+	for name, mode := range map[string]SubprocessEnvMode{
+		"ReadOnlyRouting": ReadOnlyRouting,
+		"MutationRouting": MutationRouting,
+		"ReadOnlyPinned":  ReadOnlyPinned,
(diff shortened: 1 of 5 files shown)


KayoticSully:
Closing — root cause was a **beads issue**, not gt/env.

The bd auto-import churn this targeted is bd's GH#2994 "import .beads/issues.jsonl into an empty embedded DB on upgrade" path firing on writes in a shared-server setup. It exists in the bd **1.0.4 release** but is **already fixed in beads main** (the `if serverMode { return }` + non-empty `GetStatistics` guards in cmd/bd/auto_import_upgrade.go).

This gt-side env-var approach (AgentEnv BEADS_NO_AUTO_IMPORT + watchdog-path suppression) targets the wrong layer — verified ineffective against the churn, while installing a beads build with the GH#2994 guards dropped auto-import ~25-30/min → 0/min. No gt change is needed; superseded by the beads fix (pending a beads release ≥ it).

PR #4125, opened 2026-05-26
fix(reaper): schema-aware dependency column detection

Jacob-qd:
Fixes reaper schema mismatch (hq-4p83). The reaper now detects whether the database uses the new schema (depends_on_wisp_id/depends_on_issue_id) or old schema (depends_on_id) and builds queries accordingly.

Changes:
- Add depCols struct and detectTableDepColumns() for runtime schema detection
- Add depTargetExpr() for COALESCE-based dependency expressions
- Update parentExcludeJoin() to accept *depCols
- Update Scan() to detect schema before building queries
- Update tests to match new API

### DIFF
--- a/internal/agent/provider/acp.go
+++ b/internal/agent/provider/acp.go
@@ -1,6 +1,7 @@
 package provider
 
 import (
+	"bytes"
 	"encoding/json"
 	"fmt"
 	"time"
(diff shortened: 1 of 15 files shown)


Bella-Giraffety:
Superseded by the merged cleanup-first path in #4266/#4277. Main now converges on the typed dependency schema instead of adding runtime legacy-column detection/fallbacks. Closing this stale schema-aware fallback PR.

PR #4129, opened 2026-05-27
fix: stuck-agent-dog NOTICE-only on deacon heartbeat staleness (#3877)

### LINKED ISSUE #3877: stuck-agent-dog: false-positive Deacon stuck_heartbeat escalations (~18d delta) on patrol cycles
## Repro

In a 30-minute window, dogs/kilo (once) and dogs/echo (twice) fired HIGH-severity `Deacon stuck_heartbeat_<seconds>` escalations against a deacon that is in fact alive:

- `gt status` → `🐺 deacon ● [claude]`
- `gt heartbeat status deacon` → `Heartbeat updated: state=working`

The reported delta values cluster around ~1,578,500–1,579,500 seconds (~18 days). That's wall-clock-stable across the three reports and does not reflect actual deacon idle time, so the dog is comparing against a stale stored heartbeat field rather than current liveness.

## Likely cause

The patrol scan's heartbeat lookup reads a field that isn't being refreshed (or is refreshed on a different storage path than the dog reads). Suggests:
- Heartbeat write goes to one location (e.g. agent bead or runtime state file)
- Patrol reads from a different one
- Or: the field IS updated but the dog's freshness comparison uses wrong epoch base

## Impact

- HIGH-severity escalations spam mayor inbox
- Each fires a system nudge demanding a reply, blocking real work
- Across rigs this could cause real escalations to be missed in the noise

## False-positive instances closed in this session

- hq-wisp-4z7 (dogs/kilo, 2026-05-07 20:42)
- hq-wisp-wfk (dogs/echo, 2026-05-07 20:55)
- hq-wisp-dq1 (dogs/echo, 2026-05-07 21:01)

## Proposed fix

1. Audit: which storage path `gt heartbeat` writes to vs which the dog reads.
2. If divergent: route both through the same source.
3. If unified and the field is genuinely 18 days stale, find what stopped writing.
4. Add tighter sanity check in the dog: cross-reference with `gt status` (claude session running?) before firing HIGH.

## Severity

P2 — annoying, not blocking. Real escalations still get through; this adds noise.

csauer02-personal-user:
## Summary

Closes #3877.

The `stuck-agent-dog` plugin escalated **HIGH** on Deacon heartbeat staleness
alone. But the Deacon is event-driven and parks during await-signal idle, so a
stale heartbeat usually just means there was no work to do — not that the Deacon
is stuck or dead. The Go daemon already owns the nudge/kill decision on heartbeat
staleness and correctly **idle-guards** it (skips when no work is in flight —
`internal/daemon/daemon.go:1600`; threshold rationale in
`internal/deacon/heartbeat.go`). The plugin duplicated the staleness check
*without* that guard, producing recurring false HIGH escalations.

## Changes

- **`plugins/stuck-agent-dog/run.sh`**: the heartbeat-staleness branch now logs a
  `NOTICE` only (threshold `7200`s) and no longer sets `DEACON_ISSUE`, so it never
  escalates. The crashed / zombie session+process branches and the HIGH escalation
  for those are unchanged.
- **`plugins/stuck-agent-dog/plugin.md`**: prose + decision framework updated to
  match — heartbeat age alone never escalates; only a dead tmux session or dead
  process escalate; the daemon owns the idle-guarded nudge/kill.
- **`internal/templates/roles/deacon.md.tmpl`**: added 4 durable ops lessons —
  notification-only senders (daemon, `gt-sling LIFECYCLE:Shutdown`) have no reply
  route; route escalations via `gt nudge hq-mayor` when mail/escalate hang; check
  `uptime` before restarting Dolt on `signal: killed` (load >20 = OS killed the
  subprocess, not a Dolt fault); `stuck-heartbeat`/`wisp-threshold` escalations are
  recurring noise (this PR removes the stuck-heartbeat source itself).

## Testing

- `go build -o /tmp/gt ./cmd/gt` — passes.
- `go test ./internal/daemon/...` (incl. `deacon_idle_guard_test.go`, which already
  covers the Go idle-guard — no Go change here) — passes.
- `go test ./internal/templates/...` — passes.
- `bash -n` syntax check on `run.sh` — passes. There is **no plugin/shell test
  harness** in the repo (no `.bats`, no shellcheck wiring), so no plugin test was
  added per the plan.

### DIFF
--- a/internal/templates/roles/deacon.md.tmpl
+++ b/internal/templates/roles/deacon.md.tmpl
@@ -288,6 +288,11 @@ gt mail delete <id>              # ALWAYS delete after handling
 
 **Handoff messages** (`🤝 HANDOFF:`) — read for situational awareness, then delete immediately.
 
+**Notification-only senders have NO reply route.** Mail from the daemon plugin
+and `gt-sling` `LIFECYCLE:Shutdown` is notification-only — handle it and delete
+it, do NOT reply (their inbox shows empty when you check it, and `gt mail send`
+to them errors with "no agent found"). Replying just queues a dead nudge.
+
 ## Lifecycle Request Handling
 
 **Subject format**: `LIFECYCLE: <identity> requesting <action>`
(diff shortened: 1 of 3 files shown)


Bella-Giraffety:
Closed as superseded by merged PR Sheriff replacement MR gt-wisp-k03, landed at ff8c0f510fde117fcc2d01bb0fc1a92b06c39609. Thank you for the original contribution.

PR #4131, opened 2026-05-27
fix(done): rebase even when --pre-verified to avoid stale-base CI failures (gs-4bn)

gagecane:
## Summary

Polecat branches that pushed with `--pre-verified` were skipping the auto-rebase in `gt done`, leaving the polecat branch with stale workflow files (e.g., an old `ci.yml` from before a CI fix landed on origin). The branch then arrived at GitHub and CI failed on the PR for reasons unrelated to the polecat's work. Multiple polecats hit this pattern (slit, rictus) per the gs-4bn report.

This change drops the `--pre-verified` short-circuit in `autoRebaseOnTarget`: we always rebase when behind. The pre-verified attestation is now stale after the rebase, so the caller (`runDone`) suppresses the `pre_verified` metadata on the MR bead — refinery will run gates normally instead of fast-pathing on a stale claim. `donePreVerified` itself stays set so `pushForDone` still skips the local pre-push hook, preserving the gu-d416 witness-timeout mitigation.

## Why

- **Before:** `--pre-verified` → `autoRebaseOnTarget` returns `skipReason: "--pre-verified is set"` → branch pushed with stale workflow → GitHub CI fails on unrelated dependency.
- **After:** `--pre-verified` does not short-circuit; rebase runs whenever the branch is behind. Stale workflow files are replaced by what's on origin before the push.

The refinery already handles stale pre-verification metadata correctly (`ProcessMRInfo` checks `targetHead == mr.PreVerifiedBase` and falls through to normal gates), so there is no refinery change needed — we just stop *advertising* a now-stale attestation.

## Test plan

- [x] `go test ./internal/cmd/ -run TestAutoRebase` — gating tests updated for new behavior
- [x] `go test ./internal/cmd/` — full cmd package green
- [x] `go test ./internal/refinery/` — refinery unchanged
- [x] `go build ./...`
- [x] `golangci-lint run ./internal/cmd/...` — clean
- [ ] In-the-wild verification on a polecat session whose branch base advanced during work

Refs: gs-4bn


gagecane:
Moving PR to gagecane/gastown per overseer direction.

PR #4137, opened 2026-05-28
chore: suppress empty wisp-compaction digest mail

reppam:
## Summary
Gate `sendCompactDigest` in `internal/cmd/compact_report.go` on a new `reportIsNoteworthy(report)` helper. Email is sent only when at least one of the following is true:
- any category has `Deleted > 0` or `Promoted > 0`
- `len(report.Promotions) > 0`
- `len(report.Anomalies) > 0`
- `len(report.Errors) > 0`

## Why
Daily wisp compaction always emailed mayor/ even when the cycle was a no-op, filling the inbox with routine "nothing happened" reports. The audit bead is still created on every cycle so the historical record is preserved.

When suppressed, stdout prints a dim `suppressed (no activity)` line so daemon-side observation still shows the cycle ran.

Tracks: gst-3ek.

## Test plan
- [x] `go build ./internal/cmd/...` clean
- [x] Local rebuild + daemon restart; subsequent no-op compactions no longer mail
- [ ] Consider adding a unit test for `reportIsNoteworthy` (kept out of this PR to stay minimal; happy to follow up)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,76 @@
+# Project Instructions for AI Agents
+
+This file provides instructions and context for AI coding agents working on this project.
+
+<!-- BEGIN BEADS INTEGRATION v:1 profile:minimal hash:7510c1e2 -->
+## Beads Issue Tracker
+
+This project uses **bd (beads)** for issue tracking. Run `bd prime` to see full workflow context and commands.
+
+### Quick Reference
+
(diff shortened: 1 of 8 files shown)


reppam:
Closing — landed on the user's fork (reppam/gastown:main) directly. Not pursuing upstream merge at this time.

PR #4150, opened 2026-05-30
fix(tmux): guard NewSession against socket-clobber on unresponsive servers (gt-h9z)

Rome-1:
## Summary

Closes gt-h9z. When the configured `-L` socket path is held by a server that tmux's internal liveness probe fails to reach inside its tight timeout, tmux's recovery path unlinks the socket and binds a fresh server at the same path. The original server's listener stays on the kernel fd table — so **two `tmux: server` processes end up simultaneously bound to the same socket path with different inodes**, and any client invoked from a non-tmux shell routes to the new server (which sees none of the old sessions). gt-h9z captured `ss -lx` evidence of this state where a user "lost" their existing pane because the visible server was the post-clobber one.

## Approach

Two-stage health probe in front of every `new-session` call on a named socket:

1. **Kernel-level Unix-socket dial.**
   - `ECONNREFUSED` → file is a benign leftover (tmux exited cleanly, left the inode). tmux will recreate. **Proceed.**
   - Dial succeeds → listener is alive at the kernel layer; advance to stage 2.
   - Other dial error (`ENOTSOCK`, timeout, etc.) → file at the socket path is unsafe. **Bail.**
2. **App-level `list-sessions` probe.**
   - Succeeds → server is responsive. **Proceed.**
   - `ErrNoServer` → server shut down between stages 1 and 2. **Proceed.**
   - Other error → app-layer wedge; this is the gt-h9z window. **Bail.**

Gates all three NewSession entry points (`NewSession`, `NewSessionWithCommand`, `NewSessionWithCommandAndEnv`) so every "create a new tmux session" entry point on a named socket is covered. Default-socket `Tmux` (no `-L`) is left alone — that path is shared with the user's interactive tmux and outside the gt-crew-start workflow that produced the captured evidence.

## Replaces #4041

This PR replaces gastownhall/gastown#4041, which is being closed. #4041's branch carried the same fix in commit 8f9ebab2 but accumulated 76,791 lines of accidentally-committed `.beads/backup/*.jsonl` exports plus 2,398 lines of unrelated polecat-reconcile/config-loader work from a sibling commit on the branch. This PR is the clean cherry-pick — only `internal/tmux/tmux.go` (+85) and `internal/tmux/tmux_test.go` (+139), rebased onto current `gastownhall/gastown:main`.

## Test plan

- [x] `go build ./internal/tmux/...` clean
- [x] `go vet ./internal/tmux/...` clean
- [x] `TestProbeServerHealth` (5 subtests covering default socket / absent socket / stale unix socket / non-socket file / live server) PASS
- [x] `TestNewSessionAllowsStaleUnixSocket` (the previously-broken earlier-draft failure mode — NewSession must succeed against an unbound socket so tmux can recreate the server cleanly) PASS
- [x] All `TestNewSession*` and `TestNewSessionWithCommand*` regression suite PASS
- [x] Pre-existing wall-clock-bound flakes `TestAcceptWorkspaceTrustDialog_NoDialog` and `TestAcceptBypassPermissionsWarning_NoDialog` reproduce on pristine `origin/main` and are unrelated.

### DIFF
--- a/internal/tmux/tmux.go
+++ b/internal/tmux/tmux.go
@@ -6,6 +6,7 @@ import (
 	"context"
 	"errors"
 	"fmt"
+	"net"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -15,6 +16,7 @@ import (
 	"strconv"
 	"strings"
 	"sync"
+	"syscall"
 	"time"
 
 	"github.com/steveyegge/gastown/internal/config"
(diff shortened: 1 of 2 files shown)


Bella-Giraffety:
PR Sheriff update for gt-pr-sheriff-4150:

Research and review:
- Ran 15 independent research legs. Consensus: PR is focused and policy-eligible, but not mergeable as-is due to unbounded app-level `list-sessions`, unsafe non-socket classification, missing unresponsive-listener coverage, and duplicated guard placement.
- Ran 5 pre-implementation reviews. Consensus approved a cleanup-first fix: strict `Lstat`/`ModeSocket` validation, context-bounded app probe, one shared new-session creation path, and deterministic socket tests.
- Ran final 5 post-implementation reviews after fix-ups. No blockers remained.

Fix-up implemented on the sheriff branch:
- Centralized all three `NewSession*` paths through one guarded `createNewSession` helper.
- Added fail-closed named-socket safety checks: reject non-sockets/symlinks/dirs, allow absent/stale sockets, and leave default/Windows paths untouched.
- Bounded the app-level `list-sessions` probe and used file-backed stderr/stdout so cancellation is not defeated by tmux child fd inheritance.
- Rechecked `ErrNoServer` after a live socket dial so live closing/protocol-incompatible listeners cannot be treated as benign stale sockets.
- Added tests for real stale Unix sockets, non-socket path types, live closing listeners, unresponsive listeners, and end-to-end `NewSession` refusal without replacing the socket path.

Verification:
- `go build ./...` passed.
- `go vet ./...` passed.
- `go test ./internal/tmux -run 'TestEnsureNewSessionSocketSafe|TestNewSessionAllowsStaleUnixSocket' -count=1 -timeout 120s` passed.
- Same targeted test run with `-count=3` passed.
- Full `go test ./internal/tmux` still has baseline command-detection failures in this environment (`sleep` reports as `coreutils`); two failures reproduce on pristine `origin/main`, so this branch should not be treated as full-suite pre-verified locally.

Bella-Giraffety:
Closing as superseded by merged replacement PR #4415 (merge commit c1949cd51aca915ea1569d1e43691078fb91b983). #4415 carried forward this PR's tmux NewSession socket-clobber guard intent on current main, converged all NewSession variants through one guarded creation path, passed full GitHub CI, and had PR Sheriff merge-gate evidence with merge_path_allowed=true. Rome-1 attribution/provenance was preserved in the replacement commit via Original-PR/Original-Commit trailers and Co-authored-by: Rome Thorstenson <romethorstenson@gmail.com>.

PR #4151, opened 2026-05-30
fix(release): reject unsupported bd versions

julianknutsen:
## Summary
- add an upper compatibility bound for bd at v1.0.4 on the release branch
- hard-fail normal gt startup when bd is newer than the supported schema version
- preserve hot-path/emergency exemptions so status-line, signal, estop, thaw, metrics, and heartbeat avoid bd subprocess work
- add focused tests for version classification, hard-fail behavior, and hot-path exemption

## Validation
- go test ./internal/deps ./internal/cmd -run 'Test(ParseBeadsVersion|CompareVersions|CheckBeadsVersionString|CheckBeadsVersion|PersistentPreRun)' -count=1\n- go test ./internal/cmd -run '^$'\n- git diff --check

### DIFF
--- a/internal/cmd/beads_version.go
+++ b/internal/cmd/beads_version.go
@@ -2,6 +2,7 @@
 package cmd
 
 import (
+	"errors"
 	"fmt"
 	"sync"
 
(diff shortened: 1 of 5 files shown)


julianknutsen:
Thanks again for the contribution, @julianknutsen. This was a strong PR: it added a release-branch guard so Gastown rejects unsupported `bd` versions before a newer beads schema can create release recovery problems. The review found the core version-ceiling approach was useful and worth carrying forward once the recovery paths, doctor reporting, and vet issue were fixed.

I opened #4159 as a maintainer follow-up because this workflow needed to carry the change set through a latest-base rebase and maintainer fixups, and the original PR was already closed before finalization. The follow-up preserves this PR's contribution and adds the maintainer-side fixes needed to get it through review, local gates, CI, and merge. The original PR was already closed, so #4159 is now the single active merge surface for this contribution, not a rejection of the original work.

PR #4155, opened 2026-05-30
fix(cli): correct 'gt agents state' help to plural command form (hq-8v3)

reppam:
Fixes hq-8v3. The `state` subcommand is registered on `agentsCmd` (plural) — real invocation is `gt agents state` — but the help/examples used singular `gt agent state`, which errors with 'unknown command' when copied. The deacon patrol idle-reset (`gt agent state hq-deacon --set idle=0`) consequently never ran, so the idle counter never reset and backoff stayed elevated.

Fixed all 9 examples in `agent_state.go` + the reset-hint comment in `molecule_await_signal.go`. Build green.

**Follow-up:** the `mol-deacon-patrol` formula step text lives in the beads DB formula (data, not code) and may need a separate `bd formula` update to match.

Authored directly by Mayor — dispatch was unavailable (sling-autospawn-fail: hq-8v3 was in JSONL but not the live Dolt DB, so no polecat spawned).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -1,0 +1,76 @@
+# Project Instructions for AI Agents
+
+This file provides instructions and context for AI coding agents working on this project.
+
+<!-- BEGIN BEADS INTEGRATION v:1 profile:minimal hash:7510c1e2 -->
+## Beads Issue Tracker
+
+This project uses **bd (beads)** for issue tracking. Run `bd prime` to see full workflow context and commands.
+
+### Quick Reference
+
+```bash
+bd ready              # Find available work
+bd show <id>          # View issue details
+bd update <id> --claim  # Claim work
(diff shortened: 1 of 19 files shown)


Bella-Giraffety:
PR Sheriff cleanup: closing as superseded/already fixed. The narrow bug intent is valid, but current main already uses the plural command form in the affected help surfaces: internal/cmd/agent_state.go and internal/cmd/molecule_await_signal.go. The follow-up formula text also uses gt agents state. Evidence points to superseding main commit 602893fb5965bcf07a976d7922737a21f548fbc2. This PR remains status/review-failed and dirty/conflicting with unrelated bundled commits, so it should not merge as-is or via replacement. Original contribution credited to reppam via PR #4155 / cbbb524e9d24dda80f4f071d4d4cc856be2d3b2d.

PR #4164, opened 2026-05-31
fix(tmux): wake agent's window, not session's active window, on nudge

seanbearden:
## Problem

`NudgeSessionWithOpts` resolves the correct pane target for *delivering* the nudge keystrokes, but then passes the **bare session name** to `WakePaneIfDetached` for the SIGWINCH wake:

```go
t.WakePaneIfDetached(session)
```

`resize-window` on a bare session name resizes the session's **active** window. In a multi-window session — for example one with a `gt feed -w` window open and focused (`internal/cmd/feed.go` creates and selects a `feed` window in the current session) — the agent's pane lives in a now-inactive window. The wake's SIGWINCH therefore goes to the *wrong* window, and a detached agent never wakes despite the nudge having been delivered to its pane.

## Fix

Pass the already-resolved `target` (`session:window.pane`) to `WakePaneIfDetached`:

```go
t.WakePaneIfDetached(target)
```

In the common single-window case `target` resolves to the same window, so behavior there is unchanged. This is a one-line behavioral change plus an explanatory comment.

## Verification

Confirmed empirically on tmux 3.6a:
- `resize-window -t $session` resizes the session's **active** window
- `resize-window -t $session:0.0` resizes the **pane's own** window

## Test

Adds `TestNudgeSession_WakesAgentWindowNotActiveWindow`, which:
1. Creates a session and records the agent pane via `GT_PANE_ID`
2. Opens a second window (which becomes active), placing the agent pane in a non-active window
3. Pre-sets both windows' `window-size` to `manual`
4. Nudges, then asserts only the agent's window flipped back to `latest` (the wake's final `set-option window-size latest` step), proving the correct window was woken

The test fails against the previous `WakePaneIfDetached(session)` code (agent window stays `manual`, active window becomes `latest`) and passes with the fix. Full `internal/tmux` package suite passes.

### DIFF
--- a/internal/tmux/tmux.go
+++ b/internal/tmux/tmux.go
@@ -1773,8 +1773,14 @@ func (t *Tmux) NudgeSessionWithOpts(session, message string, opts NudgeOpts) err
 		return fmt.Errorf("nudge to session %q: %w", session, err)
 	}
 
-	// 8. Wake the pane to trigger SIGWINCH for detached sessions
-	t.WakePaneIfDetached(session)
+	// 8. Wake the pane to trigger SIGWINCH for detached sessions.
+	// Use the resolved target (session:window.pane) rather than the bare
+	// session name. resize-window on a bare session name resizes the
+	// session's *active* window — in a multi-window session (e.g. one with a
+	// `gt feed -w` window open and focused) that is not the agent's window,
+	// so the agent's pane never receives SIGWINCH and stays idle despite the
+	// delivered nudge. Targeting the resolved pane wakes the correct window.
+	t.WakePaneIfDetached(target)
 	return nil
 }
 
--- a/internal/tmux/tmux_test.go
+++ b/internal/tmux/tmux_test.go
@@ -1789,6 +1789,77 @@ func TestNudgeSession_WithStoredPaneID(t *testing.T) {
 	}
 }
 
+// TestNudgeSession_WakesAgentWindowNotActiveWindow is a regression test for a
+// missed-wake bug in multi-window sessions. NudgeSessionWithOpts used to pass
+// the bare session name to WakePaneIfDetached, which resizes the session's
+// *active* window. When an agent session also has another window open and
+// focused (e.g. a `gt feed -w` window), the agent's pane lives in a now-inactive
+// window, so the SIGWINCH went to the wrong window and the agent never woke.
+//
+// The wake's resize dance ends by setting the targeted window's window-size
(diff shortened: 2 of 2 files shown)


Bella-Giraffety:
Sheriff review summary for PR #4164

Result: adopted the fix locally for the Gas Town merge queue. The code review consensus is that the proposed fix is sound, minimal, and low risk. I also added the missing status/accepted label. Existing red CI appears to match current baseline failures outside internal/tmux, not this PR.

Research legs:
1. Nudge flow: resolved target is available and correct for wake.
2. Wake semantics: pane-qualified targets wake the containing window.
3. Pane resolution: GT_PANE_ID and scan logic already define the agent pane.
4. Regression test: manual/latest window-size signal directly proves the bug.
5. CI failures: observed failures are unrelated baseline issues.
6. Target compatibility: WakePane and NudgePane already tolerate pane targets.
7. Minimality: using the existing resolved target is the cleanest fix.
8. Edge cases: fallback and attached-session behavior remain unchanged.
9. Call sites: central NudgeSession change covers callers; no propagation needed.
10. Labels: status/accepted was missing; priority/p2 and kind/bug are right.
11. Apply drift: patch still applies without touched-hunk conflict.
12. Security: no new injection or access-control risk identified.
13. Test cost: acceptable tmux integration-test cost for this package.
14. Comments: shorter comments are preferable to product-specific wording.
15. User impact: real P2 detached multi-window wake failure in Gas Town workflows.

Pre-implementation review legs:
1. Correctness review: sound, no blocker.
2. Test review: sound fail-before/pass-after coverage.
3. Simplicity review: sound, optional comment trim only.
4. CI review: proceed, but verify locally and document baseline red CI.
5. Tmux API review: pane/window target semantics support the change.

Post-implementation review legs:
1. Correctness: pass; related nudge tests passed.
2. Regression test: pass; deterministic target-window signal.
3. Scope: pass; only internal/tmux touched.
4. Build risk: pass; gofmt/imports/conventions OK.
5. Merge readiness: intended tmux diff passes; reviewer noted local fork/upstream branch divergence, so I kept the MQ branch scoped to the local refinery target and documented this here.

Local verification:
1. go test ./internal/tmux -run TestNudgeSession_(WithRetry|WithStoredPaneID|WakesAgentWindowNotActiveWindow)$ -count=1 passed.
2. go test ./internal/tmux -run TestFindAgentPane|TestNudgeSession -count=1 passed.
3. git diff --check origin/main...HEAD passed.

PR #4173, opened 2026-06-03
fix(memories): tolerate non-string values in bd kv list

rjgeng:
## Problem

`gt memories` fails outright and `gt prime`'s memory injection silently drops all memories with:

```
parsing kv list: json: cannot unmarshal number into Go value of type string
```

Both commands shell out to `bd kv list --json` and unmarshal the result into `map[string]string`. Recent `bd` (1.0.5) stores a numeric bookkeeping key in the same kv store:

```json
"schema_version": 1
```

`1` is a JSON number, so unmarshaling into a string-valued map fails. `bd memories` is unaffected because it reads the store natively.

## Fix

`bdKvListJSON()` (`internal/cmd/remember.go`) and `bdKvListJSONForPrime()` (`internal/cmd/prime.go`) now decode into `map[string]json.RawMessage` and keep only string-valued entries, skipping non-strings like `schema_version`. Downstream code already filters to `memory.*` keys, so this is safe and forward-compatible with any future non-string kv entries.

## Testing

Built and verified against a live store: `gt memories` (default), `gt memories --type project`, and `gt memories <search>` all list correctly; the numeric `schema_version` key is silently skipped.

### DIFF
--- a/internal/cmd/prime.go
+++ b/internal/cmd/prime.go
@@ -652,9 +652,22 @@ func bdKvListJSONForPrime(workDir string) (map[string]string, error) {
 		return nil, err
 	}
 
-	var kvs map[string]string
-	if err := json.Unmarshal(stdout.Bytes(), &kvs); err != nil {
+	// bd kv list may include non-string values (e.g. the numeric
+	// schema_version key), so decode into RawMessage first and keep only
+	// the string-valued entries. Memories are always stored as strings.
+	var raw map[string]json.RawMessage
+	if err := json.Unmarshal(stdout.Bytes(), &raw); err != nil {
 		return nil, fmt.Errorf("parsing kv list: %w", err)
+	}
+
+	kvs := make(map[string]string, len(raw))
+	for k, v := range raw {
+		var s string
+		if err := json.Unmarshal(v, &s); err != nil {
+			// Skip non-string values (e.g. schema_version).
+			continue
+		}
+		kvs[k] = s
 	}
 	return kvs, nil
 }
--- a/internal/cmd/prime.go
+++ b/internal/cmd/prime.go
@@ -652,24 +652,7 @@ func bdKvListJSONForPrime(workDir string) (map[string]string, error) {
 		return nil, err
 	}
 
-	// bd kv list may include non-string values (e.g. the numeric
-	// schema_version key), so decode into RawMessage first and keep only
-	// the string-valued entries. Memories are always stored as strings.
-	var raw map[string]json.RawMessage
-	if err := json.Unmarshal(stdout.Bytes(), &raw); err != nil {
-		return nil, fmt.Errorf("parsing kv list: %w", err)
-	}
-
-	kvs := make(map[string]string, len(raw))
-	for k, v := range raw {
-		var s string
-		if err := json.Unmarshal(v, &s); err != nil {
-			// Skip non-string values (e.g. schema_version).
-			continue
-		}
-		kvs[k] = s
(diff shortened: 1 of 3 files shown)


Bella-Giraffety:
## PR Sheriff Summary

Review result: the bug is real and the fix direction is sound. I applied a maintainer fix-up on the merge-queue branch that keeps the PR behavior, consolidates the parser into one helper, treats JSON `null` as non-string, and adds regression tests. No action is needed from the contributor.

### 15 Research Legs

1. Current `gt memories` fails because `bd kv list --json` includes numeric `schema_version` and the code unmarshaled `map[string]string`; tolerate heterogeneous KV values.
2. The remember path stores and consumes textual memories; skipping non-string non-memory metadata restores valid memories without rewriting data.
3. Prime memory injection has the same string-memory contract; malformed top-level JSON should remain an error while prime keeps its silent-skip behavior.
4. Live `bd kv` output can legitimately include non-string metadata; RawMessage filtering matches the consumer contract.
5. Existing tests did not cover numeric `schema_version`; regression coverage was needed.
6. Duplicate parsing loops should be consolidated in one helper while keeping command wrappers separate.
7. Edge cases: skip numbers, booleans, arrays, objects, and null; preserve real strings including empty strings; error on invalid JSON.
8. `json.RawMessage` plus explicit string filtering is cleaner than `map[string]interface{}`, CLI filtering, or changing `bd`.
9. Security risk is low; skipping non-strings does not expand prompt injection surface and avoids coercing bogus memories.
10. Silent skip is acceptable for non-memory metadata; warning on corrupt `memory.*` keys can be future work.
11. PR labels needed a `status/*` label in addition to `kind/bug` and `priority/p2`.
12. Original PR scope was narrow: prime and remember KV parsing only.
13. Compile/import risk was low because `encoding/json` was already in use.
14. Persisted memories are strings by design in `gt remember` and `bd remember`; no known non-string memory must remain externally visible.
15. Best sheriff strategy was a small consolidated helper plus regression tests, with contributor attribution preserved.

### 5 Pre-Implementation Reviews

1. Helper with `RawMessage` and `*string` decoding is correct; preserve malformed JSON errors and skip null.
2. Parser-boundary unit tests are sufficient if they cover `schema_version`, null, other non-strings, preserved strings, and invalid JSON.
3. Architecture is sound: shared parser, separate command execution wrappers, no warnings or behavior broadening.
4. Compatibility risk is low; coercion would create new behavior, while skipping unsupported values keeps textual memory semantics.
5. Scope is right: `remember.go`, reuse from existing prime/memories callers, `remember_test.go`, plus required PR status label.

### 5 Post-Implementation Reviews

1. Correctness approved: final parser preserves strings and empty strings, skips null/non-strings, and keeps caller behavior unchanged.
2. Tests approved: they catch the numeric `schema_version` regression and null handling.
3. Scope approved: final branch only changes `internal/cmd/remember.go` and `internal/cmd/remember_test.go`.
4. Build/import/lint risk approved: gofmt clean and targeted compile/test checks pass.
5. Merge readiness approved after recording summaries and labels; full local suite has unrelated environment/pre-existing failures noted below.

### Verification

Passed:

- `go build ./cmd/gt`
- `go test ./internal/cmd -run 'TestParseBdKvListJSON|TestMemTypeRank' -count=1`

Full local `go test ./...` did not pass in this workspace due unrelated environment/pre-existing failures, including missing rootless Docker/testcontainers, config preset expectation, stale socket/PID cleanup tests, and tmux command-name assumptions.

Issue #4043, opened 2026-05-16
BEADS_DOLT env leaks make rig bd/gt commands read HQ database


Bella-Giraffety:
## Bug Description

Rig-scoped `gt`/`bd` commands can silently read the HQ beads database instead of the target rig database when the caller environment contains `BEADS_DOLT_*` variables.

This breaks Dolt-native multi-rig workspaces in a way that is hard to diagnose:

- `bd` from a rig worktree reports the correct `.beads/metadata.json` `dolt_database`, but database-backed config values come from `hq`.
- `gt mq list <rig>` reports an empty queue even when merge-request wisps exist in `<rig>.wisps` with `gt:merge-request` labels.
- `gt done` / `gt mq submit` can create or discover MR beads inconsistently, stranding completed polecat branches.
- `gt doctor` reports misleading database-prefix mismatches because its `bd` subprocesses are querying the wrong database.

The issue is reproducible by unsetting inherited `BEADS_DOLT_*` variables: the same commands immediately select the correct rig database and the merge queue becomes visible.

## Steps to Reproduce

These steps use an existing Dolt-native Gas Town workspace with one town DB and at least one rig DB on the same Dolt server.

1. Start from a workspace where town and rig `.beads/metadata.json` files point to different Dolt databases on the same server. Example:

```json
// /home/coder/gt/.beads/metadata.json
{"dolt_database":"hq","dolt_mode":"server","dolt_server_host":"127.0.0.1","dolt_server_port":3307}

// /home/coder/gt/prcod/mayor/rig/.beads/metadata.json
{"dolt_database":"prcod","dolt_mode":"server","dolt_server_host":"127.0.0.1","dolt_server_port":3307}
```

2. Ensure each database has a distinct `issue_prefix` in its `config` table:

```sql
-- hq.config
issue_prefix = hq

-- prcod.config
issue_prefix = pod
```

3. Export the global Dolt environment variables that Gas Town sessions commonly inherit:

```bash
export BEADS_DOLT_DATA_DIR=/home/coder/gt/.dolt-data
export BEADS_DOLT_HOST=127.0.0.1
export BEADS_DOLT_PORT=3307
export BEADS_DOLT_SERVER_HOST=127.0.0.1
export BEADS_DOLT_SERVER_PORT=3307
```

4. From the rig worktree, ask `bd` which prefix it sees:

```bash
cd /home/coder/gt/prcod/mayor/rig
bd config show
bd config get issue_prefix
```

[...]

PR #4174, opened 2026-06-03
fix(beads): omit shared BEADS_DOLT_DATA_DIR in server mode (#4140)

### LINKED ISSUE #4140: gt sling formula bond fails for cross-rig beads: doltTargetEnvFromBeadsDir emits shared BEADS_DOLT_DATA_DIR in server mode
# Bug: polecat dispatch fails for cross-rig (routed-prefix) beads — `bd mol bond` does not honor prefix routing

**Environment:** gastown main @ 241a72c (`gt 1.2.0 dev`), beads @ main (`v1.0.5-0.20260528224138-278b24ab4bb0`), dolt sql-server on :3307, multi-rig town (town `hq` + rig `minerals`, prefix `mi-` routed to the minerals DB via `~/gt/.beads/routes.jsonl`).

## Symptom
`gt sling mi-jjzc minerals` (and the deferred scheduler path) spawns the polecat, then fails during formula instantiation and rolls back:

```
Auto-applying mol-polecat-work for polecat work...
Instantiating formula mol-polecat-work...
{ "error": "'mi-jjzc' not found (not an issue ID or formula name)", "schema_version": 1 }
⚠ Formula instantiation failed, rolling back spawned polecat rust...
Error: instantiating formula mol-polecat-work: bonding formula to bead: exit status 1
  (direct formula bond fallback failed: exit status 1
   (args: mol bond mol-polecat-work mi-jjzc --json --ephemeral --var feature=... --var issue=mi-jjzc ...))
```

The polecat hook-verification fix (commit 2a7c903) works correctly — it fails *safe* (unhook + delete branch) instead of false-closing. The real defect is one layer down, in the formula bond.

## Root cause (minimal, reproducible)
`bd mol bond <formula|wisp> <bead>` resolves `<bead>` **only against the pinned `BEADS_DIR` / cwd database — it does NOT apply prefix routing**, whereas `bd show <bead>` DOES. So a routed cross-rig bead (`mi-…`) only resolves if `BEADS_DIR` already points at that rig's DB.

Demonstration (town root = `~/gt`, `mi-` routes to `~/gt/minerals`):

| command | from `~/gt` (town) / `BEADS_DIR=~/gt/.beads` | `BEADS_DIR=~/gt/minerals/.beads` |
|---|---|---|
| `bd show mi-jjzc --json` | ✅ resolves (routing applied) | ✅ |
| `bd mol bond mol-polecat-work mi-jjzc --ephemeral --var …` | ❌ `'mi-jjzc' not found (not an issue ID or formula name)` | ✅ `bond_type: sequential, result_id: mi-jjzc` |

`strace -f` of `gt sling` shows bd is invoked with **both** `BEADS_DIR=~/gt/.beads` (town, 12×) and `BEADS_DIR=~/gt/minerals/.beads` (17×) within one dispatch; the `mol bond` that resolves the work bead lands on the town context and fails, because `mol bond` can't route `mi-` to the minerals DB.

## Why it matters
Any multi-rig town where the work bead's prefix is routed to a non-town DB cannot dispatch polecats — every dispatch dies at `mol bond`. Single-rig / town-prefix setups don't hit it (the bead is in the same DB the bond runs against), which is likely why it's gone unnoticed.

## Exact root cause (traced + fixed)
`internal/beads/database.go: doltTargetEnvFromBeadsDir(beadsDir)` builds the dolt-target env when a bd command is pinned to a `.beads` dir. It **unconditionally** appends `BEADS_DOLT_DATA_DIR=<townRoot>/.dolt-data` (the *shared, multi-database* town data dir) — even in server mode — and does **not** include a database selector (`BEADS_DOLT_SERVER_DATABASE`).

In server mode, the running dolt server already owns the data and the host/port is the connection target; handing bd the shared multi-DB data dir makes `bd mol bond`'s database resolution select a database other than the one chosen by the pinned `.beads/metadata.json`. So a routed cross-rig bead (e.g. `mi-`) fails to resolve at the bond step, even though `BEADS_DIR` is correct. (`bd show` resolves the DB differently and isn't affected — hence the asymmetry.) The bond step in gastown's `InstantiateFormulaOnBead`/`bondFormulaDirect` therefore fails, and polecat dispatch rolls back.

Bisection proof: with `BEADS_DIR=…/minerals/.beads` + server host/port, the bond **succeeds**; add `BEADS_DOLT_DATA_DIR=…/.dolt-data` and it **fails** "not found".

## Fix (verified end-to-end)
In `doltTargetEnvFromBeadsDir`, emit `BEADS_DOLT_DATA_DIR` only in **embedded** mode (no server host/port configured). In server mode the host/port is the connection target and the pinned `.beads` metadata selects the database:

```go
func doltTargetEnvFromBeadsDir(beadsDir string) []string {
	if beadsDir == "" { return nil }
	meta := readDoltMetadata(beadsDir)
	var env []string
	if meta.Host != "" { env = append(env, "BEADS_DOLT_SERVER_HOST="+meta.Host) }
	if meta.Port != "" { env = append(env, "BEADS_DOLT_SERVER_PORT="+meta.Port, "BEADS_DOLT_PORT="+meta.Port) }
	// Server owns the data in server mode; the shared multi-DB data dir would make
	// bd mol resolution pick the wrong database for routed cross-rig beads.
	if meta.Host == "" && meta.Port == "" {
		if townRoot := FindTownRoot(filepath.Dir(beadsDir)); townRoot != "" {
			env = append(env, "BEADS_DOLT_DATA_DIR="+filepath.Join(townRoot, ".dolt-data"))
		}
	}
	return env
}
```

Validated: with this change, `gt sling mi-jjzc minerals` completes the full chain — `✓ Formula bonded to mi-jjzc`, `✓ Work attached to hook (status=hooked)`, the work bead shows `status=hooked, assignee=minerals/polecats/chrome` on the server, and the polecat picks up its hook and works it (no empty-hook false-close).

## Related issues (checked — this is NOT a duplicate)
- **#1056** (closed, gastown 0.5.0) — *identical user-facing symptom* (`'<bead>' not found (not an issue ID or formula name)` → `instantiating formula mol-polecat-work: bonding formula to bead`), but its root cause was "daemon doesn't route to rig DBs." Fixed long ago on a much older version; the symptom now recurs via a *different* mechanism (below).
- **#4043** (closed 2026-05-27) — *adjacent root cause*: "BEADS_DOLT env leaks make rig commands read HQ database." Its fix added `BEADS_DOLT_DATA_DIR` to `bdTargetEnvKeys` so `StripBDTargetEnv` removes *inherited* values.

**Why this is a distinct, still-open gap:** `BuildPinnedBDEnv` strips the inherited var (the #4043 fix) and then **immediately re-adds it**:
```go
func BuildPinnedBDEnv(base []string, beadsDir string) []string {
    env := SuppressBDSideEffects(StripBDTargetEnv(base))      // #4043: strips inherited BEADS_DOLT_DATA_DIR
    ...
    env = append(env, "BEADS_DIR="+beadsDir)
    env = append(env, doltTargetEnvFromBeadsDir(beadsDir)...) // re-adds the SHARED town BEADS_DOLT_DATA_DIR
    return addGTDerivedDoltTargetEnv(env)
}
```
`doltTargetEnvFromBeadsDir` emits the shared, multi-DB town `.dolt-data` even in server mode, so `bd mol bond` resolves the routed bead against the wrong database — reproducing #1056. The fix below scopes the re-add to embedded mode only.

## Also note (separate, minor)
gastown main's `go.mod` pins `github.com/steveyegge/beads v1.0.0`, which is incompatible with the current bd schema (`ready_issues` view references the old column `depends_on_id`; current schema uses `depends_on_issue_id`) → daemon "no beads stores available". Building gt with `go get github.com/steveyegge/beads@main` (bumps to v1.0.5) compiles cleanly and fixes it; the go.mod pin should be bumped to >= the schema-matching beads version.

Rome-1:
## Summary

Closes #4140. \`doltTargetEnvFromBeadsDir\` previously emitted the **shared** multi-database town data dir (\`<townRoot>/.dolt-data\`) on every call, even when the pinned \`.beads/metadata.json\` named a server connection. In server mode the running Dolt server already owns the data and the connection target is host/port; the shared data dir overrode the pinned database selector, so \`bd mol bond\` resolved routed cross-rig beads against the wrong database:

\`\`\`
bd: '<bead>' not found (not an issue ID or formula name)
\`\`\`

This broke polecat dispatch in any multi-rig town whose work bead's prefix routes to a non-town DB. Single-rig / town-prefix setups escaped because the bead lives in the DB the bond runs against — which is why this went unnoticed.

Companion to #4043, which added \`BEADS_DOLT_DATA_DIR\` to \`bdTargetEnvKeys\` so \`StripBDTargetEnv\` removes the inherited value. \`BuildPinnedBDEnv\` then immediately re-added it via \`doltTargetEnvFromBeadsDir\`; this fix scopes the re-add to **embedded mode only** (\`host == "" && port == ""\`).

## Diagnosis credit

The issue author (#4140) did the full bisection, traced the exact emit site, named the fix, and validated end-to-end (full \`gt sling mi-jjzc minerals\` → \`✓ Formula bonded\` → polecat picks up its hook). This PR applies their proposed fix and adds regression coverage.

## Test plan

- [x] \`go vet ./internal/beads/...\` clean
- [x] \`go test ./internal/beads/...\` — full package PASS (existing + new)
- [x] **TestBuildPinnedBDEnv_ServerModeOmitsSharedDataDir** — under a real town root (\`mayor/town.json\` present) with metadata.json naming server host+port, asserts \`BEADS_DOLT_DATA_DIR\` is absent. Verified the test FAILS on \`main\` without the fix (caught the bug).
- [x] **TestBuildPinnedBDEnv_EmbeddedModeKeepsSharedDataDir** — complement: under a real town root with NO server metadata, asserts the shared data dir IS still emitted so bd can locate on-disk data.
- [x] Pre-existing \`TestBuildPinnedBDEnvUsesSelectedConnectionMetadata\` — still passes (its temp dir has no \`mayor/town.json\` above it, so \`FindTownRoot\` returned \`""\` and the bug path wasn't exercised there).
- [x] Companion env builders (\`TestBuildRoutingBDEnvStripsDatabaseButKeepsSelectedConnection\`, \`TestBuildReadOnlyBDEnv*\`) — still pass.

### DIFF
--- a/internal/beads/beads_test.go
+++ b/internal/beads/beads_test.go
@@ -162,6 +162,86 @@ func TestBuildRoutingBDEnvStripsDatabaseButKeepsSelectedConnection(t *testing.T)
 	}
 }
 
+// TestBuildPinnedBDEnv_ServerModeOmitsSharedDataDir is a regression test for
+// gastownhall/gastown#4140. When the pinned beads dir's metadata.json names a
+// server host/port (server mode), BuildPinnedBDEnv must NOT emit
+// BEADS_DOLT_DATA_DIR — the shared multi-database town data dir would override
+// the pinned database selection and make `bd mol bond` resolve routed
+// cross-rig beads against the wrong database.
+//
+// The previous tests above use a beadsDir under t.TempDir() with no
+// mayor/town.json above it, so FindTownRoot returned "" and the bug was
+// invisible. This test creates the town root explicitly so the regression
+// path is exercised.
+func TestBuildPinnedBDEnv_ServerModeOmitsSharedDataDir(t *testing.T) {
+	townRoot := t.TempDir()
+	if err := os.MkdirAll(filepath.Join(townRoot, "mayor"), 0755); err != nil {
+		t.Fatal(err)
+	}
+	if err := os.WriteFile(filepath.Join(townRoot, "mayor", "town.json"), []byte(`{}`), 0644); err != nil {
+		t.Fatal(err)
+	}
+
+	rigBeadsDir := filepath.Join(townRoot, "minerals", ".beads")
+	if err := os.MkdirAll(rigBeadsDir, 0755); err != nil {
+		t.Fatal(err)
+	}
+	metadata := []byte(`{"dolt_database":"minerals","dolt_server_host":"127.0.0.1","dolt_server_port":3307}`)
+	if err := os.WriteFile(filepath.Join(rigBeadsDir, "metadata.json"), metadata, 0644); err != nil {
+		t.Fatal(err)
+	}
+
+	env := BuildPinnedBDEnv([]string{"PATH=/usr/bin"}, rigBeadsDir)
+	got := envMap(env)
+
+	if value, ok := got["BEADS_DOLT_DATA_DIR"]; ok {
+		t.Fatalf("BEADS_DOLT_DATA_DIR should be omitted in server mode (would clobber the pinned database for routed cross-rig beads), got %q in %v", value, env)
+	}
+	if got["BEADS_DOLT_SERVER_HOST"] != "127.0.0.1" {
+		t.Fatalf("BEADS_DOLT_SERVER_HOST = %q, want 127.0.0.1 in %v", got["BEADS_DOLT_SERVER_HOST"], env)
+	}
+	if got["BEADS_DOLT_SERVER_PORT"] != "3307" || got["BEADS_DOLT_PORT"] != "3307" {
+		t.Fatalf("ports = server:%q legacy:%q, want 3307 in %v", got["BEADS_DOLT_SERVER_PORT"], got["BEADS_DOLT_PORT"], env)
+	}
+}
+
+// TestBuildPinnedBDEnv_EmbeddedModeKeepsSharedDataDir verifies the
+// complement of the regression case: when metadata.json has no server
+// host/port (embedded mode), BEADS_DOLT_DATA_DIR IS still emitted so bd can
+// locate the on-disk data.
+func TestBuildPinnedBDEnv_EmbeddedModeKeepsSharedDataDir(t *testing.T) {
+	townRoot := t.TempDir()
+	if err := os.MkdirAll(filepath.Join(townRoot, "mayor"), 0755); err != nil {
+		t.Fatal(err)
+	}
+	if err := os.WriteFile(filepath.Join(townRoot, "mayor", "town.json"), []byte(`{}`), 0644); err != nil {
+		t.Fatal(err)
+	}
+
+	beadsDir := filepath.Join(townRoot, ".beads")
+	if err := os.MkdirAll(beadsDir, 0755); err != nil {
+		t.Fatal(err)
+	}
+	// Embedded mode: metadata names a database but no server connection.
+	if err := os.WriteFile(filepath.Join(beadsDir, "metadata.json"), []byte(`{"dolt_database":"hq"}`), 0644); err != nil {
+		t.Fatal(err)
+	}
+
+	env := BuildPinnedBDEnv([]string{"PATH=/usr/bin"}, beadsDir)
+	got := envMap(env)
+
+	wantDataDir := filepath.Join(townRoot, ".dolt-data")
+	if got["BEADS_DOLT_DATA_DIR"] != wantDataDir {
+		t.Fatalf("BEADS_DOLT_DATA_DIR = %q, want %q in %v", got["BEADS_DOLT_DATA_DIR"], wantDataDir, env)
+	}
+	if _, ok := got["BEADS_DOLT_SERVER_HOST"]; ok {
+		t.Fatalf("BEADS_DOLT_SERVER_HOST should be unset in embedded mode, env=%v", env)
+	}
+	if _, ok := got["BEADS_DOLT_SERVER_PORT"]; ok {
+		t.Fatalf("BEADS_DOLT_SERVER_PORT should be unset in embedded mode, env=%v", env)
+	}
+}
+
 func TestBuildPinnedBDEnvFallsBackToGTDoltPort(t *testing.T) {
 	beadsDir := filepath.Join(t.TempDir(), ".beads")
 	if err := os.MkdirAll(beadsDir, 0755); err != nil {
--- a/internal/beads/database.go
+++ b/internal/beads/database.go
@@ -193,15 +193,25 @@ func doltTargetEnvFromBeadsDir(beadsDir string) []string {
 	}
 	meta := readDoltMetadata(beadsDir)
 	var env []string
-	if townRoot := FindTownRoot(filepath.Dir(beadsDir)); townRoot != "" {
-		env = append(env, "BEADS_DOLT_DATA_DIR="+filepath.Join(townRoot, ".dolt-data"))
-	}
 	if meta.Host != "" {
 		env = append(env, "BEADS_DOLT_SERVER_HOST="+meta.Host)
 	}
 	if meta.Port != "" {
 		env = append(env, "BEADS_DOLT_SERVER_PORT="+meta.Port)
 		env = append(env, "BEADS_DOLT_PORT="+meta.Port)
+	}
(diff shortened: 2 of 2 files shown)


blairsilverberg:
+1 — independently hit this exact symptom today on a multi-rig setup (`hq` town + `openclaw` rig with `op-` prefix routed to the rig DB, shared Dolt server on :3307, `gt 1.1.0-348-ge7949128`, `bd 1.0.4`).

Same exact bisection:
- `bd show op-bpl` ✅ resolves (prefix routing applied)
- `gt sling op-bpl openclaw` ❌ "bead 'op-bpl' not found"
- `gt bead show op-bpl` ❌ "no issue found matching op-bpl"

Verified `op-bpl` exists in Dolt directly via `dolt sql -q "USE openclaw; SELECT id, title FROM issues WHERE id='op-bpl';"` — present and correct.

Also confirmed this is NOT [beads#4239](https://github.com/gastownhall/beads/issues/4239) (auto-import-clobbers-Dolt) — applied that workaround (removing `.beads/issues.jsonl`) and the `gt sling` symptom persists. Both bugs exist in this setup; #4140 is the actual blocker for cross-rig dispatch.

Result: every Jira-tracked work item that needs a shadow `op-*` bead for `gt sling` is blocked. Currently routing around via a separate (non-`gt`) build-session dispatcher, but the divergence from the gt-rig path is causing other problems for us.

The fix in this PR (only emit `BEADS_DOLT_DATA_DIR` in embedded mode) matches the upstream report's bisection precisely. Would benefit greatly from this landing on main. Happy to test against a build of this branch on our setup if useful for verification.

Bella-Giraffety:
Thanks Rome-1. This PR had the right diagnosis: server-mode bd subprocesses must not receive the shared town BEADS_DOLT_DATA_DIR, because it can make routed rig beads resolve against the wrong database. The #4140 report, bisection, and this PR direction were valuable and directly informed the landed fix.

Sheriff review record:

Research legs 1-15:
1. Confirmed #4140 symptom is cross-rig and routed-prefix dispatch failure.
2. Confirmed inherited env stripping alone was insufficient after #4043 because generated env re-added the bad selector.
3. Confirmed generated BEADS_DOLT_DATA_DIR was the incorrect server-mode target selector.
4. Confirmed bd show and bd mol bond differ enough that env targeting exposed the failure in formula bond.
5. Confirmed --hook-raw-bead does not avoid the same env bug for hook readback.
6. Confirmed the bug reproduces in fresh stock rig scenarios, not only imported multi-rig work.
7. Confirmed stale parent env is not required; stock gt can generate the bad child env.
8. Confirmed server metadata host and port should be authoritative in server mode.
9. Confirmed pinned subprocesses are cleaner with metadata database pinning via BEADS_DOLT_SERVER_DATABASE.
10. Confirmed #4174 server-mode omission matches the root cause.
11. Confirmed #4174 embedded-mode BEADS_DOLT_DATA_DIR preservation is no longer the desired mainline behavior.
12. Confirmed #4207 overlaps and addresses the same failure path.
13. Confirmed current main consolidates env policy across related subprocess callers.
14. Confirmed CI failures shown here are stale or baseline failures, not useful blockers for this superseded PR.
15. Confirmed PR labels are now status/accepted, priority/p2, kind/bug.

Pre-implementation review legs 1-5:
1. The proposed server-mode fix is sound.
2. Useful regression coverage needs a real town root or the bug path is missed.
3. Do not land the embedded-mode expectation from this branch as-is.
4. Prefer a consolidated env-targeting fix over a narrow patch-on-patch.
5. #4207/current main is the cleaner landing vehicle because it removes shared data-dir emission and pins the metadata database.

Post-implementation review legs 1-5:
1. Current main contains the #4174-equivalent server-mode fix as commit 26776f6b.
2. The landed fix also pins BEADS_DOLT_SERVER_DATABASE from .beads/metadata.json for pinned subprocesses.
3. The landed change is broader and avoids duplicating env-policy fixes.
4. This PR should close as superseded, not rejected.
5. No changes are requested from the contributor.

Resolution: closing this PR as superseded by the #4207-equivalent fix that is already on main. The diagnosis and direction here were correct; the fix landed through the broader consolidated env-targeting change.

PR #4175, opened 2026-06-03
fix: use split target columns in wisp_dependencies queries

acrawford13:
## Summary

Two bugs fixed, both caused by the `wisp_dependencies` split-column schema change:

### Bug 1: `depends_on_id` column references (hq-isz)

`wisp_dependencies` uses split target columns (`depends_on_wisp_id`, `depends_on_issue_id`, `depends_on_external`) rather than a single `depends_on_id` column. Databases created with the new schema that didn't go through the old→new migration path in `0047` lack the `depends_on_id` STORED generated column, causing `bd compact` orphan-cleanup and `gt reaper scan` to fail every run.

**`internal/cmd/compact.go`** — `cleanOrphanedWispDeps`:
- Replace single `depends_on_id` check with explicit checks on `depends_on_wisp_id` (vs `wisps`) and `depends_on_issue_id` (vs `issues`)

**`internal/reaper/reaper.go`** — 5 sites:
- `parentExcludeJoin`: join `wisps` on `depends_on_wisp_id`, `issues` on `depends_on_issue_id`
- `danglingQuery`: same split-column joins; exclude `depends_on_external` rows from dangling check
- `Scan` `staleQuery`: use `depends_on_issue_id` for issue-to-issue blocking checks
- `AutoClose` `whereClause`: same fix for the multi-db variant
- `batchDeleteRows` `delReverse`: split into two non-fatal DELETEs (one per target column)

### Bug 2: witness falls back to rig root, breaking SetupRedirect

`witnessDir()` previously fell back to `m.rig.Path` when no `witness/` or `witness/rig/` subdirectory existed. Rig roots are only 1 level deep from the town root, causing `ComputeRedirectTarget` to fail with "must be at least 2 levels deep".

**`internal/witness/manager.go`**: create `witness/` automatically in the fallback path instead.

## Test plan

- [ ] `go build ./internal/cmd/... ./internal/reaper/... ./internal/witness/...` passes (verified)
- [ ] `bd compact` on a fresh-schema db — no `depends_on_id` error
- [ ] `gt reaper scan` on a fresh-schema db — scan completes
- [ ] `gt witness start <rig>` on a rig with no `witness/` dir — starts without depth error
- [ ] Old-schema dbs (with generated `depends_on_id`) still work

### DIFF
--- a/internal/cmd/compact.go
+++ b/internal/cmd/compact.go
@@ -277,13 +277,16 @@ func runCompact(cmd *cobra.Command, args []string) error {
 }
 
 // cleanOrphanedWispDeps removes wisp_dependencies rows where either side no
-// longer exists in the wisps table. This happens when bd delete removes a wisp
-// but leaves behind its dependency records (bd delete has no cascade logic for
-// the wisp-level tables). Runs as a post-compact sweep.
+// longer exists. This happens when bd delete removes a wisp but leaves behind
+// its dependency records (bd delete has no cascade logic for the wisp-level
+// tables). Runs as a post-compact sweep.
 func cleanOrphanedWispDeps(bd *beads.Beads, result *compactResult) {
+	// wisp_dependencies uses split target columns (depends_on_wisp_id,
+	// depends_on_issue_id, depends_on_external) — no depends_on_id column.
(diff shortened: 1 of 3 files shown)


acrawford13:
PR was created automatically, sorry!

PR #4192, opened 2026-06-06
doltserver: reduce read/write timeout from 5min to 30s

Ben-Williams-Founder:
## Summary

- Reduce `DefaultReadTimeoutMs` and `DefaultWriteTimeoutMs` from 5 minutes (300,000ms) to 30 seconds (30,000ms)
- Prevents connection accumulation cascade: at 18 conn/s burst rate, the 5-min timeout allows the pool to fill to 1,000 before hung connections are reaped. With 30s, accumulation is capped to ~540 connections in the worst case, well within the 1,000 limit.
- Root cause context in hq-60wv: Dolt watchdog was restarting ~every 1.5h due to SQL deadlock caused by connection pool exhaustion. Fix validated live — 2.5h+ stable vs previous 1–1.5h cycle.
- `auto_gc` is disabled in town config, so no legitimate long-running GC reads exist that would be cut off.

## Test plan

- [ ] CI green (Integration Tests, Test, Lint all pass on town-build-2026-06-07)
- [ ] Verified live in production via direct config.yaml edit before upstream PR — 2.5h+ stable


Ben-Williams-Founder:
Closing — keeping in Ben-Williams-Founder/gastown fork for now, not ready for upstream submission.

PR #4189, opened 2026-06-06
Fix convoy stranded blocked dependency decoding

fengning-starsend:
## Summary

- Decode both `dependency_type` and `type` dependency JSON fields for convoy stranded readiness checks.
- Reuse the command package's existing blocking dependency predicate when deciding whether fresh `bd show` details are blocked.
- Add a regression where a convoy tracks A and B, B depends on A via `type:"blocks"`, and `gt convoy stranded` reports only A as ready.

## Why

`gt convoy stranded --json` feeds the daemon's convoy manager. If `bd show --json` returns dependency edges with `type` instead of `dependency_type`, a blocked tracked issue can be reported as ready and dispatched before its blocker closes.

## Validation

Passed:

```bash
mise x go@1.25.9 -- gofmt -w internal/cmd/convoy.go internal/cmd/convoy_stranded_test.go internal/cmd/convoy_empty_test.go
mise x go@1.25.9 -- go test ./internal/cmd -run 'TestIssueDetailsIsBlocked|TestFindStrandedConvoys_BlocksDependencyTypeField|TestFindStrandedConvoys_StuckConvoy' -count=1
git diff --check
```

Attempted:

```bash
mise x go@1.25.9 -- go test ./...
```

This did not complete cleanly in my local environment due to pre-existing unrelated `internal/cmd` failures, including ambient push-hook blocks in git-push tests, existing convoy routing test expectations, an install config output mismatch, and a Dolt container/server startup failure. The targeted convoy stranded tests above pass.

PR #4206, opened 2026-06-08
fix(formulas): drop redundant required=true on vars that also set a default

popandpeek:
## Summary
Three formulas declare a `[vars.*]` entry with **both** `required = true` and a `default`, which the formula validator rejects:
```
formula validation failed:
  - vars.<x>: cannot have both required:true and default
```
This makes `bd cook` — and therefore `gt sling` — fail for these formulas on a clean, matched install (reproduced on bd v1.0.0, the version `go.mod` pins, and on 1.0.5).

## Affected formulas & provenance
| Formula | Var (default) | Introduced | Broken for |
|---|---|---|---|
| `mol-digest-generate` | `period` (`"daily"`) | `1e76bfd7c` 2026-01-05 | ~5 months |
| `mol-shutdown-dance` | `requester` (`"deacon"`) | `b649635f4` 2026-01-07 | ~5 months |
| `mol-orphan-scan` | `scope` (`"town"`) | `6ed4deaf1` 2026-02-20 | ~3.5 months |

## Root cause
A var with a `default` is never "missing" — the default supplies a value when the var isn't passed — so `required = true` can never fire. The two are contradictory, and the validator enforces them as mutually exclusive. The `required` flag is redundant and renders the formula uncookable.

## Why it went unnoticed
All three are rarely-cooked admin/infra formulas (digest generation, shutdown, orphan recovery). The ~44 core work formulas (polecat-work, feature, bug, …) are unaffected, so normal operation never trips it. It only surfaces when one of these is slung.

## Fix
Remove the redundant `required = true` from each var. Behavior is unchanged — the defaults (`daily` / `deacon` / `town`) still apply.

## Verification
All three cook cleanly after the change.

### DIFF
--- a/internal/formula/formulas/mol-digest-generate.formula.toml
+++ b/internal/formula/formulas/mol-digest-generate.formula.toml
@@ -223,7 +223,6 @@ Dog returns to available state in the pool.
 [vars]
 [vars.period]
 description = "The digest period type (daily, weekly, custom)"
-required = true
 default = "daily"
 
 [vars.date]
--- a/internal/formula/formulas/mol-orphan-scan.formula.toml
+++ b/internal/formula/formulas/mol-orphan-scan.formula.toml
@@ -322,7 +322,6 @@ Dog returns to available state in the pool.
 [vars]
 [vars.scope]
 description = "Scan scope: 'town' or specific rig name"
-required = true
 default = "town"
 
 [vars.action]
--- a/internal/formula/formulas/mol-shutdown-dance.formula.toml
+++ b/internal/formula/formulas/mol-shutdown-dance.formula.toml
@@ -515,5 +515,4 @@ required = true
 
 [vars.requester]
 description = "Who filed the warrant (deacon, witness, mayor)"
-required = true
 default = "deacon"


popandpeek:
**Note on the failing checks — these are pre-existing on `main`, not caused by this change.**

This PR only deletes three `required = true` lines from formula TOML files. The failing tests are all in unrelated areas — `TestFilterAndSortSessions_SortOrder`, `TestRunConvoyList/Status_UsesTownRootAndStripsBeadsDir`, `TestJSONOutput_*`, `TestEnsureBeadsConfigYAML_CreatesWhenMissing`, `TestManagerAgentLifecycleUsesTownBeadsDir`, `internal/rig`, `internal/refinery` — plus runtime `database locked` / `missing required vars` errors. None of them exercise formula parsing or validation.

Unrelated open PR #4189 fails the identical three checks (Integration ✗ / Lint ✗ / Test ✗, with Windows Smoke ✓), which confirms the red CI predates this branch.

The fix itself is verified: all three affected formulas (`mol-orphan-scan`, `mol-digest-generate`, `mol-shutdown-dance`) cook cleanly after the change on both bd v1.0.0 (the version `go.mod` pins) and 1.0.5.

popandpeek:
Withdrawing — closing this PR.

PR #4232, opened 2026-06-11
fix(plugins): dolt-backup silently no-oped — rc capture bug, missing remotes, stale freshness signal

rbriski:
## Problem

The dolt-backup plugin has been reporting success while never backing up anything. Three compounding bugs:

1. **Exit-code capture is broken.** `SYNC_OUTPUT=$(...) || true; SYNC_RC=${PIPESTATUS[0]:-$?}` always evaluates to 0 — `PIPESTATUS` reflects the `|| true`, not the sync. Every failed sync was recorded as success and the hash marker written, so the next run skipped the DB as 'unchanged'. Result: `.dolt-backup/<db>/` contained only `.last-backup-hash` markers — zero data.

2. **Backup remotes were never created.** `dolt backup sync <db>-backup` requires a configured backup remote; nothing ever ran `dolt backup add`, so every sync failed (invisible due to bug 1).

3. **Freshness signal never updates.** The daemon's `checkBackupFreshness` stats the per-db directory mtime, which doesn't change when syncs only touch nested files (or when the hash-skip path runs). The resulting permanent "backup patrol may be stalled" warnings pour a mol-dog-doctor molecule every 5 minutes, forever — a meaningful contributor to runaway wisp counts.

## Fix

- Capture the sync exit code directly (`|| SYNC_RC=$?`)
- Auto-add the backup remote (`file://$BACKUP_DIR/<db>/<db>-backup`) when missing, and count add-failures as failures
- `touch` the per-db dir on successful sync **and** verified-unchanged skip so the daemon freshness check sees a live patrol

## Verification (live town)

- DB with no remote: remote auto-added, real sync produced 1.4G of backup data where only a hash marker existed before
- Unchanged DB: skip path bumps dir mtime; daemon warnings clear
- Induced failure: now counted in `FAILED`/`FAILED_DBS` instead of writing a bogus hash marker

### DIFF
--- a/plugins/dolt-backup/run.sh
+++ b/plugins/dolt-backup/run.sh
@@ -117,6 +117,10 @@ for DB in "${PROD_DBS[@]}"; do
   if [[ "$CURRENT_HASH" = "$LAST_HASH" ]] && [[ "$CURRENT_HASH" != "unknown" ]]; then
     log "  $DB: unchanged ($CURRENT_HASH), skipping"
     SKIPPED=$((SKIPPED + 1))
+    # Signal liveness to the daemon's dir-mtime freshness check even when
+    # there is nothing to sync — otherwise checkBackupFreshness reports the
+    # backup patrol as stalled forever and pours doctor molecules.
+    touch "$BACKUP_DIR/$DB" 2>/dev/null || true
     continue
   fi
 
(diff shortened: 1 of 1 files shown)


Bella-Giraffety:
PR Sheriff review for #4232

Final recommendation: accept the bug fix with maintainer fixups. I kept scope to `plugins/dolt-backup`, applied the core PR fix, and tightened it so stale false `.last-backup-hash` markers cannot keep skipping an empty backup. No contributor changes requested.

15 research legs:
1. Accept: sync rc capture bug is real; direct assignment rc capture preserves failures.
2. Accept: `dolt backup add <name> file://absolute-path` is valid and matches naming/location conventions.
3. Fix: freshness touch is right, but PR trusted hash skip before remote/data verification.
4. Fix: plugin conventions support explicit failure counting and no false hash markers.
5. Accept: PR CI failures appear unrelated to this shell-only change and pre-existing Go failures.
6. Accept: labels should be `status/accepted`, `priority/p1`, `kind/bug`.
7. Accept: not superseded by dependency/schema, hook/persistence, scheduler, or recent dolt-backup changes.
8. Fix: stale hash skip and regex remote matching were robustness concerns.
9. Fix: repeat-run behavior needed remote/data validation before skip.
10. Fix: stale false hash markers needed repair by forced sync when target data is absent.
11. Fix: focused shell regression tests are warranted for this data-safety path.
12. Fix: use awk literal remote matching and direct rc capture.
13. Accept: add/sync failures flow to FAILED/FAILED_DBS and escalation; add output should be logged.
14. Fix: minimal convergent fix is validation before skip plus rc capture and remote add.
15. Accept: upstream PR is mergeable; fix was applied manually on the working branch because local fork history diverges.

5 pre-implementation reviews:
1. Pass: plan fixes rc, remotes, freshness, and stale marker repair; touch only after directory exists.
2. Pass with changes: initialize `SYNC_RC`, call remote helper only in conditionals, assert statuses and side effects.
3. Pass: scope limited to `plugins/dolt-backup` plus targeted shell tests.
4. Pass: data-safe skip requires remote plus actual backup payload; failed add/sync avoids markers.
5. Pass: temp-dir stubs for `dolt`/`bd`/`gt` cover failed sync, missing remote, stale hash repair, unchanged skip, and mtime touch.

5 post-implementation reviews:
1. Pass: runtime behavior correct for rc capture, remote add, stale marker repair, failure paths, freshness touch, and missing `--databases` value.
2. Pass: strict-mode shell behavior is safe; test cleanup uses an EXIT trap; operation-order assertions verify add before sync.
3. Fix then pass: added coverage for existing empty backup target and missing `--databases` value.
4. Pass: branch scope is limited to `plugins/dolt-backup/run.sh` and `plugins/dolt-backup/run_test.sh`; cleanup/convergence policy satisfied.
5. Ready with note: full pre-verified gates were not run because existing PR CI is already red for unrelated Go lint/test/integration failures; submit without pre-verified.

Fixup evidence:
- Sync exit code is captured directly; failed syncs no longer write `.last-backup-hash`.
- Missing backup remotes are added before sync; add failures stop before sync and do not write a marker.
- Hash-equal skip now requires both a configured remote and real local backup payload, so stale markers force repair sync.
- Per-DB backup dir mtime is touched after successful sync and verified unchanged skip.
- Added `plugins/dolt-backup/run_test.sh` covering failed sync, add failure, missing remote add-before-sync, stale marker repair, unchanged skip touch, and missing `--databases` value.

Verification passed:
- `git diff --check origin/main...HEAD`
- `bash -n plugins/dolt-backup/run.sh`
- `bash -n plugins/dolt-backup/run_test.sh`
- `plugins/dolt-backup/run_test.sh`
- `go test ./internal/plugin/...`

Labels set: `status/accepted`, `priority/p1`, `kind/bug`.

PR #4230, opened 2026-06-11
chore(deps): bump beads to v1.0.5 to match migrated dependency schema

rbriski:
## Problem

bd 1.0.5 migrates the Dolt schema (typed `depends_on_issue_id`/`depends_on_wisp_id`/`depends_on_external` columns replace `depends_on_id`; `ready_issues` view moves into versioned migrations), but gastown pins beads **v1.0.0**, whose embedded store still emits old-schema SQL.

Against a 1.0.5-migrated database the daemon's convoy manager fails schema init every cycle, on every database:

```
Convoy: <db> beads store unavailable: failed to initialize schema:
failed to create ready_issues view: Error 1105 (HY000): table "d" does not have column "depends_on_id"
```

In a live town this cascades: failing health checks → doctor-dog dispatch storms → runaway open-wisp counts (~900 wisps/hour observed) → reaper alert spam. Related: #4228 fixes the same schema skew in the reaper's raw SQL.

## Fix

Bump `github.com/steveyegge/beads` v1.0.0 → v1.0.5 so the embedded library matches the schema the bd CLI actually creates. (v1.0.4 and earlier still carry the old view SQL — 1.0.5 is the first aligned release.)

**Build note:** beads ≥1.0.5 pulls a dolt engine that needs system ICU headers (`brew install icu4c` + `CGO_CFLAGS`/`CGO_LDFLAGS` pointing at it). CI images may need that added.

## Verification

- `go build ./cmd/gt` OK (with ICU flags)
- `go test ./internal/reaper ./internal/daemon` pass
- `internal/cmd` failures are pre-existing on main (identical set fails with v1.0.0)
- Running this build in a live town: convoy schema-init errors stop; wisp creation returns to baseline

## Note on the module path

`github.com/steveyegge/beads` is used (not `gastownhall/beads`) because that is the module path beads declares in its own go.mod — including on current main. The repo's canonical home is `gastownhall/beads` (GitHub redirects the old path there), so v1.0.5 here is the gastownhall release fetched under its declared module identity. Same pattern as the previously merged renovate bumps of this dependency.

PR #4228, opened 2026-06-11
fix(reaper): use depends_on_issue_id/depends_on_wisp_id columns in dependency queries

rbriski:
## Problem

The beads schema renamed the dependency tables' `depends_on_id` column to typed `depends_on_issue_id` / `depends_on_wisp_id` / `depends_on_external` columns. The reaper still queries `depends_on_id`, so every reaper operation fails against current databases:

```
$ gt reaper scan --db hq --dry-run
hq: scan error: count reap candidates: Error 1105 (HY000): table "wd" does not have column "depends_on_id"
```

In a live town this breaks `gt reaper scan/reap/purge/auto-close` on every database, the daemon's wisp reaper, and the stuck-agent-dog plugin — wisps accumulate unbounded and the reaper dogs escalate every cycle.

## Fix

Update the five query sites in `internal/reaper/reaper.go` to the typed columns:

- `parentExcludeJoin`: join wisps on `depends_on_wisp_id`, issues on `depends_on_issue_id`
- `Scan` stale-issue exclusion subqueries: `depends_on_issue_id`
- dangling-parent anomaly query: typed columns, plus a NOT NULL guard so rows holding only a `depends_on_external` reference aren't miscounted as dangling
- `AutoClose` exclusion subqueries: `depends_on_issue_id`
- `batchDeleteRows` reverse-dependency cleanup: `depends_on_wisp_id`

## Verification

- `go build ./internal/reaper` and `go test ./internal/reaper` pass
- Running a build of this branch against a live town: `gt reaper scan --dry-run` succeeds on all 11 databases that previously failed with Error 1105

PR #4239, opened 2026-06-11
fix(convoy): bdDepListRawIDs reads migrated dependency columns (unblocks mountain wave advancement)

rbriski:
## Problem

`bdDepListRawIDs` (internal/cmd/convoy.go) queries a single `depends_on_id` column, but the beads schema migration split the dependency target into typed columns: `depends_on_issue_id` / `depends_on_wisp_id` / `depends_on_external` (the last holding cross-database refs as `external:<prefix>:<id>`).

Against a migrated database, every convoy dep lookup fails:

```
bd sql for deps of <id>: Error 1105 (HY000): column "depends_on_id" could not be found in any table in scope
```

This breaks `gt convoy status`, `gt mountain status`, and — critically — the ConvoyManager's wave advancement, so **a mountain stalls after Wave 1**. (Reproduced live: a 10-bead mountain dispatched Wave 1 then could not advance.) Same schema-skew class as #4228 (reaper) and #4230 (beads bump).

## Fix

Handle both query directions against the split columns:
- **down** (targets issueID depends on): `SELECT COALESCE(depends_on_issue_id, depends_on_wisp_id, depends_on_external) AS depends_on_id`
- **up** (issues that depend on issueID): match issueID against all three, incl. `depends_on_external LIKE '%:<id>'` for cross-DB refs

`beads.ExtractIssueID` already unwraps `external:<prefix>:<id>`. Input is still validated by `isValidBeadID` before interpolation.

## Verification

- `go build ./cmd/gt` OK (ICU flags)
- `internal/cmd` test failures are **pre-existing on main** (the same set — TestCreateAutoConvoy_BasicSuccess, TestGetTrackedIssues_FallsBackToShowTrackedDependencies, etc. — fails on unmodified origin/main)
- Live: this is the carry build now driving a real cross-DB mountain (hq convoy tracking bizsearch beads via depends_on_external)

## Follow-up (not in this PR)

Sibling `depends_on_id` sites in other subsystems remain and should get a focused, carefully-tested change (they mutate data): `compact.go` `cleanOrphanedWispDeps` (DELETE) and `doctor/misclassified_wisp_check.go` (INSERT backfill).

### DIFF
--- a/internal/cmd/convoy.go
+++ b/internal/cmd/convoy.go
@@ -486,8 +486,8 @@ func runBdJSONWithOptions(dir string, allowStale bool, args ...string) ([]byte, 
 //
 // Returns deduplicated, unwrapped issue IDs (external:prefix:id → id).
 func bdDepListRawIDs(dir, issueID, direction, depType string) ([]string, error) {
-	// Bead IDs are system-generated alphanumeric strings with hyphens and
-	// dots — validate to prevent injection before interpolating below.
+	// Bead IDs are system-generated alphanumeric strings with hyphens, dots,
+	// and underscores — validate to prevent injection before interpolating below.
 	if !isValidBeadID(issueID) {
 		return nil, fmt.Errorf("invalid bead ID: %q", issueID)
 	}
@@ -503,8 +503,8 @@ func bdDepListRawIDs(dir, issueID, direction, depType string) ([]string, error) 
 		selectExpr = "issue_id"
 		parseKey = "issue_id"
 		whereClause = fmt.Sprintf(
-			"(depends_on_issue_id = '%s' OR depends_on_wisp_id = '%s' OR depends_on_external LIKE '%%:%s')",
-			issueID, issueID, issueID)
+			"(depends_on_issue_id = '%s' OR depends_on_wisp_id = '%s' OR %s)",
+			issueID, issueID, sqlExternalDepTargetClause(issueID))
 	} else {
 		selectExpr = "COALESCE(depends_on_issue_id, depends_on_wisp_id, depends_on_external) AS depends_on_id"
 		parseKey = "depends_on_id"
@@ -541,6 +541,12 @@ func bdDepListRawIDs(dir, issueID, direction, depType string) ([]string, error) 
 		}
 	}
 	return ids, nil
+}
+
+func sqlExternalDepTargetClause(issueID string) string {
+	// Use an escape character that is not valid in bead IDs so underscores stay literal.
+	escapedID := strings.ReplaceAll(issueID, "_", "!_")
+	return fmt.Sprintf("depends_on_external LIKE '%%:%s' ESCAPE '!'", escapedID)
 }
 
 // isValidBeadID checks that a string is safe for SQL interpolation in dep queries.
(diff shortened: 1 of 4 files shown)


Bella-Giraffety:
## PR Sheriff Review

Fixup pushed to this branch: `f11f5622 fix: harden migrated convoy dependency lookup (gt-pr-sheriff-4239)`.

Labels applied: `status/accepted`, `kind/bug`, `priority/p1`.

Final recommendation: accept this bug fix, but do not merge #4239 standalone. Merge only after or stacked behind #4230, which aligns gastown with the beads v1.0.5 migrated dependency schema.

### 15 Research Legs
1. Down `COALESCE` query is correct for migrated typed columns; old-schema DBs require #4230 ordering.
2. Up query finds external refs, but the original `LIKE` needed underscore escaping.
3. Scalar call sites preserve local/external raw ID behavior; stale tests were likely to fail.
4. SQL stubs still matched old `depends_on_id` up queries.
5. Beads v1.0.5 final schema uses split target columns and drops `depends_on_id`.
6. A batch-helper concern came from divergent fork main; that helper is not present on this upstream PR branch.
7. Injection validation was sound; `_` wildcard was the remaining SQL risk.
8. `ExtractIssueID` and parse keys were correct.
9. Changed-file formatting/staticcheck looked clean; PR-specific stale-stub failures needed fixing.
10. Scalar helper change is cleanup-aligned; no legacy shim unless merge order cannot be enforced.
11. Other `depends_on_id` sites are real follow-ups but out of this convoy PR scope.
12. Issue/wisp/external target precedence and dedupe are correct for valid migrated rows.
13. `COALESCE(...) AS depends_on_id` produces the expected JSON key.
14. Operational impact is P1: convoy/mountain status and wave advancement can stall.
15. High-level review found a valid bug fix that needed targeted fixups before acceptance.

### 5 Pre-Implementation Reviews
1. PASS: actual upstream PR branch lacks the local fork batch helper; fix scalar LIKE and stale stubs only.
2. PASS: no other PR-branch production code should change; record #4230 dependency.
3. PASS: update up-direction stubs, keep down alias, add underscore escaping test.
4. PASS after using `ESCAPE '!'` instead of backslash escaping.
5. PASS: maintainer fixup is policy-allowed for this bug PR.

### 5 Post-Implementation Reviews
1. PASS: implementation is minimal and keeps underscores literal in external matches.
2. PASS: targeted fixtures match the migrated typed-column query.
3. PASS: validation plus `ESCAPE '!'` leaves no wildcard/injection issue for allowed IDs.
4. PASS: final branch is based on #4239 and contains no unrelated `.opencode` or fork-main changes.
5. PASS with merge hold: code accepted, but #4230 must land first or be stacked first.

### Verification
PASS: targeted `go test ./internal/cmd -run 'TestBatchSling_ErrorsOnAlreadyTrackedBead|TestConvoyTracksBead|TestFindConvoyByDescription_FallsBackToTrackedDeps|TestIsTrackedByConvoy_FoundViaDepList|TestPrintConvoyConflict_PrintsConflictInfo|TestBdDepListRawIDs|TestSQLExternalDepTargetClauseEscapesUnderscore' -count=1`

PASS: `go build ./cmd/gt`

BASELINE FAIL: full `go test ./internal/cmd -count=1` still fails on known unrelated main failures, including agent sorting/session tests, convoy routing fallback stubs, JSON convoy stage tests, and `TestEnsureBeadsConfigYAML_CreatesWhenMissing`.

Bella-Giraffety:
CI update after fixup push: PR head is now `f11f5622`. Windows smoke and repo guard checks passed. Lint/Test/Integration still fail, but the prior PR-specific stale-stub failure `TestBatchSling_ErrorsOnAlreadyTrackedBead` is no longer in the failure set; remaining failures match the known baseline class already noted above. Recommendation remains accepted with merge hold until #4230 lands first.

Issue #4240, opened 2026-06-11
Nudge Escape-suppression depends on fragile "esc to interrupt" status string


andrewboldi:
## Summary

Nudge delivery suppresses the vim-mode `Escape` keystroke when the target agent is actively generating, so a system nudge does not interrupt in-flight work (notably the Mayor). The "is the agent generating?" decision relies on scraping the agent TUI status bar for the literal substring `esc to interrupt` (`hasBusyIndicator` in `internal/tmux/tmux.go`).

## The fragility

This is a heuristic coupled to Claude Code's (and Codex/Gemini's) rendered status text. It is **not new** — `IsIdle` / `WaitForIdle` already depend on the same `esc to interrupt` string — but the nudge Escape-suppression now extends that assumption to a correctness-critical path:

- If the upstream status text changes, `shouldSendEscape` stops suppressing the Escape → nudges can again interrupt the agent mid-turn (the exact bug the linked PR fixes).
- `IsIdle` / `WaitForIdle` would silently misbehave at the same time.

It fails **open and silently** — nothing alerts us if the string disappears.

## Possible follow-ups

- Centralize busy/idle detection strings per-agent in the agent preset config (one place to maintain).
- Add a canary/integration check that flags when the expected status string is absent from a known-generating session.
- Prefer a more structural readiness signal than scraping the TUI status bar where one is available.

## Priority

Low / tech-debt. The current heuristic is already trusted by `IsIdle`, so the linked PR does not introduce new fragility — it just makes an existing assumption load-bearing for nudge correctness. Filing so the coupling is tracked.

andrewboldi:
Fix proposed in #4242 (**open — not yet merged**). Once it lands, the nudge Escape will be gated on the busy indicator (`esc to interrupt`) so it no longer interrupts an actively-generating agent (e.g. the Mayor); until then the bug is still live on `main`.

This issue stays open to track the **remaining** fragility of the `esc to interrupt` string heuristic itself (shared with `IsIdle`/`WaitForIdle`) — #4242 depends on that heuristic rather than resolving it.

andrewboldi:
Resolution proposed in #4244 (open): centralizes the busy-indicator marker into a single documented source of truth (`busyIndicators`), matched as a list so a variant/upstream-rename is a one-line change, with `TestBusyIndicators` pinning the known marker so changes are intentional. Per-agent preset config and a runtime canary are deliberately deferred (rationale in the PR). Will auto-close on merge via `Closes #4240`.

PR #4213, opened 2026-06-09
fix: protect remote polecat prune deletes

Bella-Giraffety:
Supersedes #3847 and replaces closed PR #4208 with a clean branch based on current upstream/main.

This keeps the already-upstreamed hash-preserving remote prune classifier behavior and adds the remaining safety guard from the sheriff review: remote prune now deletes a branch only if it still points at the advertised hash used for classification.

Original author credit preserved in the commit trailer: Co-Authored-By: Brandon Smyth <bsmyth77@gmail.com>.

Sheriff summary:
- R1-R15: root cause was remote prune classifying remote-only refs by short branch name; the narrow fix is to use ls-remote hashes from the push target and avoid broader active-polecat filtering or cleanup layers.
- P1-P5: pre-implementation reviewers approved the narrow hash-preserving classifier approach with focused regression coverage and no scope expansion.
- F1-F5: final post-review found no blockers; two reviewers noted the pre-existing summary counter can count failed remote deletes, but per-branch warnings remain clear and preserving existing counter behavior matches the prior scope decision.

Validation:
- PASS: go test -count=1 ./internal/git -run 'Test(ListPushRemoteRefsWithHashes(ClassifiesRemoteOnlyMergedBranch|UsesPushURLHash)|DeleteRemoteBranchIfAtRejectsChangedBranch)$'
- PASS: go test -count=1 ./internal/cmd -run 'TestPolecat|TestPrune'
- PASS: make build
- PASS: git diff --check upstream/main...HEAD
- FAIL unrelated/pre-existing: go test -count=1 ./internal/git ./internal/cmd failed in internal/cmd tests including TestFilterAndSortSessions_SortOrder, TestGuessSessionFromWorkerDir/different_rig, convoy tracking/JSON tests, and TestEnsureBeadsConfigYAML_CreatesWhenMissing. internal/git passed in that broader run.

PR #4242, opened 2026-06-11
fix(tmux): suppress nudge Escape while agent is generating (stop interrupting the Mayor)

andrewboldi:
## Problem

System nudges are delivered to agents via `tmux send-keys`. The delivery protocol (`NudgeSessionWithOpts` / `NudgePane`) sends a literal **Escape** keystroke (step 5) before Enter — originally to exit a vim-mode composer's INSERT mode so Enter submits (GH#307), commented as *"harmless for bash/Claude Code"*.

It is **not** harmless for Claude Code: Escape cancels in-flight generation. So when a nudge lands while the agent is mid-turn — the **Mayor** being the common victim — the Escape interrupts its current work. This affects every direct `NudgeSession` caller: witness handlers, the daemon health-check heartbeat, the mail router, broadcast, and `gt nudge`.

`EscapeCancelsRequest` already existed to skip the Escape for agents where it cancels generation (Gemini; Copilot auto-skipped), but Claude was wrongly assumed exempt.

## Fix

Gate the Escape on the agent's busy state instead of sending it unconditionally:

- `shouldSendEscape(target)` captures the pane and returns `false` when the busy indicator (`esc to interrupt`) is present — i.e. the agent is generating.
- Step 5 now runs only when `!SkipEscape && shouldSendEscape(...)`.
- Applied to **both** `NudgeSessionWithOpts` and `NudgePane` (the latter previously sent Escape with no gate at all).

This reuses the exact signal `IsIdle` / `WaitForIdle` already trust, so it adds no new external assumption. Behavior is strictly safer:

- **Idle** agent → Escape still sent (vim-mode GH#307 preserved).
- **Generating** agent → Escape suppressed (no interrupt); the typed text + Enter simply queue as the next turn.
- Non-Claude agents that don't render `esc to interrupt` are unaffected (gate returns `true` → unchanged).

### Why not the alternatives
- *Set `EscapeCancelsRequest: true` for Claude* — simplest, but drops the vim-mode INSERT-exit entirely, even when idle.
- *Sniff vim mode from the pane* — fragile and version-dependent; false negatives strand vim users, false positives reintroduce the interrupt. The busy-state signal **is** the actual harm condition and is already battle-tested.

## Design note
On pane-capture error, `shouldSendEscape` returns `false` (suppress). When we can't confirm idle, not interrupting the agent is the safer default, and it's harmless for the common non-vim case where Enter alone submits. Matches `IsIdle`'s return-false-on-error convention.

## Fragility (tracked)
Detection relies on the literal `esc to interrupt` status string. This is **pre-existing** (shared with `IsIdle` / `WaitForIdle`), now load-bearing for nudge correctness. Documented in-code and tracked in #4240.

## Tests
- `TestShouldSendEscapeForLines` — pure decision logic (busy ⇒ suppress; idle/empty ⇒ allow).
- `TestShouldSendEscape_LivePane` — real tmux pane: idle ⇒ allow, `esc to interrupt` rendered ⇒ suppress. Capture-only, so it sidesteps the `sendEnterVerified` flakiness of the full nudge integration tests.

`go build` / `go vet` / `gofmt` clean on `internal/tmux`.

> Note: 3 pre-existing `TestNudgeSession_*` real-tmux integration tests fail in my local sandbox — verified **identical with these changes stashed** (a `sendEnterVerified` environment timing issue, not introduced here).

Relates to #4240

### DIFF
--- a/internal/config/agents.go
+++ b/internal/config/agents.go
@@ -160,9 +160,9 @@ type AgentPresetInfo struct {
 
 	// EscapeCancelsRequest indicates that sending an Escape keystroke to this
 	// agent cancels its in-flight generation. NudgeSession normally sends
-	// Escape (step 5) to exit vim INSERT mode — harmless for bash/Claude Code,
-	// but destructive for agents like Gemini CLI where Escape aborts the
-	// active request. When true, NudgeSessionWithOpts skips the Escape
+	// Escape (step 5) to exit vim INSERT mode — harmless for bash, but
+	// destructive for agents where Escape aborts the active request. When true,
+	// NudgeSessionWithOpts skips the Escape
 	// keystroke and the 600ms readline timeout that follows it.
 	EscapeCancelsRequest bool `json:"escape_cancels_request,omitempty"`
 
--- a/internal/tmux/tmux.go
+++ b/internal/tmux/tmux.go
@@ -1722,6 +1722,19 @@ func (t *Tmux) NudgeSessionWithOpts(session, message string, opts NudgeOpts) err
 	// 2. Sanitize control characters that corrupt delivery
 	sanitized := sanitizeNudgeMessage(message)
 
+	if !opts.SkipEscape {
+		// Auto-skip Escape for Copilot CLI sessions. Escape cancels in-flight
+		// generation in Copilot CLI (like Gemini), leaving the nudge text
+		// stranded in the input field without Enter being processed. (hq-isz)
+		agentType, _ := t.GetEnvironment(session, "GT_AGENT")
+		if agentType == "copilot" {
+			opts.SkipEscape = true
+		}
+	}
+	// Snapshot before typing the nudge so the message text itself cannot look
+	// like the agent's busy indicator.
+	sendEscape := !opts.SkipEscape && t.shouldSendEscape(target)
+
 	// 3. Send text via send-keys -l. Messages > 512 bytes are chunked
 	//    with 10ms inter-chunk delays to avoid argument length limits.
 	if err := t.sendMessageToTarget(target, sanitized); err != nil {
@@ -1732,23 +1745,7 @@ func (t *Tmux) NudgeSessionWithOpts(session, message string, opts NudgeOpts) err
 	// enough time to process all chunks under load. (GH#gt-0b5)
 	time.Sleep(adaptiveTextDelay(len(sanitized)))
 
-	if !opts.SkipEscape {
-		// Auto-skip Escape for Copilot CLI sessions. Escape cancels in-flight
-		// generation in Copilot CLI (like Gemini), leaving the nudge text
-		// stranded in the input field without Enter being processed. (hq-isz)
-		agentType, _ := t.GetEnvironment(session, "GT_AGENT")
-		if agentType == "copilot" {
-			opts.SkipEscape = true
-		}
-	}
-
-	// Only send the vim-mode Escape when the agent is NOT actively generating.
-	// In Claude Code, Escape cancels in-flight generation — the status bar reads
-	// "esc to interrupt" while the agent is working — so sending it mid-turn
-	// would interrupt the agent's current work (e.g. the Mayor). When the agent
-	// is idle, Escape harmlessly exits a vim-mode composer's INSERT mode so the
-	// following Enter submits (GH#307).
-	if !opts.SkipEscape && t.shouldSendEscape(target) {
+	if sendEscape {
 		// 5. Send Escape to exit vim INSERT mode if enabled (harmless in normal mode)
 		// See: https://github.com/anthropics/gastown/issues/307
 		_, _ = t.run("send-keys", "-t", target, "Escape")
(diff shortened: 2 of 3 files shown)


Bella-Giraffety:
PR Sheriff review complete for #4242.

15 research legs:
R1 nudge flow: scoped bugfix, core approach sound with heuristic caveat.
R2 tmux API: reuses existing `IsIdle`/busy signal idiomatically.
R3 escape logic: found one actionable false-positive risk when checking after typed nudge text.
R4 tests: helper/live coverage useful; avoid flaky full delivery test.
R5 CI: original lint/test/integration failures match red `main`, not touched files.
R6 lint: failing lint is unrelated repo-wide debt.
R7 integration: behavior intentionally narrows Escape delivery; targeted tmux tests are the right local gate.
R8 busy signal: reuse is convergent, though long-term structured idle events would be cleaner.
R9 SkipEscape options: Copilot/Gemini paths preserved; stale Claude comment needed cleanup.
R10 NudgePane: gating is required because sling paths can target live busy panes.
R11 command safety: no new shell/quoting risk.
R12 user impact: Mayor interruption fixed while idle vim-mode behavior is preserved.
R13 style: acceptable with duplicated rationale collapsed.
R14 linked issues: real bug, must preserve GH#307 vim-mode behavior.
R15 policy: bugfix, P1, kind/bug, maintainer fixup allowed.

5 pre-implementation reviews:
P1 correctness: approved pre-text busy snapshot.
P2 regression: approved if snapshot occurs after rewind/copy-mode cleanup.
P3 tests: helper/live coverage sufficient; add fail-closed capture-error test.
P4 style: update stale `EscapeCancelsRequest` comment.
P5 policy: allowed to fix up rather than request contributor changes.

Fixup pushed to this PR branch:
- Preserves contributor's central `shouldSendEscape` approach.
- Computes `sendEscape` before injecting nudge text in both `NudgeSessionWithOpts` and `NudgePane`, so the message itself cannot match `esc to interrupt` and suppress idle vim-mode Escape.
- Keeps the 600ms readline wait and rewind recovery only when Escape is actually sent.
- Adds capture-error fail-closed coverage and updates the stale agent config comment.

5 post-implementation reviews:
V1 correctness: approved; both nudge paths consistent.
V2 tests: sufficient and non-flaky; no E2E blocker.
V3 cleanup: passes cleanup-first review; no new shim/bandaid surface.
V4 risk review: only non-blocking residual heuristic/race caveats remain.
V5 merge review: approve with baseline-CI note.

Local verification:
- PASS: `go test ./internal/tmux -run 'Test(HasBusyIndicator|ShouldSendEscapeForLines|ShouldSendEscape_LivePane|ShouldSendEscape_CaptureErrorSuppressesEscape)' -count=1`
- PASS: `go build ./...`
- BASELINE FAIL: `go test ./internal/tmux -count=1` fails on existing `sleep` vs `coreutils` tmux environment expectations, unrelated to this change.

Recommendation: approve/merge the current fixed-up PR branch. If GitHub CI remains red with the same repo-wide baseline failures seen on `main`, do not attribute those failures to this PR.

Bella-Giraffety:
CI reran after the fixup commit and remains red in unrelated baseline areas: repo-wide lint debt, convoy/session/beads runtime tests, and config expectation tests. No failure is in internal/tmux or in the new Escape-gating tests. Sheriff recommendation remains: approve/merge this fixed-up PR branch if baseline CI failures are waived or resolved separately.

andrewboldi:
_Restoring a comment I removed earlier by mistake._

## CI triage — red checks are a pre-existing baseline, not from this PR

This change touches only `internal/tmux` (two unexported helpers + one gated keystroke). The red `Test` / `Integration Tests` / `Lint` checks are inherited from `main`, verified three independent ways:

**1. This PR's own tests all pass in CI:**
- `internal/tmux.TestShouldSendEscapeForLines` (all 6 subtests) ✅
- `internal/tmux.TestShouldSendEscape_LivePane` ✅
- `internal/tmux.TestNudgeSession_WithRetry` / `WithStoredPaneID` / `WakesAgentWindowNotActiveWindow` ✅ (these only flaked in my local sandbox; CI's tmux env passes them)

**2. Every failing test/lint finding is in a package/file this PR does not touch:**
- Failing tests: `internal/cmd` (`TestGuessSessionFromWorkerDir`, `TestFilterAndSortSessions_SortOrder`, `TestRunConvoyList/Status_UsesTownRootAndStripsBeadsDir`, `TestGetTrackedIssues_*`, `TestJSONOutput_*`, `TestEnsureBeadsConfigYAML_CreatesWhenMissing`), `internal/polecat`, `internal/refinery`, `internal/rig` — all beads / town-root / session-dir tests.
- Lint findings: `internal/doltserver/doltserver.go`, `internal/cmd/statusline.go`, `internal/cmd/capacity_dispatch.go`, `internal/git/git.go` — none in `internal/tmux`.

**3. The base commit already fails these jobs.** `main @ ccd16c18` (the #4213 merge this branch forks from) shows `X Test` and `X Lint` in its own CI run — i.e. `main` is red independently of this PR. (Every recent `main` push run is `failure`.)

No action needed on this PR for the red checks; they reflect existing breakage on `main`.

andrewboldi:
Reviewed the maintainer fixup (`0ef7d665`) on the contributor side — it's a strict improvement and I agree with it:

- **R3 fix is correct.** Snapshotting `sendEscape` *before* injecting the nudge text closes the false-positive where a message containing `esc to interrupt` could suppress the idle vim-mode Escape. No inverse risk: an idle pane can't become busy before the Enter (typing doesn't start generation — only the Enter does), so this doesn't add false-negatives. Applied symmetrically in `NudgePane`. 👍
- The capture-error fail-closed test and the stale `EscapeCancelsRequest` comment cleanup are good catches.

Local verification on the fixed-up branch: `go test ./internal/tmux -run 'Test(HasBusyIndicator|ShouldSendEscape)'` passes; `go build` / `go vet` / `gofmt` clean.

No further changes from my side — good to merge. The `esc to interrupt` heuristic fragility is tracked separately in #4240.

PR #4278, opened 2026-06-15
fix(daemon): dog-molecule lifecycle — capture step IDs from pour + route closes through in-process store

rbriski:
Two related dog-molecule lifecycle fixes in `internal/daemon/dog_molecule.go`.

## 1. Capture step IDs from the pour (gt-k61r)
The daemon re-read molecule children via a separate `bd show --children` connection that, under concurrent Dolt write load, couldn't see its own just-committed molecule (bd returned its no-issue envelope) — so step wisps were never discovered or closed and leaked permanently. Capture every step wisp ID from the pour's own `--json` `id_mapping` response instead; the pour's connection always observes its own writes. Close steps (with `--force`) directly from the captured IDs.

## 2. Route dog closes through the in-process store (part of #4292)
Each `bd close` subprocess builds its own DoltStore and opens a fresh short-lived connection pool. The daemon already holds one long-lived in-process beads store (`d.beadsStores`); route dog wisp closes through `store.CloseIssue` when available, falling back to the `bd` subprocess only when no store is present (e.g. during shutdown). The store UPDATE is unconditional (matching the old `--force`) and idempotent on an already-closed wisp.

**Scope note:** measurement showed the daemon's own dog-close spawns are ~1% of total Dolt connection churn, so this is dog-molecule correctness/hygiene that *contributes to* #4292, not a standalone fix for the connection storm (dominated by external `bd`/`gt` process invocations). Adds a `beads.BDSpawnCount` measurement seam + a regression test asserting the close path spawns zero subprocesses when a store is set.

CI note: the repo's Test job currently fails for everyone because it doesn't install `bd`; the failures here are not from this change.

### DIFF
--- a/internal/beads/exec.go
+++ b/internal/beads/exec.go
@@ -5,9 +5,25 @@ import (
 	"os"
 	"os/exec"
 	"strings"
+	"sync/atomic"
 
 	"github.com/steveyegge/gastown/internal/util"
 )
+
+// bdSpawnCount counts bd subprocesses configured via ConfigureCommand (every
+// bd call that flows through the shared Gas Town bd policy). It exists to
+// measure connection-amplification reduction for gt-ye21/#4292: each bd
+// subprocess opens its own short-lived Dolt connection pool, so spawn count is
+// a direct proxy for connection churn under patrol load.
+var bdSpawnCount atomic.Int64
+
+// BDSpawnCount returns the cumulative number of bd subprocesses configured via
+// ConfigureCommand since process start (or since the last ResetBDSpawnCount).
+func BDSpawnCount() int64 { return bdSpawnCount.Load() }
+
+// ResetBDSpawnCount zeroes the spawn counter. Used by tests and per-window
+// measurement to count spawns within a bounded interval.
+func ResetBDSpawnCount() { bdSpawnCount.Store(0) }
 
 // SubprocessEnvMode describes how a bd subprocess should target Dolt and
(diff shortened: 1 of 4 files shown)


rbriski:
Follow-on commit: the --json capture fixed *discovery*, but verifying in production showed the steps still leaked on close — molecule steps carry a needs DAG (probe→inspect→report) and `bd close` refuses a step "blocked by open issues" until its deps close. close() iterates stepIDs in map order, so dependency-blocked steps failed to close. Added `--force` to the step + root closes (teardown ignores work-ordering). Verified live: a freshly-poured molecule's 3 steps now all close (0 open).

rbriski:
Heads up on the red checks here: the **Test** and **Integration Tests** failures are a CI-environment issue, not this PR. The runner doesn't have `bd` on PATH — `agent_beads_check_test.go` fails with `exec: "bd": executable file not found in $PATH`, which cascades to `TestJSONOutput_NoHumanReadableText` / `TestJSONOutput_ErrorsReturnNonZeroExit` ('unexpected end of JSON input', empty `bd show` output). The same two suites fail identically on unrelated PRs (e.g. #4311, a 2-line config change), and Lint + Windows Smoke pass here. This branch was just rebased clean onto current main. So it's the CI image missing `bd`, not the code.

Bella-Giraffety:
Closing stale draft PR in favor of bug issue #4456. We will re-check whether the bug still exists from the issue before any cleanup-first fixup.

PR #4295, opened 2026-06-17
fix(dolt): cap managed sql-server heap with GOMEMLIMIT soft ceiling

seanbearden:
## Problem

The managed Dolt `sql-server` accumulates an in-memory working set across all rig databases with no memory ceiling of its own (`auto_gc` is off and no `GOMEMLIMIT` is set), so its RSS climbs **unbounded** until the host OOM-kills it. On one town this was observed as ~2.3 GB → 24 GB over ~100 min on a ~3 GB on-disk dataset, OOM-crashing roughly every 6 hours (the daemon supervisor reboots it each time, masking the root cause).

This is distinct from two already-addressed issues:
- the Stats V2 collector leak (mitigated by `dolt_stats_enabled: 0` in the generated config), and
- reaper wisp floods.

A 90-minute monitor with open-wisp count flat while RSS climbed confirmed the growth is the server's own heap, not those subsystems.

## Fix

Set `GOMEMLIMIT=8GiB` (a Go **soft** memory limit) on the spawned `dolt sql-server` process so the runtime GCs more aggressively as the heap approaches the limit, instead of letting RSS grow until OOM. Applied at **both** spawn sites so a crash-restart is covered too, not only a manual start:

- `internal/doltserver/doltserver.go` `Start()` — the `gt dolt start` path
- `internal/daemon/dolt.go` `startLocked()` — the daemon crash-supervisor restart path

Details:
- Default `8GiB` leaves ample headroom over the few-GB dataset on a 64 GB host while capping growth.
- Tunable via `GT_DOLT_GOMEMLIMIT` (e.g. `"12GiB"`); set to `"off"` or empty to disable the soft limit entirely. This mirrors the existing `GT_DOLT_STATS_ENABLED` override philosophy.
- Any inherited `GOMEMLIMIT` is stripped before appending ours, so the Go runtime reads our value (it uses the first match on duplicate env keys) — same defensive pattern the codebase already uses for `DOLT_CLI_PASSWORD`.

## Tests

Added `TestDoltServerMemLimit` and `TestDoltServerEnv` covering the default, the `GT_DOLT_GOMEMLIMIT` override, the disable cases, inherited-value replacement, and preservation of unrelated env vars.

## Verification

Deployed to a live town: the spawned `dolt sql-server` process now carries `GOMEMLIMIT=8GiB` in its environment (confirmed via `ps eww`), and the server starts cleanly with all databases verified.

### DIFF
--- a/internal/daemon/dolt.go
+++ b/internal/daemon/dolt.go
@@ -938,6 +938,11 @@ func (m *DoltServerManager) startLocked() error {
 	cmd.Dir = m.config.DataDir
 	cmd.Stdout = logFile
 	cmd.Stderr = logFile
+	// Cap the server's heap with a GOMEMLIMIT soft ceiling so Dolt's unbounded
+	// in-memory working-set growth gets GC'd under pressure instead of climbing
+	// until OOM. The supervisor restarts dolt on crash, so the limit must be set
+	// here too (not only on the `gt dolt start` path).
+	cmd.Env = doltserver.DoltServerEnv(os.Environ())
 
 	// Detach from this process group so it survives daemon restart
 	setSysProcAttr(cmd)
--- a/internal/doltserver/doltserver.go
+++ b/internal/doltserver/doltserver.go
@@ -1687,6 +1687,47 @@ behavior:
 	return os.WriteFile(configPath, []byte(content), 0600)
 }
 
+// DefaultDoltServerMemLimit is the default soft heap ceiling (GOMEMLIMIT)
+// applied to the managed dolt sql-server. Dolt accumulates an in-memory working
+// set across all databases with no ceiling of its own; without a soft limit the
+// Go runtime lets RSS climb unbounded until the host OOM-kills the server. 8GiB
+// leaves ample headroom over the few-GB on-disk dataset on a 64GB host while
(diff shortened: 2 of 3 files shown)


seanbearden:
Closing: live verification showed GOMEMLIMIT=8GiB does NOT solve the problem.

Over a 90-min monitor the dolt sql-server RSS climbed straight through the 8GiB soft cap to ~14.8GB (1.85x) and the server still cycled (went unresponsive -> supervisor restart; not an OS OOM — host was ~80% free). GOMEMLIMIT only governs the Go-managed heap's GC; the unbounded growth is dominated by memory the GC cannot reclaim — Dolt's NBS storage layer mmaps table files across all DBs (mmap'd pages count in RSS but are invisible to GOMEMLIMIT) plus live off-heap working set. There's also a risk the soft cap induces futile GC/CPU pressure once exceeded, triggering earlier unresponsive-restarts.

So this is the wrong lever. The real fix needs to target Dolt's actual consumer (chunk/table cache sizing, the actively-mapped DB working set, or a scheduled bounded restart). Reverted the corresponding live deploy. Reopen/rework if a tuned variant proves useful.

PR #4309, opened 2026-06-21
fix(daemon): reclaim AgentDead dogs in cleanupStuckDogs (hq-54br)

munizr13:
## Summary

Fixes **hq-54br** (P1): `mol-dog-reaper` leaves its dog stuck in `state=working` after the agent exits without calling `gt dog done`.

The daemon's `cleanupStuckDogs` already returned **SessionDead** dogs (tmux session gone) to idle on each heartbeat — confirmed firing every reaper cycle in `daemon.log`, so the common case already self-heals (no runaway loop). The real remaining gap: it used a crude `HasSession` check that **missed AgentDead** — a lingering tmux session whose agent process died. Those husks sat zombied until `detectStaleWorkingDogs`' multi-hour stale timeout, holding a dog out of the 4-dog pool.

## Change

`cleanupStuckDogs` now classifies session health and reclaims **both** failure modes on the next heartbeat:
- `SessionDead` → clear work → idle (unchanged behavior)
- `AgentDead` → kill the husk session → clear work → idle (new)

Hung-but-live dogs are deliberately left to `detectStaleWorkingDogs` (a huge inactivity window ensures a slow-but-alive agent is never treated as hung here). This makes the supervisor's guaranteed-on-exit idle-return comprehensive across all agent-exit modes — robust even against SIGKILL, which an in-agent defer/trap could never catch.

Scope is intentionally limited to `handler.go` / `handler_test.go` per the bead; the separate "why does the reaper agent exit without `gt dog done` every cycle" session-exit lifecycle question is left for a future bead.

## Test
- `go build` + `go vet`: clean
- New regression tests: `TestCleanupStuckDogs_ClearsDeadSessionWorker`, `TestCleanupStuckDogs_SkipsIdleDogs`
- Full `internal/daemon` package: **pass** (219s)

### DIFF
--- a/internal/daemon/handler.go
+++ b/internal/daemon/handler.go
@@ -37,6 +37,13 @@ var (
 	maxDogPoolSize = config.DefaultMaxDogPoolSize
 )
 
+// neverHungInactivity is a deliberately enormous inactivity window passed to
+// CheckSessionHealth from cleanupStuckDogs. It ensures a live-but-slow agent is
+// never classified as AgentHung there — cleanupStuckDogs reclaims only genuinely
+// dead agents (SessionDead/AgentDead), while hung-but-alive dogs remain the job
+// of detectStaleWorkingDogs, which honors the configured grace period.
+const neverHungInactivity = 100 * 365 * 24 * time.Hour
+
 // handleDogs manages Dog lifecycle: cleanup stuck dogs, reap idle dogs, then dispatch plugins.
 // This is the main entry point called from heartbeat.
 func (d *Daemon) handleDogs() {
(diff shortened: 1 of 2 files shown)


munizr13:
Manual governance hold: closing this PR because Rodrigo explicitly stopped autonomous Gas Town self-modification. Gas Town core/source changes must not be opened, pushed, merged, or managed by CTO/Deacon/Mayor autonomously. This repair may be reviewed and reintroduced separately by Rodrigo + Codex under manual ownership.

munizr13:
Closing under manual governance hold. No merge. The branch/history is preserved for review, but autonomous continuation is stopped.

Issue #4292, opened 2026-06-16
[BUG] spawn-per-query connection storm wedges dolt sql-server listener under patrol load


flux423:
## Summary

bd/gt spawn a new Dolt connection per query. Under patrol or multi-agent load this storms the `dolt sql-server` listener until max-waiting-connections wedges. Dolt stays alive but is unresponsive to all clients (alive-but-unresponsive wedge).

## Linked Bluefly Beads
- hq-xj5 — GT_BD_CONNECTION_AMPLIFICATION (root cause)
- hq-tx4x — gt done hangs notifying witness (symptom: completion path)
- hq-21oy — wisp GC i/o timeout (symptom: batch delete)
- hq-2qe — GT_DOLT_CONNECTION_AMPLIFICATION_001 (upstream tracking)
- hq-fk0 — CRITICAL: Dolt unreachable after mail inbox failure (goroutine dump captured)
- hq-paz — CRITICAL: Dolt unreachable after unexpected EOF (goroutine dump captured)

## Goroutine Dumps (captured on Oracle via SIGQUIT)
Three dumps at /tmp/dolt-hang-\*.log on host bluefly-platform (100.103.48.75):
- dolt-hang-1781494398.log — PID 1399241, 38/1000 conns, 403.9MB disk, 0s latency
- dolt-hang-1781498980.log — PID 1399241, 16/1000 conns, 312.0MB disk, 0s latency
- dolt-hang-1781512555.log — PID 3314451, 45/1000 conns, 386.2MB disk, 0s latency

All show Dolt alive with 0s query latency at capture, then unresponsive to all bd clients.

## Reproduction Recipe
```bash
# 1. Start deacon patrol sweep (burst of bd queries)
gt sling mol-deacon-patrol deacon

# 2. Simultaneously run wisp GC (single large write transaction)
bd mol wisp gc --closed --force

# Result: dolt sql-server wedges within 30-90s
# Symptom: every bd call returns 'dolt circuit breaker is open: server appears down'
# Dolt process alive, 0s latency at capture, but listener unresponsive to all clients
```

## Connection State at Wedge
- max_connections: 1000 (regenerates on restart — config does not persist across restarts)
- Active connections at wedge: 38-45 (well below limit)
- Mechanism: bd mol wisp gc runs single large transaction (recompute_is_blocked on 6836 wisps); patrol burst spawns unbounded new connections; long-running write blocks all readers until timeout cascade

## Current Impact
[...]

PR #4311, opened 2026-06-21
fix(daemon): raise Dolt read/write timeout 30s->300s (gt-ye21)

rbriski:
## Problem

`writeDaemonDoltConfig` in `internal/daemon/dolt.go` hardcodes the
daemon-managed `config.yaml` with `read_timeout_millis: 30000` /
`write_timeout_millis: 30000` (30s).

Under `hq` connection churn (gt-ye21), the beads issue-list query
(`SELECT i.id, i.content_hash, i.title, ...` against `hq`) can exceed 30s.
When it does, the server severs the connection mid-query
("use of closed network connection"), which surfaces to `bd`/mail clients
as `i/o timeout`. Repeatedly this drove the pool toward exhaustion and
crashed the server ("server unreachable"), blocking dogs from mail/dispatch.

This also contradicts the codebase's own intent: `doltserver.go` defines
`DefaultReadTimeoutMs = DefaultWriteTimeoutMs = 5 * 60 * 1000` (5min), but
the daemon path wrote 30s regardless.

## Change

Raise both daemon-written timeouts `30000 -> 300000` (5min), matching
`DefaultReadTimeoutMs`. Slow-but-completing queries now finish instead of
being killed.

## Validation

Built locally and ran for hours: Dolt stayed up with **no crashes**, vs
crashing every few minutes before. The acute crash/sever symptom is gone.

## Scope / follow-up

This fixes the **server-severing/crash** symptom only. The underlying
query slowness — a thundering herd of identical `hq` issue-list queries
under connection churn — is the broader **gt-ye21** work (convoy/wisp/session
teardown + reducing concurrent pollers). A separate attempt to raise the
*client-side* timeouts to match was reverted: under a severe herd it made
pileups worse by holding connections ~10x longer (the 30s client timeout
acts as a relief valve).

Draft pending further soak before marking ready.

### DIFF
--- a/internal/daemon/dolt.go
+++ b/internal/daemon/dolt.go
@@ -859,8 +859,8 @@ log_level: info
 
 listener:
   port: %d%s
-  read_timeout_millis: 30000
-  write_timeout_millis: 30000
+  read_timeout_millis: 300000
+  write_timeout_millis: 300000
   max_connections: 1000
 
 data_dir: %q


rbriski:
Closing in favor of the root fix for #4292. Raising the Dolt read/write timeout treats the symptom (slow queries get severed) rather than the connection amplification that causes the storm. With the wisp table bounded, queries no longer approach 30s, so the raise is unnecessary. Pursuing the structural fix instead: routing the daemon's hot patrol paths through its long-lived in-process beads store to eliminate per-query `bd` subprocess connection spawns.

PR #4261, opened 2026-06-12
fix(mail): converge inbox query fanout

Bella-Giraffety:
Supersedes #4063 with a cleanup-first maintainer fix.\n\nWhat changed:\n- Runs inbox assignee issues, CC issues, and wisp fetches as three independent top-level lanes.\n- Collapses wisp inbox lookup to one SQL query across identity variants, preserving assignee vs CC provenance.\n- Preserves existing semantics: assignee open+hooked are included; CC open only; assignee query errors remain fatal; CC/wisp remain best-effort.\n- Adds parser coverage plus behavior coverage for status filtering, wisp handling, dedupe, and single wisp SQL query.\n\nValidation:\n- git diff --check upstream/main...HEAD\n- gofmt -l changed Go files\n- go test ./internal/mail -count=1\n- GT_TEST_NO_NUDGE=1 go test ./internal/cmd -run 'Test.*Mail' -count=1 -timeout=120s\n- go build ./cmd/gt\n- go test -race ./internal/mail -run 'Test(MailboxListFromDirConvergesWispQueryAndFiltersStatuses|ParseBeadsListOutput)' -count=1\n\nPR Sheriff evidence is recorded on bead gt-pr-sheriff-4063. Original #4063 should not merge as-is because it parallelizes per-identity subprocess fanout instead of reducing wisp query work.

Issue #3623, opened 2026-04-13
Dolt connection exhaustion and slow mail/beads queries


hefgi:
## Problem

Two related issues with Dolt in multi-agent setups:

### 1. Connection exhaustion (death spiral)

Dolt's default `wait_timeout` is 28800s (8 hours). Short-lived `bd` processes open connections that linger long after the process exits. Under load (mayor + deacon + witness + refinery + 3-5 polecats + dashboard polling), connections accumulate faster than they expire, leading to a death spiral:

- Connections grow → Dolt slows down → queries take longer → more queries pile up → 1000/1000 connections → everything blocks
- Hundreds of zombie `bd sql --json SELECT` processes accumulate (observed 481 at once)
- Dashboard polling (every 30s) accelerates the spiral
- Only fix is `pkill -f 'bd '` + Dolt restart

**Workaround**: `SET GLOBAL wait_timeout = 30` after each Dolt start. This should be the default in the Dolt config template, or exposed as `gt config set dolt.wait_timeout`.

### 2. Slow `gt mail inbox` / `bd` queries

Even with healthy Dolt (10 connections, 0s query latency), `gt mail inbox` takes 23-58 seconds for 5 messages. Raw Dolt queries take 60ms.

The bottleneck is that `gt mail` spawns multiple `bd` processes sequentially, each with its own Go binary cold start + Dolt connection setup (~1-2s per invocation). For an inbox with 5 messages this adds up quickly.

## Environment

- 1 rig (scan_agent) with witness, refinery, deacon, mayor, 3-5 polecats
- Mixed agent setup (claude + pi + kimi)
- Dolt server: single instance, port 3307, max_connections 1000

## Suggested fixes

1. **Add `wait_timeout: 30` to the Dolt config template** (`config.yaml`) or expose via `gt config set dolt.wait_timeout`
2. **Connection pooling in `bd`** — reuse connections across queries within a single `gt` command instead of spawning separate `bd` processes
3. **Query batching in `gt mail`** — fetch all messages in one query instead of N sequential `bd` invocations
4. **Dashboard query coalescing** — batch dashboard poll queries to reduce connection pressure

jeffplourde:
Confirmed this failure mode in a multi-agent setup (mayor + deacon + ~8 crew/polecats running continuously). Dolt went down overnight with no OOM kills or kernel crashes visible — the process just disappeared. Previous session log was overwritten so root cause is inferred: connection exhaustion death spiral matches exactly.

[...]

PR #4314, opened 2026-06-22
fix(reaper): lower default wisp delete_age 7d->24h (gt-ye21, #4292)

rbriski:
## Problem

`SearchIssues` with no ephemeral filter (what `gt mail inbox` uses) scans the live `wisps` table and merges results. The reaper's `delete_age` default is **7 days**, so on a high-churn town (~200 wisps/hour from convoy steps, dog molecules, plugin receipts, mail, escalations) the live table bloats with tens of thousands of closed wisps. Every mail read then scans that whole table, and under patrol/multi-agent load it thundering-herds the Dolt listener.

This is the wisp-table-size half of #4292 (the other half being per-query connection amplification). It's independent of connection count: #3623 saw `gt mail inbox` take 23-58s at only 10 connections.

## Field data

On our `hq`: wisps reached **33,647 rows (33,565 closed)** under the 7d default. Lowering `lifecycle.reaper.delete_age` to 6h and purging the backlog took the wisp scan from a 33k-row table to ~2.8ms, `gt mail inbox` 698ms -> ~190ms, and stopped the pile-up (we'd captured 339 identical `SELECT ... FROM wisps` queries stacked up, oldest 54 min). Held stable for hours.

## Change

Lower the shipped default `defaultWispDeleteAge` 7d -> 24h, plus the matching `DefaultLifecycleConfig` string and help text. Deleted-wisp history is preserved via Dolt AS OF, so a short live-table retention is safe. High-churn towns can override lower via `gt config set lifecycle.reaper.delete_age`.

Complements #4261 (cut wisp query *count*; this cuts table *size*).

## Tests

`go test ./internal/daemon/` passes (lifecycle defaults test updated to expect 24h).

Draft pending maintainer view on whether this fits the #4292 direction. Refs #4292.

### DIFF
--- a/internal/cmd/config.go
+++ b/internal/cmd/config.go
@@ -638,7 +638,7 @@ Supported keys:
   Lifecycle (Dolt data maintenance):
   lifecycle.reaper.enabled     Enable/disable wisp reaper (true/false)
   lifecycle.reaper.interval    Reaper check interval (default: 30m)
-  lifecycle.reaper.delete_age  Delete closed wisps after this duration (default: 168h / 7d)
+  lifecycle.reaper.delete_age  Delete closed wisps after this duration (default: 24h)
   lifecycle.compactor.enabled  Enable/disable compactor dog (true/false)
   lifecycle.compactor.interval Compactor check interval (default: 24h)
   lifecycle.compactor.threshold Commit count before compaction (default: 500)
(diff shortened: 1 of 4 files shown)


rbriski:
Closing as a standalone code-default change. The retention tuning is real but it's a mitigation, not the structural fix — and towns can already set this via `gt config set lifecycle.reaper.delete_age` (we run 6h operationally). Redirecting effort to the root cause (#4292 connection amplification). The default value can be revisited separately if desired.

PR #4322, opened 2026-06-22
fix(refinery): close work bead on real merge success (stop re-dispatch churn)

Ben-Williams-Founder:
## Problem

When a polecat's MR/PR merges, its **work bead can stay OPEN**. The scheduler then re-dispatches a polecat onto already-merged work — the polecat redoes the work and wedges in `NEEDS_RECOVERY`. We observed this live as recurring re-dispatch churn.

The deterministic merge-success handler `HandleMRInfoSuccess` only closed the source issue inside `if mr.SourceIssue != ""`. When the merged MR carried no resolvable `source_issue`, the merge succeeded but **nothing closed the work bead** — it stayed open and got re-slung. There was no fallback to the worker agent bead's `active_mr → work-bead` mapping, which is exactly the link that survives when `source_issue` is absent.

## Fix

Extract the close into `closeMergedWorkBead(bd, out, mr, mergeCommit)`, called from `HandleMRInfoSuccess` on **real merge success only** (`HandleMRInfoFailure` never calls it, so rejects/conflicts still leave the work bead open for re-submit). The helper:

- Resolves the work bead via `source_issue` first, then falls back to the worker agent bead's `last_source_issue` (the `active_mr → bead` mapping).
- Is **idempotent**: an already-terminal bead is a no-op, never an error (covers the polecat-already-ran-`gt done` race and post-merge re-runs).
- Force-closes to bypass open-molecule-step dependency checks, matching how `gt done` closes work beads.

A small `workBeadCloser` interface is extracted so the merge → bead-closed behaviour is unit-testable with a fake (no live Dolt server required).

## Test

`internal/refinery/engineer_bead_close_test.go` covers: close via `source_issue`; fallback via the agent bead's `last_source_issue` when `source_issue` is blank (the churn case); idempotency on an already-terminal bead; and that a failure path (`HandleMRInfoFailure`) never closes the bead. `go build ./...` and `go test ./internal/refinery/` pass against `main`.

## Methodology note

This came out of treating a recurring operational symptom (re-dispatch churn / `NEEDS_RECOVERY` after merge) as a determinism gap: the merge-success path should leave exactly one well-defined post-state (work bead terminal), not depend on a field that is sometimes empty. The fallback closes that gap deterministically rather than relying on the polecat having already self-closed.

### DIFF
--- a/internal/refinery/engineer.go
+++ b/internal/refinery/engineer.go
@@ -1205,26 +1205,15 @@ func (e *Engineer) HandleMRInfoSuccess(mr *MRInfo, result ProcessResult) {
 		}
 	}
 
-	// 1. Close source issue with reference to MR.
-	// Use ForceCloseWithReason to bypass dependency checks — the source issue
-	// may have an attached molecule (wisp) whose open steps would block a
-	// normal close. This matches how gt done handles closures.
-	if mr.SourceIssue != "" {
-		closeReason := fmt.Sprintf("Merged in %s", mr.ID)
-		if result.MergeCommit != "" {
-			closeReason = fmt.Sprintf("%s\ntarget_branch: %s\ncommit_sha: %s", closeReason, mr.Target, result.MergeCommit)
-		}
-		if err := e.beads.ForceCloseWithReason(closeReason, mr.SourceIssue); err != nil {
-			// Check if already closed (by polecat's gt done) — that's fine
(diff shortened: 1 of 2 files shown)


Bella-Giraffety:
Cleanup-first attributed replacement published as #4489. Keep this original PR open until #4489 merges; then close it as superseded with replacement merge evidence.

Bella-Giraffety:
Superseded by merged cleanup-first replacement #4489. The replacement preserves Ben-Williams-Founder attribution in the squash commit.

PR #4323, opened 2026-06-22
fix(polecat): submit uncommitted work on session Stop, not just unpushed commits

Ben-Williams-Founder:
## Problem

Polecats were observed doing the work (writing implementation files) but the session turn ended before they ran `git commit && gt done` — stranding the work as **uncommitted changes in the worktree** → `NEEDS_RECOVERY` → work lost on reap.

There is already a Stop-hook safety net for this class of bug: the `polecats` role installs a Claude Code `Stop` hook running `gt tap polecat-stop-check`, and `gt done` itself auto-commits a dirty worktree before submitting. The `Stop` hook is the correct, well-gated trigger — per the Claude Code hooks spec, `Stop` fires **only** when the agent finishes its turn normally (never on context-limit, which is `PreCompact`, nor on crash/API error), so a `Stop` is a genuine completion signal, not a snapshot of an in-flight edit.

**The gap:** `polecat-stop-check` only invoked `gt done` when there were commits *ahead* of `origin/main` (`git rev-list --count origin/main..HEAD > 0`). In the exact stranding case the work is **uncommitted**, so the ahead-count is `0`, the hook concluded "nothing to submit", and exited — leaving the uncommitted work to rot.

## Fix

Extend the Stop-check decision to detect uncommitted work too. It now runs `gt done` when **either** there are commits ahead of `origin/main` **or** the working tree is dirty (`git status --porcelain` non-empty). `gt done`'s existing auto-commit safety net then commits and submits the work instead of stranding it. The decision is extracted into a pure `polecatStopPendingWork` helper that **fails closed** (no work) on any git error, so a transient git failure never blocks session stop.

## Test

`internal/cmd/tap_polecat_stop_test.go` (`TestPolecatStopPendingWork`) covers: clean feature branch → no work; uncommitted changes → work detected (the stranding bug); committed-but-unpushed → work detected; and a bogus dir → fails closed. `go build ./...` and the test pass against `main`.

## Methodology note

The fix deliberately reuses the existing, correctly-gated `Stop` trigger and the existing `gt done` auto-commit net rather than adding a new snapshot mechanism — it only widens the *detection* predicate to match the failure mode actually observed. The conservative "any dirt → hand to `gt done`" choice defers the authoritative include/exclude decision (runtime/overlay artifact filtering) to `gt done`, keeping one source of truth for what gets committed.

### DIFF
--- a/internal/cmd/tap_polecat_stop.go
+++ b/internal/cmd/tap_polecat_stop.go
@@ -97,20 +97,23 @@ func runTapPolecatStop(cmd *cobra.Command, args []string) error {
 		return nil // On default branch — nothing to submit
 	}
 
-	// Check for commits ahead of origin/main
-	aheadCmd := exec.Command("git", "-C", cloneDir, "rev-list", "--count", "origin/main..HEAD")
-	aheadOut, err := aheadCmd.Output()
-	if err != nil {
-		return nil // Can't check — exit quietly (don't block session stop)
-	}
-	ahead := strings.TrimSpace(string(aheadOut))
-	if ahead == "0" {
-		return nil // No commits ahead — nothing to submit
+	// Decide whether the polecat has work to submit. This covers BOTH:
+	//   - committed-but-unpushed work (commits ahead of origin/main), and
+	//   - uncommitted work in the working tree (files written but never committed).
+	// The second case is the stranding bug this fix targets: the agent wrote
+	// implementation files, finished its turn (firing Stop), but never ran
+	// `git commit && gt done` — so there are 0 commits ahead and the old check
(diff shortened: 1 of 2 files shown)


Bella-Giraffety:
Closing as superseded by merged replacement PR #4421. #4421 carries forward the dirty Stop submit safety fix on current main, passed full GitHub CI, and had PR Sheriff merge-gate PASS with merge_path_allowed=true for the replacement PR head. Attribution/provenance for Ben-Williams-Founder and original head b9e00b316d05c855f3d45742814a0db0fd4a6201 was preserved in #4421 body/evidence.

Bella-Giraffety:
Verified replacement merge commit for #4421: 997f75a46028c500b70afa11f7b69eccb5a7409a.

PR #4326, opened 2026-06-22
feat(up): machine-adaptive boot via explicit --role gate (fail-open)

Ben-Williams-Founder:
## Problem

`gt up` boots the same way regardless of host. On a constrained box, the right `scheduler.max_polecats` and governor thresholds differ from a big host, and operators had to tune them by hand after every boot. There was no first-class, **explicit** way to derive those settings from host facts at boot — and any auto-tuning must never silently change behaviour on hosts that don't opt in.

## Fix

Add an explicit **role gate** to `gt up` that (optionally) invokes a machine-adaptive boot step deriving `scheduler.max_polecats` + governor thresholds from detected host facts. The integration is **fully additive and fail-open** — it never changes the `services`/`allOK` accounting and never makes `gt up` fail:

- New `--role town-host|sandbox|dev` flag; also honors `GT_TOWN_HOST_ROLE` and `GT_ROLE` env (precedence: flag > `GT_TOWN_HOST_ROLE` > `GT_ROLE`).
- `town-host` **applies** the derived tuning; `--yes` is added automatically when `gt up` is headless/non-interactive (no TTY).
- `sandbox`/`dev` are explicit acknowledgements that **propose only**.
- **No explicit role** → runs in propose-only mode and tells the operator to re-run with `--role`. This preserves the historical `gt up` behaviour by default.
- A missing script, a non-town-host, a read-only devcontainer, or macOS never makes `gt up` fail.

Pure decision logic (`resolveAdaptiveRole`, `decideAdaptiveBoot`) is split out for unit testing. JSON mode suppresses the human-facing notes.

## Operator-provided script

The adaptive step shells out to `${GOV_TOOLS:-$HOME/gt/.gov-tools}/tools/ops/adaptive-boot.sh`. **That script is operator-provided and is intentionally NOT part of this PR / this repo** — different operators derive host tuning differently. `gt up` resolves the path at runtime and **fail-open skips** (with a note) when the script is absent, so this change is safe to merge with no script present: nothing depends on it to build or boot. The PR upstreams only the `gt up` integration (the `--role` flag + the fail-open invocation helper), giving operators a standard, explicit hook to wire their own host-tuning script into boot.

## Test

`internal/cmd/up_adaptive_test.go` covers role resolution precedence, the decide-to-apply/propose/skip logic, and the script-absent fail-open path. `go build ./...` and `go test ./internal/cmd/` (adaptive tests) pass against `main`.

## Methodology note

The design principle here is **explicit-acknowledgement over silent auto-detect**: host tuning is only ever *applied* when the operator names a `town-host` role, and is otherwise propose-only — so the change can never surprise an existing deployment. Everything is fail-open and additive, so adopting it is zero-risk even before an operator writes their tuning script.

### DIFF
--- a/internal/cmd/up.go
+++ b/internal/cmd/up.go
@@ -8,9 +8,9 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"runtime"
 	"strconv"
 	"strings"
-	"runtime"
 	"sync"
 	"syscall"
 	"time"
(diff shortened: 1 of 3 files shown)


Bella-Giraffety:
Closing per Mayor PR Sheriff feature triage. This is an unapproved enhancement with only initial triage coverage, unstable merge state, and a new machine-adaptive boot/fail-open role gate surface. Feature/enhancement work needs explicit human approval before implementation or merge; without that approval this should be re-proposed only as a smaller approved design.

PR #4342, opened 2026-06-25
fix(tmux): nudge falls back to session name when pane ID is stale (gy-avjr)

kaushikNaarayan:
## Problem
`gt nudge` recurringly fails with `send-keys: can't find window: %NNNN`, forcing a manual `tmux send-keys` workaround every session (gy-avjr — reproducible across sessions when nudging crew/coach/analyst, especially after a tmux server restart).

## Root cause
`NudgeSessionWithOpts` resolves a delivery target via `canonicalPaneTarget(session, FindAgentPane(session))`. `FindAgentPane` can return a pane ID (from the declared `GT_PANE_ID` or a scan), but `canonicalPaneTarget` **fell back to returning that raw pane ID** whenever it couldn't canonicalize it. A stale pane ID — e.g. a `GT_PANE_ID` declared before a tmux **server restart**, after which pane IDs reset — then flowed straight into `send-keys -t %NNNN` and failed every time. The cached ID is never refreshed, so the failure recurs until manually worked around.

Two failure shapes were possible:
1. `display-message -t <stalePane>` errors → raw stale ID returned → `can't find window`.
2. `display-message` "succeeds" with empty fields (`:.`) for a bogus pane → garbage target returned.

Plus a latent mis-targeting hole: tmux pane IDs are **global** and get reused after a server restart, so a stale ID could resolve into a *different* session and nudge the wrong agent.

## Fix
`canonicalPaneTarget` now returns a pane-based target **only** when `display-message` resolves it to a well-formed `session:window.pane` that **belongs to the expected session**; otherwise it falls back to the stable **session NAME**. tmux resolves the session name to its active pane, which always exists if the session does. Worst case (multi-pane session, agent pane not focused) we nudge the active pane instead of failing entirely.

## Test
Adds `TestCanonicalPaneTarget_StalePaneFallsBackToSession` covering empty-pane, stale-pane (must **not** return the dead ID), and live-pane (resolved target stays addressable) cases. Full `internal/tmux` suite + `go vet` pass.

### DIFF
--- a/internal/tmux/tmux.go
+++ b/internal/tmux/tmux.go
@@ -1666,12 +1666,26 @@ func (t *Tmux) canonicalPaneTarget(session, pane string) string {
 	out, err := t.run("display-message", "-t", pane, "-p", "#{session_name}:#{window_index}.#{pane_index}")
 	if err == nil {
 		target := strings.TrimSpace(out)
-		if target != "" {
+		// Accept the resolved target ONLY when it is well-formed AND belongs to
+		// the session we intend to nudge. A stale/bogus pane can make
+		// display-message succeed with empty fields (":.") instead of erroring,
+		// and because tmux pane IDs are global and get reused after a server
+		// restart, a stale ID can also resolve into a *different* session — both
+		// would mis-target send-keys.
+		if target != "" && strings.HasPrefix(target, session+":") {
 			return target
 		}
 	}
 
-	return pane
+	// The pane ID could not be resolved to a live pane in this session — it is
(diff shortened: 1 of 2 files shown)


Bella-Giraffety:
Closed as superseded by replacement PR #4433, merged in 5ba32ea7c55597c25a57ccbcfed3dba53ac6cb05. Attribution preserved for @kaushikNaarayan / Kanaba and the original Claude co-author in the merge commit trailers and replacement PR evidence. Thank you for the original stale-pane fallback fix.

PR #4343, opened 2026-06-25
fix(beads): pin server database in pinned bd env so mol bond targets the rig DB

marvincris:
## Problem

On a multi-rig town backed by a **shared Dolt server**, scheduler dispatch of a rig bead (e.g. `mm-*`) fails and the polecat boots with no work, then false-completes. The daemon log shows:

```
Scheduler dispatch: Dispatching mm-xxx → minime...
  "error": "'mm-xxx' not found (not an issue ID or formula name)"
✓ Dispatched 0, failed 1 (reason: batch)
```

The work bead clearly exists (`bd show mm-xxx` finds it), and gt's routing is correct — `beads.ResolveHookDir` resolves the bead to the rig dir (`<town>/minime`) before the bond. The failure is in the **bd subprocess environment**.

## Root cause

`BuildPinnedBDEnv` sets `BEADS_DIR=<rig>/.beads` and, via `doltTargetEnvFromBeadsDir`, also re-adds `BEADS_DOLT_DATA_DIR=<town>/.dolt-data` — but never pins the database **by name**.

Most bd subcommands honor `BEADS_DIR`, but **`bd mol bond` / `bd mol wisp` resolve the target database from `BEADS_DOLT_DATA_DIR`** (the town data dir) and ignore `BEADS_DIR`. So the wisp/bond for a rig bead runs against the town/HQ database, can't find the `mm-` bead, and dispatch fails.

Reproduced directly (shared server, `BEADS_DIR=<rig>/.beads` set):

| env | `bd mol bond mol-polecat-work mm-xxx …` |
|---|---|
| **with** `BEADS_DOLT_DATA_DIR=<town>/.dolt-data` | ❌ `'mm-xxx' not found` |
| **without** it | ✅ bond succeeds |
| with `BEADS_DOLT_SERVER_DATABASE=<rig>` + no `DATA_DIR` | ✅ reliably succeeds |

(Relying on `BEADS_DIR` alone without an explicit `BEADS_DOLT_SERVER_DATABASE` was unreliable in server mode.)

## Fix

Add `pinnedDoltTargetEnvFromBeadsDir`. In shared-server mode it selects the database **explicitly** via `BEADS_DOLT_SERVER_DATABASE` (from the beads dir's `metadata.json`) and **omits** `BEADS_DOLT_DATA_DIR`. Embedded (non-server) mode is unchanged. `BuildRoutingBDEnv` is untouched — routed env must not pin a database.

## Tests

- Updated `TestBuildPinnedBDEnvUsesSelectedConnectionMetadata`: the pinned env now carries the correct DB name (from metadata) instead of stripping `BEADS_DOLT_SERVER_DATABASE` — the inherited *wrong* value is still stripped.
- Added `TestBuildPinnedBDEnvPinsRigDatabaseInsideTown`: exercises a rig beads dir **inside a real town** (so `FindTownRoot` resolves and `BEADS_DOLT_DATA_DIR` would otherwise be re-added). This is the case the previous test missed — it used a town-less temp dir, so `FindTownRoot` returned `""` and the re-addition path was never covered, which is how the bug shipped.

`go test ./internal/beads/` passes (the one unrelated failure, `TestRunEnv_StripsPollutedDoltEnvAndUsesRigMetadata`, fails identically on a clean checkout in this environment due to hardcoded paths).

### DIFF
--- a/internal/beads/beads_test.go
+++ b/internal/beads/beads_test.go
@@ -114,8 +114,13 @@ func TestBuildPinnedBDEnvUsesSelectedConnectionMetadata(t *testing.T) {
 	if got["BEADS_DIR"] != beadsDir {
 		t.Fatalf("BEADS_DIR = %q, want %q in %v", got["BEADS_DIR"], beadsDir, env)
 	}
-	if value, ok := got["BEADS_DOLT_SERVER_DATABASE"]; ok {
-		t.Fatalf("BEADS_DOLT_SERVER_DATABASE should be stripped, got %q in %v", value, env)
+	// In shared-server mode the database is pinned EXPLICITLY by name from the
+	// selected beadsDir's metadata: the inherited (wrong) "hq" value is stripped
+	// and replaced with the metadata's "rigdb". BEADS_DIR alone does not reliably
+	// select the server database for all bd subcommands (e.g. `bd mol bond`), so
+	// the name must be pinned. See pinnedDoltTargetEnvFromBeadsDir / hq-qzrce.
+	if got["BEADS_DOLT_SERVER_DATABASE"] != "rigdb" {
+		t.Fatalf("BEADS_DOLT_SERVER_DATABASE = %q, want %q (pinned from metadata) in %v", got["BEADS_DOLT_SERVER_DATABASE"], "rigdb", env)
 	}
 	if got["BEADS_DOLT_SERVER_HOST"] != "127.0.0.1" {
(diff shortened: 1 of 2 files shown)


Bella-Giraffety:
Closed as superseded by replacement PR #4344. Sheriff review found #4343 was useful evidence but not a standalone merge path; #4344 preserves the relevant inside-town pinned bd env regression intent and credits Marvin Cris Andrade via Co-authored-by.

PR #4356, opened 2026-06-28
feat(opencode): turn-boundary drain so OpenCode self-drives autonomously

willpartcl:
## Problem

OpenCode towns don't self-sustain. The mayor cold-starts, runs one turn, goes idle, and never picks up queued work — so polecats never spawn and the town produces nothing. (Observed running a 24/7 OpenCode + GLM synthesis-QA town: witness/refinery up, mayor sitting at "waiting for instructions", `Polecats: 0` indefinitely.)

Root cause: the OpenCode plugin has no equivalent of Claude's `UserPromptSubmit` hook, which drains the nudge queue at **every turn boundary** via `gt mail check --inject`. The OpenCode preset doesn't set `HasTurnBoundaryDrain`, so gt falls back to the background **tmux nudge-poller** — which is unreliable for OpenCode because it has no detectable ready-prompt (`ReadyPromptPrefix` is empty → delay-only guessing) and no session resume.

## Fix

Give OpenCode a real turn-boundary drain, the same role it plays for Claude:

**`internal/hooks/templates/opencode/gastown.js`**
- On `event.type === "session.idle"` (the turn boundary), for self-driving roles (everything except `witness`/`refinery`/`deacon`/`boot`, matching Claude's `UserPromptSubmit` gate), run `gt mail check --inject`.
- If it returns queued work, push it back into the session as a new prompt via the OpenCode SDK: `client.session.prompt({ path: { id }, body: { parts: [{ type: "text", text }] } })`. **No tmux send-keys** — the plugin itself is the propulsion, so it doesn't depend on ready-prompt detection.
- Long-lived roles (`mayor`) `setTimeout`-poll while empty-idle (`GT_OC_POLL_MS`, default 20s) so newly-queued work wakes them; one-shot roles (polecats) only drain at the boundary and then let gt reap them.
- The plugin factory now destructures `client` from `PluginInput` (already provided by the plugin API).

**`internal/config/agents.go`**
- Set `HasTurnBoundaryDrain: true` on the OpenCode preset so gt no longer starts the background nudge-poller for OpenCode roles.

All existing behavior (SessionStart/Compaction prime injection, `system.transform`, cost recording) is preserved.

## Validation

- With the agent's real environment, `gt mail check --inject` correctly resolves and drains the per-session nudge queue (`/gt/.runtime/nudge_queue/<GT_SESSION>`) and returns the queued nudge.
- With `HasTurnBoundaryDrain: true`, gt starts **no** nudge-poller for the mayor/polecat OpenCode roles (confirmed via process list) — so any self-driving is attributable to the plugin.
- `go build ./cmd/gt` is clean.
- End-to-end self-drive cycle (`session.idle → drain → client.session.prompt → sling polecat`) is being validated in a live OpenCode + GLM town; happy to fold in results / iterate on review feedback.

Tested against OpenCode 1.17.11 (`@opencode-ai/sdk`, `SessionPromptData`).

### DIFF
--- a/internal/config/agents.go
+++ b/internal/config/agents.go
@@ -361,11 +361,12 @@ var builtinPresets = map[AgentPreset]*AgentPresetInfo{
 			"OPENCODE_CONFIG_CONTENT": `{"lsp":true}`,
 		},
 		ProcessNames:        []string{"opencode", "node", "bun"}, // Runs as Node.js or Bun
-		SessionIDEnv:        "",                                  // OpenCode manages sessions internally
-		ResumeFlag:          "",                                  // No resume support yet
-		ResumeStyle:         "",
+		SessionIDEnv:        "",                                  // OpenCode has no session-id env; --continue resumes most-recent in the role dir
+		ResumeFlag:          "--session",                         // `opencode --session <id>` continues a specific session
+		ContinueFlag:        "--continue",                        // `opencode --continue` auto-resumes the most recent session
+		ResumeStyle:         "flag",
 		SupportsHooks:       true, // Uses .opencode/plugins/gastown.js
-		SupportsForkSession: false,
+		SupportsForkSession: true, // `opencode --fork` (with --continue/--session)
 		NonInteractive: &NonInteractiveConfig{
 			Subcommand: "run",
 			OutputFlag: "--format json",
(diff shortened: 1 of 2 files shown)


Bella-Giraffety:
Closing stale feature PR in favor of issue #4455. Please continue product/design discussion there; feature work needs fresh human approval before any implementation or merge path.

PR #4367, opened 2026-07-01
fix: resolve rig-local agent tracking from cwd

enkemmc:
Recovery PR for Gas Town bead gt-ltw after gt done could not push to the upstream remote from gastown/rust.\n\nSummary:\n- Prefer cwd-derived rig beads directory when resolving rig-local agents.\n- Route await-event agent tracking through the same cwd-aware path.\n- Add regression coverage for the bad BEADS_DIR environment case.\n\nVerification from polecat notes:\n- CGO_ENABLED=0 go test ./internal/cmd\n- Patched binary resolves gastown refinery to /home/markenkema/gt/gastown/mayor/rig/.beads even with BEADS_DIR=/home/markenkema/gt/.beads.\n\nSource bead: gt-ltw\nCommit: d274dfa6

### DIFF
--- a/internal/cmd/agent_state.go
+++ b/internal/cmd/agent_state.go
@@ -6,7 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"os"
-	"os/exec"
+	"path/filepath"
 	"strconv"
 	"strings"
 	"time"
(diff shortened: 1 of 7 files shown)


Bella-Giraffety:
Verified replacement merge commit for #4427: 11e6075d9c2a2049b54d786672474230c66babab.

PR #4369, opened 2026-07-01
Add issue tracker backend abstraction

rrnewton:
## Summary

- add an issue-tracker backend abstraction for install-time storage setup
- support minibeads as a repo-local issue tracker backend without Dolt
- use mb directly for minibeads-backed towns instead of requiring a bd alias
- persist the selected backend in mayor/town.json and infer the minibeads edition from a sibling mb binary next to gt
- route internal cmd/beads issue-tracker subprocesses through backend-aware helpers
- skip Dolt-specific status/doctor/mayor attach checks for minibeads-backed towns

## Validation

- go test ./internal/deps ./internal/beads ./internal/cmd ./internal/doctor
- fresh minibeads-edition smoke with ~/gt_mb_bin first on PATH and no bd available:
  - gt install <town> --git succeeds
  - mayor/town.json records issue_tracker_backend: minibeads
  - gt config agent list succeeds
  - gt status --json omits Dolt
  - gt doctor --no-start reports zero failures
  - gt mayor attach outside a town reports only the workspace error, with no tracker warning

## Notes

This PR is intentionally scoped to Gastown. Minibeads compatibility improvements live separately in rrnewton/minibeads branch integration commit 842fdf7.

### DIFF
--- a/internal/beads/beads.go
+++ b/internal/beads/beads.go
@@ -16,6 +16,7 @@ import (
 	"time"
 
 	beadsdk "github.com/steveyegge/beads"
+	"github.com/steveyegge/gastown/internal/deps"
 	"github.com/steveyegge/gastown/internal/runtime"
 	"github.com/steveyegge/gastown/internal/telemetry"
 	"github.com/steveyegge/gastown/internal/util"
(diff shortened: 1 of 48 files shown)


Bella-Giraffety:
Closing this PR as stale and too broad for the direction we want to take. Beads already has first-class sync integrations for GitHub, GitLab, Jira, Linear, Azure DevOps, and Notion, plus multi-repo/federation support, so Gas Town probably should not grow a separate issue-tracker backend abstraction as proposed here. If we need deeper multi-tracker support, we should design it around Beads as the tracker hub and keep Gas Town integration thin. Please open or comment on a focused issue if there is a specific tracker workflow this does not cover.

PR #169, opened 2026-01-05
[DO NOT MERGE] fix: use ResolveBeadsDir for redirect support in more commands

michaellady:
## Summary
- `gt mq submit`, `gt release`, `gt rig reset`, and merge callbacks now use `beads.ResolveBeadsDir(cwd)` to properly follow `.beads/redirect` files in polecat worktrees
- Cherry-picked from stale branch `polecat/furiosa-mk18qjeg`

## Files changed
- `internal/cmd/callbacks.go` - handleMergeCompleted uses ResolveBeadsDir
- `internal/cmd/mq_submit.go` - submit uses ResolveBeadsDir
- `internal/cmd/release.go` - release uses ResolveBeadsDir
- `internal/cmd/rig.go` - rig reset uses ResolveBeadsDir

## Test plan
- Run commands from within a polecat worktree that has a `.beads/redirect` file
- Verify beads operations correctly resolve to the rig's beads directory

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4379, opened 2026-07-02
fix(exit): exclude runtime scaffolding from gt-exit auto-commit (sbx-gastown-8awxz)

epluntze:
## Summary

The `gt exit` auto-commit safety net did `git add -A` then only unstaged the CLAUDE overlay files (`exit.go:126-131`), so polecat runtime scaffolding — `.beads/`, `.runtime/`, `.claude/commands/`, and `node_modules` — got swept into child-repo commits. Observed in **sandboxx-store PR #169** (8 non-code files, including node_modules).

Root cause: `git add -A` stages **already-tracked** runtime files regardless of ignore rules, and `node_modules` was never in the exclude set at all — so the per-worktree `.git/info/exclude` alone could not prevent the leak. The reliable exclusion point is at commit time.

## Fix (defense-in-depth at commit time)
- **`git.UnstageRuntimeArtifacts()`** strips `.beads .runtime .claude .logs __pycache__ node_modules` from the index after `git add -A`; `exit.go` calls it before the auto-commit. Working tree is left untouched (`git reset -q HEAD --`), and unmatched pathspecs are a safe no-op.
- **`isGasTownRuntimePath`** now recognizes `node_modules`.
- **`node_modules/`** added to the polecat per-worktree local-exclude set (`gasTownLocalExcludePatterns`), deliberately kept out of child repos' tracked `.gitignore`.

## Test plan
- `TestUnstageRuntimeArtifacts` — stage a mix of real work + runtime artifacts, unstage runtime, assert only `src/main.go` remains staged and the working tree is intact; also asserts the empty-index call is safe.
- `node_modules` cases added to `TestIsGasTownRuntimePath`.
- `TestGasTownLocalExcludePatterns_IncludesNodeModules` — guards node_modules in the local exclude and its absence from the tracked .gitignore.
- `go build ./...` clean; `internal/git` + `internal/rig` targeted tests pass. (Two unrelated `TestGetPushURL*`/`TestClearPushURL` failures are pre-existing, caused by the local global-git `insteadOf` URL-rewrite config; untouched by this change.)

Bead: sbx-gastown-8awxz


epluntze:
Mistargeted at fork upstream; recreating against sandboxxapp/gastown-prime.

PR #4244, opened 2026-06-12
refactor(tmux): centralize agent busy-indicator detection (closes #4240)

andrewboldi:
## Problem

The busy-state detection that powers `IsIdle`, `WaitForIdle`, and the nudge Escape-suppression (#4242) scraped the agent TUI status bar for a **hardcoded `"esc to interrupt"` literal** buried inside `hasBusyIndicator`. That made a correctness-relevant signal an undocumented, single-point coupling to upstream agent UI text — if the wording changes, detection fails **open and silently**. (Filed as #4240.)

## Change

- Hoist the marker into a single documented source of truth: `var busyIndicators = []string{"esc to interrupt"}`.
- `hasBusyIndicator` now matches **any** registered marker, so supporting a new agent's wording (or absorbing an upstream rename) is a one-line edit in one place.
- `TestBusyIndicators` pins the known marker — a change to the upstream-coupled string now has to be deliberate and reviewed, not accidental.

The **idle** side was already centralized via the agent presets' `ReadyPromptPrefix`; this gives the **busy** side the same treatment.

## Scope (deliberately proportionate)

This is the low-risk, high-value slice of #4240. Two larger options from the issue are **intentionally deferred**, and called out so the residual risk is explicit rather than hidden:

- **Per-agent config** (a `BusyIndicator` preset field mirroring `ReadyPromptPrefix`): YAGNI today — Claude Code, Codex, and Gemini all render the same marker. The centralized list already supports adding variants; graduating to per-agent is a clean follow-up if an agent ever diverges.
- **Runtime canary** (detect a live upstream rename): would need to drive a known-generating agent and is inherently flaky. TUI-scraping can't fully eliminate silent-failure risk without a structural readiness signal, which no agent exposes uniformly today.

## Tests
- `TestHasBusyIndicator` (existing) — still green against the centralized list.
- `TestBusyIndicators` (new) — guards the source of truth.
- `go build` / `go vet` / `gofmt` clean.

Independent of #4242 — both touch the `hasBusyIndicator` neighborhood but not the same lines; either can merge first.

Closes #4240

PR #4049, opened 2026-05-17
fix: gt sling --dry-run can hang after printing spawn plan

esciara:
## Summary

- `gt sling --dry-run` was calling `isTrackedByConvoy(beadID)` before checking the dry-run flag, running `bd sql` queries against Dolt that could hang under load
- Fix restructures the convoy block to short-circuit on `slingDryRun` first, skipping all Dolt queries and emitting the "Would create convoy" plan text directly
- Adds `TestSlingDryRunSkipsConvoyLookup` regression test verifying `bd sql` is never invoked during `--dry-run`

## Changes

- `internal/cmd/sling.go`: Invert the dry-run/live conditional so `isTrackedByConvoy` is only called in live mode (never in `--dry-run`)
- `internal/cmd/sling_test.go`: Regression test for GH#3903 + pane-based self-sling detection tests (GH#3839)
- `internal/cmd/sling_target.go`: Detect self-sling when named target resolves to caller's own tmux pane (GH#3839)
- `internal/cmd/fanout.go` + `fanout_test.go`: New `gt fanout` command for throttled bulk bead creation (mayor orchestration)

## Testing

- [x] `go build ./...` passes
- [x] `TestSlingDryRunSkipsConvoyLookup` passes (verifies `bd sql` skipped in dry-run)
- [x] `TestSlingDryRun*` suite passes
- [x] `go vet ./...` passes
- [x] Pre-existing `TestSlingFormulaOnBeadRoutesBDCommandsToTargetRig` failure confirmed on `main` (macOS `/var` symlink issue, not introduced by this PR)

Closes #3903

Issue #3903, opened 2026-05-08
gt sling --dry-run can hang after printing spawn plan


Bella-Giraffety:
## Bug Description

During the May 8 reproduction sweep, `gt sling <bead> gastown --dry-run` sometimes hung long enough to exceed command timeouts. This undermines safe validation because dry-run should be the non-mutating preflight for route/dispatch checks.

## Steps to Reproduce
1. Use a valid gastown-owned bead, e.g. a `gt-gh####` task in the gastown database.
2. Run `gt sling <bead> gastown --dry-run`.
3. Observe whether the command exits quickly after printing the plan.

## Expected Behavior

`--dry-run` should perform no writes, start no sessions, and return quickly after printing the planned operations.

## Actual Behavior

Observed historical output:
```
Would spawn fresh polecat in rig 'gastown'
Slinging <bead> to gastown/polecats/<new>...
```
then the command hung long enough to hit a caller timeout.

A later fresh-eyes validation also reproduced a current dry-run hang with another gastown bead, although that reproduction timed out before capturing the same plan text.

## Environment
- **OS**: Linux 6.12.79-101.147.amzn2023.x86_64
- **Go version**: go1.25.8 linux/amd64
- **Gas Town version**: `v1.0.1-129-ga395fed84`
- **tmux version**: 3.2a
- **bd version**: `1.0.3 (1b2dd2cb5)`

## Logs / Error Output
```
gt sling gt-gh2405 gastown --dry-run
# printed dry-run spawn plan, then exceeded a 120s timeout
```

## Additional Context

Related: #3900 and #3899. Workaround: use `bd --db <rig/.beads> show <bead>` for route validation and avoid relying on `gt sling --dry-run` until fixed.

PR #4380, opened 2026-07-02
fix: pgrep -P on a dead PID matches every process, hanging session kills on macOS

psbanka:
⚠️ Disclaimer: This PR was written by AI. I scanned all contributing docs and didn't see any guidelines around this, so I'm submitting this PR, constructed by Fable 5. I am not a `go` developer, but can confirm that this PR solved a problem that I was having with gastown on my mac running MasOS 26.5.1 (25F80).

I would mention that this PR complies with the **Zero Framework Cognition (ZFC)** policy. This fix is squarely on the "transport" side (walking a process tree to kill sessions — pure plumbing, no judgment logic), so it's ZFC-aligned.

---

## Problem

On macOS, `pgrep -P <pid>` where `<pid>` no longer exists returns **every process on the system** (~1000 PIDs, exit 0) instead of nothing:

```
$ pgrep -P 99999 | wc -l
    1032
$ ps -ax | wc -l
    1030
```

`getAllDescendants` walked the process tree recursively with one `pgrep -P` call per node. Agent sessions constantly spawn short-lived children (ripgrep, git, node helpers), so whenever one exited between being listed by its parent's `pgrep` and being visited by its own, the dead PID's lookup sprayed back the entire process table and the walk recursed into every process on the machine — with each further PID that died mid-walk spraying all ~1000 again. Effectively unbounded.

## Symptoms

- `gt sling <bead> <rig>` hangs forever when **reusing** an idle polecat (`ReuseIdlePolecat` → `killExistingPolecatSession` → `KillSessionWithProcesses`), while fresh spawns work — reuse is the path that kills an existing session. Hangs reproduce right after "Auto-detected integration branch:".
- `gt down` hangs (same kill path for each session).
- Intermittent by nature: only fires when a transient child dies inside the race window, so busy agent sessions hit it constantly while quiet ones don't.

A SIGQUIT goroutine dump of a hung `gt sling` shows the main goroutine 1050+ frames deep in `getAllDescendants`, still forking `pgrep`:

```
goroutine 1 [syscall]:
...
github.com/steveyegge/gastown/internal/tmux.getAllDescendants(...)
	internal/tmux/tmux.go:858 +0xe8
github.com/steveyegge/gastown/internal/tmux.getAllDescendants(...)
	internal/tmux/tmux.go:858 +0xe8
...1033 frames elided...
github.com/steveyegge/gastown/internal/tmux.(*Tmux).KillSessionWithProcesses(...)
	internal/tmux/tmux.go:655
github.com/steveyegge/gastown/internal/polecat.(*Manager).killExistingPolecatSession(...)
github.com/steveyegge/gastown/internal/polecat.(*Manager).ReuseIdlePolecat(...)
github.com/steveyegge/gastown/internal/cmd.SpawnPolecatForSling(...)
```

One more note for severity: had the walk ever completed, the descendant list would have contained **every user process on the machine**, and `KillSessionWithProcesses` would have SIGTERM/SIGKILLed all of them. The hang was the safe failure mode.

## Fix

Take a single `ps -axo pid=,ppid=` snapshot and walk the parent/child map in memory:

- `getAllDescendants`: same deepest-first ordering, one subprocess instead of one per node, no listed-vs-visited race window, cycle-proof via a seen-set.
- `hasDescendantWithNamesPosix` had the same flaw (its depth cap bounded recursion depth but not the full-table fan-out, and a dead PID could make zombie detection match any `node`/`claude` anywhere on the system). Now walks a `ps -axo pid=,ppid=,comm=` snapshot.

## Testing

- `go test -run 'Descendant' ./internal/tmux/` passes (`TestGetAllDescendants`, `TestHasDescendantWithNames`, including the nonexistent-PID case — which is exactly the bug scenario).
- End-to-end: a `gt sling` onto an idle polecat that reliably hung forever before the fix completes in ~46s after it (reuse → hook → session start).
- Other failures in `./internal/tmux/` (key bindings, startup dialogs, nudge) are environment-dependent and fail identically on the unmodified file.

## Adjacent work (not overlapping)

I searched all PRs (open + merged/closed) and issues for this root cause before opening; nothing else touches it. `getAllDescendants` was last modified in #882 (2026-01-25) for an unrelated handoff self-kill fix. Listing the near-neighbors so reviewers can see this is distinct:

- **#4049 / issue #3903** — `gt sling --dry-run` hang. A *different* hang: it stalls in the dry-run print path, not the session-kill path this PR fixes.
- **#4342** — `fix(tmux): nudge falls back to session name when stale pane ID`. Same file, unrelated function (nudge, not process-tree walking).
- **#4150** — `fix(tmux): guard NewSession against socket-clobber`. Same file, session-creation path.
- **#4244** — `refactor(tmux): centralize agent busy-indicator detection`. Same file, unrelated.
- **#3860** (closed) — `fix(mail): kill bd subprocess group on context timeout`. Different subsystem (mail/bd), but related in spirit — another place subprocess handling caused a hang.

No open issue tracks the symptom (polecat-reuse / `gt down` hangs on macOS); this PR is both the report and the fix.

### DIFF
--- a/internal/tmux/tmux.go
+++ b/internal/tmux/tmux.go
@@ -841,25 +841,44 @@ func collectReparentedGroupMembers(pgid string, knownPIDs map[string]bool) []str
 	return reparented
 }
 
-// getAllDescendants recursively finds all descendant PIDs of a process.
+// getAllDescendants finds all descendant PIDs of a process.
 // Returns PIDs in deepest-first order so killing them doesn't orphan grandchildren.
+//
+// Implemented as a single ps snapshot walked in memory, NOT recursive pgrep -P
+// calls: on macOS, pgrep -P <pid> returns EVERY process on the system when
+// <pid> no longer exists (exit 0), so a transient child exiting between the
+// parent listing and its own lookup made the old recursive walk explode into
+// an unbounded traversal of the whole process table (gt sling/down hangs) —
+// and the resulting kill list would have included every user process.
 func getAllDescendants(pid string) []string {
-	var result []string
-
-	// Get direct children using pgrep
-	out, err := exec.Command("pgrep", "-P", pid).Output()
+	out, err := exec.Command("ps", "-axo", "pid=,ppid=").Output()
 	if err != nil {
-		return result
+		return nil
 	}
 
-	children := strings.Fields(strings.TrimSpace(string(out)))
-	for _, child := range children {
-		// First add grandchildren (recursively) - deepest first
-		result = append(result, getAllDescendants(child)...)
-		// Then add this child
-		result = append(result, child)
+	childrenOf := make(map[string][]string)
+	for _, line := range strings.Split(string(out), "\n") {
+		f := strings.Fields(line)
+		if len(f) != 2 {
+			continue
+		}
+		childrenOf[f[1]] = append(childrenOf[f[1]], f[0])
 	}
 
+	var result []string
+	seen := map[string]bool{pid: true}
+	var walk func(string)
+	walk = func(p string) {
+		for _, child := range childrenOf[p] {
+			if seen[child] {
+				continue
+			}
+			seen[child] = true
+			walk(child) // grandchildren first - deepest first
+			result = append(result, child)
+		}
+	}
+	walk(pid)
 	return result
 }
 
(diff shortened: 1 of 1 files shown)


Bella-Giraffety:
Closing as superseded by replacement PR #4555, which carried the process traversal fix forward to current main with attribution preserved and merged successfully.

PR #4381, opened 2026-07-03
fix(refinery): clear agent active_mr on every MR close path

hux:
## Summary

`closeMRWithReason()` closed the MR bead but never cleared the originating
agent bead's `active_mr` reference. Only the separate `HandleMRInfoSuccess`
happy-path merge did that clear, via its own inline close logic.

Two other paths route through `closeMRWithReason` and left `active_mr` stale
forever:

- `closeIneligibleMR` (rejected/policy-ineligible MRs — e.g. missing
  `source_issue`, owned-direct, stale branch)
- `recheckMRStillMergeable`'s already-merged dedup close

Once `active_mr` pointed at a closed or later-purged MR bead,
`polecat.AssessActiveMR`'s fail-closed classification permanently blocked
`gt polecat check-recovery` from marking the polecat safe-to-reuse — even
though the underlying work was already merged or correctly rejected. This
matches the reported symptoms exactly: `fib/nitro` reporting `active_mr` for
an MR that `bd show` confirms is closed/merged, and `hci/rust` reporting
`active_mr` for an MR bead that no longer exists.

## Fix

Move the `active_mr` clear into `closeMRWithReason` itself, so every current
and future caller gets it — not just the merge-success path. Guarded by
`mr.AgentBead != ""`, and a clear failure is a non-fatal logged warning,
consistent with the existing pattern.

`polecat.AssessActiveMR`'s fail-closed read-side policy is intentionally left
untouched — its own doc comment states the conservatism is deliberate. The
actual bug was write-side (the field was never cleared), so the write-side
fix is the surgical, root-cause fix.

## Test plan

- [x] `go build ./internal/refinery/... ./internal/polecat/... ./internal/beads/...`
- [x] `go vet` on the same packages — clean
- [x] `go test ./internal/refinery/...` — all pass, including new
      `TestCloseMRWithReasonClearsAgentActiveMR` regression test (mocks the
      `bd` subprocess to assert the agent bead is looked up and updated when
      an MR is closed via `closeMRWithReason`)
- [x] `go test ./internal/polecat/...` — all pass (unchanged, confirms
      fail-closed policy untouched)

### DIFF
--- a/internal/refinery/engineer.go
+++ b/internal/refinery/engineer.go
@@ -1629,6 +1629,19 @@ func (e *Engineer) closeMRWithReason(mr *MRInfo, closeReason string) error {
 		return fmt.Errorf("close MR: %w", err)
 	}
 	_, _ = fmt.Fprintf(e.output, "[Engineer] Closed MR bead: %s (%s)\n", mr.ID, closeReason)
+
+	// Clear the agent bead's active_mr reference now that this MR has reached
+	// a terminal state. Without this, check-recovery's fail-closed
+	// classification (polecat.AssessActiveMR) can report a permanent false
+	// active_mr blocker for an MR that was closed here — e.g. rejected as
+	// policy-ineligible via closeIneligibleMR, or closed as already-merged by
+	// recheckMRStillMergeable — since those paths previously never touched
+	// active_mr (only the separate HandleMRInfoSuccess merge path did).
+	if mr.AgentBead != "" {
+		if err := e.clearAgentActiveMR(mr.AgentBead); err != nil {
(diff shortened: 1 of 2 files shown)


Bella-Giraffety:
Closing as superseded by converged replacement PR #4403, merged to main at 81e2fc8faa0b1e79dda1f7761d985601225b25f2. #4403 carries forward the active_mr cleanup intent on current main, covers both Engineer and Manager terminal MR close paths, makes cleanup retryable/idempotent for already-terminal MRs, preserves newer active_mr safety, and passed the PR Sheriff merge gate.

PR #4430, opened 2026-07-08
feat(polecat): configurable branch delimiter for CI-constrained charsets

blairsilverberg:
## What

Adds a `polecat_branch_delimiter` rig config option that selects the delimiter placed between the issue ID and the generated timestamp suffix in polecat branch names (`polecat/{name}/{issue}<delimiter>{timestamp}`). Default remains `+` — behavior is unchanged unless a rig opts in.

## Why

Downstream CI infrastructure we don't control derives a docker-compose project name from the branch name, and docker-compose project names only allow `[a-z0-9_-]`. Both the legacy `@` and the current default `+` delimiter are rejected, which breaks every CI run for polecat branches on that repo. `_` is charset-safe **and** unambiguous — issue IDs are `[a-z0-9-]` (plus `.` for subtasks) and never contain `_`, so the parser can still split the issue from the timestamp cleanly.

Rather than hardcoding `_`, this makes the delimiter a general config option so any rig facing a branch-charset constraint can pick what its CI accepts.

## Changes

- **`internal/polecat/branch_name.go`**
  - New `BranchDelimiterConfigKey` (`polecat_branch_delimiter`) and `ValidBranchDelimiter` (valid: `+`, `_`, `@`).
  - `FormatGeneratedBranchNameWithDelimiter` — formats with a chosen delimiter; invalid values fall back to `+` so a bad config can never produce an unparseable branch. `FormatGeneratedBranchName` delegates with the default.
  - `parseIssueTail` now recognizes `_` alongside `+` and `@`. Parsing accepts **all** valid delimiters regardless of configuration, so branches created under a previous delimiter setting keep resolving to the right issue after a config change.
- **`internal/polecat/manager.go`** — `buildBranchName` body extracted to package-level `buildBranchNameFor(rig, git, beads, name, issue)` (nil-tolerant), and the default (non-template) path now honors the configured delimiter.
- **`internal/polecat/session_manager.go`** — `freshBranchName` now shares `buildBranchNameFor`, so session-restart branches honor `polecat_branch_template` and `polecat_branch_delimiter` instead of always emitting the default separator. Previously a restart leaked the default form even on rigs with a custom template.
- **`internal/rig/config.go`** — `polecat_branch_delimiter` added to `SystemDefaults` (shows up in `gt rig config show` automatically).
- **`docs/reference.md`** — documents the new key under Polecat Branch Naming.

## Tests

- Delimiter format/parse round-trip (`polecat/nux/cap-5gw_mrb05j24` ⇄ issue `cap-5gw`), including `git check-ref-format` validation.
- `ValidBranchDelimiter` accept/reject table; invalid/empty/multi-char delimiters fall back to `+`.
- `buildBranchName` with configured delimiter (underscore, unset, invalid, no-issue cases), each asserting parse round-trip.
- `freshBranchName` honors both the configured delimiter and a branch template.
- Existing `freshBranchName` tests adapted for the new `(g *git.Git, ...)` parameter (pass `nil`); assertions unchanged.

`go test ./...` and `go vet ./...` clean.

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -195,6 +195,24 @@ When `polecat_branch_template` is empty or not set:
 - With issue: `polecat/{name}/{issue}+{timestamp}`
 - Without issue: `polecat/{name}-{timestamp}`
 
+**Issue/timestamp delimiter (`polecat_branch_delimiter`):**
+
+Some CI systems constrain the branch-name charset — e.g. docker-compose
+project names derived from branch names allow only `[a-z0-9_-]`, rejecting
+the default `+`. Rigs whose CI needs this can pick a different delimiter for
+the default (non-template) branch format:
+
+```bash
+gt rig config set myrig polecat_branch_delimiter _
+# → polecat/{name}/{issue}_{timestamp}
+```
+
+Valid delimiters are `+` (default), `_`, and `@` (legacy) — characters that
+can never appear in issue IDs or polecat names, so branches always parse
+back to the right issue. Invalid values fall back to `+`. Parsing accepts
+all valid delimiters regardless of configuration, so branches created under
+a previous delimiter keep working after a config change.
+
 **Example Configurations:**
 
 ```bash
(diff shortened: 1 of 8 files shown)


Bella-Giraffety:
Closing this feature PR as stale for the current infrastructure shape. The request is preserved in issue #4482. Does #4482 still match what you wanted? If so, we can rebuild a current cleanup-first implementation after the remaining bug-fix PR backlog is resolved.

PR #4431, opened 2026-07-08
feat: hard CI gate — block polecat done/reap and refinery merge until PR CI is green

blairsilverberg:
## What

Adds a hard CI gate to the polecat lifecycle: a polecat cannot complete (`gt done`) and the refinery will not merge its MR while the PR's CI checks are red or pending. Prevents half-done work from reaching the merge queue and master from going red via the autonomous pipeline.

## Behavior

- Gate is ON by default, configurable per rig
- Polecat `done` flow polls CI status with a configurable timeout (default 30m) before submitting to the merge queue
- On timeout/undeterminable CI: fails open with an explanatory comment and routes the item to human attention rather than silently merging or silently dropping the work
- Refinery independently re-verifies CI green before merge (belt and suspenders)

## Why

Downstream deployments of gastown hit repeated incidents where polecats declared done with red CI, and merges landed while required checks were failing (merge gates configured as advisory). A pipeline-level hard gate makes green CI a precondition of the autonomous path regardless of repo-side branch-protection configuration.

## Testing

Built and unit-tested on this branch (commit 4f344d95); affected packages pass locally. Draft while we finish integration verification downstream.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_0193nwJhFM7AFAzwkWXwKh6n

PR #4436, opened 2026-07-08
fix: fail-safe polecat reaper/allocator on Dolt-lookup failure + honor heartbeat threshold config

blairsilverberg:
## Summary

Two production incidents on 2026-07-08 (05:46Z and 14:45Z): the daemon/allocator killed **healthy, mid-work polecat sessions** while Dolt lookups were degraded. Forensics traced this to several places where a *failed or unverifiable* bead lookup is treated as a *definitive "no work"* answer, and the session is then killed. This PR makes every such path **fail-safe**: a tmux session whose work-state cannot be verified — especially one with a live agent process — is never killed; the reap/reclaim is deferred until beads answer again.

### Fixes

**1. Idle reaper fail-deadly on lookup failure (`internal/daemon/daemon.go`)**
- `hasAssignedOpenWork` swallowed every `bd list` error (`continue` → `return false`), so a Dolt outage read as "no assigned work". It now returns `(bool, error)`, where `(false, nil)` is only a *verified* negative.
- `reapIdlePolecat`: when both the agent-bead lookup and the work-bead verification fail, the work-state is UNKNOWN → never kill, log and defer. When "no work" is verified but the agent process is still alive, never kill either (previously the `staleness >= 3×threshold` arm killed regardless of agent liveness). Verified-no-work + agent-dead still reaps at 2× threshold.

**2. Allocation-path kills of live sessions (`internal/polecat/manager.go`)**
- `reconcilePoolInternal` derived "names with directories" from `m.List()`, which **silently drops any polecat whose bead load fails** — under Dolt degradation a live polecat vanished from the list and its session was killed as a directory-less "orphan". It now reads the directory names straight from the filesystem (the ZFC source of truth; no bead lookup involved).
- `ReconcilePoolWith`: an orphan (no-dir) session with a **live agent process** is no longer killed — it is protected and its name kept in-use so the allocator can't hand it out. Dead orphans are still reaped.
- `AllocateName` / `AllocateAndAdd`: allocation now skips any name whose session hosts a live agent (marks it in-use and picks a different name) instead of blind-killing the "lingering" session.
- `ReuseIdlePolecat`: refuses destructive reuse (`kill … "reuse"`) when the target session has a live agent process — `Issue == ""` can mean the bead lookup failed, not that the polecat finished.
- `loadFromBeads`: a running session is never classified `idle` (= destructively reusable) when the hooked-work query errored.
- `isSessionProcessDead`: a stale heartbeat alone is no longer a death verdict — heartbeats are only touched by `gt` commands, so agents deep in a long turn go stale while alive. A stale heartbeat is now confirmed with a direct agent-process probe before declaring death.

**3. Heartbeat threshold config never read (`internal/polecat/heartbeat.go`)**
- `IsSessionHeartbeatStale` used the compiled `SessionHeartbeatStaleThreshold = 3m` even though the comment said "configurable via `operational.polecat.heartbeat_stale_threshold`" and the config accessor (`HeartbeatStaleThresholdD`) already existed. Added `SessionHeartbeatStaleThresholdFor(townRoot)` and wired it through.
- Same audit for the other "Configurable via" bare constant in `internal/polecat`: `pendingMaxAge` now honors `operational.polecat.pending_max_age` via `PendingMaxAgeD()`.

**4. Rig-status misdiagnosis: "not found" ≠ "Dolt unavailable" (`internal/daemon/daemon.go`)**
- Investigation of the incident-day logs showed the daemon's rig-bead lookups did **not** time out — the rig beads genuinely don't exist on that town, but `isRigOperational` conflated `beads.ErrNotFound` with a transport failure, logging ~18k "cannot verify rig status (Dolt unavailable)" lines/day and excluding those rigs from witness/refinery patrol indefinitely. Not-found is now treated as the definitive answer it is (no global docked/parked state → proceed); genuine transport errors keep the existing fail-safe (assume not operational).

### Tests

New:
- `TestReapIdlePolecat_SkipsWhenAllBeadLookupsFail` — lookup error ⇒ no kill verdict, even at 10× staleness
- `TestReapIdlePolecat_SkipsAliveAgentWhenBeadLookupFails` — live agent ⇒ no kill even when "no work" is verified
- `TestAllocateNameSkipsLiveAgentSession` — allocator skips names with live-agent sessions, doesn't kill them
- `TestReconcilePoolWithProtectsLiveAgentOrphan` / `TestReconcilePoolWithStillKillsDeadOrphan` — protection without over-protection
- `TestIsSessionHeartbeatStale_HonorsConfiguredThreshold` / `TestSessionHeartbeatStaleThresholdFor_Defaults` — threshold honors config
- `TestIsRigBeadNotFound` — not-found vs transport-error classification

Adapted (signature-only, assertions unchanged): `TestHasAssignedOpenWork_UsesPinnedBeadsDirInsteadOfRigOrRepoFlag` (new `(bool, error)` return), `TestCleanupOrphanPolecatState` (constant `pendingMaxAge` → method `m.pendingMaxAge()`).

All existing reaper tests (`TestReapIdlePolecat_*`) pass unmodified.

### Notes

Stacked on #4430 (configurable branch delimiter) and #4431 (hard CI gate) — the first three commits are those PRs; review the last commit here.

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -84,7 +84,14 @@ town-level Gas Town beads.
     "integration_branch_polecat_enabled": true,
     "integration_branch_refinery_enabled": true,
     "integration_branch_template": "integration/{title}",
-    "integration_branch_auto_land": false
+    "integration_branch_auto_land": false,
+    "ci_gate": {
+      "enabled": true,
+      "pending_timeout": "30m",
+      "poll_interval": "30s",
+      "mayor_alert_after": "30m",
+      "escalation_cmd": ""
+    }
   }
 }
 ```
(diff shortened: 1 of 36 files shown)


Bella-Giraffety:
Closing this stale draft PR during manual cleanup after roughly two calendar weeks without author activity. If this work is still intended, it can be reopened or resubmitted when ready. If maintainers carry any part of it forward in a replacement/fixup, contributor credit will be preserved.

PR #4461, opened 2026-07-10
fix(nudge): verify submission by composer state + C-j recovery for swallowed Enter

blairsilverberg:
## Problem

Nudge delivery (`gt nudge`, nudge-poller injection, daemon health-check pokes) types the payload into the target Claude Code pane, but the trailing Enter is sometimes swallowed when the composer buffer is in a bad state — the text sits unsubmitted in the composer until something resets the input buffer. We hit this **five times in one day** on our deacon (incident evidence: openclaw beads `op-03ke`, `hq-zrpgf`, `hq-ygybv`), each stranding costing a 27–44 min stale-heartbeat gap and an escalation.

Live debugging of one occurrence showed:
- printable keys still echo → the Ink event loop is alive
- Enter / C-m / CSI-u `13u` / Escape+Enter are **all** swallowed
- **C-j (LF) clears the parked buffer** without submitting
- after that clear, retyping the text + plain Enter submits normally

Consistent with a lost/duplicated bracketed-paste boundary wedging the composer so CR is dropped until the buffer is reset.

The existing `sendEnterVerified` guard cannot catch this: its "did the pane change" heuristic passes while the pane repaints (spinner/clock) around a wedged composer, and its remedy — more Enters — is exactly the keystroke being swallowed.

## Fix

At the delivery choke point (`NudgeSessionWithOpts` / `NudgePane`), layered on the existing Enter-retry guard:

1. **`submitComposer`** (new, `internal/tmux/submit_verify.go`): verify the message actually *left the composer* — `capture-pane -e`, locate the live composer line (bottom-most `❯` prompt line), and classify:
   - busy indicator visible → turn started ✓
   - typed text absent, or present but **dim (SGR 2)** → cleared ✓ (dim is Claude Code's ghost/draft placeholder rendered over an *empty* composer; treating it as stranded input caused a documented false-positive rescue)
   - typed text present at normal intensity → **stranded**: Enter was swallowed
2. **C-j recovery**: send C-j to reset the wedged buffer; if it cleared the text, retype + Enter; re-verify. Still stranded → sentinel `ErrSubmitNotVerified`.
3. **Queue fallback** so unverified submits are never lost:
   - `--mode=wait-idle` direct delivery falls back to the nudge queue (+ idle watcher) on `ErrSubmitNotVerified`.
   - The idle watcher and `gt nudge-poller` re-enqueue drained nudges when injection fails (previously a failed injection silently discarded them).

Agents without prompt-based detection degrade gracefully: no composer line found → `probeUnknown` → the legacy `sendEnterVerified` verdict stands.

## Tests

- Table-driven unit tests for the ANSI dim tracker (incl. `38;5;2` / truecolor args not reading as SGR-2 dim), needle extraction, and the composer-state classifier (stranded / ghost / cleared / wrapped input / bottom-most-prompt selection / busy).
- `TestProbeSubmission_LivePane`: renders stranded, dim-ghost, and busy composers in a real tmux pane and asserts the probe classification end-to-end (skips when tmux is unavailable).

Related prior work: #1216 (nudge delivery reliability), GH#gt-0b5 (`sendEnterVerified`), GH#gt-8el (Rewind recovery) — this addresses the remaining swallowed-CR mode those don't cover.

### DIFF
--- a/internal/cmd/nudge.go
+++ b/internal/cmd/nudge.go
@@ -245,7 +245,25 @@ func deliverNudge(t *tmux.Tmux, sessionName, message, sender string) error {
 				Message:  message,
 				Priority: nudgePriorityFlag,
 			}})
-			return t.NudgeSessionWithOpts(sessionName, formatted, tmux.NudgeOpts{TownRoot: townRoot})
+			deliverErr := t.NudgeSessionWithOpts(sessionName, formatted, tmux.NudgeOpts{TownRoot: townRoot})
+			if !errors.Is(deliverErr, tmux.ErrSubmitNotVerified) {
+				return deliverErr
+			}
+			// The payload was typed but submission could not be verified —
+			// the composer swallowed the Enter (op-03ke). Fall back to the
+			// queue so the message is not lost, and watch for a delivery
+			// window (the C-j remedy has already reset the composer, so a
+			// later attempt has a fresh buffer).
+			fmt.Fprintf(os.Stderr, "wait-idle: %v; queueing for %s\n", deliverErr, sessionName)
+			if qErr := nudge.Enqueue(townRoot, sessionName, nudge.QueuedNudge{
+				Sender:   sender,
+				Message:  message,
+				Priority: nudgePriorityFlag,
+			}); qErr != nil {
+				return fmt.Errorf("queue fallback after unverified submit failed: %v (original: %w)", qErr, deliverErr)
+			}
+			watchAndDeliver(t, townRoot, sessionName)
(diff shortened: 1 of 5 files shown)


Bella-Giraffety:
Closed as superseded by replacement PR #4481, merged as f97ed211bdcf7d8ad686f4d5f62233961aaa71a3. The replacement preserved the accepted nudge submit-verification bug intent with PR Sheriff checker PASS, green CI, and co-author attribution for blairsilverberg.

PR #4463, opened 2026-07-10
fix(patrol): stamp deacon heartbeat on patrol report + inline full step bodies into root-only patrol wisps

blairsilverberg:
## Summary

Fixes three linked defects observed in production (a Deacon's `heartbeat.json` went stale for 80 minutes across 11 healthy patrol cycles, and the daemon's staleness watchdog nearly killed a healthy Deacon):

1. **`gt patrol report` now stamps the deacon heartbeat on success.** A successful report is proof-of-life by construction. After the patrol root closes, the deacon role stamps `deacon/heartbeat.json` + `.deacon-heartbeat` + the agent-bead label via the same store-sync path as `gt deacon heartbeat` (`syncDeaconHeartbeatStores`). Skipped when paused (matching `runDeaconHeartbeat` semantics); best-effort, never fails the report.

2. **`gt patrol new` root-only wisps are now self-describing.** The wisp cook is root-only by design ("steps are read inline at prime time"), but the per-step instructions — including the mandatory `gt deacon heartbeat` command whose omission gets the Deacon killed by the daemon — never reached the executing session: the wisp description carried only the formula root text, and `gt mol progress` errors on child-less roots. `autoSpawnPatrol` now inlines the formula root description + full step bodies (vars substituted, overlays applied) into the root wisp description, in the same `bd update` call that hooks the wisp (no extra Dolt commit). Applies to all patrol roles (deacon/witness/refinery).

3. **Deacon prime renders full step bodies.** `outputDeaconPatrolContext` used `showFormulaSteps` (120-char truncated one-liners), which dropped the heartbeat command from the step body. Now uses `showFormulaStepsFull`, matching the refinery path.

## Tests

- `TestStampDeaconHeartbeatOnReport_StampsAllStores` / `_SkipsWhenPaused` — file stores written with the report action, legacy file touched, bead sync invoked; paused Deacon skips entirely (bead-label sync stubbed via a package var: it shells out to bd).
- `TestRenderPatrolWispDescription_*` — root text preserved, step bodies + heartbeat command inlined, `{{vars}}` substituted, ExtraVars override defaults, unknown formula falls back to empty (keeps bd's default description).
- `TestPatrolRigName` — rig extraction from assignee.

`go build ./...`, `go vet`, and targeted tests green.

### DIFF
--- a/internal/cmd/heartbeat.go
+++ b/internal/cmd/heartbeat.go
@@ -93,9 +93,13 @@ func syncDeaconHeartbeatStores(townRoot, action string) error {
 	} else {
 		err = deacon.Touch(townRoot)
 	}
-	syncDeaconAgentBeadHeartbeat(townRoot)
+	deaconAgentBeadHeartbeatSync(townRoot)
 	return err
 }
+
+// deaconAgentBeadHeartbeatSync is swappable in tests: the agent-bead label
+// refresh shells out to bd (Dolt), which unit tests must not touch.
+var deaconAgentBeadHeartbeatSync = syncDeaconAgentBeadHeartbeat
 
 // syncDeaconAgentBeadHeartbeat refreshes the heartbeat:EPOCH label on the
 // Deacon's agent bead — the third heartbeat store, read by Witness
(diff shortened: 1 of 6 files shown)


Bella-Giraffety:
Superseded by merged replacement PR #4496 (merge commit 2bd80946ddf1f3314d880c68800f44205a0a5f5e). The replacement preserves the valid intent and contributor attribution while rebuilding from current main without stale WIP history. Thank you for the original contribution.

PR #4502, opened 2026-07-14
fix(daemon): execute plugin run.sh directly — exit code gates the AI agent step

blairsilverberg:
## Problem

When a plugin ships a `run.sh`, the daemon's dog dispatch path never executes it — it only feeds `plugin.md` to an AI dog as a prompt (with `FormatMailBody` asking the dog to run the script). In practice dogs skim past that instruction: a correct code guard added to `plugins/stuck-agent-dog/run.sh` protected nothing for two days, and six false RESTART_POLECAT notifications fired on deliberately-stopped polecats across two prompt-level fix attempts. Instruction placement must match the execution layer: code guards need a code execution path.

## Fix

**The daemon executes `run.sh` itself; the exit code gates the AI agent step.**

- `internal/plugin/runner.go` (new): `RunScript()` executes `plugins/<name>/run.sh` — cwd = plugin dir, `GT_TOWN_ROOT` set, `execution.timeout` enforced (default 5m, process-group kill so children can't hold the output pipe past the deadline).
- Exit-code contract (documented in the plugins README template, `docs/design/plugin-system.md`, and the stuck-agent-dog EXECUTION NOTE):
  - `0` — script completed the plugin run; no agent step
  - `10` — pre-checks passed; dispatch an AI dog with the plugin.md instructions + script output tail
  - anything else / timeout — failure, recorded `result:failure`; agent step NOT dispatched (fail-safe)
- `internal/daemon/handler.go`: `dispatchPlugins` routes `HasRunScript` plugins through `startScriptPlugin` — async so the patrol loop never blocks behind a script; a per-plugin in-flight set prevents overlap across heartbeats; the dispatch receipt is recorded at launch so the cooldown gate holds while the script runs. `dispatchPluginToDog` is extracted and shared with the post-script agent step; `FormatAgentStepMailBody` tells the dog the script already ran and must not be re-run.
- Defense-in-depth in `plugins/stuck-agent-dog/run.sh`: both RESTART_POLECAT mailer loops re-check the hook bead's DEFERRED-TRACKER marker immediately before mailing and refuse it — the mailer step is where every historical bypass fired (the detection-time guard already upstream stays as the first layer).

## Tests

- 10 new tests in `internal/plugin/runner_test.go` (exit-code contract, timeout fail-safe, cwd/env, agent-step mail body).
- 4 new tests in `internal/daemon/handler_script_plugin_test.go` (script-instead-of-dog, fail-safe on nonzero exit, exit-10 agent-step attempt, in-flight dedupe).
- 4 new cases in `plugins/stuck-agent-dog/run_test.sh` (detection-time guard crashed+zombie; mailer-gate race crashed+zombie via plain-`bd show` call counting in the fake).

Gates: `go build ./...` + `go vet` clean; `internal/plugin`, `internal/rig`, `internal/daemon` (217s) suites pass; `run_test.sh` 49/49.

### DIFF
--- a/docs/design/plugin-system.md
+++ b/docs/design/plugin-system.md
@@ -91,6 +91,31 @@ Benefits:
 - Plugin failures don't stall patrol
 - Consistent with Dogs' purpose (infrastructure work)
 
+### Execution Model: run.sh Script Contract (op-omcx)
+
+When a plugin directory contains a `run.sh` alongside `plugin.md`, the daemon
+executes the script DIRECTLY (background goroutine; working directory = the
+plugin directory; `GT_TOWN_ROOT` set; `execution.timeout` enforced, default
+5m). The AI dog dispatch above becomes conditional on the script's exit code:
+
+| exit code | meaning | agent step |
+|-----------|---------|------------|
+| `0` | script completed the plugin run | NOT dispatched |
+| `10` | pre-checks passed; plugin requests AI judgment | dispatched with plugin.md instructions + script output tail |
+| other / timeout | failure (recorded as `result:failure`) | NOT dispatched (fail-safe) |
+
+**Why**: instruction placement must match the execution layer. A correct code
+guard in `run.sh` protected nothing for two days because the dispatch path
+only ever fed `plugin.md` to a dog as a prompt — the dog "ran" the plugin by
+reading about it, and six RESTART_POLECAT false fires resulted (op-omcx).
+Deterministic logic and guards belong in `run.sh`, which code executes;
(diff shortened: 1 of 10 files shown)


Bella-Giraffety:
PR Sheriff completed 15 independent research legs and 5 approving reject reviews. This exact head leaves plugin execution authority split across pre-execution success receipts, later failures, cooldown/history, manual shell dispatch, exit-10 dog dispatch, non-atomic assignment, and an unrestricted daemon environment. Closing as-is rather than requesting contributor changes; the valid intent will be carried by a cleanup-first maintainer replacement with attribution.

PR #4509, opened 2026-07-15
fix(dolt): re-enable non-blocking auto_gc to bound sql-server RSS

seanbearden:
## Problem

The managed Dolt sql-server has `auto_gc` disabled in both config generators (after lockups on the old blocking `call dolt_gc()` path). With storage GC off, every commit's chunks accumulate and Dolt keeps + mmaps each table file's index, so RSS grows far beyond on-disk size until the host OOM-cycles it — currently band-aided by an RSS watchdog. (e.g. hq was ~2.1GB on disk for ~27MB of live data.)

## Fix

Re-enable `auto_gc_behavior {enable: true, archive_level: 1}` in both generators (`internal/doltserver/doltserver.go`, `internal/daemon/dolt.go`). Dolt's **non-blocking** auto-GC is default-on since 1.75 (we run 1.83.2) — distinct from the old blocking path that caused the lockups, so it does not reintroduce the `GCSafepointWaiter` panic.

Validated off-prod on a copy of hq: a one-time `dolt gc --full --archive-level=1` reclaimed **98.6%** on disk; `auto_gc` under sustained churn kept RSS in a bounded sawtooth, non-blocking, no lockups.

Adds a **`GT_DOLT_AUTO_GC`** kill-switch (`on`|`off`) to both generators so auto_gc can be disabled at the next Dolt restart without a source revert+rebuild — a runtime escape hatch given the old lockup history (mirrors `GT_DOLT_STATS_ENABLED`).

The separate stats-V2 fix (`dolt_stats_enabled: 0`) is untouched.

## Deploy note

Enabling in-generator only regenerates config on the next Dolt restart. Realizing the reclaim needs a one-time `dolt gc --full` on the live DBs in a maintenance window (operationally gated).

## Test

Updated the config-content assertion and added `TestWriteServerConfig_AutoGCDisabled` (kill-switch). `go test ./internal/doltserver/ ./internal/daemon/` passes.

(hq-excy9g)

### DIFF
--- a/internal/daemon/dog_molecule.go
+++ b/internal/daemon/dog_molecule.go
@@ -272,12 +272,25 @@ func parseChildrenJSON(raw string) ([]childInfo, error) {
 		return arr, nil
 	}
 
-	var wrapped map[string][]childInfo
-	if err := json.Unmarshal(data, &wrapped); err == nil {
-		for _, children := range wrapped {
-			return children, nil
+	// Map form keyed by parent ID: {"hq-wisp-abc": [{...}, ...]}. bd may also
+	// include non-array metadata keys (e.g. "schema_version": 1), so decode into
+	// RawMessage and skip any value that is not a JSON array — otherwise the typed
+	// unmarshal into map[string][]childInfo fails on the metadata key and the whole
+	// parse is lost, pouring a 0-step molecule (gt-dep3).
+	var rawMap map[string]json.RawMessage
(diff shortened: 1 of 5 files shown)


Bella-Giraffety:
Superseded by replacement PR #4533, merged as 86fc560c. The replacement carries forward Sean Bearden's scoped auto-GC fix with attribution and exact-head PR Sheriff evidence.

PR #4522, opened 2026-07-17
fix(hook): recognize patrol roots as formula work

### LINKED ISSUE #4518: gt patrol new creates hooked wisp without molecule; gt mol attach requires pinned status — mutually incompatible
## Summary

`gt patrol new` creates a patrol wisp with `status=hooked` but does **not** attach the patrol molecule to it. `gt hook` then reports "No molecule attached" and suggests running `gt mol attach <bead-id> <molecule-id>`. But `gt mol attach` rejects any bead that is not `pinned`:

```
Error: attaching molecule: issue gt-wisp-tqq is not pinned (status: hooked)
```

Since patrol wisps are born `hooked`, the two commands are mutually incompatible: the wisp can never get its molecule attached through the advertised path. The tool's own error-recovery advice (`gt hook` → "run `gt mol attach`") leads directly into a hard error.

## Reproducer (deterministic)

```bash
cd <town-root>/<rig>
gt patrol new --role witness
# → 🔥 Burned 1 previous patrol wisp(s)
# → gt-wisp-tqq

bd show gt-wisp-tqq
# → ◇ gt-wisp-tqq · mol-witness-patrol   [● P2 · HOOKED]
# → Assignee: ops/witness · Type: molecule
# → DESCRIPTION
# →   (none)            ← no molecule steps materialized

gt mol attach gt-wisp-tqq mol-witness-patrol
# → Error: attaching molecule: issue gt-wisp-tqq is not pinned (status: hooked)
# → exit code 1
```

Reproduced 3× on the same host (wisps `gt-wisp-c05`, `gt-wisp-q35` by our Witness during regular patrol, `gt-wisp-tqq` in a controlled re-run).

## Expected behaviour

Either (in order of preference):

1. **`gt patrol new` attaches the molecule atomically at wisp creation** — the wisp already knows its formula (`mol-witness-patrol` is in the title), so there is no reason to leave it molecule-less and require a second manual step; or
2. **`gt mol attach` accepts `status=hooked`** — if hooked wisps are a legitimate pre-attach state, the attach command must not reject the only status `gt patrol new` produces.

Additionally, `gt hook`'s advice should never point at a command that is guaranteed to fail for the bead's current status.

## Actual behaviour

- Patrol wisp exists, `hooked`, no molecule, no step checklist materialized.
- `gt mol attach` hard-fails with `is not pinned (status: hooked)`.
- No supported way to get the molecule onto the wisp.

## Workaround

Patrol is **not blocked**: `gt prime` renders the full patrol formula checklist inline from the formula catalog regardless of whether a molecule is attached to the wisp. Our Witness works through the 9-step `mol-witness-patrol` checklist from `gt prime` output and closes the wisp with `gt patrol report` / `bd close` at cycle end. The wisp then serves only as the per-cycle digest record, which still works.

## Environment

- `gt version 1.2.1 (319d33a9: HEAD@319d33a91b2d)` — same behavior expected anywhere on this commit
- macOS Darwin 25.5.0 (Apple Silicon), town with 5 rigs, Dolt-backed beads
- Note: on our Ubuntu 24.04 server with the identical binary the repro aborts earlier with `Error: proto mol-witness-patrol not found in catalog` (empty formula catalog), so the incompatibility only manifests on hosts that have the patrol formulas installed.

## Related

- #4094 — prior report from the same deployment (different subsystem, listed for context)

coyaSONG:
## Summary

Treat root-only patrol wisps as formula work in `gt hook` / `gt mol status` instead of suggesting an incompatible `gt mol attach` command.

## Related Issue

Fixes #4518

## Changes

- recognize deacon, witness, and refinery patrol roots using the established patrol-title prefix convention
- preserve explicit `attached_formula` and `attached_molecule` metadata as authoritative
- surface the inferred patrol formula through the existing formula-work status path
- add regression coverage for patrol inference and explicit-metadata precedence

## Root Cause

Patrol wisps intentionally use `status=hooked` without an `attached_molecule`; the root itself is the formula work. Hook status treated every bead without attachment metadata as naked work, so it recommended `gt mol attach`, which accepts pinned handoff beads rather than hooked patrol roots.

## Testing

- [x] Focused regression passes:
  `CGO_ENABLED=0 go test -ldflags='-X github.com/steveyegge/gastown/internal/cmd.BuiltProperly=1' ./internal/cmd -run 'Test(OutputMoleculeStatus|PopulateHookedWorkMetadata)' -count=1`
- [x] `CGO_ENABLED=0 go vet ./internal/cmd`
- [x] `git diff --check`
- [ ] Full `internal/cmd` suite is not green in this environment. The attempted run fails in unrelated macOS `/var` vs `/private/var` path assertions, external-tool call-log fixtures, and unavailable Dolt/tmux integration dependencies; the changed tests pass.

## Checklist

- [x] Documentation updated (not applicable; this corrects status classification and adds regression tests)
- [x] No breaking changes

## AI assistance

Implementation, testing, and PR preparation were performed with OpenAI Codex.

### DIFF
--- a/internal/cmd/molecule_status.go
+++ b/internal/cmd/molecule_status.go
@@ -12,6 +12,7 @@ import (
 	"github.com/spf13/cobra"
 	"github.com/steveyegge/gastown/internal/beads"
 	"github.com/steveyegge/gastown/internal/config"
+	"github.com/steveyegge/gastown/internal/constants"
 	"github.com/steveyegge/gastown/internal/git"
 	"github.com/steveyegge/gastown/internal/style"
 	"github.com/steveyegge/gastown/internal/workspace"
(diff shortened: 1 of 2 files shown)


coyaSONG:
To keep maintainer review load within two concurrent PRs from me in this repository, I am closing this validated draft for now. The branch is synchronized with current main and all checks pass; it can be reopened after one of #4517 or #4525 leaves the queue.

PR #4524, opened 2026-07-17
fix(polecat): route identities to their rig database

jfabre:
## Summary
- route prefixed polecat identities to the configured rig database rather than HQ
- apply consistent routing to hooks, completion metadata, Ralph mode, Witness patrol, Refinery cleanup, and identity listing
- support configured long and hyphenated prefixes, and correctly bind and close routed in-process stores

## Reproduction
A `gst-*` polecat task could be hooked in its rig database while the polecat identity was created or read from HQ. `gt hook show` and `gt prime --hook` then reported no work, causing the polecat to run `gt done` prematurely.

## Validation
- focused routing, lifecycle, Witness, Refinery, Ralph mode, store ownership, and identity-list regression tests
- `gofmt`
- `git diff --check`
- `CGO_ENABLED=0 go vet ./...`
- attempted `CGO_ENABLED=0 go test ./...`; unrelated baseline/environment-sensitive failures remain (including missing ICU `unicode/regex.h` and existing integration/tmux failures)

### DIFF
--- a/internal/beads/beads.go
+++ b/internal/beads/beads.go
@@ -16,6 +16,7 @@ import (
 	"time"
 
 	beadsdk "github.com/steveyegge/beads"
+	"github.com/steveyegge/gastown/internal/constants"
 	"github.com/steveyegge/gastown/internal/runtime"
 	"github.com/steveyegge/gastown/internal/telemetry"
 	"github.com/steveyegge/gastown/internal/util"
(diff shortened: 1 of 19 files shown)


jfabre:
Closing this PR because live polecat validation found an incomplete prime/hook routing path after submission. The branch is preserved for investigation; a replacement will be proposed only after complete end-to-end validation.

PR #4544, opened 2026-07-21
fix: context-check-interval preempts backoff, freezes idle counter

joshuaguyervs:
## Summary
- \`await-event\`'s \`--context-check-interval\` yielded at a fixed wall-clock interval regardless of backoff timeout, so once backoff exceeded the interval (e.g. 8m wait vs 5m interval), the context check fired first and the idle counter never incremented — freezing patrol backoff progression.
- Fix: only yield for context check when the backoff timeout meaningfully exceeds the interval; otherwise let backoff complete naturally so idle increments and tiers advance.

Fixes ci-73d.


joshuaguyervs:

PR #4545, opened 2026-07-21
fix: allow town-level agent beads in rig-scoped resolution

joshuaguyervs:
Fixes ci-lx8. `gt agents resolve --role refinery --rig <rig>` rejected agent beads found only in town-issues, blocking await-event backoff/idle tracking for refinery agents whose bead isn't rig-local. Removed the inconsistent check so patrol can use the available town-level bead for backoff tracking.


joshuaguyervs:

PR #4546, opened 2026-07-21
fix: mark agent beads as ephemeral when creating

joshuaguyervs:
Fixes hq-2a4 (wisp-visibility bug). Patrol/handoff wisps created via gt patrol new / gt mol attach never surfaced via bd show/bd list, breaking witness patrol self-report/close cycles and witness bootstrap on fresh rigs. Root cause: agent beads weren't marked ephemeral at creation time, so the write-visibility path treated them inconsistently from lookup. Marks them ephemeral on creation.


joshuaguyervs:

PR #4547, opened 2026-07-21
fix(polecat): increase doltStateRetries to prevent stalled fresh dispatches

joshuaguyervs:
Fixes hq-uhd. Fresh polecat/dog dispatches consistently started stalled (state=stopped, 0 commits), 100% hit rate on last 5 dispatches per witness report, always recoverable via manual 'gt prime --hook' nudge. Root cause: doltStateRetries was 3 (~3.5s backoff) — too short for Dolt to respond after high-concurrency dispatches, causing state transitions to fail and leaving fresh sessions stuck in 'spawning'. Increased to 10 (~2min) to match the existing hook retry pattern.

PR #4559, opened 2026-07-22
fix(await-event): preserve backoff across context yields

Bella-Giraffety:
Replacement for PR #4547, refreshed on current main fcbfc475.

Carries forward the accepted context-yield regression intent with a narrower cleanup-first fix: preserve the persisted await-event backoff window across context-yield re-entry, and clear it only when the wait completes by event or timeout.

Intentionally not carried forward from #4547: increasing doltStateRetries from 3 to 10, because agent_state is a monitor-only write and extending the synchronous retry loop can block sling/pool-init without focused proof.

Evidence: bead gt-pr-sheriff-4547 records 15 research legs, 5 pre-implementation reviews, 5 post-implementation reviews, and checker-passing merge_replacement evidence. Contributor attribution is preserved via Co-authored-by for dog <joshua.guyer@vergesense.com>.

Local verification:
- git diff --check upstream/main..HEAD
- CGO_ENABLED=0 go test ./internal/cmd -run 'Test(AwaitEventContextYieldPreservesBackoffWindow|AwaitEventTimeoutClearsBackoffWindow|AwaitEventBackoff|AwaitEventResult)$' -count=1
- CGO_ENABLED=0 go test ./internal/cmd ./internal/polecat -count=1 -timeout=10m

Supersedes #4547.

### DIFF
--- a/internal/cmd/molecule_await_event.go
+++ b/internal/cmd/molecule_await_event.go
@@ -200,11 +200,13 @@ func runMoleculeAwaitEvent(cmd *cobra.Command, args []string) error {
 
 	// Resume from backoff-until if interrupted (same pattern as await-signal)
 	timeout := fullTimeout
+	resumed := false
 	now := time.Now()
 	if awaitEventAgentBead != "" && !backoffUntil.IsZero() && backoffUntil.After(now) {
 		remaining := backoffUntil.Sub(now)
 		if remaining <= fullTimeout {
 			timeout = remaining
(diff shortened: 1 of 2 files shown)


Bella-Giraffety:
Closed as superseded by current-main replacement PR #4568, which merged at exact head a6aede3 as merge commit 649b832b. This branch was based on an older release main and is not the landed replacement.

PR #4620, opened 2026-07-30
fix(si-n7vl): sling to a named polecat resumes ITS worktree instead of replacing it

JennyMink:
The compounding half of si-d6kw. #4619 makes the dangerous route (`--branch`
onto a ref another worktree holds) refuse instead of corrupt. It does not make
recovery possible — this does.

## The defect, two layers

`gt sling <bead> <rig>/<polecat>` could not resume an idle polecat:

1. **The 2-part `silicon/keeper` form was only accepted with `--create`.**
   Without it the raw pane error surfaced: `resolving target: getting pane for
   si-keeper: exit status 1`. That is the exact message the operator hit.
2. **When the dead-polecat fallback *did* fire, it replaced the polecat.** It
   called `SpawnPolecatForSling`, which allocates a fresh polecat on a fresh
   branch off main. The named polecat's worktree and branch — the work being
   recovered — were never reattached.

So naming a branch was the only route back onto the original work, and naming a
branch is precisely what put two worktrees on one ref.

## The change

- `polecatTargetRigAndName` (was `missingPolecatTargetRig`) returns the polecat
  **name**, not just the rig, and accepts the shorthand for a polecat that
  **exists** whether or not `--create` was passed. `--create` means "make one
  that does not exist"; this one does.
- `ResumePolecatForSling` reattaches to that polecat: its own worktree, its own
  branch, no `git worktree add`, no new clone. With no `--branch` and no
  `--base-branch` it carries over the branch the worktree is currently on — read
  off the worktree, not off the polecat record, because the record can lag it.
  A worktree parked on the rig default branch or on a detached HEAD has nothing
  to resume and takes the normal fresh-branch path.
- The reuse block `SpawnPolecatForSling` already had is **extracted, not
  copied**, so the pool path and the named path compute the base branch and the
  reuse result identically.

Nothing routes through `git worktree add`, so this path cannot create the
collision #4619 guards against: the polecat already holds its own branch in its
own worktree.

## Behaviour that is deliberately unchanged

An unresumable polecat still gets a fresh one — now with a printed reason,
because a silent substitution is how "resume keeper" became "here is toast".
`--create` still mints a polecat that does not exist. A crew member and a role
name are still not polecat targets. One tightening: a failure *after* reuse
succeeded (missing polecat record, unverifiable worktree) is now a hard error
rather than being papered over with a fresh allocation.

## Verification

`CGO_ENABLED=0 go test ./internal/cmd/` — two failures, both pre-existing and
environmental (macOS `/private/var` symlink in `agent_tracking_beads_test.go`);
both reproduce identically on a clean `origin/main` worktree at 649b832b, which
was run as a baseline before making the claim.

Eight mutations, eight kills — each turned a specific test red:

| mutation | killed by |
|---|---|
| drop the existing-polecat shorthand acceptance | `ShorthandResolvesExistingPolecatWithoutCreate` |
| accept any 2-part target regardless of `--create` | `ShorthandRefusesNonexistentPolecatWithoutCreate` |
| never attempt resume (always spawn fresh) | `TestResolveTargetResumesNamedIdlePolecat` |
| hard-fail instead of falling through to fresh spawn | `SpawnsFreshWhenNamedPolecatCannotResume` |
| never adopt the polecat's own branch | `DefaultsToThePolecatsOwnBranch` |
| drop the default-branch exclusion | `DefaultBranchIsNotResumed` |
| compute the branch decision and discard it | `PassesThePolecatsOwnBranchToReuse` |
| ignore a missing worktree | `FailsWhenThereIsNoWorktreeToResume` |

The first version of this change did **not** kill the "computed and discarded"
mutation: the decision was pure-tested but its call site was not. That is why
`ResumePolecatForSling` is seamed — the hole was found by running the mutation,
not by reading the diff.

### DIFF
--- a/internal/cmd/polecat_spawn.go
+++ b/internal/cmd/polecat_spawn.go
@@ -268,16 +268,35 @@ func reuseIdlePolecatForSling(env *slingPolecatEnv, polecatName string, opts Sli
 // resumeOptionsForPolecat fills in the branch a resume of a NAMED polecat should
 // land on, given the branch its worktree is currently sitting on.
 //
-// With no --branch and no --base-branch, that branch is the polecat's own. This
-// is the point of si-n7vl: recovering an idle polecat must not require the
-// caller to name a branch, because naming one is what put two worktrees on a
-// single ref (si-d6kw). A worktree parked on the rig default branch, or on a
-// detached HEAD, has nothing to resume and takes the normal fresh-branch path.
+// With no --branch and no --base-branch, that branch is the polecat's own — but
+// only when it holds THIS bead's work. This is the point of si-n7vl: recovering
+// an idle polecat must not require the caller to name a branch, because naming
+// one is what put two worktrees on a single ref (si-d6kw).
+//
+// The test is "does the branch encode the bead being slung", not "does the
+// branch look like work". Those differ, and the difference is a live failure:
+// idle keeper sits on polecat/keeper/si-aka.37+ms43gm2z, so slinging a NEW bead
+// si-aka.99 to keeper under the looser test would land si-aka.99's work on
+// si-aka.37's branch and contaminate its MR — with the sling reporting success
+// and no symptom but two beads on one ref (Mayor's ruling, 2026-07-29).
+//
+// Ambiguity resolves to fresh: an unparseable branch, a detached HEAD, the rig
+// default branch, or a parent/child id that does not match exactly. A fresh
+// branch is recoverable; contamination is not.
 func resumeOptionsForPolecat(currentBranch, defaultBranch string, opts SlingSpawnOptions) SlingSpawnOptions {
 	if opts.ResumeBranch != "" || opts.BaseBranch != "" {
 		return opts
 	}
 	if currentBranch == "" || currentBranch == defaultBranch {
+		return opts
+	}
+	// Decode with the inverse of the producer (FormatGeneratedBranchName), so a
+	// change to the branch convention cannot leave this reading the old one.
+	meta, ok := polecat.ParseBranchName(currentBranch)
+	if !ok {
+		return opts
+	}
+	if opts.HookBead != "" && meta.Issue != opts.HookBead {
 		return opts
 	}
 	opts.ResumeBranch = currentBranch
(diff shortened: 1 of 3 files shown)


JennyMink:
## Updated for the Mayor's ruling (hq-z0fxy): the carry-over predicate is narrowed

`1a521fbc`. The ruling is right and the failure it names is real: "carry over
unless it is the rig default branch" tests for **has work**, not for **has *this*
work**.

    keeper  polecat/keeper/si-aka.37+ms43gm2z
    dag     polecat/dag/si-aka.9+ms43eli5
    nux     polecat/nux/si-aka.12+ms41h57u
    furiosa master

Sling a new bead `si-aka.99` to idle `keeper` and si-aka.99's work lands on
si-aka.37's branch, contaminating its MR. The sling reports success and the
tests pass — the only symptom is two beads' work on one ref, which is the same
silence this bead exists to remove.

**Now:** carry over iff the branch's embedded bead equals the bead being slung,
or no bead is being slung. `--branch` / `--base-branch` still win. Ambiguity
resolves to fresh — unparseable branch, detached HEAD, rig default branch, or a
parent/child id that is not an exact match (`si-aka.9` vs `si-aka.9.1`). Fresh
is recoverable; contamination is not.

The bead is decoded with `polecat.ParseBranchName`, the existing inverse of
`FormatGeneratedBranchName`, so the read cannot drift from the convention that
wrote the name. Both worked cases survive unchanged: dag/si-aka.9 recovery
matches and resumes; furiosa on master has no embedded bead and goes fresh.
Acceptance D is unaffected — the caller still names no branch.

Four more mutations, four kills: drop the bead-match check; prefix-match instead
of exact; require a bead when none is slung; carry over an unparseable branch.

**The last one survived at first.** With a bead present, the bead-match test
already rejects an undecodable branch, so every test I had passed a bead and not
one of them could tell whether the unparseable check existed at all. It needed
the no-bead case — where the ruling's two rules overlap ("no bead slung → carry
over" vs "ambiguous → fresh") and ambiguity has to win. Twelve mutations now,
twelve kills.

Full `CGO_ENABLED=0 go test ./internal/cmd/` unchanged: the same two
pre-existing macOS `/private/var` symlink failures that reproduce on clean
`origin/main` at 649b832b.

PR #4636, opened 2026-08-03
fix(patrol): refuse to cook a patrol wisp that cannot name its own rig (dbt-as8)

blairsilverberg:
Closes dbt-as8.

Every rig-scoped patrol formula declares `rig` with a default of `UNSET_RIG` so it parses standalone. A minting caller that never supplies the var therefore does not fail — it renders ~25KB of shell instructions addressed to a rig that does not exist:

```
YOUR_AGENT_BEAD=$(gt agents resolve --role witness --rig UNSET_RIG --json | jq -r .id)
```

`gt agents resolve` exits 1, but the pipe means `$?` is jq's, so the assignment succeeds with the literal string `"null"`. The formula's next line — *"This must resolve exactly one bead ID. If it fails, STOP and report the error"* — is prose addressed to the agent, not a shell test, and `"null"` is non-empty so even a `[ -z ]` guard would pass it. `await-signal` then warns three times and carries on with idle tracking, backoff persistence and the agent heartbeat all silently no-op'd, so the patrol never reaches abbreviated effort. The formula states its own intent 20 lines above the defect ("idle cycles should burn ~10% of the tokens a full patrol uses"), so an idle patrol on an affected build burns roughly ten times that, indefinitely. **It does not fail safe.**

## What this changes

1. `gt patrol new` and `gt patrol report` build the witness `PatrolConfig` with no `ExtraVars` at all, while `gt prime` passes `buildWitnessPatrolVars`. The wisp is *created* from those vars, so on those two paths the stored var set carries neither `rig` nor `prefix`.
2. The wisp was CREATED from `cfg.ExtraVars` but its description was RENDERED from a different, role-derived set. `patrolFormulaVars` is now the single source for both, so stored vars and the instructions the agent actually reads cannot disagree about which rig the patrol belongs to. Duplicate keys collapse last-wins — the precedence `buildFormulaVarMap` already applied.
3. The fail-safe, and **on current `main` this is the only part that changes rendered output**: `assertPatrolRigSubstituted` refuses a rendered patrol description still carrying the sentinel, and `autoSpawnPatrol` now renders *before* it burns the previous cycle's wisp or creates a new one, so a patrol that cannot name its rig leaves nothing behind. Other render failures stay non-fatal exactly as before; only the unsubstituted rig is fatal, via a distinct sentinel error.

The guard keys on the **rendered text**, not on the formula name, so it says SAFE for formulas with no rig var at all — the deacon patrol renders unchanged, and the tests assert that arm in the same run as the refusing arm.

## Scope — narrower than the bug report, and worth stating plainly

Measured rather than inherited from the report:

- **Against this PR's base (`649b832b`), the four real role/rig arms already render correctly.** `2bd80946 fix(patrol): rebuild heartbeat and root-wisp flow (#4496)` added `rig=` to `buildRefineryPatrolVars` and moved the render path to derive role vars from the assignee. So this PR does **not** repair a broken rendering on `main`. What it repairs is the residual **fail-open**: a config from which no rig can be derived at all renders 25,886 bytes carrying the sentinel with `err=nil`, plus the create/render var-set split in (2).
- **The reported symptom is nonetheless live in a running town, because the deployed build predates `2bd80946`.** One probe run against the real formulas on three trees: on the deployed commit all four arms carry `UNSET_RIG`; on this PR's base none do. That half is a rebuild, not a code change, and nothing here can clear already-minted wisps.
- **It is neither rig-specific nor witness-specific.** On the deployed tree the witness arm fails because two of three callers pass no vars; the refinery arm fails for an unrelated reason — `buildRefineryPatrolVars` never emitted `rig=` there at all — so refinery was broken on every path in every rig. A fix that only corrected the caller list would have left every refinery patrol broken.

## Verification

Both verdicts in one run, against the live 25KB/42KB formulas rather than fixtures:

| arm | base `649b832b` | this branch |
|---|---|---|
| witness + refinery × 3 real rigs | renders `--rig <rig>`, 0 sentinel | same |
| rig-less assignee | **25,886 B, `err=nil`, sentinel present** | **refused with a named error** |
| rig-less **deacon** (control) | 50,158 B, `err=nil` | 50,158 B, `err=nil` |

The deacon control is load-bearing: a guard that could only refuse would fail that arm. Also checked against the live town's 47 formula files — `UNSET_RIG` occurs exactly twice, both as `default = "UNSET_RIG"` in the two rig-scoped patrol formulas — so no legitimate description can trip the refusal.

`go build ./...` and `go vet ./...` clean; `internal/cmd` and `internal/constants` pass. `internal/doctor` and `internal/doltserver` fail **identically on this branch and on an unmodified base checkout** (`TestGetServerAddr_UsesConfigYAMLPort`, `TestWaitForCatalog_NoServer`, `TestFindBrokenWorkspaces_MultipleRigs`) — pre-existing and environment-dependent, untouched here.

New tests only; no pre-existing test was modified. Code authored by `nitro`, whose session was terminated by account quota before it could open this PR; re-verified end to end and the commit message corrected where it described the deployed tree rather than this base.

### DIFF
--- a/internal/cmd/patrol_helpers.go
+++ b/internal/cmd/patrol_helpers.go
@@ -36,6 +36,12 @@ type PatrolConfig struct {
 // Remaining stale beads are cleaned by burnPreviousPatrolWisps at cycle end.
 const maxStalePurgePerRun = 5
 
+// errPatrolRigUnsubstituted marks a patrol description that still carries the
+// UNSET_RIG sentinel. It is a distinct sentinel error so that autoSpawnPatrol can
+// treat it as fatal while leaving every other render failure as non-fatal as
+// before (dbt-as8).
+var errPatrolRigUnsubstituted = errors.New("patrol rig var was never substituted")
+
 // findActivePatrol finds an active patrol molecule for the role.
 // Returns the patrol ID, display line, and whether one was found.
 // Returns an error if discovery fails (e.g. transient bd failure),
@@ -198,6 +204,18 @@ func autoSpawnPatrol(cfg PatrolConfig) (string, error) {
 		return "", refinery.NewSafetyStoppedError(stop)
 	}
 
+	// Render the description BEFORE mutating anything, so a patrol that cannot
+	// name its own rig burns nothing and creates nothing (dbt-as8).
+	//
+	// A render FAILURE stays non-fatal, as it was: it is warned about where the
+	// description is written, below. An UNSUBSTITUTED RIG is fatal, because that
+	// wisp is not merely bare — it is ~25KB of instructions naming a rig that
+	// does not exist, and every command inside it fails in a way nothing checks.
+	desc, descErr := renderPatrolWispDescription(cfg)
+	if descErr != nil && errors.Is(descErr, errPatrolRigUnsubstituted) {
+		return "", descErr
+	}
+
 	// Resolve the beads directory following redirects.
 	// This ensures bd targets the correct database (e.g., rig database
 	// instead of HQ) regardless of inherited BEADS_DIR. See gt-ctir.
@@ -244,8 +262,11 @@ func autoSpawnPatrol(cfg PatrolConfig) (string, error) {
 	// Create the patrol wisp (root only — steps are read inline at prime time,
 	// not tracked as individual DB rows). Child wisps are reserved for pour=true
 	// formulas like releases where checkpoint recovery matters.
+	// The wisp is created with the SAME var set the description is rendered from
+	// (patrolFormulaVars), so the stored vars and the instructions the agent
+	// reads can never disagree about which rig this patrol belongs to (dbt-as8).
 	spawnArgs := []string{"mol", "wisp", "create", protoID, "--root-only", "--actor", cfg.RoleName}
-	for _, v := range cfg.ExtraVars {
+	for _, v := range patrolFormulaVars(cfg) {
 		spawnArgs = append(spawnArgs, "--var", v)
 	}
 	cmdSpawn := BdCmd(spawnArgs...).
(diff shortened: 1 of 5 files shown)


blairsilverberg:
### The red `Integration Tests` is an infrastructure timeout, not a test result — no test ever ran

Recording the evidence rather than asking anyone to take my word for it, because a red X on a CI job reads as "this change breaks integration tests" and here it cannot mean that.

The job died in the **`Pre-pull Dolt Docker image`** step, before any test binary was invoked:

```
##[group]Run docker pull dolthub/dolt-sql-server:2.0.7
Error response from daemon: Get "https://registry-1.docker.io/v2/": context deadline exceeded
##[error]Process completed with exit code 1.
```

Grepping that job's own log for any sign of a test run — `gotestsum|=== RUN|PASS:` — returns **0**, against a positive control of **294** readable log lines, so the zero is a true zero and not a blind read. No `junit-*.xml` was produced either; the `junit-report.py` lines that follow in the log are the next step running against files that do not exist.

**I cannot re-run it.** Both the run-level and job-level rerun endpoints return `403 Must have admin rights to Repository`, since this PR comes from a fork. (`gh run rerun --failed` reports "its workflow file may be broken", which is misleading — the underlying error is the 403.) A maintainer re-running that one job should clear it.

Every other check is green: `Test`, `Lint`, `Windows Smoke Test`, `Reject go.mod replace directives`, `Reject issues.jsonl`, `add-triage-label`. `Block Internal PRs` and `[code]smith` are skipped.

**What I could and could not verify locally**, stated separately so the gap is visible:

- Verified: `go build ./...` and `go vet ./...` clean; `internal/cmd` (the only package with behaviour changes) and `internal/constants` pass.
- Verified pre-existing, by running both arms rather than asserting it: `internal/doctor` and `internal/doltserver` fail identically on this branch and on an unmodified `649b832b` checkout — `TestGetServerAddr_UsesConfigYAMLPort`, `TestWaitForCatalog_NoServer`, `TestFindBrokenWorkspaces_MultipleRigs`. Environment-dependent, untouched here.
- **Not verified, and I want to be explicit about it:** the integration suite itself. This machine has no Docker, so `go test -tags=integration ./internal/cmd/...` dies with `integration TestMain: dolt setup: Docker not available` — identically on this branch and on a clean base checkout. That reproduces the *environment* failure, not the suite, so it discriminates nothing and I am not offering it as evidence the suite passes.

So the honest status of the integration suite against this change is **unknown**, not green. Given the diff is confined to patrol var construction plus a render-time guard, with the new unit tests covering both verdicts, I do not expect integration fallout — but that is an expectation, not a measurement.

PR #4673, opened 2026-08-10
health: stop reporting green over absent databases and absent backups

morganvenable:
Two independent false greens in `gt health`. Both were found on a town where the affected subsystem was **completely absent** and health reported it fine, so neither is a cosmetic reporting nit — in both cases the instrument reads identically whether or not the thing it measures works.

## 1. Phantom databases, and dolt.log pollution

`checkDatabaseHealth` and `checkPollution` iterated a hardcoded `{hq, gt, mo}` and opened a connection per name. A town with only `hq` queries two databases that do not exist. Every query error is swallowed by the `_ =` assignments, so the missing databases are reported as healthy all-zero rows — indistinguishable from real but empty ones.

Each invocation also writes **44 `database not found` lines into dolt.log** (22 per absent database). On one idle town that reached 2376 lines, **62% of the log**. That log is the first artifact the operational runbook says to capture when diagnosing a hung Dolt server, so the noise degrades the diagnostic that matters most, exactly when it matters.

Fixed by enumerating `SHOW DATABASES` and intersecting with the known set, falling back to the full list on any enumeration error so behaviour is unchanged if enumeration fails.

Measured on a live town:

```
baseline:      4752
after old gt:  4796   (delta 44)
after fixed:   4796   (delta  0)
```

and the report now lists only `hq`, the database that exists.

## 2. A missing backup reported as not-stale

`DoltStale` and `JSONLStale` were assigned only inside their "backup exists" branches, so the zero value — `false`, meaning fresh — survived whenever no backup was found. **"No backup has ever been taken" and "backed up seconds ago" produced identical output.**

This is not hypothetical. A town ran a full day with zero backups on either leg — no Dolt backup ever taken, and the JSONL ticker skipping every firing against a missing archive dir — while `gt health` reported both legs not-stale throughout. Nobody could have noticed from health output.

Fixed by defaulting both `Stale` fields to `true` and adding explicit `dolt_present` / `jsonl_present`, so "no backup" is distinguishable from "old backup" without overloading a single bool. The text report now prints `NO BACKUP FOUND` with a warning marker rather than a dim `not found` that reads as neutral.

Verified on a live town — `dolt_stale` flips to `true` with `dolt_present:false`, correctly reporting a backup that does not exist:

```
BEFORE                          AFTER
"name": "hq"                    "name": "hq"
"name": "gt"      <- phantom
"name": "mo"      <- phantom
"dolt_stale": false  <- lie     "dolt_stale": true
                                "dolt_present": false
"jsonl_stale": true             "jsonl_stale": true
                                "jsonl_present": true
```

## Note on a related mismatch, not fixed here

`checkBackupHealth` looks for the Dolt backup at `townRoot/.dolt-backup`, but `bd backup init` writes its destination to `.beads/dolt-backup.json` and can point anywhere. On the town above, backups exist and verify by restore, yet `dolt_present` is correctly `false` because they are not at the path health probes. Left alone deliberately — reading the bd config is a larger change and a separate decision.

## Testing

Built and run against a live town; both behaviours verified by value (error deltas and JSON field flips shown above) rather than by inspection. No existing tests cover `checkDatabaseHealth` / `checkBackupHealth`; happy to add coverage if you'd like it in this PR.

### DIFF
--- a/internal/cmd/health.go
+++ b/internal/cmd/health.go
@@ -8,6 +8,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"sort"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


morganvenable:
Pushed a follow-up commit — the original fix was only half of this defect.

The hardcoded `{hq, gt, mo}` list is wrong in **both** directions, and I only addressed one:

- **Too many** (fixed in the first commit): absent databases queried on every invocation, reported as healthy zeros, 44 `database not found` lines per call.
- **Too few** (this commit): **every rig gets its own database, and none of them are named `hq`, `gt` or `mo`.** On a town with four rigs, health reported one database and silently omitted the other four. Rig data had no health coverage at all — and the omission was invisible, because nothing reports a database as *missing*.

That second half is arguably the worse one: the phantom databases at least showed up as suspicious all-zero rows. An omitted rig database produces no output whatsoever.

`discoverProductionDBs` now enumerates `SHOW DATABASES` and excludes only server-internal schemas, sorted for stable output. The hardcoded list survives purely as a fallback if enumeration fails, so a broken query degrades to the old behaviour rather than to reporting nothing.

Verified on a live town with four rigs:

```
before this commit:   hq
after this commit:    coparent, gastown, hello_world, hq, svalscripts
dolt.log errors added per invocation: 0   (unchanged)
```

Credit where due — I had shipped the half-fix and believed it complete. The omission direction was caught by the Deacon patrol that originally reported the phantom databases, after new rig databases appeared and it noticed they never showed up in health output.

PR #3849, opened 2026-05-06
fix: exclude agent beads from AutoClose to prevent polecat state deletion

dah81:
Fixes issue hq-qeae: AutoClose was missing the agent bead filter that Reap and Scan have, causing stale agent beads to be auto-closed.

This led to 5 raybar polecat beads being deleted when they should have been preserved.

Changes:
- Add 'AND i.issue_type != 'agent'' filter to AutoClose WHERE clause
- Update AutoClose docstring to mention agent bead exclusion  
- Add TestAutoCloseExcludesAgentBeads to verify the filter

This matches the existing filters in Reap() and Scan() functions.

Related work:
- Issue: hq-ahhy (Merge gastown fix for reaper AutoClose agent bead filter)
- Root cause: hq-qeae (5 raybar polecat agent beads disappeared)

---
<a href="https://app.blacksmith.sh/gastownhall/codesmith/gastown/pr/3849"><picture><source media="(prefers-color-scheme: dark)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://pr-comments-assets.blacksmith.sh/codesmith/view-in-codesmith-light.svg">View in Codesmith</picture></a>
<sup>Need help on this PR? Tag <code>@codesmith</code> with what you need.</sup>

- [ ] Let Codesmith autofix CI failures and bot reviews

PR #4871, opened 2026-09-15
fix(reaper): protect gt:agent beads from AutoClose

### LINKED ISSUE #4630: AutoClose omits gt:agent from its protected-label list: agent registry beads are auto-closed on a 7-day updated_at clock that heartbeats do not reset
## Summary

`AutoClose`'s protected-label list does not contain `gt:agent` — the label that *defines* an agent registry bead. Every agent registry bead is therefore an auto-close candidate on a 7-day `updated_at` clock, and **heartbeats do not reset that clock** because they write to the `labels` table and never touch `issues.updated_at`.

In our deployment this silently closed 5 agent registry beads at 13–15 days stale before anyone noticed.

All source references below were read from `main` at the time of filing (`raw.githubusercontent.com`, HTTP 200), not inferred from behaviour.

## The predicate

`internal/reaper/reaper.go`, `AutoClose`:

```
i.status IN ('open','in_progress')
AND i.updated_at < NOW() - <staleAge>
AND i.priority > 1                       -- P0/P1 exempt
AND i.issue_type NOT IN ('epic','convoy')
AND i.id NOT IN (SELECT ... FROM labels l
                 WHERE l.label IN ('gt:standing-orders', 'gt:keep', 'gt:role', 'gt:rig'))
AND <no open dependency in either direction>
```

- `internal/reaper/reaper.go:741` — the four-label list above.
- `internal/daemon/wisp_reaper.go:30` — `defaultStaleIssueAge = 7 * 24 * time.Hour`, passed at `wisp_reaper.go:310`.
- `grep -c 'gt:agent' internal/reaper/reaper.go` → **0**. Positive control in the same run: `grep -c 'gt:keep'` → **1**, so the probe can express presence.

The same four labels are duplicated in `internal/beads/beads.go:295`, `ProtectedIssueLabel`:

```go
func ProtectedIssueLabel(label string) bool {
	switch strings.ToLower(strings.TrimSpace(label)) {
	case "gt:standing-orders", "gt:keep", "gt:role", "gt:rig":
		return true
	default:
		return false
	}
}
```

which is what `IsProtectedBead` (`beads.go:338`) delegates to, and which documents itself as guarding "AutoClose, unassign on polecat removal, etc."

Meanwhile `IsAgentBead` (`beads.go:323`) identifies agent beads primarily **by the label**:

```go
func IsAgentBead(issue *Issue) bool {
	if issue == nil {
		return false
	}
	// Check legacy type field first for backward compatibility
	if issue.Type == "agent" {
		return true
	}
	// Check for gt:agent label (current standard)
	return HasLabel(issue, "gt:agent")
}
```

So the codebase has one concept of "this is an agent bead" and a second, disjoint concept of "this bead is protected", and agent beads are absent from the second.

## Why heartbeats don't save them

Heartbeats write `heartbeat:<epoch>` / `idle:<n>` rows into the **labels** table. They never write a column on `issues`, so `updated_at` reflects only real column writes and a continuously-live agent can look arbitrarily stale to the reaper.

Demonstrated from a single AutoClose batch in our deployment, two independent instances, no waiting required:

| registry bead | heartbeat label last written | outcome |
| --- | --- | --- |
| `op-openclaw-witness` | 2 days before the batch | **closed** |
| `cap-capital-witness` | 5 days before the batch | **closed** |

If heartbeat writes touched `updated_at`, neither could have satisfied `updated_at < NOW() - 7d`. Both did.

The two survivors in that batch are also explained by `updated_at` alone: one was the only registry bead carrying notes (i.e. real column writes), the other was created inside the 7-day window. Note for anyone reproducing this: `started_at` is a **correlate, not a cause** — it appears nowhere in the predicate. A started bead gets worked, and being worked bumps `updated_at`. We initially mis-attributed the survivorship to `started_at` and it does not hold up.

## Impact

Beyond losing the registry row: `gt agents resolve` matches only **open** beads. When a registry bead closes, resolution returns null, the patrol `loop-or-exit` step loses its idle signal, and the patrol runs at **full effort instead of backing off** — for every affected role in every rig. That is a recurring token-cost multiplier with no visible cause, which is why we'd rate this above a tidiness bug.

Population in our (single) deployment: 60 `gt:agent` beads, none carrying any protected label before we intervened; 5 already closed; a further ~39 within 1–2 days of the line.

## Prior art, and why the earlier approach would not work today

PRs #3848 / #3849 ("fix: exclude agent beads from AutoClose to prevent polecat state deletion") targeted this exact function and failure class and were both closed unmerged. They fixed it with a **type** predicate, `i.issue_type != 'agent'`.

That predicate would be a no-op on current `main`: per `IsAgentBead` above, the type field is explicitly "legacy … for backward compatibility" and the `gt:agent` label is "current standard", so a type filter no longer matches a newly-created agent bead. We mention this only so the earlier attempt isn't revived as-is.

Possibly related, not duplicates as far as we can tell: #4397, #4535, #4513 (closed by #4530), #1035.

## Suggested fix

Add `'gt:agent'` to both copies of the protected-label list:

- `internal/reaper/reaper.go:741` — the `AutoClose` `whereClause` label list
- `internal/beads/beads.go:295` — `ProtectedIssueLabel`

Separately worth considering, but a wider behaviour change that we'd suggest **not** bundling: whether a heartbeat should touch `issues.updated_at`. That arguably fixes the root cause rather than the symptom, but it changes staleness semantics for everything that reads `updated_at`.

## Suggested acceptance test

A stored count won't work here, because the population moves — "the registry is all open" passes on a broken guard. We'd suggest a two-arm behavioural test in **one** run:

- a freshly created `gt:agent` bead with **no** `gt:keep` label and `updated_at` forced past the stale cutoff is **refused** by AutoClose; **and**
- a non-agent P2 bead of the same age in the same run **is** taken.

A guard that cannot say SAFE isn't a guard; a guard that cannot say TAKEN isn't being exercised.

## Workaround for anyone hitting this now

Add `gt:keep` to existing `gt:agent` beads. Two caveats: the label write itself bumps `updated_at`, which destroys the staleness evidence the diagnosis rests on (snapshot first if you want to confirm), and beads created afterwards are unprotected from birth.

keppy:
## Summary

`AutoClose` auto-closes agent registry beads. Its protected-label exclusion lists `gt:standing-orders`, `gt:keep`, `gt:role`, `gt:rig` — but not `gt:agent`, the label that *defines* an agent bead (`IsAgentBead`). Heartbeats write `heartbeat:<epoch>` / `idle:<n>` rows to the `labels` table and never touch `issues.updated_at`, so a continuously live agent looks stale to the 7-day clock and gets closed. Once closed, `gt agents resolve` (which only matches open beads) returns nothing and the role's patrol loses its idle/backoff signal and runs at full effort.

This adds `gt:agent` to both copies of the list and pins them together with a test so they cannot drift again.

## Related Issue

Fixes #4630

## Changes

- `internal/reaper/reaper.go`: add `'gt:agent'` to the `AutoClose` protected-label `IN (...)` list; document why in the function comment.
- `internal/beads/beads.go`: add `gt:agent` to `ProtectedIssueLabel`; introduce `ProtectedIssueLabels()` (returns a copy) as the single source of truth the predicate iterates over.
- `internal/reaper/reaper_test.go`: `TestAutoCloseProtectedLabelsMatchBeads` parses the label list out of the `AutoClose` SQL and asserts set-equality with `beads.ProtectedIssueLabels()`. Test-only import — the reaper package still has no runtime dependency on `internal/beads`.
- `internal/beads/beads_test.go`: table tests for `ProtectedIssueLabel` (incl. normalization and non-protected `gt:*` labels), `ProtectedIssueLabels()` copy semantics, and `IsProtectedBead` with agent beads carrying heartbeat labels.

Behavior change is limited to AutoClose. The two existing `IsProtectedBead` callers already skip agent beads via `IsAgentBead` before consulting it (`polecat/manager.go`, `cmd/polecat_inventory.go`), so nothing else changes.

Deliberately **not** bundled: making heartbeats touch `issues.updated_at`. That would fix the root cause but changes staleness semantics for everything that reads `updated_at`; the issue suggests keeping it separate and I agree.

Note on #3848 / #3849: those earlier attempts used `issue_type != 'agent'`, which is a no-op for current agent beads because the type field is legacy and identification is now by label.

## Testing

- [x] Unit tests pass for the touched packages (full `go build ./...` and `go vet ./...` also clean; I did not run the entire `go test ./...` suite locally)
  - `go test ./internal/reaper/` — ok (new `TestAutoCloseProtectedLabelsMatchBeads` passes, and fails if either list is edited alone)
  - `go test ./internal/beads/` — new tests pass; the package has 10 pre-existing failures on Windows (`pwsh` not on PATH, bd stub scripts). I ran the suite against `main` with the change stashed and the failure set is byte-identical, so none are introduced here.
  - `go build ./...` and `go vet ./...` — clean (Windows, cgo + MSYS2 ICU per README)
- [x] Manual testing performed — no live town available, so I reproduced it against a local Dolt SQL server with the beads v1.0.5 schema and ran `gt reaper auto-close --stale-age 168h` from both `main` (649b832b) and this branch. On `main` the live `gt:agent` bead (with `heartbeat:`/`idle:` labels, `updated_at` 8 days old) is closed; on this branch it stays open. Details in the comment below.

## Checklist

- [x] Documentation updated (if applicable) — function comments updated; no user-facing docs reference the list


keppy:
Manual repro (I've checked the box in the description now):

**Setup:** local `dolt sql-server` (Dolt 2.3.5, Windows), database created from the beads v1.0.5 migrations (`internal/storage/schema/migrations/*.up.sql`). Three open P2 beads, all with `updated_at` = now − 8 days:

| bead | labels |
|---|---|
| `gt-agent-live` | `gt:agent`, `heartbeat:<now>`, `idle:3` |
| `gt-keep-agent` | `gt:agent`, `gt:keep` (the reporter's workaround) |
| `gt-plain-stale` | none (control) |

**Run:** `gt reaper auto-close --db gt_repro --stale-age 168h` (168h is the daemon's `defaultStaleIssueAge`). I ran it with a binary built from `main` @ 649b832b, reset the rows, then ran it with one built from this branch.

| bead | `main` | this PR |
|---|---|---|
| `gt-agent-live` | **closed** (`stale:auto-closed by reaper`) | open |
| `gt-keep-agent` | open | open |
| `gt-plain-stale` | closed | closed |

So on `main`, a live agent bead with a current heartbeat gets closed, which is the bug in #4630. With this PR it stays open. Ordinary stale beads are still closed.

